PR #519, opened 2014-12-10
Articles/#380 reply reply to originator differences

mauroservienti:
* [ ] add to the index

@Particular/documentation can you review?


mauroservienti:
@indualagarsamy @johnsimons would you mind taking a look at my comments, so that I can move on?

johnsimons:
@mauroservienti as I said before I'm actually on the fence whether this clarifies the mess or not.

mauroservienti:
@SimonCropp @andreasohlund 

* given what we discussed is this valuable as a sample?
* Should we drop it entirely?
* Is it OK to merge as docu?

SimonCropp:
I dont mind having this as doco. But i dont think it is finished. The image confuses things more than anything. And the text doesnt explain the differences, it just points to the image and does a mic drop. For this to be good doco we need to have a definition for each and then compare/contrast the two. Only after that can we have a diagram that accompanies the text. 

And using a si image for doco is not good. Si serves a different purpose and screen captures from si will never make good doco.

mauroservienti:
OK, I'll move on, refactor the original article, that was a note in an email and create some well looking diagram incorporating Indu's feedback.





        Sent by Outlook for Android




On Mon, Apr 20, 2015 at 1:17 PM -0700, "Simon Cropp" <notifications@github.com> wrote:












I dont mind having this as doco. But i dont think it is finished. The image confuses things more than anything. And the text doesnt explain the differences, it just points to the image and does a mic drop. For this to be good doco we need to have a definition for each and then compare/contrast the two. Only after that can we have a diagram that accompanies the text. 



And using a si image for doco is not good. Si serves a different purpose and screen captures from si will never make good doco.

mauroservienti:
[Direct view](https://github.com/Particular/docs.particular.net/blob/articles/%23380-Reply-ReplyToOriginator-differences/nservicebus/sagas/reply-replaytooriginator-differences.md)

SimonCropp:
@mauroservienti this looks good. a few minor things

- [ ] Can you commit the raw assets you used to create the diagram
- [ ] Add it to the menu

mauroservienti:
@SimonCropp both done, please check the what I did with the menu.yaml is correct. Thanks.

SimonCropp:
@andreasohlund this looks ok to me. but since, int he past few days, u have been deep diving into how ReplyToOriginator works can u please review this for correctness or any possible additions?

andreasohlund:
Looks good, we're missing docs on how auto correlate works (and how it doesn't work btw sagas) I'll open a new pull for that

+1 to pull this

Issue #460, opened 2014-10-25
Add information about remote queues


sfarmar:
Raised by @udidahan 
Add information about remote queues with regards to MSMQ, 
specifically explaining that if they want remote queues that can be done with brokers like RabbitMQ.

SimonCropp:
looks like this is done http://docs.particular.net/nservicebus/msmq-information#remote-queues

PR #522, opened 2014-12-12
added comment for remote queues

sfarmar:
https://github.com/Particular/docs.particular.net/issues/460


sfarmar:
Thanks @mauroservienti :+1:

PR #524, opened 2014-12-12
ServiceMatrix - Publish/Subscribe stand alone article

dprothero:
@indualagarsamy OK, here is the reworked standalone pub/sub article


SimonCropp:
no longer needed

PR #548, opened 2015-01-27
Update SagaBasics.cs

fafachd:
This page only shows the NHibernate example of a MySagaFinder.FindBy() method. I added a RavenDb example since I couldn't find any example of this online. People shouldn't have to dig up NServiceBus.RavenDb source code to figure out how to do this (I had to)! The paragraph below this snippet needs to be updated a bit once you merge this example.


SimonCropp:
thanks

PR #551, opened 2015-01-28
callback use for request/response

SeanFeldman:
Added section on callbacks usage for request/response scenario to address users that need it [usergroup questions](https://groups.google.com/d/msg/particularsoftware/-YmEoyof9U0/GxKdqNdqZ8wJ) mostly in the context of web applications.

### DIFF
--- a/nservicebus/how-do-i-handle-responses-on-the-client-side.md
+++ b/nservicebus/how-do-i-handle-responses-on-the-client-side.md
@@ -32,9 +32,11 @@ To trigger a callback, you need to return an `enum` or `int` value.
 
 <!-- import TriggerCallback -->
 
-## Using callbacks in a web application
+## When should you use callbacks?
 
-When should you use callbacks? When the data returned is **not business critical and data loss is acceptable** and **blocking operation** is acceptable. Otherwise, use [request/response](/samples/fullduplex) with a message handler for reply messages.
+Use callbacks when the data returned is **not business critical and data loss is acceptable**. Otherwise, use [request/response](/samples/fullduplex) with a message handler for reply messages.
+
+When using callbacks in ASP.NET web application, NServiceBus is not blocking executing thread and allowing processing of other requests. When response is recieved, it is handled and returned to the client side. Client will be blocked while waiting for response. This scenario is common when migrating from traditional blocking request/response to messaging.
 
 To access response message through callback, the following code can be used
 <!-- import CallbackToAccessMessageRegistration -->
--- a/nservicebus/how-do-i-handle-responses-on-the-client-side.md
+++ b/nservicebus/how-do-i-handle-responses-on-the-client-side.md
@@ -34,7 +34,7 @@ To trigger a callback, you need to return an `enum` or `int` value.
 
 ## Using callbacks in a web application
 
-When should you use callbacks? When the data returned is **not business critical and data loss is acceptable**. Otherwise, use [request/response](/samples/fullduplex) with a message handler for reply messages.
+When should you use callbacks? When the data returned is **not business critical and data loss is acceptable** and **blocking operation** is acceptable. Otherwise, use [request/response](/samples/fullduplex) with a message handler for reply messages.
 
 To access response message through callback, the following code can be used
 <!-- import CallbackToAccessMessageRegistration -->
--- a/nservicebus/how-do-i-handle-responses-on-the-client-side.md
+++ b/nservicebus/how-do-i-handle-responses-on-the-client-side.md
@@ -34,10 +34,9 @@ To trigger a callback, you need to return an `enum` or `int` value.
 
 <!-- import TriggerCallback -->
 
-## Handling callbacks in the context of web application
+## Using callbacks in a web application
 
-Callbacks can be used to provide an acknowledgement for successfully dispatched command, but should not be used instead of message handlers. Message handlers should handle asynchronous communication using [Full Duplex](/samples/fullduplex). Request/response with callback is possible, through discurraged and reason is outlined above. 
+When should you use callbacks? When the data returned is **not business critical and data loss is acceptable**. Otherwise, use [request/response](/samples/fullduplex) with a message handler for reply messages.
 
 To access response message through callback, the following code can be used
 <!-- import CallbackToAccessMessageRegistration -->
-
--- a/nservicebus/how-do-i-handle-responses-on-the-client-side.md
+++ b/nservicebus/how-do-i-handle-responses-on-the-client-side.md
@@ -25,10 +25,19 @@ When sending a message/request, you can register a callback that is invoked when
 
 
 ```C#
-bus.Send( messageInstance ).Register( asyncCallback, state );
+bus.Send(messageInstance).Register(asyncCallback, state);
 ```
 
-DANGER: If the server process returns multiple responses, NServiceBus cannot know which response message will be the last. To prevent memory leaks, the callback is invoked only for the first response. Callbacks won't survive a crash as they are held in memory, so they are less suitable for server-side development where fault-tolerance is required. In those cases, [sagas are preferred](sagas-in-nservicebus.md) .
+DANGER: If the server process returns multiple responses, NServiceBus cannot know which response message will be the last. To prevent memory leaks, the callback is invoked only for the first response. Callbacks won't survive a crash as they are held in memory, so they are less suitable for server-side development where fault-tolerance is required. In those cases, [sagas are preferred](sagas-in-nservicebus.md).
 
-If your client is a web application, use the `RegisterWebCallback` method.
+To trigger a callback, you need to return an `enum` or `int` value.
+
+<!-- import TriggerCallback -->
+
+## Handling callbacks in the context of web application
+
+Callbacks can be used to provide an acknowledgement for successfully dispatched command, but should not be used instead of message handlers. Message handlers should handle asynchronous communication using [Full Duplex](/samples/fullduplex). Request/response with callback is possible, through discurraged and reason is outlined above. 
+
+To access response message through callback, the following code can be used
+<!-- import CallbackToAccessMessageRegistration -->
 
--- a/Snippets/Snippets_3/Callback/Callback.cs
+++ b/Snippets/Snippets_3/Callback/Callback.cs
@@ -1,0 +1,56 @@
+﻿namespace MyServer.Callback
+{
+    using NServiceBus;
+
+    public class Callback
+    {
+        class PlaceOrder : ICommand
+        {
+        }
+
+        class PlaceOrderResponse : IMessage
+        {
+            public object Response { get; set; }
+        }
+
+        public void CallbackSnippet()
+        {
+            var placeOrder = new PlaceOrder();
+            IBus bus = null;
+            // ReSharper disable once NotAccessedVariable
+            PlaceOrderResponse message; // get replied message
+
+            #region CallbackToAccessMessageRegistration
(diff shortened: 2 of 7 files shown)


SeanFeldman:

nservicebus/how-do-i-handle-responses-on-the-client-side.md
```diff
@@ -25,10 +25,19 @@ When sending a message/request, you can register a callback that is invoked when
...
+DANGER: If the server process returns multiple responses, NServiceBus cannot know which response message will be the last. To prevent memory leaks, the callback is invoked only for the first response. Callbacks won't survive a crash as they are held in memory, so they are less suitable for server-side development where fault-tolerance is required. In those cases, [sagas are preferred](sagas-in-nservicebus.md).
 
-If your client is a web application, use the `RegisterWebCallback` method.
+To trigger a callback, you need to return an `enum` or `int` value.
+
+<!-- import TriggerCallback -->
+
+## Handling callbacks in the context of web application
+
+Callbacks can be used to provide an acknowledgement for successfully dispatched command, but should not be used instead of message handlers. Message handlers should handle asynchronous communication using [Full Duplex](/samples/fullduplex). Request/response with callback is possible, through discurraged and reason is outlined above.
```
johnsimons: What do you mean by ?
> through discurraged and reason is outlined above

nservicebus/how-do-i-handle-responses-on-the-client-side.md
```diff
@@ -25,10 +25,19 @@ When sending a message/request, you can register a callback that is invoked when
...
+To trigger a callback, you need to return an `enum` or `int` value.
+
+<!-- import TriggerCallback -->
+
+## Handling callbacks in the context of web application
+
+Callbacks can be used to provide an acknowledgement for successfully dispatched command, but should not be used instead of message handlers. Message handlers should handle asynchronous communication using [Full Duplex](/samples/fullduplex). Request/response with callback is possible, through discurraged and reason is outlined above. 
+
+To access response message through callback, the following code can be used
+<!-- import CallbackToAccessMessageRegistration -->
```
johnsimons: This is a hack, yes it works but I don't think was ever intentional.

nservicebus/how-do-i-handle-responses-on-the-client-side.md
SeanFeldman: The fact that it's not resilient to web app restarts as mentioned in `Danger` note

nservicebus/how-do-i-handle-responses-on-the-client-side.md
SeanFeldman: Don't know if it was intentional or not, but it is provided as an answer http://stackoverflow.com/questions/3668648/using-nservicebus-with-asp-net-mvc-2/3673664#3673664

nservicebus/how-do-i-handle-responses-on-the-client-side.md
johnsimons: what is "discurraged", is that misspelled ?

nservicebus/how-do-i-handle-responses-on-the-client-side.md
johnsimons: I've asked the question on slack

nservicebus/how-do-i-handle-responses-on-the-client-side.md
SeanFeldman: :) indeed

nservicebus/how-do-i-handle-responses-on-the-client-side.md
SeanFeldman: +1

johnsimons:
I think it looks good now, but lets wait for @andreasohlund and @udidahan to chime in on whether accessing a message from the callback is something we should be endorsing or not.

nservicebus/how-do-i-handle-responses-on-the-client-side.md
```diff
@@ -17,18 +17,25 @@ public class MyHandler : IHandleMessages<MyMessage>
...
 
 ```C#
-bus.Send( messageInstance ).Register( asyncCallback, state );
+bus.Send(messageInstance).Register(asyncCallback, state);
 ```
 
-DANGER: If the server process returns multiple responses, NServiceBus cannot know which response message will be the last. To prevent memory leaks, the callback is invoked only for the first response. Callbacks won't survive a crash as they are held in memory, so they are less suitable for server-side development where fault-tolerance is required. In those cases, [sagas are preferred](sagas-in-nservicebus.md) .
+DANGER: If the server process returns multiple responses, NServiceBus cannot know which response message will be the last. To prevent memory leaks, the callback is invoked only for the first response. Callbacks won't survive a crash as they are held in memory, so they are less suitable for server-side development where fault-tolerance is required. In those cases, [sagas are preferred](sagas-in-nservicebus.md).
+
+To trigger a callback, you need to return an `enum` or `int` value.
```
andreasohlund: They will trigger on bus.Reply as well but perhaps we should leave that out here to avoid getting users in trouble?

The syntax would be:

```
 .Register( (AsyncCallback)delegate(IAsyncResult asyncResult)
                                                {
                                                    var Result = asyncResult.AsyncState as CompletionResult;
                                                    replyMessage = Result.Messages[0] as TReply;
                                                },
                                                null);

```

johnsimons: Well if this is something we do not want users doing it, yes let's leave it
out.
Wouldn't this kind of rpc a style that we should discourage/hide users away?
I actually always thought that the syntax to get the messages out on a
callback was a hack that user found out and eventually got out.

andreasohlund: Agreed, @udidahan thoughts?

SeanFeldman: Based on response from @udidahan on slack
>yes - leave as is.
Any objections to merge PR?

nservicebus/how-do-i-handle-responses-on-the-client-side.md
```diff
@@ -17,18 +17,25 @@ public class MyHandler : IHandleMessages<MyMessage>
...
+
+<!-- import TriggerCallback -->
+
+## Handling callbacks in the context of web application
```
andreasohlund: Is "Using callbacks in a web application" a better heading?

nservicebus/how-do-i-handle-responses-on-the-client-side.md
```diff
@@ -17,18 +17,25 @@ public class MyHandler : IHandleMessages<MyMessage>
...
+
+Callbacks can be used to provide an acknowledgement for successfully dispatched command, but should not be used instead of message handlers. Message handlers should handle asynchronous communication using [Full Duplex](/samples/fullduplex). Request/response with callback is possible, although discouraged, see reason outlined above.
```
andreasohlund: This doesn't read well to me. Acks are valid even outside a web app?

How about make this section more along the lines of "When should you use callbacks" ?

* When the data returned isn't business critical and its ok if it gets lost
* If not, then use request/response (bus.Reply) with a message handler for the reply etc etc

What we could mention in regards to webapps is that RegisterWebCallback hooks up to the asp.net async support and there by frees up threads on the webserver.

What do you think?

nservicebus/how-do-i-handle-responses-on-the-client-side.md
SeanFeldman: Yes

nservicebus/how-do-i-handle-responses-on-the-client-side.md
SeanFeldman: I have removed mentioning of `RegisterWebCallback` as it was removed as of v3.
As for the rest of the comment, completely makes sense. Will update.

SeanFeldman:
@andreasohlund have simplified that section.

nservicebus/how-do-i-handle-responses-on-the-client-side.md
```diff
@@ -17,18 +17,24 @@ public class MyHandler : IHandleMessages<MyMessage>
...
+
+<!-- import TriggerCallback -->
+
+## Using callbacks in a web application
 
-If your client is a web application, use the `RegisterWebCallback` method.
+When should you use callbacks? When the data returned is **not business critical and data loss is acceptable**. Otherwise, use [request/response](/samples/fullduplex) with a message handler for reply messages.
 
+To access response message through callback, the following code can be used
+<!-- import CallbackToAccessMessageRegistration -->
```
johnsimons: I thought we all agreed that we are **not** promoting this kind of hack?

nservicebus/how-do-i-handle-responses-on-the-client-side.md
```diff
@@ -17,18 +17,24 @@ public class MyHandler : IHandleMessages<MyMessage>
...
+DANGER: If the server process returns multiple responses, NServiceBus cannot know which response message will be the last. To prevent memory leaks, the callback is invoked only for the first response. Callbacks won't survive a crash as they are held in memory, so they are less suitable for server-side development where fault-tolerance is required. In those cases, [sagas are preferred](sagas-in-nservicebus.md).
+
+To trigger a callback, you need to return an `enum` or `int` value.
+
+<!-- import TriggerCallback -->
+
+## Using callbacks in a web application
 
-If your client is a web application, use the `RegisterWebCallback` method.
+When should you use callbacks? When the data returned is **not business critical and data loss is acceptable**. Otherwise, use [request/response](/samples/fullduplex) with a message handler for reply messages.
```
johnsimons: Another reason is also, if you are ok with waiting for things to come back

SeanFeldman: Good point @johnsimons, added.

nservicebus/how-do-i-handle-responses-on-the-client-side.md
udidahan: @johnsimons how about "although other architectural approaches should be preferred over this, if you need to access a response message through a callback, the following code can be used:"

nservicebus/how-do-i-handle-responses-on-the-client-side.md
```diff
@@ -17,18 +17,24 @@ public class MyHandler : IHandleMessages<MyMessage>
...
+DANGER: If the server process returns multiple responses, NServiceBus cannot know which response message will be the last. To prevent memory leaks, the callback is invoked only for the first response. Callbacks won't survive a crash as they are held in memory, so they are less suitable for server-side development where fault-tolerance is required. In those cases, [sagas are preferred](sagas-in-nservicebus.md).
+
+To trigger a callback, you need to return an `enum` or `int` value.
+
+<!-- import TriggerCallback -->
+
+## Using callbacks in a web application
 
-If your client is a web application, use the `RegisterWebCallback` method.
+When should you use callbacks? When the data returned is **not business critical and data loss is acceptable** and **blocking operation** is acceptable. Otherwise, use [request/response](/samples/fullduplex) with a message handler for reply messages.
```
andreasohlund: This is relevant for all endpoints not just web apps. Perhaps promote `When should you use callbacks?` as a heading?

Then you could use the "Using callbacks in a web application" to explain how to hook up to the async asp.net support in order to free up web server threads?

SeanFeldman:
@andreasohlund updated based on our conversation. Please review.

PR #559, opened 2015-02-05
Update home.md

sergioc:
Modified html to allow for a 2x2 layout for products + SC

Need to add servicecontrol.png (https://www.dropbox.com/s/aewg6y73ilve106/serviceControl.png?dl=0) + change links.


sergioc:
@SimonCropp is there a way for us to preview this?

SimonCropp:
not really. but if you have modified the html in the same way as you have on your local copy it should not be necessary.

but it is trivial to roll back or tweak if something is wrong

sergioc:
I suggest we pull this while we're both online, so that we can quickly roll
back.

Still need to add the right SC links and image, though.

Sergio

SimonCropp:
closing this since it is out of date and will need to be re-done

Issue #134, opened 2014-01-11
Author comparison table between transports 


SimonCropp:
Raised by @dannycohen
Migrated from by https://github.com/Particular/NServiceBus/issues/1455

Follow-up on #1454, including features, and pros and cons of each.

SimonCropp:
decided this should not be in our doco.

brianbolger:
Are there any plans to share the earlier content/docs on this... future blog post?

PR #562, opened 2015-02-09
Articles/#134 transport comparison

mauroservienti:
This PR is intended to start a discussion related to the transport comparison article we discussed in #134.

Currently the draft document in this PR was designed, w/ @dannycohen, with comparison in mind, so the goal was to highlight, in a matrix, transports features and NServiceBus features across transports, for each feature give a brief description and link to the relevant doco page.

Each feature description, currently listed under the matrix, is intended to be hosted in a light-box and not shown as a list.

What's missing?

* [ ] a general review of the overall idea;
* [ ] review of the transports matrix;
* [x] details for the SqlTransport;
* [ ] most of the descriptions of the features;
* [ ] strengths/weaknesses review per transport;

@Particular/documentation chime in.


SimonCropp:
I have gone over this a few times in my head. I think I have settled on "lets do this as a blog/story and not as a docs page"

The reasoning is as follows

- It is a moving target. versions and features of transports change and we would need to constantly keep this up to date if it is a docs page
- it is very difficult to create this doco without some of it becoming opinion based
- it is difficult to target this kind of comparison without having mor context of the business domain. often the definition of a benefit/disadvantage is dependent on the perspective. thie is perspective we do not have
- it is simply not "nservicebus doco". in a similar way that a comparison between oracle and sqlserver (which is possible since we support both via nhibernate) is not a domain we want to document.
- it is a very deep topic. if we go this deep in the transports we should also put similar effort into comparisons of containers/logging/serializers/transports

So I think a blog post would be better here since it does not have the above negatives

SimonCropp:
assuming people are ok with my previous comment so closing this

PR #569, opened 2015-02-16
minor grammatical update

bbrandt:
**types** is the subject, so **is** would be the verb.
http://www.grammarbook.com/grammar/subjectVerbAgree.asp


mauroservienti:
thanks, merged.

PR #570, opened 2015-02-17
Prerequisites added for the browser

danielmarbach:
@sergioc please review

### DIFF
--- a/servicepulse/introduction-and-installing-servicepulse.md
+++ b/servicepulse/introduction-and-installing-servicepulse.md
@@ -15,7 +15,7 @@ Particular Software ServicePulse is the operational monitoring tool for distribu
 ### Prerequisites
 
 * .NET Framework 4.5 or later
-* Internet Explorer 9+  or similar modern browser
+* Internet Explorer 9 or higher, Chrome 35 or higher, Firefox 31 or higher, Safari 7 or higher
 * Particular Software ServiceControl 
 * Monitored NServiceBus endpoints must use NServiceBus 4.0.0 or higher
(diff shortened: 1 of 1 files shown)


SimonCropp:
I think "similar modern browser" is too vague. Sounds like "meh we could not be bothered testing"

sergioc:
Why do we only support IE9+? IE8 comes standard with Win7 - do you see IE8 presenting any major restrictions?

SimonCropp:
I thought ie9 was standard with win 7?

SimonCropp:
Either way ie9 is 3years old and 3 versions old. It is a reasonable stance to only support at far back

sergioc:
hehe, nope IE8 is standard with Win7. Maybe Win7 SP1 installers add IE9, but pre-SP1 installers definitely install IE8...

sergioc:
I'd say IE9+ is probably a good idea, because:

- Microsoft dropping support for IE8 completely in Jan. 2016 - no point in restricting front-end because of another 10 months for IE8
- IE8 global market share is relatively high, but probably mostly due to being the highest IE version supported on WinXP
- also difficult to know for sure what's IE8's enterprise market share
- IE9 already provides some basic HTML5 support

Regarding other browsers, I'd say focus on detailing major browser versions:
- support Chrome 35+ (older versions still used, but market share extremely small)
- support Firefox 31+ (current ESR (extended support release) version is 31)
- support Safari 7+ (compatible with Mac OS 10.8+, still receives security updates from Apple)

danielmarbach:
@sergioc I added the reviews. Please review again and merge

SimonCropp:
@sergioc @danielmarbach is this good to merge?

danielmarbach:
From my side yes

PR #575, opened 2015-02-24
Documented the need to have unique QMIds

janovesk:
@indualagarsamy: Is this a good location to document the duplicate QMIds issue?


mauroservienti:
Looks good, a couple of comments.

* Should we add an external link to MSDN pointing to "Sysprep"?
* We have another page that speaks about clustering with MSMQ we should link from that page to the paragraph here adding a not there (http://docs.particular.net/nservicebus/deploying-nservicebus-in-a-windows-failover-cluster)

janovesk:
Good suggestions, @mauroservienti. Added them to the PR. Merging it now.

PR #583, opened 2015-03-06
WIP Pipelinestream

SimonCropp:


johnsimons:
:+1: to send this to the customer

johnsimons:
@SimonCropp do you want to package this and give it to the customer or want me to do it ?

johnsimons:
I had a go at seeing what could be done by replacing existing behaviours.
I do have something working, and the customer would only need to modify the sender endpoints.
My version is more generic, it support all types, the existing `IDataBusProperty`, `byte[]` and `Stream`, but, it is not as optimised as this version.
Without using `Stream` for everything we will always be doing allocations which rely on memory, so even though my version works for all the above types, it puts pressure on memory and depending on load and sizes of data it will eventually break.

SimonCropp:
Thoughts on how well this approach would work with any future native stream support?

-----Original Message-----

PR #585, opened 2015-03-17
hosting samples

mauroservienti:
adding a set of samples to show various hosting scenarios:

* basic self-hosting (v3, v4, v5)
* NServiceBus.Host (v3, v4, v5)
* Self hosting with the Bootstrapper (v3, v4, v5)
* multi hosting (v5)

@SimonCropp can you please review?


samples/hosting/multi-hosting/Version_5/Sample/packages.config
```diff
@@ -0,0 +1,4 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
```
SimonCropp: Why 5.1 and not 5.2?

mauroservienti: I started from another sample that was based on 5.1, I'll update it to the latest. Thanks.

SimonCropp:
I dont think the windows service sample should contain the word bootstrapper. Yes u can use the bootstrapper nuget to create it but the result is simply a windows service host. I think u just need a note about the v5 bootstrapper in the content. Also while u cant use the bootstrapper nuget against v4 u can copy the basics of the code to createa v 4 version.

samples/hosting/multi-hosting/Version_5/Sample/Program.cs
```diff
@@ -0,0 +1,56 @@
...
+	{
+		#region multi-hosting
+
+		var bus1 = StartInstance1();
+		var bus2 = StartInstance2();
+
+		Console.WriteLine( "\r\nBus instances created and configured; press any key to stop program\r\n" );
+		Console.ReadKey();
+
+		bus1.Dispose();
```
SimonCropp: These should be usings.

mauroservienti:
Good point, I'll update the name and add a v4 sample. Shouldn't it work for v3 as well?

samples/hosting/multi-hosting/Version_5/Sample/Program.cs
mauroservienti: I didn't like to nest using a one inside the other, do you think I should?

SimonCropp:
Yep should work for v3

SimonCropp:
but i think the rest is good

mauroservienti:
@SimonCropp all fixed

mauroservienti:
merged

PR #591, opened 2015-03-23
asa_client (and ada_server) have been deprecated

pashute:
Added remark about deprecated Asa_Client and Asa_Server


SimonCropp:
I have updated this sample and this PR is no longer relevant. But thanks for the contribution

PR #597, opened 2015-04-09
small grammar corrections

pashute:
are (not: is) for plural. 
hyphenated: publish-and-subscribe


mauroservienti:
merged, thanks.

PR #599, opened 2015-04-20
Describing outbox ravendb and including how to configure timetokeep

danielmarbach:
I'm not yet sure whether we should really document the access to the `Outbox.*` settings. Really on the fence here. But I added time for completeness reasons.

Additionally I have a few questions about the current structure:

We have the outbox doco page which is describing outbox in general and then primarily NH persistence and Sql transport related things (before). Now I just added as an extension the ravendb outbox settings as well. My question is:

Should we move all persistence specific doco into a page in the corresponding persister? And if we do so also align the navigation there? For example currently the persister in the NH case is called `NHibernate persister` but in the RavenDB case simple `RavenDB`.

### DIFF
--- a/Snippets/Snippets_5/Outbox/RavenDB/TimeToKeep.cs
+++ b/Snippets/Snippets_5/Outbox/RavenDB/TimeToKeep.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using NServiceBus;
-using NServiceBus.Configuration.AdvanceExtensibility;
 
 public class TimeToKeep
 {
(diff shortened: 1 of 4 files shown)


Snippets/Snippets_5/Outbox/RavenDB/TimeToKeep.cs
```diff
@@ -0,0 +1,16 @@
...
+using NServiceBus.Configuration.AdvanceExtensibility;
+
+public class TimeToKeep
+{
+    public void Customize()
+    {
+        BusConfiguration configuration = null;
+
+        #region OutboxRavendBTimeToKeep
+        configuration.GetSettings().Set("Outbox.TimeToKeepDeduplicationData", TimeSpan.FromDays(7));
```
SimonCropp: Shouldnt these be explicit apis?

danielmarbach: @SimonCropp we decided not to do it ecause of this discussion: https://github.com/Particular/NServiceBus/pull/2712#issuecomment-93671459

danielmarbach:
Updated doco. Now Using extension methods

PR #610, opened 2015-04-30
Inform NServiceBus.RavenDB v3.x not available yet

pashute:
It is not clear that NServiceBus.RavenDB versions 3.x have not been released yet, causing confusion.


SimonCropp:
@pashute while you are correct about the timeline for v3 of raven that "version 3" text you modified is actually referring to version 3 of NServicebus, not Ravendb. I have added some extra text to clarify that

PR #624, opened 2015-05-10
grammar

pashute:
grammar!!


pashute:
also, if I understand correctly, you do not need to inherit ITimeoutData anymore for timeouts...

SimonCropp:
@pashute thanks

And I assume you are referring to ITimeoutState which is no longer required

pashute:
correct.

2015-05-12 11:04 GMT+03:00 Simon Cropp <notifications@github.com>:

PR #627, opened 2015-05-12
Updated assembly scanning docu and snippets

ramonsmits:
I have updated the scanning docu because of some desk cases related to assembly scanning.

* Mentioning that using `AllAssembly` helper can be used to either to whitelist or blacklist assemblies and that its methods behave like `.StartsWith(string)`
* ScanningIncludeByPattern snippet now includes the pattern to include all NServiceBus assemblies.
* ScanningMixingIncludeAndExclude snippet previously didn't make any sense as the exclusion was not included in the first place. Updated to reflect a more common scenario.


SimonCropp:
@ramonsmits this PR seems to contradict this doco http://docs.particular.net/nservicebus/hosting/assembly-scanning

> During the scanning process, the core dlls for NServiceBus namely NServiceBus.Core.dll, NServiceBus.dll (in versions prior to Version 5) and if in use NServiceBus.Host.exe are automatically included since the endpoint needs them to function properly.

@johnsimons can u weigh in

SimonCropp:
merged manually

PR #630, opened 2015-05-15
Corrections to setting up MarkdownPad

SeanFeldman:
Discrepancy between this doco and internal one at https://github.com/Particular/Operations.DocsWeb/blob/master/ReadMe.md#code-snippets-in-markdownpad (internal works, this one doesn't)

@SimonCropp this is something you might have a little more insight into? Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,9 +163,9 @@ For editing markdown on your desktop (after cloning locally with Git) try [Markd
 
 #### Markdown flavor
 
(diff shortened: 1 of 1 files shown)


SimonCropp:
One is for the general public who dont have access to the markdown pad hack and one is for us. The doco is correct

SeanFeldman:
Good to know that those have to be different. It's not stated in the private one, so not clear. // cc @indualagarsamy 
Thanks.

PR #637, opened 2015-05-22
Updated setting-custom-hostname.md

froko:
Small improvement: The port number in the documentation should be 33333 instead of 80


gbiellem:
Thanks for the update. I've included your change into a larger update of that page

PR #640, opened 2015-05-26
WIP: Issue 612 azure sample - shared hosting

SeanFeldman:
I have tried to bring azure multi-host sample to a bare minimum and running within context of Visual Studio only w/o need to use a 3rd party tool or command line. Had to use AzCopy to allow pushing zip files into emulated storage in order not to resort to powershel scripts.

@Particular/documentation @Particular/azure please review


SimonCropp:
can u update to Azure sdk 2.6

SimonCropp:
dont think the cost of checking in AZCopy is worth it

Instead point them to this http://azure.microsoft.com/en-us/documentation/articles/storage-use-azcopy/#download-and-install-azcopy

samples/azure/shared-host/sample.md
```diff
@@ -0,0 +1,84 @@
...
+ 1. Run the solution 
+ 1. Inspect Azure Storage Emulator Tables for `MultiHostedEndpointsOutput` table and its content for something like the following:
+
+| PartitionKey | RowKey | Timestamp | Message |
+|:--|:--|:--|:--|
+|Sender	|2015-05-25 14:58:33	|5/25/2015 8:58:33 PM	|Pinging Receiver |
+|Receiver	|2015-05-25 14:58:34	|5/25/2015 8:58:34 PM	|Got Ping and will reply with Pong |
+|Sender	|2015-05-25 14:58:35	|5/25/2015 8:58:35 PM	|Got Pong from Receiver |
+
+WARNING: When running sample, do not stop execution from Visual Studio, but instead stop cloud service from Azure Compute Emulator UI. Do to an issue with Azure Compute Eulator not invoking OnStop method, multi hosted endpoints are not terminated.
```
SimonCropp: Eulator

samples/azure/shared-host/sample.md
```diff
@@ -0,0 +1,84 @@
...
+
+NOTE: To inspect multi-host [role emulated file system](https://msdn.microsoft.com/en-us/library/azure/hh771389.aspx), navigate to c:\Users\<username>\AppData\Local\dftmp\Resources
```
SimonCropp: dftmp is a well known string and not some short guid that will be diff on diff machines?

SimonCropp:
since zips are being built in this sample make sure they are excluded from git so no one accidentally commits them

SeanFeldman:
Yep, zips are all ignored here https://github.com/Particular/docs.particular.net/pull/640/files#diff-a084b794bc0759e7a6b77810e01874f2R96

SeanFeldman:
In regards to AZCopy @SimonCropp, I thought to go that route. And frankly, that would be several times easier than what I ended up doing. The issue is that then it's back to square one with no F5-kind-of experience and frankly AzCopy syntax is not making it easier for someone who just wants to run it. I'd keep it for now, and if down the road it taxes too much, then change approach.

SimonCropp:
well the user is going to have to learn to use azcopy anyway so just document the syntax in the .md and let them run it. it is simply not worth the update hassle and perf impact on rendering to check in.

For an idea of how much larger it makes the repo. that samples (with the binaries) increases the size of the entire docs codebase by 30%.

SeanFeldman:
>well the user is going to have to learn to use azcopy anyway so just document the syntax in the .md and let them run it. i

Not really. Users might not need to touch AzCopy ever. All depends on type of deployment and scenario they'll have.

>For an idea of how much larger it makes the repo. that samples (with the binaries) increases the size of the entire docs codebase by 30%.

Yes, that is huge. I still think there's a way to automate this w/o forcing users to go back to VideoStore type of execution. It is not fun to execute AzCopy each time after compilation. I'll try another approach that is more of a hybrid. Will cause users to install AzCopy and configure solution, but will keep repo slim.

SimonCropp:
> It is not fun to execute AzCopy each time after compilation

Then just force them to have AzCopy in the path. u can leave everything else the same

SeanFeldman:
shhh @SimonCropp , don't reveal my secret :)

SeanFeldman:
@SimonCropp give it a try. Thanks.

SimonCropp:
did u update to the 2.6 sdk?

.gitignore
```diff
@@ -90,4 +90,7 @@ Fody/.DS_Store
 *.userprefs
 *.jtconfig
 
-*nsb_log*
\ No newline at end of file
+*nsb_log*
+
+samples/azure/shared-host/Version_5/tools/AzCopy/*
```
SimonCropp: do we still need this ?

SeanFeldman: Yes, we need it. 
Similar to `_zips` folder, we don't want anyone accidentally commit and push AzCopy files.

SeanFeldman:
@SimonCropp thank you fixing MSBuild to take in account compilation on TC :smiley: 
Looks like this is done and can be merged.

SimonCropp:
@SeanFeldman what was your final decision on this?

SeanFeldman:
@SimonCropp as we talked about it, I am going to split F5/debugging aspect from deployment aspect so that upon download users can execute, set breakpoints and evaluate. Deployment would be outside of VS and scripted.

SeanFeldman:
@SimonCropp I have updated sample the following way
- Removed dependency of AzCopy. Taking dependency on PowerShell instead.
- Split sample doco into running in dev mode, deployment, and running with cloud service

Could you review and let know if this approach would be developer friendlier? Thanks.

SimonCropp:
@SeanFeldman i still think the whole hosting story has problems. but that is a problem with the host code not with your sample and hence should not stop you from finishing the sample. so i think go with what you have

SeanFeldman:
@Particular/azure I'd like to merge this one. Any final comments?

PR #643, opened 2015-05-28
Documentation on custom nhibernate mappings

ramonsmits:
It describes what options are available, their differences and how they can be used with NServiceBus.


nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,153 @@
...
+Custom mapping options:
+
+* Create a `*.hbm.xml` file for your saga data class
+* Use Fluent NHibernate (popular seperate fluent api)
+* Use NHibernate.Mapping.Attributes
+* Loquacious Configuration (native fluent api)
+
+There are probably even other options but these are the most frequently used.
+
+See http://docs.particular.net/nservicebus/nhibernate/configuration on how you can pass this configuration object to the NServiceBus NHibernate configuration.
```
SimonCropp: @ramonsmits links need to be relative or fully qualified by starting with a "/" https://github.com/Particular/docs.particular.net#linking  
if you install the markdownpad hack it will perform these checks for you

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,153 @@
...
+
+NOTE: When you create a custom mapping then you are responsible for the correct behavior assigning the correct primary key.
+
+NOTE: The `[Unique]` attribute will not used to generate a index. You are responsible to add this to your custom mapping. This is especially importent when using optimistic concurrency control and a transaction isolation level different from serializable.
+
+
+## Custom *.hbm.xml mapping
```
SimonCropp: An `*` means italic in markdown. which means this is not rendering nice in github

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,153 @@
...
+### Mappings as embedded resources
+
+It is not required to create your own NHibernate configuration object and pass it to the NServiceBus NHibernate configuration.
+
+
+### Read mappings from file system instead of embedded resources
+
+If you do not want to embed the `*.hbm.xml` files then you can also load mappings from the file system. Create a custom NHibernate configuration object and use the following example to add mappings from the file system.
+
+```
```
SimonCropp: suggest you use a snippet for this https://github.com/Particular/docs.particular.net#defining-snippets  let me know if u need help getting started

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,153 @@
...
+2. Create a custom NHibernate configuration
+** via FluentNHibernate
+** by creating a new Configuration instance and pass it to FluentNHibernate
+4. Pass it to the NServiceBus NHibernate configuration.
+
+
+Example of a possible implementation:
+
+```
+static Configuration BuildConfiguration(Configuration nhConfiguration)
```
SimonCropp: use a snippet

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,153 @@
...
+```
+
+
+## Use Fluent NHibernate
+
+Fluent NHibernate gives you a type-safe mapping approach where the mapping is seperate from the classes. The benefit is that you get compile time feedback when a mapping is not valid any more when you have breaking changes in your mapping.
+
+To use it with NServiceBus:
+
+1. Install `FluentNHibernate package` via Nuget.
```
SimonCropp: so `FluentNHibernate package` isnt really "code"

SimonCropp: should there be a link to the package here?

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,153 @@
...
+
+
+References:
+* http://nhibernate.info/doc/nhibernate-reference/mapping-attributes.html
+* https://github.com/nhibernate/NHibernate.Mapping.Attributes
+
+
+## Use the Loquacious API
+
+The Loquacious api is NHibernate its native fluent api. It not only can help you create a type-safe configuration but you can also create your custom mappings. The benefit is that this api is already available via the NHibernate package.
```
SimonCropp: > The Loquacious api is NHibernate its native fluent api.

I think this has too many words

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,153 @@
...
+ - Persistence
+ - Configuration
+ - Mapping
+related:
+ - nservicebus/nhibernate/accessing-data
+ - nservicebus/nhibernate/index
+---
+
+
+Sometimes the default schema that gets generated for your saga does work. For example, you might need to store strings with a different string length, want so use complex types or want to store DateTime values with a high precision or maybe want to tweak the eager or lazy loading to improve performance. You will need to create a custom mapping to achieve this.
```
SimonCropp: > Sometimes the default schema that gets generated for your saga does work. For example, you might n

You are missing some text here? it read like you are giving examples of when the default schema is correct when you are actually giving examples of the opposite

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,153 @@
...
+
+Custom mapping options:
+
+* Create a `*.hbm.xml` file for your saga data class
+* Use Fluent NHibernate (popular seperate fluent api)
```
SimonCropp: `separate`

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,153 @@
...
+* Use NHibernate.Mapping.Attributes
+* Loquacious Configuration (native fluent api)
+
+There are probably even other options but these are the most frequently used.
+
+See http://docs.particular.net/nservicebus/nhibernate/configuration on how you can pass this configuration object to the NServiceBus NHibernate configuration.
+
+NOTE: When you create a custom mapping then you are responsible for the correct behavior assigning the correct primary key.
```
SimonCropp: > responsible for the correct behavior assigning the correct primary key.

is this missing some words?

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,153 @@
...
+
+NOTE: The `[Unique]` attribute will not used to generate a index. You are responsible to add this to your custom mapping. This is especially importent when using optimistic concurrency control and a transaction isolation level different from serializable.
```
SimonCropp: "responsible for adding"

SimonCropp: > This is especially importent when using optimistic concurrency control and a transaction isolation level different from serializable.

Perhaps

> This is especially important when using both optimistic concurrency and a transaction isolation level different from serializable.

Also can we have some links for context here?

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,153 @@
...
+
+
+## Custom *.hbm.xml mapping
+
+Using NHibernate mapping files is the original way to customize your mapping. It just needs to be created and needs to be embedded as a resource in your assembly or available on the file system.
```
SimonCropp: > It just needs to be created and needs to be embedded as a resource in your assembly or available on the file system.

perhaps

> It needs to be created and then either embedded as a resource in your assembly or available on the file system.

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,153 @@
...
+{
+	return Fluently.Configure(nhConfiguration)
+	    .Mappings(cfg =>
+	    {
+	        cfg.FluentMappings.AddFromAssemblyOf<MySagaData>();
+	    }).BuildConfiguration();
+}
+```
+
+Its best to read the [Getting started]() article from the FluentNHibernate project to see how you create mappings.
```
SimonCropp: link?

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,153 @@
...
+1. Add the nuget package `NHibernate.Mapping.Attributes`
+2. Create a custom NHibernate configuration object.
+3. Initialite the attribute mapping (see sample below).
+3. Pass it to the NServiceBus NHibernate configuration.
+
+
+Initialize the NHibernate attribute based mappings:
+
+```
+static void AddAttributeMappings(Configuration nhConfiguration)
```
SimonCropp: make this a snippet

ramonsmits:
@SimonCropp Thanks for the detailed feedback.

nservicebus/nhibernate/custom-mappings.md
ramonsmits: That is only when something is wrapped between asterisks right on the same line? I don't see issues while previewing.

ramonsmits:
@SimonCropp Please let me know if the NOTE regarding the [Unique] attributes is now better and if this is what you had in mind with the Snippets.

Also, I now put the snippets in the v5 project although its the same for all as the snippets are related to NHibernate and the referenced projects. So they are not really bound to a specific NHibernate version.

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,125 @@
...
+
+References:
+* http://nhibernate.info/doc/nhibernate-reference/mapping-attributes.html
+* https://github.com/nhibernate/NHibernate.Mapping.Attributes
+* http://www.nuget.org/packages/NHibernate.Mapping.Attributes/
+
+
+## Use the Loquacious API
+
+The Loquacious api is NHibernate its native fluent mapping ability. You declare your mapping in code just like FluentNHibernate but using a different syntax. It can help you create a type-safe configuration but you can also create your custom mappings. The benefit is that this api is already available via the NHibernate package so requires no additional downloads.
```
SimonCropp: `additional downloads` isnt really a negative. better phrased as `additional dependencies`

SimonCropp:
there are a lot of raw links in the PR. I would prefer to see them at formatted links. ie instead of
`http://www.nuget.org/packages/FluentNHibernate/` use `[Fluent Nhibernate Package](http://www.nuget.org/packages/FluentNHibernate/)`

Snippets/Snippets_5/NHibernate/NHibernateLoadMappingsFromFileSystem.cs
```diff
@@ -0,0 +1,51 @@
...
+    #endregion
+
+    #region NHibernateInitWithFluentNHibernate
+
+    static NHibernate.Cfg.Configuration BuildConfiguration(NHibernate.Cfg.Configuration nhConfiguration)
+    {
+        return FluentNHibernate.Cfg.Fluently.Configure(nhConfiguration)
+            .Mappings(cfg =>
+            {
+                cfg.FluentMappings.AddFromAssemblyOf<MySagaData>();
```
SimonCropp: so does the user have to do this for all sagas? as in they take full control over the NH config? if so the doco is unclear in that regard?

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,125 @@
...
+title: NHibernate Custom mapping with hbm.xml, attributes or fluent api
+summary: Create a custom mapping to change your NHibernate creates the database schema using different techniques.
+tags:
+ - NHibernate
+ - Persistence
+ - Configuration
+ - Mapping
+related:
+ - nservicebus/nhibernate/accessing-data
+ - nservicebus/nhibernate/index
```
SimonCropp: the index is redundant here

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,125 @@
+---
+title: NHibernate Custom mapping with hbm.xml, attributes or fluent api
+summary: Create a custom mapping to change your NHibernate creates the database schema using different techniques.
```
SimonCropp: > Create a custom mapping to change your NHibernate creates the database schema using different techniques.

That doesnt make sense

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,125 @@
...
+tags:
+ - NHibernate
+ - Persistence
+ - Configuration
```
SimonCropp: so commonly used as to be redundant

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,125 @@
...
+Custom mapping options:
+
+* Create a `*.hbm.xml` file for your saga data class
+* Use Fluent NHibernate (popular separate fluent api)
+* Use NHibernate.Mapping.Attributes
+* Loquacious Configuration (native fluent api)
+
+There are probably even other options but these are the most frequently used.
+
+See [NHibernate configuration](configuration) on how you can pass this configuration object to the NServiceBus NHibernate configuration.
```
SimonCropp: suspect this link should be `configuration.md`??

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,125 @@
...
+
+NOTE: When you create a custom mapping then you are responsible for mapping the primary key and unique indexes correctly.
```
SimonCropp: ok is the more info on this ? link or example?

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,125 @@
...
+
+
+NOTE: The `[Unique]` attribute will not be used to generate a unique index. You are responsible to add this to your custom mapping. This is especially important when using both optimistic concurrency and a transaction isolation level different from serializable and allow concurrent processing of messages. Not adding this can result in multiple saga entities be persisted as the second insert will not fail when inserting a duplicate saga key (different from the primary key column) when multiple message  are processed concurrently that target the same saga instance.
```
SimonCropp: @andreasohlund can u review this in the context of the V6 `[Unique]` changes

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,125 @@
...
+Using NHibernate mapping files is the original way to customize your mapping. It needs to be created and then either embedded as a resource in your assembly or available on the file system.
+
+Look at these pages to see how you can create these mappings:
+
+* http://nhibernate.info/doc/tutorials/first-nh-app/your-first-nhibernate-based-application.html
+
+
+### Mappings as embedded resources
+
+It is not required to create your own NHibernate configuration object and pass it to the NServiceBus NHibernate configuration.
```
SimonCropp: this statement confuses me. why are we highlighting something that doesnt need to be done?

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,125 @@
...
+
+
+### Read mappings from file system instead of embedded resources
+
+If you do not want to embed the `*.hbm.xml` files then you can also load mappings from the file system. Create a custom NHibernate configuration object and use the following example to add mappings from the file system.
+
+<!-- import NHibernateLoadMappingsFromFileSystem -->
```
SimonCropp: isnt loading mappings from resources or files an implementation detail in NH. why do we need to explain the destitution instead of just linking to the NH doco? perhaps we should just talk about whatever the default approach is for NH

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,125 @@
...
+
+## Use Fluent NHibernate
+
+Fluent NHibernate gives you a type-safe mapping approach where the mapping is specified as c# code and not as xml but still seperate from the classes. The benefit is that you get compile time feedback when a mapping is not valid any more when you have breaking changes in your mapping but also that your classes and mappings are not part of the same .net type.
```
SimonCropp: >  but still seperate from the classes

i think this is redundant.

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,125 @@
...
+1. Install `FluentNHibernate` package via Nuget.
+2. Create a custom NHibernate configuration
+a. via FluentNHibernate
+b. or by creating a new Configuration instance and pass it to FluentNHibernate
+3. Pass it to the NServiceBus NHibernate configuration.
+
+
+Example of a possible implementation:
+
+<!-- import NHibernateInitWithFluentNHibernate
```
SimonCropp: this import is missing the trailing `-->`

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,125 @@
...
+
+Its best to read the [Getting started](https://github.com/jagregory/fluent-nhibernate/wiki/Getting-started) article from the FluentNHibernate project to see how you create mappings.
```
SimonCropp: isnt this redundant since we proceed to list the same link in the next few lines

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,125 @@
...
+
+* http://www.fluentnhibernate.org
+* http://www.nuget.org/packages/FluentNHibernate/
+* https://github.com/jagregory/fluent-nhibernate
+* https://github.com/jagregory/fluent-nhibernate/wiki/Getting-started
+
+
+## Use NHibernate.Mapping.Attributes
+
+With NHibernate.Attributes you can decorate your saga data classes. This keeps your classes, mapping and schema data very close. Your saga types have a dependency on the NHibernate.Attributes assembly.
```
SimonCropp: the title is `NHibernate.Mapping.Attributes` but on this line it says  `NHibernate.Attributes`. which one is it?

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,125 @@
...
+* http://nhibernate.info/doc/nhibernate-reference/mapping-attributes.html
+* https://github.com/nhibernate/NHibernate.Mapping.Attributes
+* http://www.nuget.org/packages/NHibernate.Mapping.Attributes/
+
+
+## Use the Loquacious API
+
+The Loquacious api is NHibernate its native fluent mapping ability. You declare your mapping in code just like FluentNHibernate but using a different syntax. It can help you create a type-safe configuration but you can also create your custom mappings. The benefit is that this api is already available via the NHibernate package so requires no additional downloads.
+
+To use it:
```
SimonCropp: the 3 steps below seem so gloss over a fair amount of complexity

SimonCropp:
@ramonsmits while i think this is valuable content i think it is reaching too far into the realm of NH implementation to be good doco pages. What do you think about moving this to a samples where we can have a project showing an example of each implementation?

Snippets/Snippets_5/NHibernate/NHibernateLoadMappingsFromFileSystem.cs
ramonsmits: That is a good remark. I know that when using `*.hmb.xml` that it is not required to customize all sagas. For the other, don't know. I don't know what NServiceBus does with the NHibernate configuration. Needs some investigation.

ramonsmits:
@SimonCropp I agree that this needs to be migrated to a sample project where these mappings can be demonstrated as then customers have working samples to use as a start.

I think I have processed all comments that are currently in this PR and also merged master into it as it was out of sync.

Will move this doco page to a readme.md for a nhibernate custom mapping sample to have a good introduction on what that sample project is trying to explain.

ramonsmits:
@mauroservienti I'll try do make the sample project this evening so we can send it to the customer that is having a mapping issue.

Snippets/Snippets_5/NHibernate/NHibernateLoadMappingsFromFileSystem.cs
ramonsmits: Will test this in the sample project and add the result in the readme.

nservicebus/nhibernate/custom-mappings.md
ramonsmits: @SimonCropp Added a note that embedding xml mappings is the default. I don't understand 'destitution'.

nservicebus/nhibernate/custom-mappings.md
ramonsmits: @SimonCropp Would it make sense to make a comparison table in the top where a few characteristics are compared? I added this keeps classes clean compared to the attributes.

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,126 @@
...
+* Use Fluent NHibernate (popular separate fluent api)
+* Use NHibernate.Mapping.Attributes
+* Loquacious Configuration (native fluent api)
+
+There are probably even other options but these are the most frequently used.
+
+NOTE: **Mention here if it is required to add custom mappings for all sagas**
+
+
+See [NHibernate configuration](configuration.md) on how you can pass this configuration object to the NServiceBus NHibernate configuration.
```
mauroservienti: `this`? what configuration object?

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,126 @@
...
+
+NOTE: When you create a custom mapping then you are responsible for mapping the primary key and unique indexes correctly.
+
+
+NOTE: When using custom mappings you are responsible for creating unique indexes for all mapped saga properties. This is especially important when using both optimistic concurrency and a transaction isolation level different from serializable and allow concurrent processing of messages. Not adding this can result in multiple saga entities be persisted as the second insert will not fail when inserting a duplicate saga key (different from the primary key column) when multiple message  are processed concurrently that target the same saga instance.
```
mauroservienti: does not read very well, at least given my English knowledge

mauroservienti: Both NOTE(s) should be merged into a paragraph, IMO.

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,126 @@
...
+
+
+## Custom .hbm.xml mapping
+
+Using NHibernate mapping files is the original way to customize your mapping. It needs to be created and then either embedded as a resource in your assembly or available on the file system.
```
mauroservienti: there is a mismatch, Starts with a plural subject and continues with a singular one.

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,126 @@
...
+
+NOTE: Using xml mappings is the default approach for NHibernate.
```
mauroservienti: Already said at the beginning:

> Using NHibernate mapping files is the original way to customize your mapping.

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,126 @@
...
+
+Look at these pages to see how you can create these mappings:
```
mauroservienti: What about?

`For more information refer to the following documents on the official NHibernate documentation`

nservicebus/nhibernate/custom-mappings.md
mauroservienti: I'd say no, we should not include any detail from the NH side of things unless strictly necessary, otherwise we end up maintaining NH documentation on not ours.

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,126 @@
...
+* [Fluent NHibernate - Nuget package](http://www.nuget.org/packages/FluentNHibernate/)
+* [Fluent NHibernate - GitHub Repository](https://github.com/jagregory/fluent-nhibernate)
+* [Fluent NHibernate - Getting Started](https://github.com/jagregory/fluent-nhibernate/wiki/Getting-started)
+
+
+## Use NHibernate.Mapping.Attributes
+
+With NHibernate.Mapping.Attributes you can decorate your saga data classes. This keeps your classes, mapping and schema data very close. Your saga types have a dependency on the NHibernate.Mapping.Attributes assembly.
+
+How NHibernate.Mapping.Attributes works is that it needs to know which types it needs to scan, then it create a regular NHibernate mapping configuration that gets passed to your NHibernate configuration.
```
mauroservienti: `NHibernate.Mapping.Attributes` need to know which types to scan so to create a regular NHibernate mapping configuration that gets passed to your NHibernate configuration.

nservicebus/nhibernate/custom-mappings.md
```diff
@@ -0,0 +1,126 @@
...
+* [NHibernate.Mapping.Attributes - Documentation](http://nhibernate.info/doc/nhibernate-reference/mapping-attributes.html)
+* [NHibernate.Mapping.Attributes - GitHub repository](https://github.com/nhibernate/NHibernate.Mapping.Attributes)
+* [NHibernate.Mapping.Attributes - NuGet package](http://www.nuget.org/packages/NHibernate.Mapping.Attributes/)
+
+
+## Use the Loquacious API
+
+The Loquacious api is NHibernate its native fluent mapping ability. You declare your mapping in code just like FluentNHibernate but using a different syntax. It can help you create a type-safe configuration but you can also create your custom mappings. The benefit is that this api is already available via the NHibernate package so requires no additional downloads.
+
+Creating mappings with this API is very similar to its native xml mapping. Meaning that for each entity a mappings needs to be created.
```
mauroservienti: Creating mappings with the Loquacious API is similar to the native `xml` mappings: for each entity a mapping needs to be created.

nservicebus/nhibernate/custom-mappings.md
ramonsmits: this -> a

nservicebus/nhibernate/custom-mappings.md
ramonsmits: Too much in one sentence.

Issue #579, opened 2015-03-02
Document usages of XElement or XDocument options in xml serialization


johnsimons:
Mostly already done by @danielmarbach in http://www.planetgeek.ch/2013/01/20/nservicebus-embedding-raw-xml-into-messages/

danielmarbach:
@johnsimons should I bring part of this blogpost over?

johnsimons:
:+1:

mauroservienti:
+1

On Tue, Mar 24, 2015 at 9:36 AM, John Simons <notifications@github.com>
wrote:

> [image: :+1:]
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Particular/docs.particular.net/issues/579#issuecomment-85399177>
> .
>



-- 
Mauro Servienti | Solution Architect | Particular Software

danielmarbach:
As a sample or documentation on the XmlSerializer?

mauroservienti:
an integration sample would be great, we can add guidance later if
required. Thoughts?

On Tue, Mar 24, 2015 at 9:44 AM, Daniel Marbach <notifications@github.com>
wrote:

> As a sample or documentation on the XmlSerializer?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Particular/docs.particular.net/issues/579#issuecomment-85404091>
> .
>



-- 
Mauro Servienti | Solution Architect | Particular Software

danielmarbach:
Done as part of #644

PR #644, opened 2015-05-29
[WIP] Raw Xml Doco and Snippets

danielmarbach:
#579

I added the sample to the integration subfolder which I newly created. Needed to add the Namespace to the messages. If you remove it then the XmlSerializer fails to reflect the namespace.

Doco is missing.


SimonCropp:
This is one of the messaging approaches that will be going in the "dont do this" category. Why would we want to add a sample that enables this approach?

danielmarbach:
@mauroservienti @ramonsmits @jdrat2000 Any thoughts on this? before I remove the sample. @SimonCropp pointed out that it shouldn't be a sample but documentation with snippets only.

SimonCropp:
@danielmarbach looks good

SimonCropp:
@danielmarbach thanks

PR #647, opened 2015-06-02
Logging usage snippets

ramonsmits:
Improvements on logging usage. Included snippets in snippet projects, added a note about log levels and passing a type to the LogManager which is more common.

### DIFF
--- a/Snippets/Snippets_3/Logging/UsingLogging.cs
+++ b/Snippets/Snippets_3/Logging/UsingLogging.cs
@@ -8,6 +8,6 @@ public class ClassUsingLogging
         //your code
         Logger.Debug("Something interesting happened.");
     }
-    static ILog Logger = LogManager.GetLogger("Name");
+    readonly static ILog Logger = LogManager.GetLogger(typeof(ClassUsingLogging)); // Can also use a string
 }
 #endregion
--- a/Snippets/Snippets_3/Logging/UsingLogging.cs
+++ b/Snippets/Snippets_3/Logging/UsingLogging.cs
@@ -8,6 +8,6 @@ public class ClassUsingLogging
         //your code
         Logger.Debug("Something interesting happened.");
     }
-    readonly static ILog Logger = LogManager.GetLogger(typeof(ClassUsingLogging)); // Can also use a string
+    static ILog Logger = LogManager.GetLogger(typeof(ClassUsingLogging)); // Can also use a string
 }
 #endregion
--- a/nservicebus/logging/usage.md
+++ b/nservicebus/logging/usage.md
@@ -12,6 +12,6 @@ Writing to logging from your code is straightforward. Set up a single static fie
 <!-- import UsingLogging -->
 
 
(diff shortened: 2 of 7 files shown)


Snippets/Snippets_3/Logging/UsingLogging.cs
```diff
@@ -8,6 +8,6 @@ public void SomeMethod()
         //your code
         Logger.Debug("Something interesting happened.");
     }
-    static ILog Logger = LogManager.GetLogger("Name");
+    readonly static ILog Logger = LogManager.GetLogger(typeof(ClassUsingLogging)); // Can also use a string
```
SimonCropp: readonly is a redundant noise keyword

nservicebus/logging/usage.md
```diff
@@ -10,6 +10,8 @@ redirects:
 Writing to logging from your code is straightforward. Set up a single static field to a `ILog` in your classes, and then use it in all your methods, like this:
 
 <!-- import UsingLogging -->
- 
 
-Note: Since `LogManager.GetLogger("Name");` is an expensive call it is important that the field is static so that the call only happens once per class.
+
+NOTE: Make sure that the log level is set to DEBUG when calling `.Debug(..)` or this log statement will not be added to the log file. See [Change settings via configuration](logging#changing-settings-via-configuration).
```
SimonCropp: this will be a broken link since it does not qualify the file extension with a .md

nservicebus/logging/usage.md
```diff
@@ -10,6 +10,8 @@ redirects:
...
+
+NOTE: Since `LogManager.GetLogger(..);` is an expensive call it is important that the field is `readonly static` so that the call only happens once per class and have the best possible performance.
```
SimonCropp: using `readonly` is only valuable if you think a class is so complex that you will forget that a certain variable should not be modified. there is no performance improvement in using it. so dont bother adding it

Snippets/Snippets_3/Logging/UsingLogging.cs
ramonsmits: Read-only makes sure that the the logger is jitted in the method or is todays .net compiler smart enough to see that it is only assigned in the static contructor? I also like its intent otherwise static would be noisy too right?

SimonCropp: It does not nor has it ever added any performance improvement. http://stackoverflow.com/a/277117/53158

it does make the variable pointer immutable. but those benefits are mostly void in the context of a private variable. And even the mutability argument is difficult since people often make things readonly that are mutable like dictionaries or lists. The only approach to immutability is either going full immutable types (pulling in nuget dependencies which we cant do) or moving over to F# which would be a full re-write.

reaonly is only a reminder to someone that you should not change the pointer to a variable other than in the constructor. but tbh if you find yourself forgetting the desired usage of a static variable then the class is either too large or too complicated, or both

static has actual perf benefits since it means it is only instantiate once per type.

SimonCropp:
merged manually

PR #650, opened 2015-06-05
Clarify how ServicePulse looks for its license in a couple places.

DavidBoike:
Note that I'm making the recommendation that using the registry is the preferred method of installing the license. That could be controversial I guess.

Related to Desk case 11165.

### DIFF
--- a/nservicebus/licensing/license-management.md
+++ b/nservicebus/licensing/license-management.md
@@ -13,6 +13,8 @@ There are several ways to make sure that your NServiceBus endpoints pick up and 
 
 ## Using the Registry
 
+Using the registry to store your license information is preferred as all platform tools can access this information easily. This includes NServiceBus endpoints, ServiceControl, ServiceInsight, and ServiceMatrix. (ServicePulse determines licensing status by querying the ServiceControl API.) Using the registry ensures that all the platform tools can access the license status without requiring additional complexity on every deployment.
+
 ### NServiceBus 4.5 and above
 
 The standalone NServiceBus PowerShell V5.0 includes a commandlet for importing the Platform License into the `HKEY_LOCAL_MACHINE` registry. See [Managing Using PowerShell](/nservicebus/operations/management-using-powershell.md) for more details and installation instructions for the PowerShell Module.
--- a/nservicebus/licensing/license-management.md
+++ b/nservicebus/licensing/license-management.md
@@ -11,9 +11,11 @@ redirects:
 
 There are several ways to make sure that your NServiceBus endpoints pick up and use your license. 
 
+
 ## Using the Registry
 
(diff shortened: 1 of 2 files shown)


servicepulse/introduction-and-installing-servicepulse.md
```diff
@@ -50,4 +50,10 @@ For details on ServiceControl and ServicePulse configuration options, see:
...
\ No newline at end of file
+NOTE: ServiceControl consumes messages from the Audit queue and stores it temporarily (by default, for 30 days) in its embedded database. You can set the message storage timespan by [setting automatic expiration for ServiceControl data](/servicecontrol/how-purge-expired-data.md).
+
+### ServicePulse license
+
+ServicePulse will check your current licensing status by querying the ServiceControl API, located by default at `http://localhost:33333/api`. Therefore, if ServicePulse indicates that the license is invalid or has expired, it is ServiceControl that you must update the license for, by following the instructions at [How to install your NServiceBus license file](/nservicebus/licensing/license-management.md).
+
+Note that because of browser caching, it's possible for ServicePulse to continue reporting an expired license for a time after the license has been updated. This can be addressed by clearing your browser cache.
```
SimonCropp: isnt this a bug in SC? should it  be adding no cache headers to the response?

gbiellem: AFAIK the license caching issue was fixed in 1.3 so I assume that note could be removed

gbiellem: @DavidBoike  the SC installer we're building provides a UI to import license and stores it in the registry so the recommendation is a good one as far as I'm concerned

DavidBoike: I added that because Indu said she'd been burned by it. How can we confirm it's fixed? I did a quick issue + code search and couldn't find anything. It would be better to not include this as it would say "Hey we realize there's an error but we're not fixing it."

SimonCropp: not sure who fixed it. perhaps @johnsimons ?

gbiellem: No it was me .  It's was fixed here - https://github.com/Particular/ServiceControl/commit/363ae6890a3664648a0a0b68bf6688655e020911
We had caching applied  to the root module and that's what reports back the license.

DavidBoike:
Removed the chunk about caching since @gbiellem fixed it.

Good to go here?

nservicebus/licensing/license-management.md
```diff
@@ -13,6 +13,8 @@ There are several ways to make sure that your NServiceBus endpoints pick up and
 
 ## Using the Registry
 
+Using the registry to store your license information is preferred as all platform tools can access this information easily. This includes NServiceBus endpoints, ServiceControl, ServiceInsight, and ServiceMatrix. (ServicePulse determines licensing status by querying the ServiceControl API.) Using the registry ensures that all the platform tools can access the license status without requiring additional complexity on every deployment.
```
SimonCropp: Since SC, SM and Endpoints are most likely on different machines adding the registry to the mix actually makes it more complex. The least deployment steps for enspoints is to ship the license on the filesystem

PR #652, opened 2015-06-08
changes for the removal of the multiple RabbitMQ hosts anti-feature in the 3.0 transport

adamralph:


nservicebus/rabbitmq/cluster.md
```diff
@@ -0,0 +1,10 @@
...
+
+If each of your clients is deployed to a machine which is running at least one RabbitMQ host from the cluster, another approach you may consider is to configure your client to only connect to the local RabbitMQ instance, i.e. using `localhost` as the hostname. Note that the RabbitMQ hosts must still be clustered, otherwise each machine will only have access to the messages which originated locally.
+
+In version 2.x of the Rabbit MQ transport, we provided the facility to specify multiple hosts in the connection string. We removed this facility in version 3.0 for the reasons stated in the guide. We also found that this feature a) encouraged a misconception that the RabbitMQ transport was providing load balancing by sending messages to each host in a round-robin fashion and b) led to the belief that the RabbitMQ hosts themselves did not need to be clustered.
```
andreasohlund: I know that MassTransit is calling the underlying broker a "host", but for us a host is the thing that host a endpoint, win service, website, console app what not. So 

>b) led to the belief that the RabbitMQ hosts themselves did not need to be clustered.
Should be b) led to the belief that the RabbitMQ brokers themselves did not need to be clustered?

nservicebus/rabbitmq/connection-strings.md
```diff
@@ -29,15 +29,3 @@ For remote host provide username and password because remote hosts don't accept
                           password=mypassowrd"/>
```
andreasohlund: Nitpick: mypassword :)

adamralph: lol - when I saw the email notification I thought that perhaps *your* password had been commited :stuck_out_tongue_winking_eye: 

Fixed.

adamralph:
@andreasohlund I made the changes in reaction to your comments.

Good spot regarding host/broker. That prompted me to have another look through https://www.rabbitmq.com/clustering.html to see how they use these terms. It turns out that

> broker is a *logical* grouping of one or several Erlang nodes

With that in mind I rehashed the wording across the whole of the new page.

Can you take a look and see what you think?

nservicebus/rabbitmq/cluster.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Using a RabbitMQ cluster 
+summary: Guidelines for using RabbitMQ in a cluster
+---
+
+The [RabbitMQ Clustering Guide](https://www.rabbitmq.com/clustering.html) recommends that you do not bake in the hostnames or IP addresses of the machines hosting nodes in a RabbitMQ cluster in your client applications and instead use an approach such DNS or load balancing. We make the same recommendation when using the NServiceBus RabbitMQ transport. For more details see the guide.
+
+If each of your clients is deployed to a machine which is running at least one RabbitMQ node in the cluster, another approach you may consider is to configure your client to only connect to the local node, e.g. using `localhost` as the hostname. Note that the RabbitMQ nodes must still be clustered, otherwise clients on each machine will only have access to the messages which originated locally.
+
```
andreasohlund: >Note that the RabbitMQ nodes must still be clustered

Should we mention the Federation/Shovel plugins here?

andreasohlund:
Looks good, +1 to merge

nservicebus/rabbitmq/cluster.md
adamralph: Federation and shovel are very different to clustering so they are out of scope for this page.

We *could* have a separate page for RabbitMQ distribution in general (clustering, federation and shovel) but I think this is probably out of scope for our docs. The important thing from a client perspective is connection to a logical broker which may or may not be clustered. Federation and shovel distribution are implemented at a lower level (exchange and queue) and are specifically configured server side, typically for distribution over a WAN, so I don't think this is something that will ever be a client concern. However, I've no direct experience of either, so please correct me if I'm wrong.

andreasohlund: Ok, make sense

PR #655, opened 2015-06-11
Raven custom saga finder

danielmarbach:
* Simplified to only have one executable
* Added SagaNotFoundHandler in case people are trying this against their own RavenDB instance and forget to install the server side of the UniqueConstraints bundle
* Updated doco accordingly


danielmarbach:

danielmarbach:
@SimonCropp good changes. merge?

SimonCropp:
was wondering if we are spinning up an embedded store why not just pass that into the NSB.RavenDB. no need to go via http???

danielmarbach:
I asked myself that as well. But if we do it like it is now you better see in the sample how a custom store needs to register the client listener

SimonCropp:
i think here we are trying to illustrate custom saga finding. so the least moving pieces in all other areas is a win

danielmarbach:
Fair. I’ll change



*From:* Simon Cropp [mailto:notifications@github.com]
*Sent:* Donnerstag, 11. Juni 2015 14:09
*To:* Particular/docs.particular.net
*Cc:* Daniel Marbach
*Subject:* Re: [docs.particular.net] Raven custom saga finder (#655)



i think here we are trying to illustrate custom saga finding. so the least
moving pieces in all other areas is a win

—
Reply to this email directly or view it on GitHub
<https://github.com/Particular/docs.particular.net/pull/655#issuecomment-111109650>
.

danielmarbach:
@SimonCropp pushed the change. I slightly changed the doco as well and moved the region to only include the necessary stuff in the snippet

SimonCropp:
looks good. merging

PR #675, opened 2015-06-17
Extended doco on self hosting.

ramonsmits:
Doco on self-hosting extended with information about configuring, initializing, creating, starting, stopping and disposing of the bus.

### DIFF
--- a/nservicebus/hosting/ibus-instance.md
+++ b/nservicebus/hosting/ibus-instance.md
@@ -1,0 +1,113 @@
+---
+title: Managing IBus instance
+summary: Fluent configuration API to get transnational one-way messaging
+tags:
+- Hosting
+- IOC
+- Dependency Injection
+- Service Locator
+---
+
+
+When you have created a `IBus` instances then you must register it somewhere so that your application can use it to send messages.
+
+There are various ways in managing and storing the `IBus` instance.
+
+* Global static
+* Using dependency injection
+
+The first option creates tight coupling the last loose coupling.
+
+
+
+## Global static
+
+After you have created the bus you assign the object to a global static variable so that the bus can be used in other locations
+
+
+Initializing:
+
+```
+public class Global
+{
+    public static IBus Bus { get; set; }
+}
+
+Global.Bus = NServiceBus.Create(busConfiguration).Start();
+```
+
+Using it:
+
+```
+Global.Bus.Send(new MyCommand());
+```
+
+Cleanup:
+
+```
+Global.Bus.Dispose();
+```
+
+This is very simple to use but creates a lot of tight coupling which makes unit testig pretty hard.
+
+## Dependency injection container
+
+Here is the description of the dependency injection pattern:
+
+>In software engineering, dependency injection is a software design pattern that implements inversion of control for software libraries. Caller delegates to an external framework the control flow of discovering and importing a service or software module specified or "injected" by the caller.
+
+Source: https://nl.wikipedia.org/wiki/Dependency_injection
+
+
+### Registering
+
+When you have created the `IBus` instance then you need to register it in a dependency injection container so that it knows what and how it can inject it into objects that require an `IBus` instance.
+
+### Single instance
+
+The following example uses Autofac but looks very similar with other frameworks
+
+
+Registering bus instance:
+```
+ContainerBuilder builder = new ContainerBuilder();
+builder.RegisterInstance(bus);
+IContainer container = builder.Build();
+```
+
+This will make sure that when a object is created via the container and it is dependent on `IBus` that this instance will be injected.
+
+NServiceBus will still use its own container to resolve its own objects. NServiceBus can also use an external container.
+
+See: [Containers](/nservicebus/containers/index.md)
+
+You need to use this when you want types that are registered in the external controller to be injected in to objects created by NServiceBus like message handlers and message mutators.
+
+
+
+### Disposing
+
+When the application done it only needs to dispose the container. The container should take care of disposing all objects in the correct order.
+
+There should be no need in calling `IBus.Dispose` when you have registered the bus instance in the container.
+
+NOTE: Some containers will not dispose registered instances unless explicitly configured to do so.
+
+## Custom container
+
+In t
+
+Dependency Injection is used internally by NServiceBus and is referred to as the 'object builder'. Sometimes yo
+
+
+
+Samples 
+
+http://docs.particular.net/samples/web/asp-mvc-injecting-bus/
+http://docs.particular.net/nservicebus/containers/
+http://docs.particular.net/samples/containers/
+
+
+
+
+
--- a/nservicebus/hosting/self-hosting.md
+++ b/nservicebus/hosting/self-hosting.md
@@ -11,13 +11,13 @@ redirects:
  - nservicebus/hosting-nservicebus-in-your-own-process
 ---
 
-Lighter-weight than BizTalk and more powerful than WCF, NServiceBus comes with its own host process and allows you to host it in your own process.
+NServiceBus comes with its own host process but you can host it in your own process.
 
 Requiring just a reference to the NServiceBus NuGet package the configuration API can get you up and running with transactional one-way messaging in a snap.
 
-## Configuring the bus
+## Initializing and configuring the bus
 
-In the `ApplicationStart` method of your Global.asax file in a web application, or in the `Main` method of your `Program` file for console or Windows Forms application, include the following initialization code:
+Initializing the bus requires a few lines of code depending on the features that you want to have enabled.
 
 <!-- import MinimumConfiguration -->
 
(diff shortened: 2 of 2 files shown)


nservicebus/hosting/ibus-instance.md
```diff
@@ -0,0 +1,113 @@
...
+
+Initializing:
+
+```
+public class Global
+{
+    public static IBus Bus { get; set; }
+}
+
+Global.Bus = NServiceBus.Create(busConfiguration).Start();
```
SimonCropp: please use snippets so this type of code can be shown for multiple versions

nservicebus/hosting/ibus-instance.md
```diff
@@ -0,0 +1,113 @@
...
+
+In t
+
+Dependency Injection is used internally by NServiceBus and is referred to as the 'object builder'. Sometimes yo
+
+
+
+Samples 
+
+http://docs.particular.net/samples/web/asp-mvc-injecting-bus/
```
SimonCropp: these are invalid links  https://github.com/Particular/docs.particular.net#linking

nservicebus/hosting/ibus-instance.md
```diff
@@ -0,0 +1,113 @@
+---
+title: Managing IBus instance
+summary: Fluent configuration API to get transnational one-way messaging
```
SimonCropp: i think this summary is wrong

nservicebus/hosting/ibus-instance.md
```diff
@@ -0,0 +1,113 @@
...
+
+
+
+### Disposing
+
+When the application done it only needs to dispose the container. The container should take care of disposing all objects in the correct order.
+
+There should be no need in calling `IBus.Dispose` when you have registered the bus instance in the container.
+
+NOTE: Some containers will not dispose registered instances unless explicitly configured to do so.
```
SimonCropp: `Some containers` is not really explicit enough

nservicebus/hosting/ibus-instance.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Custom container
+
+In t
```
SimonCropp: is this missing something?

nservicebus/hosting/ibus-instance.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Global static
+
+After you have created the bus you assign the object to a global static variable so that the bus can be used in other locations
+
+
+Initializing:
+
+```
+public class Global
```
SimonCropp: this is kind of out of context. clearly this is copied from a web application. but anywhere else why would you name something as generic as global.

nservicebus/hosting/ibus-instance.md
```diff
@@ -0,0 +1,113 @@
...
+- Dependency Injection
+- Service Locator
+---
+
+
+When you have created a `IBus` instances then you must register it somewhere so that your application can use it to send messages.
+
+There are various ways in managing and storing the `IBus` instance.
+
+* Global static
```
SimonCropp: this is an anti pattern and is only used in some samples for the sake of simplicity. i dont think we should be documenting this as a viable option

nservicebus/hosting/ibus-instance.md
```diff
@@ -0,0 +1,113 @@
+---
```
SimonCropp: why do we need a new page? adding this doco here http://docs.particular.net/nservicebus/containers/injecting-ibus should suffice

nservicebus/hosting/ibus-instance.md
```diff
@@ -0,0 +1,113 @@
...
+
+This is very simple to use but creates a lot of tight coupling which makes unit testig pretty hard.
+
+## Dependency injection container
+
+Here is the description of the dependency injection pattern:
+
+>In software engineering, dependency injection is a software design pattern that implements inversion of control for software libraries. Caller delegates to an external framework the control flow of discovering and importing a service or software module specified or "injected" by the caller.
+
+Source: https://nl.wikipedia.org/wiki/Dependency_injection
```
SimonCropp: nl?

nservicebus/hosting/ibus-instance.md
```diff
@@ -0,0 +1,113 @@
...
+
+### Single instance
+
+The following example uses Autofac but looks very similar with other frameworks
+
+
+Registering bus instance:
+```
+ContainerBuilder builder = new ContainerBuilder();
+builder.RegisterInstance(bus);
```
SimonCropp: this should not be required?? doesnt it happen as part of the call to `busConfiguration.UseContainer`

nservicebus/hosting/ibus-instance.md
```diff
@@ -0,0 +1,113 @@
...
+
+See: [Containers](/nservicebus/containers/index.md)
+
+You need to use this when you want types that are registered in the external controller to be injected in to objects created by NServiceBus like message handlers and message mutators.
+
+
+
+### Disposing
+
+When the application done it only needs to dispose the container. The container should take care of disposing all objects in the correct order.
```
SimonCropp: > When the application done

missing some words?

SimonCropp: > The container should 

"should" is ambiguous. in what cases does it not?

nservicebus/hosting/ibus-instance.md
```diff
@@ -0,0 +1,113 @@
...
+
+There should be no need in calling `IBus.Dispose` when you have registered the bus instance in the container.
```
SimonCropp: remove "should"

nservicebus/hosting/ibus-instance.md
```diff
@@ -0,0 +1,113 @@
...
+
+NOTE: Some containers will not dispose registered instances unless explicitly configured to do so.
+
+## Custom container
+
+In t
+
+Dependency Injection is used internally by NServiceBus and is referred to as the 'object builder'. Sometimes yo
```
SimonCropp: > Sometimes yo

?

nservicebus/hosting/ibus-instance.md
pashute: I agree with Simon. Add the following at the top: 

In typical NServiceBus projects the IBus is obtained by the automatic configured injection as explained below. 
However in some applications one may wish to use a different approach :
* Typically for web applications, the project can be configured with an IBus for one-way messaging <link needed to explain this> 
** Ad hoc IBus creation - for a quick sample < code: using (...){}  >, 
** Manual configuration of the IBus injection in a typical web app 
*In SignalR web applications configure the IBus injection manually    

***Automatically Configured Injection***
The following code is needed for acquiring an injected IBus in your handler classes. 
Note: The configuration code is automatically added, if you use the Nuget for the NServiceBus Host package. 

< here goes all the current injecting-bus documentation >

** Manually setting IBus injection **
< here goes you new doc >

nservicebus/hosting/ibus-instance.md
pashute: sp: transactive (not transnational)
summary: Configuring the application for acquiring the IBus instance

nservicebus/hosting/ibus-instance.md
pashute: The Global static way of doing this should be moved to the end. An example is needed of when you would recommend its usage.

nservicebus/hosting/ibus-instance.md
pashute: * replace:  This is very simple...
* with: We do not recommend using the static IBus definition, as it introduces tight coupling, and creates difficulties when unit testing. 

* remove the Dependency inection container header and its section. This is not needed. Note the Dutch wikipedia link instead of the English one...

nservicebus/hosting/ibus-instance.md
pashute: * replace: When you have ... then you need to...
* with: The 'IBus' instance registration is automatically set by calling busConfiguration.UseContainer. UseContainer() is invoked automatically if the NServiceBus Nuget package was added to the project. You may wish to register the IBus manually, for example < insert example of why this would be so >  in which case you would use code similar to the following (sample code for Autofac but similar for other injection containers):

 < your sample code >

nservicebus/hosting/ibus-instance.md
pashute: corrections: 

this will ensure that 'IBus' instance will be injected in an object dependent on 'IBus', which is created via the container
+
+NServiceBus will still use its own container to resolve its own objects. 

+ NServiceBus can also use an external container: See: [Containers](/nservicebus/containers/index.md)
+
//removed the lines explaining when to use, will move those up to the beginning of the section
++
+### Disposing
+
+When the application done it only needs to dispose the container. The container should take care of disposing all objects in the correct order.

nservicebus/hosting/ibus-instance.md
pashute: replace: there should be ... in calling
with: There is no need to call 'IBus.Dispose' after registering the bus instance in the container.

nservicebus/hosting/ibus-instance.md
pashute: I cannot edit the original so rewriting the whole thing as a single comment below

pashute:
I'm forking the original page (I cannot edit directly here) and adding the corrected code... I'll write when I'm done

SimonCropp:
this PR is now obsolete. Following pages have been updated 
http://docs.particular.net/nservicebus/hosting/
http://docs.particular.net/nservicebus/hosting/windows-service

Issue #171, opened 2014-02-11
Document perf counters


synhershko:
How to work with perf counters - setting up, explaining the differnt counters, what they can be used for

SeanFeldman:
done

SimonCropp:
@SeanFeldman the pr u reference only adds descriptions not how to set them up

SimonCropp:
done http://docs.particular.net/nservicebus/operations/performance-counters http://docs.particular.net/samples/performance-counters/

PR #678, opened 2015-06-19
Document all performance counters

SeanFeldman:
PR to address performance counters doco issue https://github.com/Particular/docs.particular.net/issues/171


SimonCropp:
From the discussion on slack i thought better guidance on how to install them was also going to be added

indualagarsamy:
@simoncropp - I merged it. As of now there's no dock on counters/ I can add that as an enhancement. I needed to send the info. I'll get to it in a day or two.

PR #688, opened 2015-06-25
Child containers are now supported by spring

danielmarbach:
I'm working on the release. I'll update on this PR as soon as it is released


danielmarbach:
@SimonCropp here we go, I read through the child container page again and think the explanation is good. Don't think we need to change it

PR #691, opened 2015-07-01
Update sample.md

ryanrdl:
I spent several hours and had to dig into the source to figure out it was doing a simple Type.GetType  and by including the namespace I was changing the type name.  This is not called out very clearly anywhere in the documentation or the sample.


SimonCropp:
@ryanrdl sorry for the delay. Your feedback is valid and i think i addressed it as part of this https://github.com/Particular/docs.particular.net/commit/730be12cca388d42c8eedc8ec998e468a252a21c

you can see the new rendered content here http://docs.particular.net/samples/rabbitmq/native-integration/#code-walk-through-message-serialization

PR #701, opened 2015-07-09
remove backwards compatibility note in auditing doc

timbussmann:
The documentation states that a warning recommends to use the `AuditConfig` section when using `ForwardReceivedMessagesTo`. Verified that this warning does not appear when using `ForwardReceivedMessagesTo` in NSB 5.2.3, therefore removed the note.


danielmarbach:
Verified. Merging

PR #708, opened 2015-07-17
Snippets for Request/Response API

danielmarbach:
In progress for new NServiceBus.Callback API


Snippets/Snippets_6/Callback/Usage.cs
```diff
@@ -0,0 +1,32 @@
...
+
+    public class Usage
+    {
+        public async Task Describe()
+        {
+            IBus bus = null;
+            var options = new SendOptions();
+
+            #region CallbackWithMessageAsResponse
+            var responseMessage = await bus.RequestWithTransientlyHandledResponseAsync<PlaceOrderResponse>(new PlaceOrder(), options);
```
SimonCropp: are we planning on suffix all async member in NSB V6 with `Async`?

danielmarbach: Good question. Might be overkill if most APIs become async

Am Freitag, 17. Juli 2015 schrieb Simon Cropp :

danielmarbach: Started an internal poll. I think we should comply to the "rules"

Quote from Steven Toub, PFX team

> If a public method is Task-returning and is asynchronous in nature (as opposed to a method that is known to always execute synchronously to completion but still returns a Task for some reason), it should have an “Async” suffix. That’s the guideline. The primary goal here with the naming is to make it very obvious to a consumer of the functionality that the method being invoked will likely not complete all of its work synchronously; it of course also helps with the case where functionality is exposed with both synchronous and asynchronous methods such that you need a name difference to distinguish them. How the method achieves its asynchronous implementation is immaterial to the naming: whether async/await is used to garner the compiler’s help, or whether types and methods from System.Threading.Tasks are used directly (e.g. TaskCompletionSource) doesn’t really matter, as that doesn’t affect the method’s signature as far as a consumer of the method is concerned.

> Of course, there are always exceptions to a guideline. The most notable one in the case of naming would be cases where an entire type’s raison d’etre is to provide async-focused functionality, in which case having Async on every method would be overkill, e.g. the methods on Task itself that produce other Tasks.

> As for void-returning asynchronous methods, it’s not desirable to have those in public surface area, since the caller has no good way of knowing when the asynchronous work has completed. If you must expose a void-returning asynchronous method publicly, though, you likely do want to have a name that conveys that asynchronous work is being initiated, and you could use the “Async” suffix here if it made sense. Given how rare this case should be, I’d argue it’s really a case-by-case kind of decision.

> I hope that helps, Steve

[Official naming convention:](https://msdn.microsoft.com/en-us/library/hh191443.aspx#BKMK_NamingConvention)

SimonCropp: @danielmarbach i think that advice is out of date since 

> it of course also helps with the case where functionality is exposed with both synchronous and asynchronous methods such that you need a name difference to distinguish them

Is now generally recognized as a antipattern

danielmarbach: @SimonCropp I know. I just wanted to add it here for public reference.

danielmarbach:
Not super happy with the [subtiltle containing NServiceBus 6](https://github.com/Particular/docs.particular.net/pull/708/files#diff-494493c47dc8017f8d9b35bac859c108R39) any ideas how to better structure it?

Is there something missing in the doco for the new Callback API?

@Particular/documentation please review // cc @johnsimons

johnsimons:
:+1: it looks good

Snippets/Snippets_6/Callback/Usage.cs
```diff
@@ -0,0 +1,46 @@
...
+            #endregion
+
+            #region CancelCallback
+            var cs = new CancellationTokenSource();
+            options.RegisterCancellationToken(cs.Token);
+            try
+            {
+                responseMessage = await bus.RequestWithTransientlyHandledResponseAsync<PlaceOrderResponse>(new PlaceOrder(), options);
+            }
+            catch (OperationCanceledException ex)
```
SimonCropp: empty catch. perhaps a inline comment to explain why or what the user should do?

nservicebus/messaging/handling-responses-on-the-client-side.md
```diff
@@ -36,6 +36,29 @@ To access the response message through callbacks, the following code can be used
 
 DANGER: If the server process returns multiple responses, NServiceBus cannot know which response message will be the last. To prevent memory leaks, the callback is invoked only for the first response. Callbacks won't survive a process restart (common scenarios are a crash or an IIS recycle) as they are held in memory, so they are less suitable for server-side development where fault-tolerance is required. In those cases, [sagas are preferred](/nservicebus/sagas/).
 
+## NServiceBus 6: Handling responses in the context of the request
```
SimonCropp: i would split this into functionality headings.

eg   "where to get callback functionality: in v5 an lower it it build in, in v6 get the nuget"

Snippets/Snippets_6/Callback/Usage.cs
danielmarbach: I was really on the fence how to express that an `OperationCancelledException` is raised. I decided to go with the empty catch. So maybe add a comment like

`// Exception that is raised when the CancellationTokenSource is canceled`

thoughts?

SimonCropp: yep the comment is good

danielmarbach:
@SimonCropp Pushed changes

danielmarbach:
@SimonCropp ping

danielmarbach:
Awesome job @SimonCropp !

Issue #713, opened 2015-07-27
change raven links to use search


SimonCropp:
bascally the raven docs url structure is no suitable for us to link to https://github.com/ravendb/docs/issues/112

current plan is to change all raven links to be of the search form eg http://ravendb.net/search?q=unique-constraints

PR #717, opened 2015-07-27
closes change raven links to use search #713

sfarmar:
a stab at issue #713


SimonCropp:
@sfarmar thanks. perfect

PR #718, opened 2015-07-28
ServiceControl MaxBodySizeToStore configuration instructions

petersgiles:
I added a snippet about how to configure Service Control for #296

### DIFF
--- a/servicecontrol/creating-config-file.md
+++ b/servicecontrol/creating-config-file.md
@@ -76,110 +76,110 @@ net stop particular.servicecontrol
 
 ### Configuration Options
 
-##### ServiceControl/LogPath (string)
+#### ServiceControl/LogPath (string)
 The path for the ServiceControl logs. 
 
 Default: `%LOCALAPPDATA%\Particular\ServiceControl\logs`
 ___
 
-##### ServiceControl/Port  (int)
+#### ServiceControl/Port  (int)
 
 The port to bind the embedded http server. 
 
 Default: `33333`.
 ___
-##### ServiceControl/Hostname  (string)
+#### ServiceControl/Hostname  (string)
 
 The hostname to bind the embedded http server to, modify if you want to bind to a specific hostname, eg. sc.mydomain.com. 
 
 Default: `localhost`
 ___
-##### ServiceControl/VirtualDirectory  (string)
+#### ServiceControl/VirtualDirectory  (string)
 
 The virtual directory to bind the embedded http server to, modify if you want to bind to a specific virtual directory.
  
 Default: `empty`
 ___
-##### ServiceControl/HeartbeatGracePeriod  (timespan)
+#### ServiceControl/HeartbeatGracePeriod  (timespan)
 
 The period that defines whether an endpoint is considered alive or not. 
 
 Default: `00:00:40` (40 secs)
 ___
-##### ServiceControl/MaximumMessageThroughputPerSecond (int)
+#### ServiceControl/MaximumMessageThroughputPerSecond (int)
 
 This setting was introduced in version 1.5. The setting controls the maximum throughput of messages ServiceControl will handle per second and is necessary to avoid overloading the underlying messages database. An apropriate limit ensures that the database can cope with number of insert operations. Otherwise the query performance would drop significantly and the message expiration process would stop working when under heavy insert load. Make sure to concudct thorough performance tests on your hardware before increasing this value.  
 
 Default: `350`. 
 ___
-##### ServiceControl/ForwardAuditMessages (bool `true`/`false`)
+#### ServiceControl/ForwardAuditMessages (bool `true`/`false`)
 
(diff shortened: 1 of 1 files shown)


servicecontrol/creating-config-file.md
```diff
@@ -142,4 +142,9 @@ net stop particular.servicecontrol
...
 * `ServiceBus/AuditLogQueue` (string)
 
-  The audit queue name to use for forwarding audit messages. This only works if `ServiceControl/ForwardAuditMessages` is true. Default `<AuditQueue>.log`
\ No newline at end of file
+  The audit queue name to use for forwarding audit messages. This only works if `ServiceControl/ForwardAuditMessages` is true. Default `<AuditQueue>.log`
+
+
+* `ServiceControl/MaxBodySizeToStore` (int)
+
+  By default we only store bodies of audit messages that are smaller than 100Kb. Increase this number to store messages with larger bodies.  Default `102400`
```
johnsimons: Should we mention that the size is in bytes?

SimonCropp: so reading this i would automatically set it to 99999999999 so i wont loose any data

SimonCropp: > we only

not "we". "ServiceControl"

SimonCropp: what kind of logging do we do when it is discarded

johnsimons: > o reading this i would automatically set it to 99999999999 so i wont loose any data

nothing wrong with that, the only issue is that u may hit the max int value

SimonCropp: so if there is nothing wrong with that why do we have a limit at all?

johnsimons: > so if there is nothing wrong with that why do we have a limit at all?

Good point, @petersgiles the same way you did in SI, we should explain here why such limit exists

SimonCropp: @johnsimons yep :)

SimonCropp: also does this work in all versions? if not some version info needs to be added

servicecontrol/creating-config-file.md
```diff
@@ -76,70 +76,110 @@ net stop particular.servicecontrol
 
 ### Configuration Options
 
-* `ServiceControl/LogPath` (string)
+##### ServiceControl/LogPath (string)
+The path for the ServiceControl logs. 
 
-  The path for the ServiceControl logs. Default: `%LOCALAPPDATA%\Particular\ServiceControl\logs`
+Default: `%LOCALAPPDATA%\Particular\ServiceControl\logs`
+___
```
SimonCropp: whats with the horizontal rules?

servicecontrol/creating-config-file.md
```diff
@@ -76,70 +76,110 @@ net stop particular.servicecontrol
 
 ### Configuration Options
 
-* `ServiceControl/LogPath` (string)
+##### ServiceControl/LogPath (string)
```
SimonCropp: why jump from H3 to H5?

servicecontrol/creating-config-file.md
```diff
@@ -76,70 +76,110 @@ net stop particular.servicecontrol
...
-* `ServiceBus/ErrorLogQueue` (string)
+The audit queue name to use for forwarding audit messages. This only works if `ServiceControl/ForwardAuditMessages` is true. 
 
-  The error queue name to use for forwarding error messages. Default `<ErrorQueue>.log`
+Default: `<AuditQueue>.log`
+___
+##### ServiceControl/MaxBodySizeToStore (int) 
 
-* `ServiceBus/AuditLogQueue` (string)
+Up until version _1.6_ ServiceControl only stores bodies of audit messages that are smaller than 100Kb by default. After version _1.6_ Increase this number to store messages with larger bodies. Messages that have a larger message body in bytes than MaxBodySizeToStore are not stored for audit. This is to ensure that the majority of our users enjoy the best level of performance. For users with special analysis needs, edit MaxBodySizeToStore in ServiceControl.exe.config to increase the size of storeable audit messages.
```
SimonCropp: we generally dont have any formatting for version numbers in other pages. so no need to make them italic

petersgiles: ok

servicecontrol/creating-config-file.md
petersgiles: looked better, easily fixed

servicecontrol/creating-config-file.md
petersgiles: @johnsimons and @SimonCropp suggested a table for this. It looked pretty ordinary in a table because I couldn't control the column width.
You said the look and feel was up to me when I asked if you guys had a particular aesthetic in mind. this is the result

johnsimons:
:+1: to merge

PR #720, opened 2015-07-30
Adding more info on configuring MSMQ permissions

hmemcpy:


hmemcpy:
Per https://github.com/Particular/FeatureDevelopment/issues/303

PR #723, opened 2015-07-30
Added a comment about why Azure Service Bus transport is hard wired t…

### LINKED ISSUE #722: Add a comment about NServiceBus.AzureServiceBus transport's isolation level
After a call with @yvesgoeleven and @danielmarbach we agreed that we need to document the reason why:

- NServiceBus.AzureServiceBus Default Isolation level is `Serializable`
- Why is is coded such that the users can not override the [setting](https://github.com/Particular/NServiceBus.AzureServiceBus/blob/support-6.3/src/Transport/AzureServiceBusTransport.cs#L31)

sfarmar:
…o Serializable isolation level

Closes #722

### DIFF
--- a/nservicebus/azure/azure-servicebus-transport.md
+++ b/nservicebus/azure/azure-servicebus-transport.md
@@ -76,6 +76,14 @@ NOTE: `QueueName` and `QueuePerInstance` are obsoleted. Instead, use bus configu
 
 Defaults are just starting values. You should always measure and test these values against your solution and adjust those accordingly.
 
+## Transactions
+
+Azure Service Bus only supports the `Serializable` isolation scope, therefore the transport's configuration is hard coded to `Serializable`  so users won't be able and override it.
+
+(https://github.com/Particular/NServiceBus.AzureServiceBus/blob/support-6.3/src/Transport/AzureServiceBusTransport.cs#L31)
+(NServiceBus's default isolation level is ReadCommit).
+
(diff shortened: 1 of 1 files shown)


nservicebus/azure/azure-servicebus-transport.md
```diff
@@ -76,6 +76,14 @@ NOTE: `QueueName` and `QueuePerInstance` are obsoleted. Instead, use bus configu
 
 Defaults are just starting values. You should always measure and test these values against your solution and adjust those accordingly.
 
+## Transactions
+
+Azure Service Bus only supports the `Serializable` isolation scope, therefore the transport's configuration is hard coded to `Serializable`  so users won't be able and override it.
+
+(https://github.com/Particular/NServiceBus.AzureServiceBus/blob/support-6.3/src/Transport/AzureServiceBusTransport.cs#L31)
```
SimonCropp: Avoid  linking to source code. This link is likely to become out of date as soon as we do another support branch

SimonCropp:
This pr doesnt actually explain why. It just says that it is? Is this missing something?

SeanFeldman:
Azure ServiceBus library used underneath requires any transactional scope to use `IsolationLevel.Serializable` and nothing else. If another `IsolationLevel` value is used, MSFT ASB library will throw `InvalidOperationException` exception stating that `Serializable` should be used. I have personally ran into this with SignalR (link: https://github.com/SignalR/SignalR/issues/2947)

PR #729, opened 2015-08-07
Headers for integration

weralabaj:
Relates to https://github.com/Particular/FeatureDevelopment/issues/194
@andreasohlund @gbiellem @SimonCropp @danielmarbach Could you please review and see if we missed anyting?

### DIFF
--- a/menu/menu.yaml
+++ b/menu/menu.yaml
@@ -24,7 +24,7 @@
         - "nservicebus/messaging/non-durable-messaging"
         - "nservicebus/messaging/handling-responses-on-the-client-side"
         - "nservicebus/messaging/message-headers"
-        - "nservicebus/messaging/message-headers-for-third-party-integrations"
+        - "nservicebus/messaging/third-party-integration"
         - "nservicebus/messaging/messages-as-interfaces"
         - "nservicebus/messaging/reply-to-a-message"
         - "nservicebus/messaging/send-a-message"
--- a/menu/menu.yaml
+++ b/menu/menu.yaml
@@ -24,6 +24,7 @@
         - "nservicebus/messaging/non-durable-messaging"
         - "nservicebus/messaging/handling-responses-on-the-client-side"
         - "nservicebus/messaging/message-headers"
+        - "nservicebus/messaging/message-headers-for-third-party-integrations"
         - "nservicebus/messaging/messages-as-interfaces"
         - "nservicebus/messaging/reply-to-a-message"
         - "nservicebus/messaging/send-a-message"
--- a/nservicebus/messaging/message-headers-for-third-party-integrations.md
+++ b/nservicebus/messaging/message-headers-for-third-party-integrations.md
@@ -10,27 +10,27 @@ related:
 - nservicebus/messaging/message-headers
 ---
 
-# Integrating with external systems? How to get more out of your monitoring?
+# How to integrate external systems with NServiceBus?
 
-When an NServiceBus endpoint receives messages from external systems such as BizTalk, TIBCO, etc directly, the message itself will not contain regular NServiceBus headers. Without these headers, ServiceControl cannot provide sufficient information to ensure effective monitoring of these endpoints. 
+When an NServiceBus endpoint receives messages from external systems such as BizTalk, TIBCO, etc directly, the message itself will not contain regular NServiceBus headers. Some headers are required by NServiceBus to handle those messages (i.e. enable deserialization of the message). Others are essential to provide effective monitoring and troubleshooting experience. 
 
-To ensure that your Ops team is able to monitor your endpoints and have valuable meta data from ServiceControl, it becomes important as to how you integrate these messages. 
-
-When NServiceBus endpoints send messages, it includes a certain set of headers such as the name of the endpoint sending the message, the machine name etc, which is quite useful when analyzing this information from an Ops point of view in ServicePulse or when debugging situations using ServiceInsight. Therefore to get the best use of your monitoring and debugging tools, please ensure that the following headers are being added to the actual message by the third party endpoints when sending a message. 
+When NServiceBus endpoints send messages, they include a certain set of headers such as the type of message, the name of the sending endpoint, the machine name etc. To get the best use of your monitoring and debugging tools, please ensure that the following headers are being added to the actual message by the third party endpoints when sending a message. 
 
 ### Mandatory Headers
 
+Required for deserializing the message:
+
 Key  | Value
 ------------- | -------------
-NServiceBus.EnclosedMessageTypes  | Fully Qualified Message type including the assembly name, For e.g. IntegrationSample.Messages.Commands.ProcessOrder, IntegrationSample.Messages
+NServiceBus.EnclosedMessageTypes  | Fully Qualified Message type including the assembly name, for e.g. IntegrationSample.Messages.Commands.ProcessOrder, IntegrationSample.Messages
 
 
 ### Additional Headers 
 
-Adding the following headers will make the diagrams in ServiceInsight much more meaningful and for Ops when these messages fail and show up in FailedMessages in ServicePulse.
+Enable better monitoring in ServicePulse and debugging in ServiceInsight by providing additional data:
 
 Key  | Value
 ------------- | -------------
 NServiceBus.ConversationId  | Valid Guid, useful to tie in the whole message flow to get a much accurate view in ServiceInsight, for example in message flow diagram
 NServiceBus.OriginatingEndpoint  | Name of the third party endpoint sending the message, for e.g. BizTalk.ProcessOrder
-NServiceBus.OriginatingMachine  | Server where the third party endpoint is located for e.g. BizTalkServer
+NServiceBus.OriginatingMachine  | Server where the third party endpoint is located, for e.g. BizTalkServer
--- a/nservicebus/messaging/message-headers-for-third-party-integrations.md
+++ b/nservicebus/messaging/message-headers-for-third-party-integrations.md
@@ -18,7 +18,7 @@ To get the best use of your monitoring and debugging tools, please ensure that t
 
 ### Required metadata
 
-In order to enable NServiceBus to deserialize a message coming from a third party system, the message must contain its **full type name**. It can either be provided in the message body by ensuring appropriate serialization or by adding an additional header.
+In order to enable NServiceBus to deserialize a message coming from a third party system, the message must contain its **full type name**. It can either be provided in the message body by ensuring appropriate serialization or by adding an additional header (in NServiceBus v3 and above).
 
 Header key  | Value
 ------------- | -------------
@@ -29,7 +29,7 @@ For more information refer to integration samples for [RabbitMQ](/samples/rabbit
 
 ### Additional metadata 
 
-By providing the following additional headers you will gain a better debugging experience in ServiceInsight. This information is required for including third party message in a flow diagram.
+For NServericeBus v4 and above by providing the following additional headers you will gain a better debugging experience in ServiceInsight. This information is required for including the external message in a flow diagram.
 
 Header key  | Value
 ------------- | -------------
--- a/nservicebus/messaging/message-headers-for-third-party-integrations.md
+++ b/nservicebus/messaging/message-headers-for-third-party-integrations.md
@@ -12,24 +12,26 @@ related:
 
 # How to integrate external systems with NServiceBus?
 
-When an NServiceBus endpoint receives messages from external systems such as BizTalk, TIBCO, etc directly, the message itself will not contain regular NServiceBus headers. Some headers are required by NServiceBus to handle those messages (i.e. enable deserialization of the message). Others are essential to provide effective monitoring and troubleshooting experience. 
+When an NServiceBus endpoint receives messages from external systems such as BizTalk, TIBCO, etc directly, the message itself might not contain regular NServiceBus headers. Headers provide additional information necessary for proper serialization of the message, better monitoring and troubleshooting experience.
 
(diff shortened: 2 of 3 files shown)


nservicebus/messaging/message-headers-for-third-party-integrations.md
```diff
@@ -0,0 +1,36 @@
...
+- Third Party Integration
+- ServiceInsight
+- ServicePulse
+related:
+- nservicebus/messaging/message-headers
+---
+
+# How to integrate external systems with NServiceBus?
+
+When an NServiceBus endpoint receives messages from external systems such as BizTalk, TIBCO, etc directly, the message itself will not contain regular NServiceBus headers. Some headers are required by NServiceBus to handle those messages (i.e. enable deserialization of the message). Others are essential to provide effective monitoring and troubleshooting experience.
```
danielmarbach: will not contain --> might not contain ?

nservicebus/messaging/message-headers-for-third-party-integrations.md
```diff
@@ -0,0 +1,36 @@
...
+- ServicePulse
+related:
+- nservicebus/messaging/message-headers
+---
+
+# How to integrate external systems with NServiceBus?
+
+When an NServiceBus endpoint receives messages from external systems such as BizTalk, TIBCO, etc directly, the message itself will not contain regular NServiceBus headers. Some headers are required by NServiceBus to handle those messages (i.e. enable deserialization of the message). Others are essential to provide effective monitoring and troubleshooting experience. 
+
+When NServiceBus endpoints send messages, they include a certain set of headers such as the type of message, the name of the sending endpoint, the machine name etc. To get the best use of your monitoring and debugging tools, please ensure that the following headers are being added to the actual message by the third party endpoints when sending a message.
```
danielmarbach: Feels a bit like repetition of the previous statements.

nservicebus/messaging/message-headers-for-third-party-integrations.md
```diff
@@ -0,0 +1,36 @@
...
+
+When NServiceBus endpoints send messages, they include a certain set of headers such as the type of message, the name of the sending endpoint, the machine name etc. To get the best use of your monitoring and debugging tools, please ensure that the following headers are being added to the actual message by the third party endpoints when sending a message. 
+
+### Mandatory Headers
+
+Required for deserializing the message:
+
+Key  | Value
+------------- | -------------
+NServiceBus.EnclosedMessageTypes  | Fully Qualified Message type including the assembly name, for e.g. IntegrationSample.Messages.Commands.ProcessOrder, IntegrationSample.Messages
```
danielmarbach: I have weak memories but I thought this one is only required depending on the serializer. Because the XmlSerializer if I recall correctly adds that metadata to the namespace of the xml and therefore doesn't need it (not sure)

SimonCropp:
how does NSB Version effect all this?

nservicebus/messaging/message-headers-for-third-party-integrations.md
andreasohlund: I think we need to check how older versions behave because this one isn't needed currently (in fact no headers are needed) -  see:

http://docs.particular.net/samples/sqltransport/native-integration/#code-walk-through-serialization (json)

and

http://docs.particular.net/samples/rabbitmq/native-integration/#code-walk-through-message-serialization (xml)

weralabaj: @andreasohlund @danielmarbach Maybe what I did in my MSMQ sample was not correct, but if I just send the message like this:
var msg = new ProcessOrder {  OrderId = Guid.NewGuid(),  ProductId = 123456 };
var m = new Message(msg);
messageQueue.Send(m, myTransaction);
It passes type name, but excluding namespace. In such case I get serialization exception, because of missing namespace (NSB xml serializer can't find the type then in mapper).
In RabbitMQ sample the type node contains namespace name, but can we assume it's always the case after serialization or should we document that it's required? The same goes for SQL sample.

Btw if we need to check that across versions, up to which version should I check?

andreasohlund: Is the sample you mention online somewhere?

How do you serialize the data? xml? json?

danielmarbach: The problem is Xml != Xml. Our default `XmlSerializer` requires a certain schema. If this is fulfilled then you don't need the headers

weralabaj:
@SimonCropp Didn't check, I assumed this is only for the newest (at least in the beginning).
@indualagarsamy Which ones should we check? Do we go through all combinations (version, serializer, transport) or what would be the safe subset?

nservicebus/messaging/message-headers-for-third-party-integrations.md
weralabaj: So maybe that's exactly what I need to say - either you provide full name in serialized message or you need to add the header. I'll do that.

SimonCropp: Ensure u test any of your assumptions

weralabaj: @andreasohlund This is just sample I used for testing, you can see it here https://github.com/weralabaj/Work/tree/master/IntegrationSample.

weralabaj:
@danielmarbach @andreasohlund Is the new version better?

weralabaj:
@distantcam Question regarding SI - the additional headers for endpoint info are required because of the check in MessageFlowViewModel.LinkConversationNodes(). You mention there some issue with multiple parent messages in a comment.
So you're ok with adding info about additional metadata or you think we should drop the check in SI and don't require user to provide that extra info?

indualagarsamy:
@SimonCropp - I think the OriginatingEndpoint and OriginatingMachine name headers were added for only v4 and above. Not sure if v3 has these headers. But, in general, not sure how the flow diagram will be for v3 endpoints for SI. But adding the headers won't harm and might display better diagrams. But if the headers aren't available for v3 endpoints, SI might not be showing some of those diagrams properly. That's something we can verify.

andreasohlund:
@weralabaj I think we need to create subsection for each serializer.

In your sample you're (I assume) using the .net binary serilizer since you just pass the object to the `System.Messaging.Message` class (I didn't know that even worked!)

https://github.com/weralabaj/Work/blob/master/IntegrationSample/SendRawMessage/Program.cs#L34

So here a header is needed. 


* If you use json.net you need to set the `$type` attribute (json.net specific) see the sql integration sample


* If you use NServiceBus XmlSerializer the root node needs to match the typename (either fully qualified or just the class name) see the RabbitMQ integration sample


Should we also consider a section that describes how to set the raw headers for each transport if the users should have to do that?

* MSMQ is using xml in the `Extension` property https://github.com/weralabaj/Work/blob/master/IntegrationSample/SendRawMessage/Program.cs#L36

* Sql is using json in a column called `Headers` https://github.com/Particular/NServiceBus.SqlServer/blob/develop/src/NServiceBus.SqlServer/TableBasedQueue.cs#L201

* RabbitMQ has a client that allows you to `.SetHeader("name","value")` and so on

Perhaps the above should be added to the docs for each transport and we just link to it?

distantcam:
@weralabaj It should be ok to just rely on conversation ID. I think retried messages were the problem. SI should just ignore multiple messages (because they should be the same anyway) and just show the data for the first result.

weralabaj:
@distantcam So you suggest changing SI? Right now flow diagram won't properly draw arrows for the first (external) message. It will be on the diagram, just on the side.

weralabaj:
@andreasohlund Updated according to comments, please see if it's clearer now. I raised a separate issue for documenting adding headers here https://github.com/Particular/docs.particular.net/issues/733.

andreasohlund:
+1 to merge this one now

nservicebus/messaging/third-party-integration.md
```diff
@@ -0,0 +1,41 @@
+---
+title: Third Party Integration
+summary: Lists the necessary headers when receiving messages from non-NSB endpoints for better monitoring.
```
SimonCropp: the summary doesnt match the title. ie this is more than just headers

nservicebus/messaging/third-party-integration.md
```diff
@@ -0,0 +1,41 @@
...
+
+Header key  | Value
+------------- | -------------
+NServiceBus.EnclosedMessageTypes  | [FullName](https://msdn.microsoft.com/en-us/library/system.type.fullname) of your message type, e.g. `IntegrationSample.Messages.Commands.ProcessOrder`
+
+If set NServiceBus will instruct the serializer to deserialize the payload into the type specified.
+
+### Additional information 
+
+For NServericeBus v4 and above by providing the following additional headers you will gain a better debugging experience in ServiceInsight. This information is required for including the external message in a flow diagram.
```
SimonCropp: and what about NSB 3?

nservicebus/messaging/third-party-integration.md
```diff
@@ -0,0 +1,41 @@
...
+
+Header key  | Value
+------------- | -------------
+NServiceBus.ConversationId  | Valid Guid, useful to tie in the whole message flow to get an accurate view in ServiceInsight, for example in a message flow or sequence diagrams. If not provided, then the external message won't be included in the conversation it started. ConversationId will be generated though for all the following messages sent using NServiceBus.
```
SimonCropp: > generated though

SimonCropp: this makes no sense 

> If not provided, then the external message won't be included in the conversation it started

If this message started the conversation and "ConversationId will be generated" then how will it not be included?

nservicebus/messaging/third-party-integration.md
```diff
@@ -0,0 +1,41 @@
...
+- Third Party Integration
+- ServiceInsight
+- ServicePulse
+related:
+- nservicebus/messaging/message-headers
+---
+
+# How to integrate external systems with NServiceBus?
+
+When an NServiceBus endpoint receives messages from external systems such as BizTalk, TIBCO, etc. directly, the message itself might not contain regular all information that NServiceBus provides. That information is necessary for proper serialization of the message, as well as better monitoring and troubleshooting experience.
```
SimonCropp: add some brackets

> When an NServiceBus endpoint receives messages from external systems (such as BizTalk, TIBCO, etc.)...

SimonCropp: isnt the word directly redundant? or at leas out of place?

nservicebus/messaging/third-party-integration.md
```diff
@@ -0,0 +1,41 @@
...
+
+To get the best use of your tools ensure that the following information is included in the message sent by the third party endpoints. 
+
+### Required information
```
SimonCropp: this makes no sense with the prevous statement.

>To get the best use of your tools ensure

implies it is optional but this title is "Required"

nservicebus/messaging/third-party-integration.md
weralabaj: @SimonCropp Maybe that's wrong wording but this is exactly what happens - the message that normally starts a conversation if it comes from 3rd party endpoint won't get the same conversation id as other messages in the flow. If the same message type comes from NSB endpoint though, it'll be part of the conversation.
I asked about this yesterday in the meeting and @andreasohlund said it's correct and expected behaviour. If I describe it in the wrong way, then please, correct. Seems like I misunderstood.

nservicebus/messaging/third-party-integration.md
weralabaj: I'm not sure, if I got @indualagarsamy correctly then v3 is supported only to some minimal extent in SP/SI. I'll build another sample for v3 and verify.

nservicebus/messaging/third-party-integration.md
andreasohlund: Perhaps we can actually fix this by changing the core to use the `MessageId` as `ConversationId` should the message not have a explicit `ConversationId` ?

weralabaj: @andreasohlund In your opinion should we add this to 194 scope? Or raise a separate bug and document for now as it is?

nservicebus/messaging/third-party-integration.md
weralabaj: @SimonCropp I need some help here. The diagrams use "CorrelationId" header to decide which messages are part of the conversation. But I think it was only introduced in v4 (https://github.com/Particular/NServiceBus/commit/bf30e08cb62b655654702605582f8c0f73d96e3f). 
If I run sample from v3 messages will be displayed on the list in SI but it doesn't have information required for conversations.
Can I just say that it's a concept introduced in v4 and therefore some features of SI are not available for earlier versions?

SimonCropp: > Can I just say that it's a concept introduced in v4 and therefore some features of SI are not available for earlier versions?

I am ok with that. unless @distantcam can think of something better

nservicebus/messaging/third-party-integration.md
andreasohlund: Raise a new issue

weralabaj:
So can we merge or does it need anything more?

weralabaj:
Raised issue https://github.com/Particular/NServiceBus/issues/2797.

menu/menu.yaml
```diff
@@ -24,6 +24,7 @@
         - "nservicebus/messaging/non-durable-messaging"
         - "nservicebus/messaging/handling-responses-on-the-client-side"
         - "nservicebus/messaging/message-headers"
+        - "nservicebus/messaging/message-headers-for-third-party-integrations"
```
SimonCropp: the name in the menu does not match the new page name

weralabaj:
Thanks @SimonCropp! I missed that one.

SimonCropp:
@weralabaj no worries. the engine would have notified us anyway

PR #738, opened 2015-08-12
Remove delivery and critical times from si description

weralabaj:
Refers to https://github.com/Particular/ServiceInsight/pull/477


weralabaj:
@johnsimons Missed that one before release, please review.

johnsimons:
good catch @weralabaj

Issue #731, opened 2015-08-09
update showcase


SimonCropp:
http://docs.particular.net/samples/show-case/on-premise/

Since this is our only demo of how to use NSB with SignalR we should ensure it is representative of the current version of signalr. currently it uses V 1.0.0	 from February 2013

Should also update bootstrap, jquery etc

SeanFeldman:
@SimonCropp I've used this just recently, don't mind to do it

SimonCropp:
@SeanFeldman ok. ping @petersgiles or @weralabaj if u need help on the js/css changes

PR #741, opened 2015-08-13
WIP: upgrade show-case dependencies 

SeanFeldman:
This PR is to address issue #731 

@weralabaj & @petersgiles could you please help with JS/CSS dependencies upgrade?


petersgiles:
@SeanFeldman I misunderstood what you wanted and I just went and did this in another branch and your changes pretty much matched mine. So I discarded mine in preference for yours.

SimonCropp:
also should the dependencies be pulled in through nuget or through a cdn eg

   <link rel="stylesheet" href="//cdnjs.cloudflare.com/ajax/libs/twitter-bootstrap/3.3.5/css/bootstrap.min.css" />

petersgiles:
Which would you prefer?

SimonCropp:
i prefer the cdn. although i am not sure it is possible with signalr and it dependencies??

SeanFeldman:
I don't think we should pull from CDN. This is a showcase, sample, not
production code.

petersgiles:
OK, We can look into that. In short, you pretty much want the latest of everything, served from the cloud where possible.

SimonCropp:
I just want the least moving pieces and the simplest to update moving forward.

nuget is in the process of deprecating content nuget packages. which means will JS nugets are dead and just dont know it yet

SimonCropp:
but i will defer to the guys doing the work ;)

petersgiles:
Ok CDN it is. :)

petersgiles:
@SeanFeldman I've made it use CDN's for the static assets and I've made it work a little nicer in bootstrap 3.

@SimonCropp  If you're happy with this I'm happy with it.

PR #745, opened 2015-08-14
Doco for timeout storage split and timeout persistence options

danielmarbach:
Belongs to https://github.com/Particular/NServiceBus/pull/2802


danielmarbach:
@DavidBoike here the doco

PR #747, opened 2015-08-17
Update unit-of-work.md

pashute:
changed obsolete code
added licensing issue and static datastore code

### DIFF
--- a/nservicebus/ravendb/unit-of-work.md
+++ b/nservicebus/ravendb/unit-of-work.md
@@ -6,21 +6,22 @@ redirects:
  - nservicebus/unit-of-work-implementation-for-ravendb
 ---
 
-When using a framework like NServiceBus you usually need to create your own unit of work to avoid repeating code in your message handlers. Following is one approach on how to implement the NServiceBus Unit of Work for RavenDB.
+When using a framework like NServiceBus many times you may need to create your own unit of work to avoid repeating code in your message handlers. Following is one approach on how to implement the NServiceBus Unit of Work for RavenDB.
 
 ## Sharing the session
 
-Share the session between the message handler(s) and the actual unit of work implementation. Do not use thread static, which has issues mentioned by [Andreas Öhlund](http://andreasohlund.net/) in a [blog post](http://andreasohlund.net/2010/03/25/thread-static-caching-in-nservicebus/).
+In order to share the RavenDB session between the message handler(s), it is best to share it through a 'unit of work' implementation. It is recommended NOT to use thread static variables, which are not cleared when the request ends, and may cause various connectivity or corrupted data problems in the next request, for example if another type of message was handled between the handling of two handlers for a subscribed event. 
 
-Instead, beginning with NServiceBus version 3, use the new [support for child containers](/nservicebus/containers/child-containers.md). This means that all dependencies configured as a single call effectively become static within the context of one transport message.
+Instead, beginning with NServiceBus version 3, use the new [support for child containers](/nservicebus/containers/child-containers.md). This means that all dependencies configured as a single call effectively become static within the context of one transport message, and thus can be shared by that message's handlers, but the next message will have cleared and renewed the information.
 
 To resolve a RavenDB document session from the container, add the following configuration (StructureMap is used but any of the other containers except Spring and Unity would work):  
 
 ```C#
+// Note: The 'store' variable's contents will become statically available to message hanlers
+// through injection by the container and the 'Unit of Work' mechanism 
 var store = new DocumentStore {Url = "http://localhost:8080", DefaultDatabase = "MyDatabase"};
 store.Initialize();
  
-//ObjectFactory.Configure(c =>  // objectFactory has been deprecated in structureMap
 Container container = new Container( c => 
                     {
                         c.For<IDocumentStore>().Singleton()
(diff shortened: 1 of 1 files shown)


nservicebus/ravendb/unit-of-work.md
```diff
@@ -14,31 +14,55 @@ Share the session between the message handler(s) and the actual unit of work imp
 
 Instead, beginning with NServiceBus version 3, use the new [support for child containers](/nservicebus/containers/child-containers.md). This means that all dependencies configured as a single call effectively become static within the context of one transport message.
 
-To resolve a RavenDB document session from the container, add the following configuration (StructureMap is used but any of the other containers except Spring and Unity would work):
-
+To resolve a RavenDB document session from the container, add the following configuration (StructureMap is used but any of the other containers except Spring and Unity would work):
```
andreasohlund: Pretty sure both spring and unity works now (the latest versions). Please confirm @danielmarbach and @SzymonPobiega ?

nservicebus/ravendb/unit-of-work.md
```diff
@@ -14,31 +14,55 @@ Share the session between the message handler(s) and the actual unit of work imp
...
+                        c.For<IDocumentSession>()
+                            .Use(ctx => ctx.GetInstance<IDocumentStore>()
+                                        .OpenSession());
+                        c.For<IManageUnitsOfWork>()
+                            .Use<RavenUnitOfWork>();
+                    });
+configuration.UseContainer<StructureMapBuilder>(c =>
+                       c.ExistingContainer(container));
+
+// old version:
```
andreasohlund: We want to document the latest versions so just remove the commented out code

nservicebus/ravendb/unit-of-work.md
```diff
@@ -14,31 +14,55 @@ Share the session between the message handler(s) and the actual unit of work imp
...
+// rest of contiguration goes here...
+// configuration.UseSerialization<JsonSerializer>(); // default is XmlSerializer...
+configuration.UsePersistence<RavenDBPersistence>();
+
+
 ```
 
 The above code tells the container to create a new `IDocumentSession` using the specified lambda. The fact that all message processing is done using a child container means that all message handlers processing the message get the same session instance.
 
+If using the NServiceBus host via Nuget, you can add the above code inside the Customize() method in the EnpointConfig. In the case of StructureMap, you'll need get NServiceBus.StructureMap through Nuget, and add at the top of the class:
```
andreasohlund: Do we really need this, the tooling would just prompt the user and add the usings?

andreasohlund: Actually, shouldn't we just link to the sample for this?

SzymonPobiega:
@andreasohlund Unity does support the child containers. Are there any other requirements for UoW to work?

andreasohlund:
No that should be it

nservicebus/ravendb/unit-of-work.md
pashute: You'll have to fix the sample for that...

nservicebus/ravendb/unit-of-work.md
pashute: the links to andreas' blog don't work anymore...

andreasohlund: Blog is back now

I'll rework the page to not point to link to my blog since we have all
content we need on docs.particular.net

nservicebus/ravendb/unit-of-work.md
pashute: You won't believe how much time I wasted finding out that I can use the Container container = new Container...

There are tons of documentation asking about the obsolete ObjectFactory and all giving obscure and complex answers (by which you have to create your own ObjectFactory...  or by which you "shouldn't be using anything like it, that's why it went obsolete...)

When moving back to the non Unit of Work (because of the licensing issue) I forgot to remove the constructor in the handler, and my handler stopped working. After 3 hours of panic and two minutes after posting to NServiceBs support someone at NServiceBus pinpointed the problem. Yay! (and whew!)

andreasohlund: I definitely understand your frustration but this doco site needs to focus
on NServiceBus and including doco for all our various dependencies just
doesn't scale. Doco like this should be located and maintained be the
relevant donwstream library, StructureMap in this case.

Since this is StructureMap specific documentation we can't really do much
more other than linking to the SM doco like we do here
http://docs.particular.net/nservicebus/containers/structuremap

Does this make sense?

pashute: No, I'm not frustrated. I just told about the problem when old sample code showing how to use the Unit-of-Work was outdated. Its not StructureMap code because they have no interest in an NServiceBus Unit-of-Work in an NServiceBus handler (that has to do with RavenDB, which is also a good choice of example)

So no. This documentation should be simply corrected, and also made simple to understand (I think that is what I did). As you requested, I removed all the old code instead of remarking it. 

It is an important peace of code and important documentation for NServiceBus. 

It uses the built in NServiceBus container builder class, and I wish I knew if there's a way to set the "internal" default container type to  use this container. 

The only structuremap directive is in the last line...   

I tried it and it worked. I would spend some time testing with the other containers, but don't have time for this.

andreasohlund:
Hi @pashute we decided to remove that specific page and instead link to the sample

https://github.com/Particular/docs.particular.net/pull/785

really sorry for the wasted effort on your part!

pashute:
Nothing wasted. All study and learning of your great product. 

Andreas, is there any documentation about the satellite and how it is initiated? 
Do I just write one and it automatically runs, or do I have to instantiate it somewhere (in a nservicebus iwanttorunatstartup class?)

andreasohlund:
We do not have any doco on satellites yet since its really only meant for
internal usage. We're also rethinking it for v6 so there might be changes.

But yes, just implement it and register it in the container and it will run

PR #748, opened 2015-08-17
Clarify serializers configuration.

weralabaj:
In progress...


SimonCropp:
@weralabaj isnt that covered here http://docs.particular.net/nservicebus/serialization/xml

weralabaj:
@SimonCropp I don't see it mentioning that xml is the default or that we need to use the same on sending/receiving endpoints. That relates to desk case 12702 btw.
If you think there's a better place for that information I can move it, but that's where I'd go first to understand serialization better.

SimonCropp:
@weralabaj happy with 

> The default serializer is XmlSerializer. In order to use another one configure the endpoint accordingly, 

but perhaps something similar to this http://docs.particular.net/nservicebus/containers/#supported-containers instead of the json example?

weralabaj:
@SimonCropp Ok. i'll rework it.

weralabaj:
@SimonCropp better now?

SimonCropp:
Looks good to me. Pull the trigger

PR #749, opened 2015-08-17
link to Publishing Events from webapp article

pashute:
added link as to recommendations of when to use this and how.

### DIFF
--- a/samples/web/asp-web-application/sample.md
+++ b/samples/web/asp-web-application/sample.md
@@ -8,6 +8,9 @@ related:
 - nservicebus/hosting
 ---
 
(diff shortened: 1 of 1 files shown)


SimonCropp:
these should be put in a related header https://github.com/Particular/docs.particular.net/blob/master/README.md#an-example-header-for-an-article

on another note fully qualified links are not supported https://github.com/Particular/docs.particular.net/blob/master/README.md#linking

pashute:
So please do so. 

If I would be doing it, how do I find the relative path?

SimonCropp:
done here https://github.com/Particular/docs.particular.net/commit/426c224803e2da3c68d9132cc9eb4dc5e2491d68

PR #750, opened 2015-08-17
further reading

pashute:
links to the relevant how-to articles if one DOES wish to create a web application with messaging (and even events, in some cases according to this article)


SimonCropp:
these should be put in a related header https://github.com/Particular/docs.particular.net/blob/master/README.md#an-example-header-for-an-article

on another note fully qualified links are not supported https://github.com/Particular/docs.particular.net/blob/master/README.md#linking

pashute:
same questions as before. 
a. Could you do it?
b. How do you know the relative path?

SimonCropp:
done here https://github.com/Particular/docs.particular.net/commit/239496037006d62ded34af80ac9d58bed502195d

did not add a link to self hosting since that is the recommended approach for all hosting and it is not worth adding that ling to all samples and hosting pages

PR #751, opened 2015-08-17
Note about changing to JSonSerializer

pashute:
important because once some copies from the example they need their app to work correctly with the other projects communicating with this one through NServiceBus.

### DIFF
--- a/samples/web/asp-web-application/sample.md
+++ b/samples/web/asp-web-application/sample.md
@@ -27,6 +27,8 @@ In `WebApplication`, open `Global.asax.cs` and look at the code in the `Applicat
 
 <!-- import ApplicationStart --> 
 
+Note that setting the JSonSerializer will change the default behavior of NServiceBus. In order for this to work you must have all your message handling and message sending projects with the same method of serialization. So either set them all to use this serializer, or remove that line of code so they all use the default (XmlSerializer). 
+
 The rest of the code is typical for hosting NServiceBus in your own process. Read more about [the other lines](/nservicebus/hosting/self-hosting.md).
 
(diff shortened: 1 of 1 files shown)


SimonCropp:
this is not specific to this sample but a general understanding of how message serialization works in a bus. while it might be helpful the cost of manually maintaining that text on every single sample is not worth it

pashute:
Meanwhile the example choosing jsonSerializer is up, and causing frustration with whoever uses it. 
They make a web app, which talks to other NSB hosted apps. The other apps are all supposed to be super simple to create and run. (Ask Udi, or read the docs). Then it doesn't work and you have no idea why not. (I personally spent a few hours on that)   

So until you have a specific link about this (and since it is code that still works) its worth the extra 12 or so extra words to save 12 or so extra hours from your future clients....

SimonCropp:
@pashute the samples are not meant to be production ready drop in replacements endpoints for your existing solution's. they are each designed to illustrate specific a feature (or subset) of nsb. 

the hosting choices they make are chosen for the least friction of writing samples. when taking generic sample code you need to ensure all the settings match what you want ie the serializer, transport, logging. 

I agree we do need some more context on the defaults samples use, so i opened this https://github.com/Particular/docs.particular.net/issues/758

> The other apps are all supposed to be super simple to create and run. (Ask Udi, or read the docs

The are all super simple. they are all downloadable and runable.

weralabaj:
@SimonCropp @pashute Rather than "intro" would it help if the new content will be in the form of FAQ, "Troubleshooting" or "common issues"? 
I'm afraid nobody would look at intro (just like nobody but me reads introductions in books), but if we explicitly say "here are solutions for most common problems" it might catch attention when needed.

Then we can link to that content from every sample. On the one hand that might look excessive. But on the other samples are used mainly to learn, quite often by people very new to NSB, so it is a potential time and frustration saver.

pashute:
YES!!!!

2015-08-18 10:20 GMT+03:00 Weronika Łabaj <notifications@github.com>:

SimonCropp:
new samples intro can be seen here http://docs.particular.net/samples/

PR #754, opened 2015-08-17
Multi-deserialization support

hmemcpy:


nservicebus/serialization/multi-deserialization.md
```diff
@@ -0,0 +1,13 @@
+---
+title: How to specify additional deserializers
+summary: Specifying additional deserializers.
+tags:
+- serialization
+redirects:
+- nservicebus/specifying-additional-deserializers
```
SimonCropp: why the redirect? this url doenst currently exist

hmemcpy: @SimonCropp sorry, copypaste. I'm not finished with this yet.

SimonCropp:
in the PR on the core some new concepts have been added. eg `ConfigureSerialization`, changes in `SerializationFeatureHelper`. do these need to be documented?

SimonCropp:
so people implementing a custom serializer now need to implement `ConfigureSerialization` instead of `Feature`?  does that need to be captured?

weralabaj:
> When registering a default serializer to be used when serializing messages, additional deserializers may be specified, in order to support deserializing incoming messages, sent from endpoints using different serialization.

That sounds a bit complex, why not something like:
"Starting from NSB v6 it's possible to use different Serializers for sending and receiving endpoints. That requires specifying additional deserializers."

I didn't follow this issue at all (haven't even read description), so there are a few things I don't get. I think other readers also might have similar questions:
- What are "additional serializers"? How this thing works? Will it try to deserialize each message by iterating through the list of registered serializers unless it succeeds? If not, how?
- Will be the default serializer still used for serializing all messages?
- Why do we call them "default" and "additional"? The "default" means something else in the paragraph above, can we find another word ("primary"?)?
- "Specifying additional deserializers" > But below you say "to configure multiple deserializers". In fact "multiple" makes more sense to me than "additional + default" combination, does it also describe the idea better?
- I think it'd be better structure if we move "NOTE" to "Specifying additional deserializers" section and reorganize it to say: up to version 5 you have to use the same, from v6... and then what you have now. Then we could also update the section title, e.g. "Serializers configuration" or something better indicating that this is about configuration across endpoints.

hmemcpy:
Sorry it took me a while to get here.

@SimonCropp it's currently not a requirement to wrap a serialization in a feature to use a custom serializer. I understand this exists in order to decouple serialization impl. from the actual capability. If someone wants to register a custom serializer, they need only to [implement and register](http://docs.particular.net/nservicebus/serialization/custom-serializer) it.
Also, correct me if I'm wrong, are Features documented? I wasn't able to find anything specific.

@weralabaj 
I agree the language isn't ideal, but the goal was still to keep just one **serializer** to send/receive, but support additional **deserializers** to process messages from other endpoints. The reason they are called "additional" is that they don't replace the main/default serializer, but can be optionally configured in, well... addition to it.
Let me try answering your points:

 > * What are "additional serializers"? How this thing works? Will it try to deserialize each message by iterating through the list of registered serializers unless it succeeds? If not, how?  

As I mentioned, those are additional **deserializers**. I agree the language could work better, but the idea is: 1 serializer (two-way), additional (multiple) deserializer (for incoming messages)

>  * Will be the default serializer still used for serializing all messages?  

Yes. Since the serialization and deserialization functionality is coupled, essentially I am using existing serializers to deserialize only. So only the default/primary one will do both.

> * Why do we call them "default" and "additional"? The "default" means something else in the paragraph above, can we find another word ("primary"?)?  

I will be happy if we can find better, unambiguous language

>  * "Specifying additional deserializers" > But below you say "to configure multiple deserializers". In fact "multiple" makes more sense to me than "additional + default" combination, does it also describe the idea better?

I used "specifying" and configuring "interchangeably" here, since it's possibly to only use the built-in serializers as deserializers (I currently don't support using custom serializers as deserializers)

>  * I think it'd be better structure if we move "NOTE" to "Specifying additional deserializers" section and reorganize it to say: up to version 5 you have to use the same, from v6... and then what you have now. Then we could also update the section title, e.g. "Serializers configuration" or something better indicating that this is about configuration across endpoints.

I wanted to have a separate page for it, but I now think it is more suitable in the Serialization page. Need to think about it some more...

weralabaj:
@hmemcpy it makes sense now. Shout if you need help with writing.

PR #762, opened 2015-08-20
Troubleshooting common issues with samples.

weralabaj:
Draft content for first two issues.


SimonCropp:
took some of the advice and write a samples intro here http://docs.particular.net/samples/

PR #763, opened 2015-08-20
Elaborate on snippets in various file types

SeanFeldman:
to allow quick lookup w/o searching repo for example


SeanFeldman:

PR #784, opened 2015-08-26
Adding a v6 version of the mutator sample

andreasohlund:
Note that this sample is still referencing an unstable core version from MyGet. 

To make this work for our users we need to include a `nuget.config` in the downloaded zip. I will create a seprate pull against the docs engine to handle this


samples/encryption/message-body-encryption/Version_6/Endpoint1/app.config
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8"?>
```
SimonCropp: This file is redundant

samples/encryption/message-body-encryption/Version_6/Endpoint2/app.config
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8"?>
+<configuration>
```
SimonCropp: Redundant file

SimonCropp:
Apart from the redundant files this is good for a merge

andreasohlund:
Ok, I'll remove them, funny I just copy and pasted the v5 sample, wonder
where that app.config came from. Nuget update?

SimonCropp:
@andreasohlund vs adds them when u update to 451. But it is a deployment setting. So useless for samples

PR #790, opened 2015-08-27
Sample Message Mutators v6

petersgiles:
Change messages by plugging custom logic in to a couple of interfaces,

@SimonCropp review please


samples/messagemutators/Version_6/Sample/app.config
```diff
@@ -0,0 +1,3 @@
+<?xml version="1.0" encoding="utf-8"?>
```
SimonCropp: this file is redundant

PR #792, opened 2015-08-27
[WIP] Lifecycle documentation

wolfbyte:
Documenting Lifecycle interfaces and their usage.

INeedInitialization #779 
IWantToRunBeforeConfigurationIsFinalized #780
IWantToRunWhenConfigurationIsComplete #781
IWantToRunWhenTheBusStartsAndStops #782

PR #794, opened 2015-08-27
[WIP] Unobtrusive configuration via INeedInitialization

ramonsmits:
The V5 sample did not work due to a missing encryption key. So there is a fix for that.

Instead of calling the message convention extension method from the client and server projects I refactored this to a `ConfigureMessageConventions` that inherits `INeedInitialization`.

Result is cleaner bus initialization code and easier to use across more projects.

Currently only applied to the V5 project as this is a bit opinionated. If we agree that this improves the sample then I'll apply it to the other versions too.

@mauroservienti What do you think?

### DIFF
--- a/samples/unobtrusive/Version_5/Client/Program.cs
+++ b/samples/unobtrusive/Version_5/Client/Program.cs
@@ -12,8 +12,6 @@ class Program
             .BasePath(@"..\..\..\DataBusShare\");
         busConfiguration.RijndaelEncryptionService("gdDbqRpqdRbTs3mhdZh8qCaDaxJXl+e7");
 
-        busConfiguration.ApplyCustomConventions();
-
         using (IBus bus = Bus.Create(busConfiguration).Start())
         {
             CommandSender.Start(bus);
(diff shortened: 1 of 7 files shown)


SimonCropp:
i fixed the v5 sample manually

this is not a good thing to leak into the sample. Where our configuration APIs are called is up to the consumer. You stating "INeedInitialization Result is cleaner bus initialization code and easier to use across more projects" is fine but if the consumer doesnt need this then is is just obscuring how to use the API the sample is trying to illustrate with no value. Where someone calls our API will be decided by many factors that a sample "illustrating unobtrusive" should no have an opinion on. As such the API used by all the samples are the "least obscure and opinionated". Feel free to add some of this content to the doco about INeedInitialization https://github.com/Particular/docs.particular.net/pull/792

FWIW INeedInitialization should almost always be avoided for the following reasons

- adds complexity due to the requirement of assembly scanning
- order of execution is determined bu assembly scanning and hence, in most cases, is non determinitic
- do not support passing in extra context (other than the config) forcing people to use static state to get access to more things. or pollute the state of the busconfig 
- dont support dependency injection so no DI
- must have a default constructor
- dont support async

All of the above restrictions/problems do not exist for the case where user call our api explicitly

PR #813, opened 2015-09-01
Features

timbussmann:
Documents the Feature class

### DIFF
--- a/Snippets/Snippets_5/Features/FeatureConfiguration.cs
+++ b/Snippets/Snippets_5/Features/FeatureConfiguration.cs
@@ -37,7 +37,7 @@
 #region EnableDisableFeatures
             var configuration = new BusConfiguration();
             
-            // enable Saga feature since our custom feature relies on it
+            // enable delayed delivery feature since SLR relies on it
             configuration.EnableFeature<DelayedDeliveryFeature>();
 
             // this is not required if the feature uses EnableByDefault()
--- a/Snippets/Snippets_5/Features/FeatureConfiguration.cs
+++ b/Snippets/Snippets_5/Features/FeatureConfiguration.cs
@@ -1,0 +1,72 @@
+﻿namespace Snippets5.Features
+{
+    using System;
+    using NServiceBus;
+    using NServiceBus.Features;
+    using NServiceBus.Pipeline;
+
+    class MyMagicFeature : Feature
+    {
+
+#region FeatureConfiguration
+        public MyMagicFeature()
+        {
+            EnableByDefault();
+
+            DependsOn<Sagas>();
+
+            Prerequisite(c => c.Settings.EndpointName().StartsWith("Magical"), "This feature only works on magical endpoints");
+        }
+#endregion
+
+#region FeatureSetup
+        protected override void Setup(FeatureConfigurationContext context)
+        {
+            // retrieve some global settings
+            var supervised = context.Settings.Get<bool>("SupervisorActive");
+
+            if (!supervised)
+            {
+                // register a more performant cleanup process with the container
+                context.Container.ConfigureComponent<UnsafeCleanupComponent>(DependencyLifecycle.InstancePerCall);
+            }
+
+            // add a custom behavior to the message processing pipeline which does our cleanup.
+            context.Pipeline.Register<CleanupBehaviorRegistration>();
+        }
+#endregion
+
+        void ConfigureFeatures()
+        {
+#region EnableDisableFeatures
+            var configuration = new BusConfiguration();
+            
+            // enable Saga feature since our custom feature relies on it
+            configuration.EnableFeature<Sagas>();
+
+            // this is not required if the feature uses EnableByDefault()
+            configuration.EnableFeature<MyMagicFeature>();
+
+            // magical endpoints obviously need no Second Level Retries
+            configuration.DisableFeature<SecondLevelRetries>();
+
+            var bus = Bus.Create(configuration);
+        }
+#endregion
+    }
+
+    class UnsafeCleanupComponent
+    {
+    }
+
+    class CleanupBehaviorRegistration : RegisterStep
+    {
+        public CleanupBehaviorRegistration(string stepId, Type behavior, string description) : base(stepId, behavior, description)
+        {
+        }
+
+        public CleanupBehaviorRegistration() : base(string.Empty, typeof(object), "does some custom cleanup tasks")
+        {
+        }
+    }
+}
--- a/Snippets/Snippets_5/Features/FeatureConfiguration.cs
+++ b/Snippets/Snippets_5/Features/FeatureConfiguration.cs
@@ -2,71 +2,78 @@
 {
     using System;
     using NServiceBus;
+    using NServiceBus.Config;
     using NServiceBus.Features;
     using NServiceBus.Pipeline;
 
-    class MyMagicFeature : Feature
+    class SecondLevelRetries : Feature
     {
 
 #region FeatureConfiguration
-        public MyMagicFeature()
+        public SecondLevelRetries()
         {
             EnableByDefault();
 
-            DependsOn<Sagas>();
+            DependsOn<DelayedDeliveryFeature>();
 
-            Prerequisite(c => c.Settings.EndpointName().StartsWith("Magical"), "This feature only works on magical endpoints");
+            Prerequisite(context => !context.Settings.GetOrDefault<bool>("Endpoint.SendOnly"), "Send only endpoints can't use SLR since it requires receive capabilities");
         }
 #endregion
 
 #region FeatureSetup
         protected override void Setup(FeatureConfigurationContext context)
         {
-            // retrieve some global settings
-            var supervised = context.Settings.Get<bool>("SupervisorActive");
+            var retriesConfig = context.Settings.GetConfigSection<SecondLevelRetriesConfig>();
 
-            if (!supervised)
-            {
-                // register a more performant cleanup process with the container
-                context.Container.ConfigureComponent<UnsafeCleanupComponent>(DependencyLifecycle.InstancePerCall);
-            }
+            var retryPolicy = new SecondLevelRetryPolicy(retriesConfig.NumberOfRetries, retriesConfig.TimeIncrease);
+            context.Container.RegisterSingleton(typeof(SecondLevelRetryPolicy), retryPolicy);
 
-            // add a custom behavior to the message processing pipeline which does our cleanup.
-            context.Pipeline.Register<CleanupBehaviorRegistration>();
+            context.Pipeline.Register<SecondLevelRetriesBehavior.Registration>();
         }
 #endregion
 
-        void ConfigureFeatures()
+        void EndpointConfiguration()
         {
 #region EnableDisableFeatures
             var configuration = new BusConfiguration();
             
             // enable Saga feature since our custom feature relies on it
-            configuration.EnableFeature<Sagas>();
+            configuration.EnableFeature<DelayedDeliveryFeature>();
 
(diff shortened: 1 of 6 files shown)


nservicebus/pipeline/features.md
```diff
@@ -0,0 +1,61 @@
...
+### Feature setup
+`Setup` has to be implemented in your feature and will be called if the feature is enabled. It allows you to complete configuration of your feature, hook into the pipeline or registering services on the container. The `FeatureConfigurationContext` parameter on the method contains:
+* Settings: read or write settings which should be available to other components or read access settings provided by NServiceBus
+* Container: register services with the dependency injection container which can be injected to other components
+* Pipeline: register your own behavior into the message processing pipeline or replace/remove existing ones
+
+<!-- import FeatureSetup -->
+
+
+Note: Features are automatically detected and registered by NServiceBus when the assembly is scanned.
```
wolfbyte: But they are not resolved from the IoC so will require a default constructor (at least in NSB5)

nservicebus/pipeline/features.md
```diff
@@ -0,0 +1,61 @@
...
+
+
+### Feature startup tasks
+
+If you need to execute some logic after the feature has been started or stopped you can do that by providing a `FeatureStartupTask` which comes with an `OnStart` and `OnStop` method. Also, the task will always be disposed after stopping the feature if it implements `IDisposable`.
```
wolfbyte: Should this be "execute some feature related logic when the bus starts and stops"?

nservicebus/pipeline/features.md
```diff
@@ -0,0 +1,61 @@
...
+
+<!-- import FeatureSartupTaskDefinition -->
+
+To associate a `FeatureStartupTask` with your feature, register it in the constructor of your feature using `RegisterStartupTask`. The task will only be created and called if the feature is enabled.
```
wolfbyte: You can register in the `Setup` method as we rely on the contents of the Feature and not the FeatureDiagnosticData to register startup tasks. Actually this may be a bug in NSB5.

nservicebus/pipeline/features.md
```diff
@@ -0,0 +1,61 @@
...
+* Container: register services with the dependency injection container which can be injected to other components
+* Pipeline: register your own behavior into the message processing pipeline or replace/remove existing ones
+
+<!-- import FeatureSetup -->
+
+
+Note: Features are automatically detected and registered by NServiceBus when the assembly is scanned.
+
+
+### Feature startup tasks
```
wolfbyte: Is it worth mentioning that while Features get partially ordered according to their Dependencies, there is no explicit order to StartupTasks?

wolfbyte:
Should we mention that `FeaturesReport` get registered into the IoC so you can resolve it if you want to interrogate the Features that are active at runtime?

Snippets/Snippets_5/Features/FeatureConfiguration.cs
```diff
@@ -0,0 +1,72 @@
...
+#region EnableDisableFeatures
+            var configuration = new BusConfiguration();
+            
+            // enable Saga feature since our custom feature relies on it
+            configuration.EnableFeature<Sagas>();
+
+            // this is not required if the feature uses EnableByDefault()
+            configuration.EnableFeature<MyMagicFeature>();
+
+            // magical endpoints obviously need no Second Level Retries
```
wolfbyte: is it obvious :)

Snippets/Snippets_5/Features/FeatureConfiguration.cs
```diff
@@ -0,0 +1,72 @@
...
+        void ConfigureFeatures()
+        {
+#region EnableDisableFeatures
+            var configuration = new BusConfiguration();
+            
+            // enable Saga feature since our custom feature relies on it
+            configuration.EnableFeature<Sagas>();
+
+            // this is not required if the feature uses EnableByDefault()
+            configuration.EnableFeature<MyMagicFeature>();
```
wolfbyte: If Sagas were not enabled, does this also enable Sagas or does it prevent MyMagicFeature from becoming enabled?

nservicebus/pipeline/features.md
timbussmann: I wasn't aware of that. But does it make sense do document that option? Currently all "base class" methods are used in the constructor. Is there a case where we want to register a task conditionally from the Setup?

nservicebus/pipeline/features.md
timbussmann: Everything we document basically becomes part of the API and makes it more difficult to change. That's why I didn't state anything about the order of features and startup tasks. I think users should avoid to rely on any order in startup tasks.

Snippets/Snippets_5/Features/FeatureConfiguration.cs
timbussmann: it prevents MyMagicFeature from becoming enabled

nservicebus/pipeline/features.md
timbussmann: added a note that users can't use DI and need to provide an empty ctor.

SimonCropp:
just a suggestions. whenever building something or writing guidance it is better to avoid words like "Magical", "Smart", "Awesome". If someone is frustrated (perhaps trying to diognose a productin bug) and is consuming this Magical work they will be in the context of "this product/doco is terrible". For them the overuse of highly positive emotive words will just annoy them. Better to pick fake, and innocuous,  problem domain: "Order", "Shipping" or build a real pieces of infrastructure eg "authorization" or  "logging"

Snippets/Snippets_5/Features/FeatureConfiguration.cs
```diff
@@ -0,0 +1,72 @@
...
+        }
+#endregion
+
+#region FeatureSetup
+        protected override void Setup(FeatureConfigurationContext context)
+        {
+            // retrieve some global settings
+            var supervised = context.Settings.Get<bool>("SupervisorActive");
+
+            if (!supervised)
```
SimonCropp: minor coding style suggestion: invert this. if a negated and a non negated if result in roughly equivalent code u should always chose the non negated since it is easier for the developer reading to not need to do the extra mental operation.

Snippets/Snippets_5/Features/FeatureConfiguration.cs
```diff
@@ -0,0 +1,72 @@
...
+
+            Prerequisite(c => c.Settings.EndpointName().StartsWith("Magical"), "This feature only works on magical endpoints");
+        }
+#endregion
+
+#region FeatureSetup
+        protected override void Setup(FeatureConfigurationContext context)
+        {
+            // retrieve some global settings
+            var supervised = context.Settings.Get<bool>("SupervisorActive");
```
SimonCropp: couldnt u use tryget here and avoid the if below?

Snippets/Snippets_5/Features/FeatureConfiguration.cs
SimonCropp: shouldnt the enabling of saga be owned by the endpoint config code not a feature? if i was in a real large apps dev team and i saw this i would recommend instead the feature throw an exception if it detects sagas no being enable. this way it would avoid the unintended side effect of someone using the feature and sagas unexpectantly enabling

nservicebus/pipeline/features.md
wolfbyte: The `AutoSubscribe` does it but arguably it's doing it wrong

Snippets/Snippets_5/Features/FeatureConfiguration.cs
SzymonPobiega: I am :+1: for not enabling a feature from another feature (in future). Looking at the code I can see that we are using `EnableFeatureByDefault` only in parent-child feature relationship. 

If we has this parent-child (or sub-feature) relationship explicit, we could limit the enabling only to the parent where it actually makes sense.

The other flow we have with regards to dependencies is following:
 * Sagas depend on DelayedDelivery
 * DelayedDelivery depends on TimeoutManager
 * I call `config.EnableFeature<Sagas>()`
 * I call `config.DisableFeature<TimeoutManager>()`

The result is that the Sagas feature is disabled. I would argue that the desired result is an exception describing why the bus is unable to fulfill feature dependencies.

nservicebus/pipeline/features.md
SzymonPobiega: I think it is limiting the power of a startup task. As a user I would expect that if my feature X depends on feature Y, my startup task would be executed after Y's startup task. I think that is a reasonable assumption that people will make, no matter what we write in the doco.

timbussmann: yes but we don't need to support assumptions if we want to change the behavior. I think it would be safer to give no guarantees which users can rely on. what if we want to start the tasks in parallel? We could state that we guarantee no reliable order?

Snippets/Snippets_5/Features/FeatureConfiguration.cs
timbussmann: actually, that snipped should represent the endpoint config code and not the feature. I just wrote it into the same class. Does it suffice to make that more clear in the markdown file or is there a real value in extracting that method in a new class & file?

timbussmann:
@SimonCropp fair point about the naming. Will change the snippets.

Snippets/Snippets_5/Features/FeatureConfiguration.cs
timbussmann: Wouldn't you still have to check for SupervisorActive = false and therefore end up having 2 if statements?

nservicebus/pipeline/features.md
wolfbyte: I think this pull request should state that consumers cannot make assumptions about Startup Task ordering and we should have a separate task to talk about it. I had a similar expectation to Szymon and I definitely think that customers will as well.

Snippets/Snippets_5/Features/FeatureConfiguration.cs
```diff
@@ -0,0 +1,79 @@
...
+    {
+
+#region FeatureConfiguration
+        public SecondLevelRetries()
+        {
+            EnableByDefault();
+
+            DependsOn<DelayedDeliveryFeature>();
+
+            Prerequisite(context => !context.Settings.GetOrDefault<bool>("Endpoint.SendOnly"), "Send only endpoints can't use SLR since it requires receive capabilities");
```
SeanFeldman: Nitpicking here, but would error message like this be clearer? "SecondLevelRetries requires endpoint with receive capabilities and cannot be used with send-only endpoint."

Snippets/Snippets_5/Features/FeatureConfiguration.cs
```diff
@@ -0,0 +1,79 @@
...
+            context.Pipeline.Register<SecondLevelRetriesBehavior.Registration>();
+        }
+#endregion
+
+        void EndpointConfiguration()
+        {
+#region EnableDisableFeatures
+            var configuration = new BusConfiguration();
+            
+            // enable Saga feature since our custom feature relies on it
```
SeanFeldman: Is comment correct? Code enables`DelayedDeliveryFeature`, not `Sagas` :)

timbussmann: oops missed that one, good catch!

Snippets/Snippets_5/Features/FeatureConfiguration.cs
timbussmann: actually that message is copied from the productive code :D I think it's ok to keep that message, since it seems to be what is displayed to the user in case the condition fails.

PR #821, opened 2015-09-03
Clarified bypassing outbox when sending from outside

SzymonPobiega:


SimonCropp:
@SzymonPobiega looks good

PR #823, opened 2015-09-03
Outbox store and forward sample

SzymonPobiega:
Shows how to use Outbox to achieve store-and-forward kind of behavior on non-store-and-forward transport (SQL Server) in order to provide better UX in multi-broker scenarios.


samples/sqltransport-outbox-store-and-forward/Version_2.1/Receiver/App.config
```diff
@@ -0,0 +1,37 @@
...
+      connectionString="Data Source=.\SQLEXPRESS;Initial Catalog=receiver;Integrated Security=True" />
+  </connectionStrings>
+  <UnicastBusConfig>
+    <MessageEndpointMappings>
+      <add 
+        Assembly="Shared" 
+        Endpoint="Sender" />
+    </MessageEndpointMappings>
+  </UnicastBusConfig>
+  <TransportConfig
```
andreasohlund: Do we need this setting at all? (not relevent for the sample?)

SzymonPobiega: I've removed the redundant config.

SzymonPobiega:
@andreasohlund do you think this is now OK?

PR #824, opened 2015-09-03
Resolving RyuJIT issues

ramonsmits:
@SimonCropp As Aaron confirmed, disabling RyuJIT helped resolve the issue. Having **KB3083185** installed does not resolve the issue.

### DIFF
--- a/issues/net-framework-ryujit.md
+++ b/issues/net-framework-ryujit.md
@@ -1,0 +1,49 @@
+---
+title: System.TypeInitializationException when starting
+summary: Howto resolve System.TypeInitializationException: The type initializer for '<Module>' threw an exception. ---> System.AccessViolationException: Attempted
+ to read or write protected memory. This is often an indication that other memory is corrupt.
+tags: 
+- Service Control
+redirects:
+related:
+---
+
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


SimonCropp:
So this is realy good content. however i am not sure this should be doco? so "doco should long lived" depending on the outcome of this https://connect.microsoft.com/VisualStudio/feedback/details/1749339/typeinitializationexception-when-ryujit-is-enabled and this https://groups.google.com/forum/#!topic/particularsoftware/VfBfwRiznyE it may result in a windows update release to .net in which case this page might become irrelevant. or we might do a patch in which case this page would become irrelevant. 

I think we should hold off on documenting this till we have more info. However in the short term i think the content of this should be reworked to an announcement that is sent to all clients via email and posted on the group.

@ramonsmits thoughts?

andreasohlund:
+1 to convert to a customer communication for now.

andreasohlund:
@ramonsmits ping, are you onboard the customer communication idea?


Closing this one for now (reopen if needed)

ramonsmits:
Announcement: https://github.com/Particular/Announcements/issues/36

ramonsmits:
Although this is indeed not real documention, it would fit a knowledge base like system. MSDN also has list of knowledge base articles. Maybe something to consider adopting?

PR #826, opened 2015-09-03
Encryption documentation improvements

ramonsmits:


nservicebus/security/encryption.md
```diff
@@ -47,8 +52,15 @@ Property encryption is enabled via the configuration API.
 
 <!-- import EncryptionServiceSimple -->
 
+### Generating encryption keys
+
+In order to use the RijndaelEncryptionService a primary and secondary keys need to be generated. An option is to use `RijndaelManaged.GenerateKey`, please refer to the following article to generate new keys:
+
+- https://msdn.microsoft.com/en-us/library/system.security.cryptography.rijndaelmanaged.generatekey(v=vs.110).aspx
```
DavidBoike: Don't need to generate primary and secondary. The secondary keys are for transitioning from an expired key to a new one, since even if you could change all configs in the system at once (doubtful) there still may be in-flight messages.

Also pointing to GenerateKey is not helpful. That generates a byte array. We read keys as ASCII, but you can't convert the generated key to ASCII because it will have unprintable characters. If you Base64 encode the byte array, then you get a string that's too long, so you need to take the first (can't remember how many) characters.

nservicebus/security/encryption.md
```diff
@@ -96,12 +108,19 @@ This allows you to explicitly handled the encryption and decryption of each valu
 
 ## Message Encryption
 
-Message encryption leverages the pipeline to apply encryption to the whole message body.
+Message encryption encrypts the complete message body. It can even include the message headers if these would contain sensitive data.
```
DavidBoike: This makes it sounds like message encryption is built into NSB. I know in a few lines we say we don't offer but could we reword slightly? "Message encryption is the process of encrypting the entire message payload."

Should we mention this makes ServiceControl ... interesting?

nservicebus/security/encryption.md
```diff
@@ -96,12 +108,19 @@ This allows you to explicitly handled the encryption and decryption of each valu
...
 
-The this class can be then injected into the container using the following
+This class can be then injected into the container using the following:
 
 <!-- import UsingMessageBodyEncryptor -->
+
+
+## Transport Encryption
+
+However, using message encryption is often used to prevent intercepting data during transport. If this is the requirement then you can maybe on the transport for encryption. RabbitMQ, SQL Server, Azure Storage Queues and Azure Service Bus support SSL.
```
DavidBoike: Starting a section under a header with "However" is odd. I think we should add here that we recommend using SSL anytime you go over the public Internet.

nservicebus/security/encryption.md
SimonCropp: Perhaps better in a SC caveats page and link to from here. Also SC doesnt support custom serializers either.

nservicebus/security/encryption.md
```diff
@@ -96,12 +108,19 @@ This allows you to explicitly handled the encryption and decryption of each valu
...
+
```
SimonCropp: So we dont support native msmq encryption? https://msdn.microsoft.com/en-us/library/ms699845(v=vs.85).aspx

nservicebus/security/encryption.md
SimonCropp: There are many snippets that have no dependency on NSB. The value is having them compilable

SimonCropp: Remove the vs.110 from the link to make it not hard coded to verion

SimonCropp: +1 to showing the the code

nservicebus/security/encryption.md
ramonsmits: Agree with the however.

Not entirely true, message encryption makes a lot of sense if you rely on message infrastructure for communication. If the data is encrypted than it does not have a performance impact if it has multiple hops and maybe is even stored. That is why I explicitly mention 'if its used to prevent intercepting data during transport'.

 I hope you understand what I'm trying to say. I think I lack skills to summarize this correctly.

nservicebus/security/encryption.md
ramonsmits: @SimonCropp As far as I understand, that is not supported by our implementation. I think it required AD too, but not sure.

By the way, I was not aware that MSMQ has native transport encryption as mentioned in your link. I only knew transport security via HTTPS and most samples that I've seen use WCF SOAP encryption to encrypt and sign data.

nservicebus/security/encryption.md
ramonsmits: @DavidBoike message encryption is mentioned to list the encryption possibilities. I agree that reading this should make the user aware that we do not offer this out of the box but that we have a sample to demonstrate that it is possible to implement message encryption.

Regarding Service Control, yes it is interesting, as it cannot do anything with a fully encrypted message without knowing how to decrypt it. I don't know enough of SC to write something correct about its behavior.

nservicebus/security/encryption.md
ramonsmits: @DavidBoike Agreed, the primary/secondary line needs to be rewritten. What I wanted to say that all keys should to be generated using a secure pseudorandom generator.

@SimonCropp Agree, actual code is needed.

SimonCropp:

PR #836, opened 2015-09-08
heartbeat plugin does not support send-only endpoints

timbussmann:
added note to plugin documentation page.

### DIFF
--- a/servicecontrol/plugins/heartbeat.md
+++ b/servicecontrol/plugins/heartbeat.md
@@ -29,6 +29,7 @@ NOTE: It is essential that you deploy this plugin to your endpoint in production
  * NServiceBus version 4.x: [ServiceControl.Plugin.Nsb4.Heartbeat](https://www.nuget.org/packages/ServiceControl.Plugin.Nsb4.Heartbeat)
  * NServiceBus version 3.x: [ServiceControl.Plugin.Nsb3.Heartbeat](https://www.nuget.org/packages/ServiceControl.Plugin.Nsb3.Heartbeat)
 
+WARN: The Heartbeat plugin version 2 for NServiceBus 5 is currently not supporting Send-Only endpoints.
 
 ### Deprecated Nuget
 


andreasohlund:
I thought we did support this? cc @gbiellem

servicecontrol/plugins/heartbeat.md
```diff
@@ -14,6 +14,8 @@ The Heartbeat plugin enables endpoint health monitoring in ServicePulse. It send
 
 An endpoint that is marked for monitoring (by ServicePulse) will be expected to send a heartbeat message within the specified time interval. As long as a monitored endpoint sends heartbeat messages, it is marked as "active". Marking an endpoint as active means it is able to properly and periodically send messages using the endpoint-defined transport.
 
+NOTE: The Heartbeat plugin is currently not supported for Send-Only endpoints.
+
 Note that even if an endpoint is able to send heartbeat messages and it is marked as "active", other failures may occur within the endpoint and its host that may prevent it from performing as expected. For example, the endpoint may not be able to process incoming messages, or it may be able to send messages to the ServiceControl queue but not to another queue. To monitor and get alerts for such cases, develop a custom check using the CustomChecks plugin.
```
indualagarsamy: I think we should also clarify behavior change - i.e. with v2.0 endpoints will no longer be shut down if they can't communicate with ServiceControl.

timbussmann: actually there is a warning on the plugins main page stating the shutdown behavior. What if we remove that general warning and add a warning to the plugins which continue to shutdown the endpoint when SC queues are unavailable?

timbussmann:
@andreasohlund we do in the current v5 1.1 version (don't know about other version, it think some versions do while others don't).
While I still think we should support send-only clients, we just can't at the moment while maintaining proper dispose behavior (stop sending heartbeats after dispose).

andreasohlund:
Ok, so the note needs to specify which versions we are talking about right?

timbussmann:
ok I checked v3/v4 and it seems they currently should support Send-Only endpoints. On the other hand this behavior causes a bug which does not stop sending endpoints on bus.Dispose.

so in essence:
* v3: supports Send-Only, doesn't stop sending heartbeats on dispose
* v4: supports Send-Only, doesn't stop sending heartbeats on dispose
* v5-1.1: supports Send-Only, does't stop sending heartbeats on dispose
* v5-2.0: doesn't supports Send-Only, stops sending heartbeats on dispose
* v6: supports Send-Ony, stops sending heartbeats on dispose :tada:

andreasohlund:
Nice forensic work. Is this anything we need to fix?

timbussmann:
@andreasohlund will check on other FeatureStartupTasks in v5 and raise an issue.

timbussmann:
here's the issue about FeatureStartupTask in v5: https://github.com/Particular/NServiceBus/issues/2869

SimonCropp:
so this PR doesnt reflect what is here https://github.com/Particular/docs.particular.net/pull/836#issuecomment-138633671

timbussmann:
slightly changed the warning to mention that only version 2 of the plugin doesn't support Send-Only endpoints

@Particular/documentation The Plugin is already released. I'm happy with skipping this warning altogether (then close the PR) but if you're not happy with the text please provide some suggestions.

PR #845, opened 2015-09-21
Startup and Shutdown sequence sample upgraded to 6.0.0-unstable0878

danielmarbach:
@Particular/documentation See the thread changes introduced. Would this require to remove the info box 

> In some versions of NServiceBus certain extension points are executed on separate threads.

Thoughts?

### DIFF
--- a/samples/startup-shutdown-sequence/Version_6/Sample/Program.cs
+++ b/samples/startup-shutdown-sequence/Version_6/Sample/Program.cs
@@ -1,6 +1,8 @@
 ﻿using System;
 using System.Threading;
 using NServiceBus;
+using NServiceBus.Config;
+using NServiceBus.Config.ConfigurationSource;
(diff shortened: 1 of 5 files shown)


samples/startup-shutdown-sequence/Version_6/Sample/Program.cs
```diff
@@ -33,4 +36,13 @@ static void Main()
         Console.WriteLine("Press any key to exit");
         Console.ReadKey();
     }
+
+    // Remove when https://github.com/Particular/NServiceBus/issues/2913 is fixed
```
SimonCropp: make this a `//TODO:`

danielmarbach: Will do

danielmarbach:
@SimonCropp done. Go ahead

PR #848, opened 2015-09-22
MarkAsComplete no longer virtual

danielmarbach:
@andreasohlund do we need more?


andreasohlund:
Should we have a "if you where using this extension point please contact us"? cc @SimonCropp

danielmarbach:
Any decision?

PR #849, opened 2015-09-22
AutoSubscribe behavior breaking change

danielmarbach:

relates to https://github.com/Particular/NServiceBus/pull/2907


DavidBoike:
> which the necessary indicators what went wrong

Huh?

Is this all assuming, for instance, MSMQ and local queues for publishers? Because for msg-driven publishing, what Subscribe/Unsubscribe are trying to do is send a message to the publisher. So what will happen for a remote publisher? Message is sent, async subscribe returns, and all is well?

danielmarbach:
@DavidBoike if the remote queue is not available the retry kicks in as well. Don't understand what you mean. Going to correct the spelling there.

DavidBoike:
So with MSMQ:

* Local queue name, would fail and retry
* Remote queue name: would succeed but message would get stuck in outgoing queue, no retries

Other message-driven publishing transports (Azure storage queues?) would try the web request, but the queue would be invalid, and so THAT would also fail and retry.

Correct?

danielmarbach:
Correct. The code behaves the same as before. in error conditions, which
are discoverable at the time of the dispatch operation, it will retry as
before but rethrow to the caller and not silently swallow (v5). Ok?

On Tuesday, September 22, 2015, David Boike <notifications@github.com>
wrote:

> So with MSMQ:
>
>    - Local queue name, would fail and retry
>    - Remote queue name: would succeed but message would get stuck in
>    outgoing queue, no retries
>
> Other message-driven publishing transports (Azure storage queues?) would
> try the web request, but the queue would be invalid, and so THAT would also
> fail and retry.
>
> Correct?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/Particular/docs.particular.net/pull/849#issuecomment-142363081>
> .
>


-- 
*#NSBcon 2015 • The "All About NServiceBus" conference is back *
*• Learn more <http://go.particular.net/l/27442/2015-08-21/5d4wzp> →*

PR #853, opened 2015-09-28
Custom timeouts persistance doco update

weralabaj:
Related to https://github.com/Particular/NServiceBus/issues/2885


weralabaj:
@SimonCropp Do you think it's ok to use snippets in that case or should I follow some other example we have?
We discussed with @timbussmann whether it makes sense to provide more details in doco or in the issue, we both think that documenting exception in doco doesn't make sense. Hence the reference to the issue in doco, do you think that is a good idea?

SimonCropp:
@weralabaj i am ok with snippets in this case. and i dont mind referencing an issue.

nservicebus/persistence/authoring-custom.md
```diff
@@ -113,50 +113,20 @@ The in-memory implementation of `ISagaPersister` can be found [here](https://git
...
 
-The in-memory implementation of `IPersistTimeouts` can be seen [here](https://github.com/Particular/NServiceBus/blob/4.6.5/src/NServiceBus.Core/Persistence/InMemory/TimeoutPersister/InMemoryTimeoutPersistence.cs).
-
+In order to provide a custom timeout persister implementation: 
+- In versions 4.x and 5.x you need to implement interfaces `IPersistTimeouts` and `IPersistTimeoutsV2`. 
+    - The reference in-memory implementation of timeouts persistence for NServiceBus v4.x can be seen [here](https://github.com/Particular/NServiceBus/blob/support-4.4/src/NServiceBus.Core/Persistence/InMemory/TimeoutPersister/InMemoryTimeoutPersistence.cs). 
+    - The reference in-memory implementation of timeouts persistence for NServiceBus v5.x can be seen [here](https://github.com/Particular/NServiceBus/blob/support-5.0/src/NServiceBus.Core/Persistence/InMemory/TimeoutPersister/InMemoryTimeoutPersister.cs).
+ 
+The differences in interfaces definitions across versions are related to a patch fix preventing a potential message loss. You can find more details in the [issue description](https://github.com/Particular/NServiceBus/issues/2885).
```
timbussmann: does that really add value to the doco? Shouldn't we just note that V2 is design to solve issues with non-transactional transports and replaces V1. V1 is just there to maintain backward compatibility.

timbussmann:
we should also add v5 snippets for the interfaces?

nservicebus/persistence/authoring-custom.md
weralabaj: I also wanted to cover v6 in this sentence, but I'm happy to change if you think it could be worded better.

weralabaj:
@timbussmann Since they are the same we don't need to specify it. In that case snipped will be shown with label v4-v5 like v3-v5 here http://docs.particular.net/nservicebus/msmq/subscription-persistence. At least that's what I expect.

timbussmann:
@weralabaj are you sure? because those snippets exist in all the version directories.

weralabaj:
@timbussmann Checked with Simon about versions, you were right, I updated according to both comments.

timbussmann:
:+1:

weralabaj:
@SimonCropp good to merge?

Issue #856, opened 2015-09-29
Remove ServiceMatrix from Getting Started with the Particular Service Platform


petersgiles:
Task force: @petersgiles @indualagarsamy 

### Description

Add guidance on building distributed systems on the documentation site. 

Relates to https://github.com/Particular/PlatformDevelopment/issues/454

Fixes this document http://docs.particular.net/platform/getting-started-with-particular-service-platform

This will tie in  the existing documentation including information from the following URLs

- http://docs.particular.net/platform/ 
- http://docs.particular.net/nservicebus/concept-overview 
- http://docs.particular.net/samples/endpoint-configuration/ 
- http://docs.particular.net/samples/step-by-step/
- http://docs.particular.net/nservicebus/messaging/publish-subscribe

### Customer Communication

We have improved Particular's guidance for building distributed systems.

### Alignment with Vision

Making Developers Better

### Plan of Attack

- [ ] Create a page that takes users on a journey through the current documentation
- [ ] Create an approach for gradual improvement of related samples

petersgiles:
The PR was closed 15 days ago and no one has indicated if we even need a getting started page in this format anymore. 

My thought is that the samples are instructive enough and the we should just delete the page(s) in question.

indualagarsamy:
@petersgiles - You are right. We are going to leave the existing documentation for SM with a discontinued tag / warning for a little while pointing to the the announcement page. I believe we can close this issue. I am working on a separate branch for these changes. I'll raise a PR when ready. 
The announcement PR is here: https://github.com/Particular/Announcements/pull/58
I believe we can close this issue.

petersgiles:
done :)

PR #859, opened 2015-09-30
WIP: Initial Draft of Remove ServiceMatrix from Getting Started with the Particular Service Platform

petersgiles:
Related to https://github.com/Particular/PlatformDevelopment/issues/454
Related to https://github.com/Particular/docs.particular.net/issues/856

I looking at this I think we are going to have to make a sample based on
what service matrix did.

@indualagarsamy There's a fair bit of work here and I wanted to check it in where I was before I finished today.

I'm working on a sample project that will setup a Nancy web site I don't think I'll get into how that's done in the documentation but there will be doco on a PostHandler method that would demonstrate how to load a message onto the bus

### DIFF
--- a/platform/getting-started-with-particular-service-platform.md
+++ b/platform/getting-started-with-particular-service-platform.md
@@ -56,8 +56,72 @@ In Visual Studio add a new Empty ASP.NET MVC Project called `ECommerce`
 Open up Package Manager Console making sure that the highlighted project is `OrderProcessing` and run this command 
 
 `Install-Package NServiceBus`
+`Install-Package Autofac`
+`Install-Package Autofac.Mvc5`
+`Install-Package NServiceBus.Autofac `
 
 Also reference `System.Configuration` and `OnlineSales.Internal`
+
+Replace the code in `Global.asax` with this
+
+```C#
+using System.Web.Mvc;
+using System.Web.Optimization;
+using System.Web.Routing;
+using Autofac;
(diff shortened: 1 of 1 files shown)


platform/getting-started-with-particular-service-platform.md
```diff
@@ -321,4 +213,7 @@ What's so exciting about that?  After all inter-process communication has been d
 
 One answer is ***fault tolerance***.  Next we'll explore the fault tolerance and durability features NServiceBus offers.
 
-Continue on and let's review the fault tolerance in our [next article](/servicematrix/getting-started-with-nservicebus-using-servicematrix-2.0-fault-tolerance.md "Getting Started with Fault Tolerance").  
+
+
+#mORE sTUFF TO FIX
```
SimonCropp: > mORE sTUFF TO FIX 

?

petersgiles:
@SimonCropp & @indualagarsamy This document is beginning to feel like this sample http://docs.particular.net/samples/step-by-step/

petersgiles:
@SimonCropp @indualagarsamy any comments? I guess when I made the previous comment I was asking do we need this document all all if the step by step sample does the same thing? what value does this add now? should we take this in another direction?

petersgiles:
After reviewing again I'm certain there is very little value in having this page now at all.

SimonCropp:
ok closing this. can be revisited ltr

PR #864, opened 2015-10-06
Doco on new routing APIs

SzymonPobiega:
Related to https://github.com/Particular/PlatformDevelopment/issues/133

### DIFF
--- a/nservicebus/messaging/routing.md
+++ b/nservicebus/messaging/routing.md
@@ -10,7 +10,7 @@ tags:
 - reply
 ---
 
-One of the things NServiceBus takes care of is the routing of messages. Usually the only thing a developer has to do is call `Send`, `Publish` or `Reply` and the actual message destination is calculated by the framework.
+One of the things NServiceBus takes care of is the routing of messages. Usually the only thing a developer has to do is call `Send`, `Publish` or `Reply` and the actual message destination is calculated by the framework. Flexible message routing is a feature added in Version 6. In previous versions NServiceBus used fixed [message ownership mappings](/nservicebus/messaging/message-owner.md) that allowed only to map the message type (or all types in an assembly or a namespace) to physical addresses (queue names). The flexible routing in V6 breaks this mapping down into a series of individually configurable steps. NServiceBus 6 still understands the old configuration in form of message-endpoint mappings for backwards compatibility.
 
 ## Endpoint and endpoint instances
 
--- a/nservicebus/messaging/routing.md
+++ b/nservicebus/messaging/routing.md
@@ -12,33 +12,40 @@ related:
 - nservicebus/messaging/message-owner
 ---
 
-One of the things NServiceBus takes care of is the routing of messages. Usually the only thing a developer has to do is call `Send`, `Publish` or `Reply` and the actual message destination is calculated by the framework. Flexible message routing is a feature added in Version 6. In previous versions NServiceBus used fixed [message ownership mappings](/nservicebus/messaging/message-owner.md) that allowed only to map the message type (or all types in an assembly or a namespace) to physical addresses (queue names). The flexible routing in V6 breaks this mapping down into a series of individually configurable steps. NServiceBus 6 still understands the old configuration in form of message-endpoint mappings for backwards compatibility.
+One of the core features of NServiceBus is routing of messages. The only thing required is to perform a Send, Publish or Reply and the actual message destination is calculated by the framework. Flexible message routing is a feature added in Version 6. In previous versions NServiceBus used fixed [message ownership mappings](/nservicebus/messaging/message-owner.md) that allowed only to map the message type (or all types in an assembly or a namespace) to physical addresses (queue names). The flexible routing in Version 6 splits this mapping up into a series of individually configurable steps. NServiceBus 6 still understands the old configuration in form of message-endpoint mappings for backwards compatibility.
+
 
 ## Endpoint and endpoint instances
 
 Endpoint is a logical concept that relates to a program that uses NServiceBus to communicate with other similar programs. Each endpoint has a name e.g. `Sales` or `OrderProcessing`.
 
-During deployment each endpoint might be materialized in form of one or many instances. Each instance is an identical copy of binaries resulting from building the endpoint program code base. Usually each endpoint instance is placed in separate directory or on separate machine. Each instance has its unique name that consists of the name of the endpoint (e.g. `Sales`) and up to two *discriminator* values:
- * User-provided discriminator is configured via `BusConfiguration` APIs
+During deployment each endpoint might be materialized in form of one or many instances. Each instance is an identical copy of binaries resulting from building the endpoint program code base. Usually each endpoint instance is placed in separate directory or separate machine. Each instance has its unique name that consists of the name of the endpoint (e.g. `Sales`) and up to two *discriminator* values:
+
+ * User-provided discriminator is configured at startup time via the Bus Configuration APIs
  * Transport-provided discriminator is wired by the transport if a particular transport supports it
 
 Both discriminators are optional and if they are absent, the resulting instance name is based only on the endpoint name. This means that in the absence of discriminators, there can be only one instance of an endpoint because there is one possible instance name value. In order to scale out an endpoint and deploy multiple instances, either a user-provided or transport-provided discriminator has to be used.
 
+
 ## Send, Publish and Reply
 
-NServiceBus offers three top-level API calls for sending messages, namely `Send`, `Publish` and `Reply`. Each of these calls can have different behavior in terms of message routing. 
+NServiceBus offers three top-level API calls for sending messages, namely Send, Publish and Reply. Each of these calls can have different behavior in terms of message routing. 
+
 
 ## Types of routing
 
 There are two types of routing recognized by NServiceBus, unicast routing and multicast routing.
 
+
 ### Unicast routing
 
 Unicast routing uses queues which are inherently point-to-point channels. Each endpoint instance has an input queue. When sending a message to multiple destinations, a message is copied in-memory at sender endpoint and a copy of the message is dispatched to each destination's input queue.
 
+
 ### Multicast routing
 
 Multicast routing uses topics. Multiple receivers can subscribe to a single topic and each message published to the topic is copied by the messaging infrastructure and delivered to each interested receiver.
+
 
 ### Transport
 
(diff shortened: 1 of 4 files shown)


nservicebus/messaging/routing.md
```diff
@@ -0,0 +1,165 @@
...
+tags:
+- routing
+- message
+- route
+- send
+- publish
+- reply
+---
+
+One of the things NServiceBus takes care of is the routing of messages. Usually the only thing develop has to do is call `Send`, `Publish` or `Reply` and the actual message destination is calculated by the framework.
```
SeanFeldman: Typo: "a developer" instead of "develop"

nservicebus/messaging/routing.md
```diff
@@ -0,0 +1,165 @@
...
+
+## Endpoint and endpoint instances
+
+Endpoint is a logical concept that relates to a program that uses NServiceBus to communicate with other similar programs. Each endpoint has a name e.g. `Sales` or `OrderProcessing`.
+
+During deployment each endpoint might be materialized in form of one or many instances. Each instance is an identical copy of binaries resulting from building the endpoint program code base. Usually each endpoint instance is placed in separate directory or on separate machine. Each instance has its unique name that consists of the name of the endpoint (e.g. `Sales`) and up to two *discriminator* values:
```
SeanFeldman: Does this mean both user-provided and transport provided discriminators can be combined together?

nservicebus/messaging/routing.md
```diff
@@ -0,0 +1,165 @@
...
+ * User-provided discriminator is configured via `BusConfiguration` APIs
+ * Transport-provided discriminator is wired by the transport if a particular transport supports it
+
+Both discriminators are optional and if they are absent, the resulting instance name is based only on the endpoint name. This means that in the absence of discriminators, there can be only one instance of an endpoint because there is one possible instance name value. In order to scale out an endpoint and deploy multiple instances, either a user-provided or transport-provided discriminator has to be used.
```
SeanFeldman: What happens if two endpoints run with the same name w/o discriminator on different machines? Will that cause messages processing failure?

nservicebus/messaging/routing.md
```diff
@@ -0,0 +1,165 @@
...
+
+### Transport
+
+The transport selected by the user via
+
+```
+config.UseTransport<T>();
+```
+
+decides, for each of the API calls (`Send`, `Publish`, `Reply`), which routing type is used. For example, MSMQ transport uses the unicast routing for all the APIs while the RabbitMQ transport uses the unicast routing for `Send` and `Reply` but multicast routing for `Publish`.
```
SeanFeldman: Typo: "the RabbitMQ" -> "RabbitMQ"

nservicebus/messaging/routing.md
```diff
@@ -0,0 +1,165 @@
...
+
+```
+config.UseTransport<T>();
+```
+
+decides, for each of the API calls (`Send`, `Publish`, `Reply`), which routing type is used. For example, MSMQ transport uses the unicast routing for all the APIs while the RabbitMQ transport uses the unicast routing for `Send` and `Reply` but multicast routing for `Publish`. 
+
+## Unicast routing
+
+Unicast routing in NServiceBus has a lax layered architecture
```
SeanFeldman: lax layered architecture - perhaps adding a reference?

nservicebus/messaging/routing.md
```diff
@@ -0,0 +1,165 @@
...
+      |                 |     |
+|--------------------|  |     |
+| Instance mapping   |  |     |
+|--------------------|  |-----|
+
+### Type mapping layer
+
+In an ideal world the topmost layer of unicast routing would decide, given a message type, which endpoints should receive the message. In practice such an ideal model is not possible, mostly due to wire-compatibility problems it would cause. A more practical but less strict approach is used where a number of user- and system- defined rules is applied to generate a collection of destinations for a given message type. Each rule operates independently and the result is a logical sum of results of individual rules. Destinations can be expressed either via endpoint name, endpoint instance name or directly via transport address. In last two cases the second and/or third mapping layer is skipped.
+
+#### Dynamic routes
```
SeanFeldman: If common scenario is static routing, why not to explain it firs, and then dynamic?

nservicebus/messaging/routing.md
```diff
@@ -0,0 +1,165 @@
...
+
+To add a dynamic route you need to pass a function that takes a `Type` and a `ContextBag` containing the context of current message processing and returns a collection of destinations. The function is called each time any message is being send so you need to take performance into account (e.g. use caching if the destinations are to be fetched from a database).
```
SeanFeldman: By looking at configuration, both would need to be defined on busConfiguration and it's not clear what's the use case for dynamic routing? Would an example help?

nservicebus/messaging/routing.md
```diff
@@ -0,0 +1,165 @@
...
+
+```
+EndpointName sales = new EndpointName("Sales");
+busConfig.Routing().UnicastRoutingTable.AddDynamic(sales, e => new EndpointInstanceName[] {
+	new EndpointInstanceName(e, "1", "MachineA"),
+	new EndpointInstanceName(e, "2", "MachineB")
+});
+```
+
+In this example the rule returns two instances passing both user (`1`, `2`) and transport (`MachineA`, `MachineB`) discriminators.
```
SeanFeldman: Missing: "both user" -> "both, user-provided discriminator" ?

nservicebus/messaging/routing.md
```diff
@@ -0,0 +1,165 @@
...
+EndpointName sales = new EndpointName("Sales");
+busConfig.Routing().UnicastRoutingTable.AddStaticUsingTransportDiscriminators(sales, "MachineA", "MachineB");
+```
+
+
+### Instance mapping
+
+The bottom layer is responsible for mapping between the instance name and a transport address. Usually it does not require intervention from the user as the selected transport automatically registers its translation rule. The only times where user is required to configure the instance mapping is when the default translation violates some naming rules implied by the transport infrastructure (e.g. the generated transport address is too long for a queue name in MSMQ). Different types of instance mapping have different importance.
+
+#### Exceptions
```
SeanFeldman: Exceptions or "handling routing exceptions"?

nservicebus/messaging/routing.md
```diff
@@ -0,0 +1,165 @@
...
+
+User-provided exceptions have the highest importance, overriding the other sources. In order to register an exception you need to use following API:
```
SeanFeldman: Reads funky :)
Is idea here to say that routes defined for faulted address translations have the highest importance?

nservicebus/messaging/routing.md
SzymonPobiega: No, I don't think it would. Not sure if this is a valid behavior. We had an internal discussion in the TF about that. A good example worth considering is a scaled-out RabbitMQ endpoint using a single shared queue for all the instances. In such case we don't really need to distinguish between these instances. The only reason it might be useful is to provide SC with some unique ID of the thing that is running so that heartbeats work.

nservicebus/messaging/routing.md
SzymonPobiega: :+1:

nservicebus/messaging/routing.md
SzymonPobiega: Yes.

nservicebus/messaging/routing.md
SzymonPobiega: :+1:

nservicebus/messaging/routing.md
SzymonPobiega: Good point.

nservicebus/messaging/routing.md
```diff
@@ -0,0 +1,165 @@
...
+#### Rules
+
+User-provided rules have medium importance, overriding only the transport defaults. In order to register a rule you need to use following API:
+
+```
+EndpointName sales = new EndpointName("Sales");
+busConfig.UseTransport<TransportX>().AddAddressTranslationRule(i => someRule(i));
+```
+
+A rule is a function that takes the instance name as the input and return a transport address or null if it does not provided translation for that particular instance name. All rules registered via this API have equal importance. It is expected that the user is responsible for providing a set of mutually exclusive rules so that for each endpoint instance name there is only one not-null translation result. In case there is more than one, an exception is raised.
```
SeanFeldman: Typo: "if it does not provided" -> provide

nservicebus/messaging/routing.md
SzymonPobiega: OK.

nservicebus/messaging/routing.md
SzymonPobiega: Reordered.

nservicebus/messaging/routing.md
SzymonPobiega: Fixed. Exceptions -> Special cases

nservicebus/messaging/routing.md
SzymonPobiega: Reworded.

nservicebus/messaging/routing.md
SzymonPobiega: Fixed.

nservicebus/messaging/routing.md
SzymonPobiega: Yeah, working on that.

SzymonPobiega:
Thanks @SeanFeldman ! I think I applied all the changes you suggested (hope I haven't missed anything).

SimonCropp:
many things in here need to be converted to imported snippets

SzymonPobiega:
@SimonCropp true. I forgot I wrote this before I merged the routing PRs to develop.

SimonCropp:
@SzymonPobiega so this one is now redundant?

SzymonPobiega:
No, the PR is fine. The actual code PR was not yet merged when I wrote this doco and that's why I inlined the snippets and now forgot that I did that.

SzymonPobiega:
OK. I've migrated the snippets over to the solution. It should be now OK to merge.

nservicebus/messaging/routing.md
```diff
@@ -0,0 +1,130 @@
...
+
+#### Send via
+
+*This is implemented in a separate pull*
+
+The type mapping layer allows to configure the itinerary a given message is suppose to be routed. By default, a message is dispatched to its destination calculated as a result of a routing process. If a *send via* is configured, a message is physically dispatched to the destination specified by *send via* and a `NServiceBus.UltimateDestination` header is attached. Should there be more intermediate hops in the route, `NServiceBus.SendVia.N` headers are attached where `N` is the 1-based index of hop. 
+
+```
+string proxy = "ProxyQueueName";
+busConfig.Routing().UnicastRoutingTable.AddStatic(typeof(OrderAccepted), new EndpointName("Sales"), proxy);
```
SimonCropp: make this a snippet

SimonCropp:
this PR seems to make no distinction between what APIs to use in what version. so based on that all the code and doco here are applicable to all versions of NSB?

SzymonPobiega:
@SimonCropp this applies only to V6.

SimonCropp:
@SzymonPobiega yes but nothing in the content conveys that to the user. Even the title " How NServiceBus routes messages between the endpoints" seems to explain a standard thing that would span versions. So the content need to be updated to explain that routing is a feature added in v6

SzymonPobiega:
@SimonCropp right. Let me try to make it more universal.

SimonCropp:
something along the lines of a blurb at the top. perhaps in a `NOTE:`

> Routing is a feature added  in Version 6. It differs from [specifying a message owner](http://docs.particular.net/nservicebus/messaging/message-owner) in that.... 

it can be something really short. then in http://docs.particular.net/nservicebus/messaging/message-owner  add a link to the routing.

PR #867, opened 2015-10-08
Adding notes about changes in XmlSerialization from v3 to v4

hmemcpy:


nservicebus/upgrades/3to4.md
```diff
@@ -146,6 +146,7 @@ NServiceBus PowerShell cmdlets have moved to NServiceBus.PowerShell.dll. See [Ma
 
 `configure.JsonSerializer()` has been obsoleted, please use `Configure.Serialization.Json();` instead. More info on supported serializers [here](/nservicebus/serialization/).
 
+The `XmlSerializer` will now automatically escape outgoing messages containing invalid characters. Version 3 endpoints will not be able to process those messages correctly.
```
andreasohlund: >Version 3 endpoints will not be able to process those messages correctly.

That reads scary, I think we need to qualify it. How about:

NOTE: This means that messages emitted by Version 4 endpoint containing invalid XML characters will not be processable by Version 3 endpoints

hmemcpy: That's what I had initially, but it read like tautology...
Because it sounds odd: v4 endpoints can handle invalid characters, but v3 cannot, so I wanted to emphasize that v3 is at fault here, not v4. Which is why you should upgrade :)

andreasohlund: @simoncropp we need some help from the doco god here :)

SimonCropp: for starters time based statements like "will now" should never be used in doco. I read it and think "now? you mean when that text was added"?

And another question. in version 3 if there are "invalid characters" will they be correctly encoded and decoded withing Version 3 endpoints?

hmemcpy: @SimonCropp In v3 endpoints, messages with invalid characters will throw an exception during serializing/deserializing. It won't be able to read messages with encoded invalid chars from v4.

I assumed that since this is an upgrade guide between two specific versions, it's time-constrained, "will now" has a specific meaning here. It mentions what changed in this version vs the previous.

But I'm all for better text from a native English speaker :)

PR #868, opened 2015-10-09
spike a replacement for the WCF host

SimonCropp:


samples/web/wcf-callbacks/Version_5/WcfCallbacks/CallbackService.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            ICallback intLocal = bus.SendLocal(request);
+            object intValue = await intLocal.Register();
+            return (TResponse)intValue;
+        }
+        return await bus.SendLocal(request).Register(ar =>
+        {
+            object[] messages = ar.Messages;
+            return (TResponse)messages[0];
+        });
```
DavidBoike: So because these are the only 3 types of callbacks in the NServiceBus API, these 3 cases are all that would be needed? In other words, the customer would just use this class as is and wouldn't have to add additional boilerplate here. Correct?

DavidBoike:
I love the API for adding an additional command/response pair. Much better than what we have currently.

This sample cheats a bit by referencing both the Messages and the WcfCallbacks assemblies, and having namespace-less messages. So, there's the assumption that code is going to be shared on each side of the equation, which may be false. If I controlled both sides of the equation, I'd be using an NServiceBus message, not a web service.

What would this look like if you were enabling true 3rd-party access with metadata generation, and then using the Add Service Reference in Visual Studio to generate a proxy?

If that method can be shown to be easier than the current awfulness of stringifying the generic type into a contract name, then I will be sold that this is superior to our current implementation.

samples/web/wcf-callbacks/Version_5/WcfCallbacks/CallbackService.cs
SimonCropp: Yep

PR #869, opened 2015-10-09
[WIP] How to run the distributor and workers on a single machine

ramonsmits:
**This is a work-in-progress**

Separate document explaining which options are available to run both the distributor and the worker on the same machine for testing and demonstration purposes.

It reuses a bit of the distributor sample and the distributor main page to be complete.

### DIFF
--- a/nservicebus/scalability-and-ha/distributor/test-distributor-on-one-machine.md
+++ b/nservicebus/scalability-and-ha/distributor/test-distributor-on-one-machine.md
@@ -1,0 +1,112 @@
+---
+title: Howto deploy distributor and workers on a single machine for testing
+summary: How to run the distributor and workers on a single machine for testing and demonstration purposes.
+tags:
+- Distributor
+related:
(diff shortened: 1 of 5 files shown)


SimonCropp:

SimonCropp:
@SzymonPobiega @ramonsmits given the distributor is being deprecated in v6 can this be killed?

SimonCropp:
@SzymonPobiega @ramonsmits ping

SzymonPobiega:
I mean, there is still value for users who are on 5 and older and continue to be there. So if it is not a significant effort to finish this guide, maybe we should?

Probably a more important thing now would be a guide I mentioned in https://github.com/Particular/NServiceBus/issues/3437 that shows users how to switch from distributor+worker in single endpoint to a stand-alone distributor.

SimonCropp:
@ramonsmits if u no longer think this is required i will delete the PR.

PR #873, opened 2015-10-15
Added encryption service key identifier documentation.

ramonsmits:
- key identifier header
- key identifier configuration
- key format in ascii and base64 encoding

Connects to Particular/PlatformDevelopment#423

### DIFF
--- a/Snippets/Snippets_3/Encryption/Configuration/encryption.config.xml
+++ b/Snippets/Snippets_3/Encryption/Configuration/encryption.config.xml
@@ -1,13 +1,16 @@
 ﻿<?xml version="1.0" encoding="utf-8" ?>
 <configuration>
-  <!-- startcode EncryptionFromAppConfig-->
+  <!-- startcode EncryptionFromAppConfig -->
   <configSections>
-    <section name="RijndaelEncryptionServiceConfig" 
-             type="NServiceBus.Config.RijndaelEncryptionServiceConfig, NServiceBus.Core"/>
+    <section
+      name="RijndaelEncryptionServiceConfig"
+      type="NServiceBus.Config.RijndaelEncryptionServiceConfig, NServiceBus.Core"/>
   </configSections>
-  <RijndaelEncryptionServiceConfig Key="gdDbqRpQdRbTs3mhdZh9qCaDaxJXl+e6">
+  <RijndaelEncryptionServiceConfig Key="gdDbqRpQdRbTs3mhdZh9qCaDaxJXl+e6"
+                                   KeyIdentifier="2015-10">
     <ExpiredKeys>
-      <add Key="abDbqRpQdRbTs3mhdZh9qCaDaxJXl+e6" />
+      <add Key="abDbqRpQdRbTs3mhdZh9qCaDaxJXl+e6"
+           KeyIdentifier="2015-09" />
       <add Key="cdDbqRpQdRbTs3mhdZh9qCaDaxJXl+e6" />
     </ExpiredKeys>
   </RijndaelEncryptionServiceConfig>
--- a/Snippets/Snippets_4/Encryption/Configuration/encryption.config.xml
+++ b/Snippets/Snippets_4/Encryption/Configuration/encryption.config.xml
@@ -2,14 +2,15 @@
 <configuration>
   <!-- startcode EncryptionFromAppConfig -->
   <configSections>
-    <section 
+    <section
       name="RijndaelEncryptionServiceConfig"
       type="NServiceBus.Config.RijndaelEncryptionServiceConfig, NServiceBus.Core"/>
   </configSections>
-  <RijndaelEncryptionServiceConfig 
-    Key="gdDbqRpQdRbTs3mhdZh9qCaDaxJXl+e6">
+  <RijndaelEncryptionServiceConfig Key="gdDbqRpQdRbTs3mhdZh9qCaDaxJXl+e6"
+                                   KeyIdentifier="2015-10">
     <ExpiredKeys>
-      <add Key="abDbqRpQdRbTs3mhdZh9qCaDaxJXl+e6" />
+      <add Key="abDbqRpQdRbTs3mhdZh9qCaDaxJXl+e6"
+           KeyIdentifier="2015-09" />
       <add Key="cdDbqRpQdRbTs3mhdZh9qCaDaxJXl+e6" />
     </ExpiredKeys>
   </RijndaelEncryptionServiceConfig>
--- a/Snippets/Snippets_5/Encryption/Configuration/encryption.config.xml
+++ b/Snippets/Snippets_5/Encryption/Configuration/encryption.config.xml
@@ -1,13 +1,19 @@
 ﻿<?xml version="1.0" encoding="utf-8" ?>
 <configuration>
-  <!-- startcode EncryptionFromAppConfig-->
+  <!-- startcode EncryptionFromAppConfig -->
   <configSections>
-    <section name="RijndaelEncryptionServiceConfig"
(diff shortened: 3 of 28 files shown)


nservicebus/security/encryption.md
```diff
@@ -43,7 +43,40 @@ You can also use the `WireEncryptedString` type to flag that a property should b
...
 ### Enabling property encryption
 
-Property encryption is enabled via the configuration API.
+Property encryption is enabled via the configuration API. 
+
+#### Key identifier
+
+Each key needs an unique key identifier (`KeyIdentifier`). The key identifier is communicated in the message header meta data and provides the receiving endpoint information on which key to use for decryption.
+
+NOTE: If a key identifier is not set then no messages can be send but received messages without a key identifier header will be decrypted using the keys that have no or an empty KeyIdentifier set.
```
DavidBoike: What if you have endpoints A and B. You deploy the fix to both but don't change configs.

At this point, does everything run the same?

Now you update the config on A to include the key identifier on the existing key. It starts sending messages out. Even though B has the correct key, it won't be attempted, and the message can therefore not be processed?

nservicebus/security/encryption.md
```diff
@@ -43,7 +43,40 @@ You can also use the `WireEncryptedString` type to flag that a property should b
...
+Add an *empty* string (0 length) identification to the KeyIdentifier attribute in order to use a key for decrypting messages with and without key identifier meta info.
+
+
+#### Key format
+
+The key format can be specfied in either ASCII or Base64 format.
+
+With ASCII its not possible to use the full 8bit range of a byte as its a 7bit encoding and even then some characters need to be escaped which is not done resulting in even less characters. Meaning per byte only about 100 values are used. When you use 16 byte / 128 bit keys this means only about 100^16 combinations are possible. 
+If you use a 
+ An alternative is to specifify the key as Base64 encoded string.
```
DavidBoike: Looks like something got screwy here on Line 78?

Also, I would lead with Base64 as it is preferred. Right now it sounds like an also-ran. Make it clear that ASCII is not preferred and is included for backwards compatibility only.

DavidBoike:
Instead of key identifiers like "A", would it be better to use something more meaningful to give the idea of how they should treat these identifiers, like a date string? Also a section on naming of key identifiers would be good.

Snippets/Snippets_3/Encryption/Configuration/encryption.config.xml
```diff
@@ -1,15 +1,16 @@
...
-             type="NServiceBus.Config.RijndaelEncryptionServiceConfig, NServiceBus.Core"/>
+    <section
+      name="RijndaelEncryptionServiceConfig"
+      type="NServiceBus.Config.RijndaelEncryptionServiceConfig, NServiceBus.Core"/>
   </configSections>
-  <RijndaelEncryptionServiceConfig Key="gdDbqRpQdRbTs3mhdZh9qCaDaxJXl+e6">
+  <RijndaelEncryptionServiceConfig Key="gdDbqRpQdRbTs3mhdZh9qCaDaxJXl+e6" KeyIdentifier="B">
     <ExpiredKeys>
-      <add Key="abDbqRpQdRbTs3mhdZh9qCaDaxJXl+e6" />
+      <add Key="abDbqRpQdRbTs3mhdZh9qCaDaxJXl+e6" KeyIdentifier=";A" />
```
SimonCropp: why the colon separator at the front?

Snippets/Snippets_5/Encryption/Configuration/encryption.config.xml
```diff
@@ -1,15 +1,17 @@
...
+  <!-- startcode EncryptionFromAppConfig -->
   <configSections>
-    <section name="RijndaelEncryptionServiceConfig"
-             type="NServiceBus.Config.RijndaelEncryptionServiceConfig, NServiceBus.Core"/>
+    <section
+      name="RijndaelEncryptionServiceConfig"
+      type="NServiceBus.Config.RijndaelEncryptionServiceConfig, NServiceBus.Core"/>
   </configSections>
-  <RijndaelEncryptionServiceConfig Key="gdDbqRpQdRbTs3mhdZh9qCaDaxJXl+e6">
+  <RijndaelEncryptionServiceConfig Key="tWIrVnPy3Smd2r/xcSb9r/1wcVY9/3KWbrzIaemx5yw=" KeyIdentifier="B" KeyFormat="Base64"
```
SimonCropp: since this will render inline can we newline the attributes to reduce dodgy wrapping in the html?

Snippets/Snippets_4/Encryption/Configuration/encryption.config.xml
```diff
@@ -6,12 +6,11 @@
       name="RijndaelEncryptionServiceConfig"
       type="NServiceBus.Config.RijndaelEncryptionServiceConfig, NServiceBus.Core"/>
   </configSections>
-  <RijndaelEncryptionServiceConfig 
-    Key="gdDbqRpQdRbTs3mhdZh9qCaDaxJXl+e6">
+  <RijndaelEncryptionServiceConfig Key="gdDbqRpQdRbTs3mhdZh9qCaDaxJXl+e6" KeyIdentifier="B">
```
SimonCropp: since this will render inline can we newline the attributes to reduce dodgy wrapping in the html?

Snippets/Snippets_5/Encryption/Configuration/encryption.config.xml
```diff
@@ -1,15 +1,17 @@
...
+    >
     <ExpiredKeys>
-      <add Key="abDbqRpQdRbTs3mhdZh9qCaDaxJXl+e6" />
+      <add Key="YWJEYnFScFFkUmJUczNtaGRaaDlxQ2FEYXhKWGwrZTY=" KeyIdentifier=";A" KeyFormat="Base64"/>
```
SimonCropp: since this will render inline can we newline the attributes to reduce dodgy wrapping in the html?

samples/encryption/basic-encryption/Version_3/Shared/ConfigureEncryption.cs
```diff
@@ -9,7 +9,8 @@ public RijndaelEncryptionServiceConfig GetConfiguration()
     {
         return new RijndaelEncryptionServiceConfig
                {
-                   Key = "gdDbqRpQdRbTs3mhdZh9qCaDaxJXl+e6"
+                   Key = "do-not-use-this-encryption-key!!",
```
SimonCropp: LOL

nservicebus/security/encryption.md
ramonsmits: That is explained in the section 'Using the same key with and without a key identifier'.

Snippets/Snippets_3/Encryption/Configuration/encryption.config.xml
ramonsmits: The semi-colon is used to specify multiple key identifiers for a single key.

In this example the semi-colon results in a 0 length key identifier so it will be used when decrypting data where no key identifier is set as mentioned in 'Using the same key with and without a key identifier'

nservicebus/security/encryption.md
ramonsmits: Regarding the upgrading behavior. That is also mentioned in the announcement specific for the hotfix.

ramonsmits:
> Also a section on naming of key identifiers would be good.

@DavidBoike Your wish is my command.

ramonsmits:
@DavidBoike @SimonCropp @indualagarsamy I'm pretty happy with the current doco.

PR #875, opened 2015-10-15
Sample fix for timeout to execute after 5 seconds and dependencies update

SeanFeldman:
Original code set timeout to 50,000 seconds :)


SeanFeldman:
@SimonCropp this could be a perfect smoke test code indeed

PR #876, opened 2015-10-16
Dependency updates for azure projects

SeanFeldman:
Also fixes a PS script - creates a container if such is missing for uploaded endpoints. If container is missing (new machine) and PS script fails and sample is becoming useless.


SeanFeldman:
@SimonCropp this makes all azure samples up-to-date on dependencies (NSB.Azure & Core).

SimonCropp:
nice

PR #880, opened 2015-10-17
Update transactions document for NServiceBus/issues/2752

colin-higgins:
Connects to https://github.com/Particular/NServiceBus/issues/2752

How is the wording here? I assumed we'd go with documenting current behavior and reference that previous versions behave differently in a note.

@indu @ramon

### DIFF
--- a/nservicebus/messaging/transactions.md
+++ b/nservicebus/messaging/transactions.md
@@ -25,7 +25,7 @@ Based on transaction handling mode, NServiceBus offers three levels of guarantee
 
 In this mode the transport in use does not attempt to wrap the receive operation in any kind of transaction. 
 
-Once a message has been received by an Endpoint, if the message processing fails because of an exception within the message handler, it will be put into the error queue. If there is a critical failure, including system or endpoint crash, the message is **permanently lost**. It will not be retried or added to the error queue.
+Once a message has been received by an Endpoint, if the message processing fails because of an exception within the message handler, it will be put into the error queue. There will be no first level or second level retries. If there is a critical failure, including system or endpoint crash, the message is **permanently lost**. It will not be retried or added to the error queue.
 
 WARNING: In version 5 and below, when transactions are disabled, no retries will be performed and messages will not be forwarded to the error queue in the event of any failure.
 
--- a/nservicebus/messaging/transactions.md
+++ b/nservicebus/messaging/transactions.md
@@ -21,13 +21,13 @@ It does not discuss the transaction isolation aspect which only applies to the p
 Based on transaction handling mode, NServiceBus offers three levels of guarantees with regards to message processing
 
 
-### Unreliable
+### Unreliable (Transactions Disabled)
 
 In this mode the transport in use does not attempt to wrap the receive operation in any kind of transaction. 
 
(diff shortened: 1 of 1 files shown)


nservicebus/messaging/transactions.md
```diff
@@ -23,9 +23,11 @@ Based on transaction handling mode, NServiceBus offers three levels of guarantee
 
 ### Unreliable
 
-In this mode the transport in use does not attempt to wrap the receive operation in any kind of transaction. Once the message has been received, it can't be put back into the queue. Should the message processing fail for any reason (including system crash), the message is **permanently lost**. 
+In this mode the transport in use does not attempt to wrap the receive operation in any kind of transaction. 
 
-NOTE: No retries will be performed in this mode and messages will not be forwarded to the error queue in case of failures
+Once a message has been received, if the message processing fails it should still be put into the error queue. If there is a critical failure, including system or endpoint crash, the message is **permanently lost**. 
+
+NOTE: Before version 6, no retries will be performed in this mode and messages will not be forwarded to the error queue in case of any failure.
```
indualagarsamy: This should be WARN instead of NOTE

SimonCropp: @indualagarsamy fixed

nservicebus/messaging/transactions.md
```diff
@@ -23,9 +23,11 @@ Based on transaction handling mode, NServiceBus offers three levels of guarantee
 
 ### Unreliable
 
-In this mode the transport in use does not attempt to wrap the receive operation in any kind of transaction. Once the message has been received, it can't be put back into the queue. Should the message processing fail for any reason (including system crash), the message is **permanently lost**. 
+In this mode the transport in use does not attempt to wrap the receive operation in any kind of transaction. 
 
-NOTE: No retries will be performed in this mode and messages will not be forwarded to the error queue in case of failures
+Once a message has been received, if the message processing fails it should still be put into the error queue. If there is a critical failure, including system or endpoint crash, the message is **permanently lost**.
```
indualagarsamy: I think it should be the other way around. Document current behavior as is with what we intend to do in V6. 
The behavior change should be in the upgrade guide. 
Since v6 isn't currently released, it might be confusing to someone who's on v5.

SimonCropp: > Document current behavior as is with what we intend to do in V6. 

Nope. if you do that and merge the PR then we have doco that is the wrong tense and becomes historically incorrect once we release v6. It should be document is a way that is unambiguous the version. eg `version 5 and below` and `Version 6 and higher`. this way we can merge and not need to revisit later.

SimonCropp: But @indualagarsamy is correct in that as it stands it is confusing as the behavior in various versions .

indualagarsamy: I like the idea of removing ambiguity by anchoring per version.

andreasohlund:
Side note: I've spiked a new api for v6 , that aims to clean this up

https://github.com/Particular/NServiceBus/pull/2987/files#diff-1359ca33231b96e8015249cd85e81584

Comments are welcome

nservicebus/messaging/transactions.md
SimonCropp: > should still be put into the error queue

why "should"?

Issue #886, opened 2015-10-27
Native integration with ASB document (+sample)


SeanFeldman:
Replaces  #810 

Documentation to provide more details about the scenario in general that [native integration sample](http://docs.particular.net/samples/azure/native-integration-asb/) provides

yvesgoeleven:
Reopening to add a page & update to v7

SimonCropp:
@SeanFeldman @yvesgoeleven why is this tied to v6?

SeanFeldman:
No idea @SimonCropp. The v7 sample is there, but is not up to date due to our changes we're still undergoing with the transport.

SimonCropp:
ok fair enough. if there is already a pending sample i guess the v6 label makes sense

PR #888, opened 2015-10-27
[WIP] Native integration with ASB

SeanFeldman:
Addresses #886 
Connects to Particular/docs.particular.net#886


SeanFeldman:
@Particular/azure please review

@Particular/documentation native integration with ASB

/cc @andreasohlund win that customer over! :)

samples/azure/native-integration-asb/sample.md
```diff
@@ -0,0 +1,85 @@
...
+
+Once the message is received by the NServiceBus endpoint, its contents will be presented.
+
+<!-- import NativeMessageHandler -->
+
+Things to note:
+
+ * The use of the `AzureServiceBus.ConnectionString` environment variable mentioned above.
+ * The use of `UseSingleBrokerQueue` prevents the Azure Service Bus transport individualizing queue names by appending the machine name.  
+ * Execute `Receiver` first to create destination queue `NativeSender` will need to send native messages.
```
DavidBoike: Why are these things important to note? Feels like if I were a first-time driver and you said "Things to note: that red sign over there" and neglected to say "because you should stop in front of it so we don't get run into."

SeanFeldman: @DavidBoike sample is not going to concentrate on `UseSingleBrokerQueue`, but it is needed to run properly. I draw attention, but don't go into details. For that we have doco.

`AzureServiceBus.ConnectionString` is noted to draw reader attention again. It's required to run the sample and also listed in Prerequisites section.

As for the `Receiver` to be executed first, explanation is there. Or did you mean for something else?

DavidBoike: Seems like a weird amalgamation of things.

* If it's "Hey be sure to notice as a takeaway" then I'd say create a summary with sentences.
* If it's "Hey make sure you did this" then that should be described better above.
* If it's "Hey here's a weird detail" (2nd bullet) then maybe that should just be a NOTE up above?

SeanFeldman:
@DavidBoike I turned it into a header to have a section of it's own as you suggested.

SeanFeldman:
@SimonCropp could you please review? Thank you.

SeanFeldman:
Thank you Simon

PR #891, opened 2015-10-27
Update iwanttorunwhenthebusstartsandstops.md

dvdstelt:
Throughout this page IWantToRunWhenTHEBusStartsAndStops is mentioned, but it's only correct in the code example, which remove the word "the" and just calls it IWantToRunWhenBusStartsAndStops. :)


SimonCropp:
@dvdstelt good catch. and sorry for any confusion

PR #895, opened 2015-10-29
Describe ServiceControl setup with multi DB SQL transport

SzymonPobiega:


SzymonPobiega:
@Particular/engineers please review.

DavidBoike:
Made some tiny changes. :+1:

SzymonPobiega:
@johnsimons can you do final review and merge and then reply on that desk case and point them to this new piece of documentation?

nservicebus/sqlserver/multiple-databases.md
```diff
@@ -65,3 +72,16 @@ and the following connection strings:
 <!-- import sqlserver-multidb-connectionstrings -->
 
 the messages sent to `billing` will go to database `Billing` on server `DbServerB` while the messages to `sales` will go to the database and server set by default i.e. `MyDefaultDB` on server `DbServerA`.
+
+
+### ServiceControl
+
+In order to use a single instance of ServiceControl for the whole system when there are multiple databases used for queues, the `audit` and `error` queues need to be present in a single dedicated database (the ServiceControl database). In addition to configuring the mapping between the user endpoints and connection strings (as described above), there must be a mapping that redirects `audit` and `error` pseudo-endpoints to the ServiceControl database.
```
SimonCropp: how about this

> ServiceControl requires the error and audit queues that it ingests from to be presented on a single dedicated database. In a system where there are multiple databases, used by various endpoints, this can be achieved by redirecting the queues using the following configuration endpoint configuration:

johnsimons:
@SzymonPobiega but with this setup, won't the transaction always escalate to 2PC ?

andreasohlund:
>won't the transaction always escalate to 2PC ?

Don't we explicitly make sure that there is no 2PC?

https://github.com/Particular/ServiceControl/blob/develop/src/ServiceControl/Bootstrapper.cs#L52

SzymonPobiega:
I assumed (my bad I did not check) SC does not use distributed transactions.

johnsimons:
> Don't we explicitly make sure that there is no 2PC?

@andreasohlund yes that code is for SC, but what I am refering to are the user endpoints

johnsimons:
> I assumed (my bad I did not check) SC does not use distributed transactions.

@SzymonPobiega yes that is true, I was refering to the user endpoints, since now at the user endpoint we are dequeing from database A but forwarding to database B, it is my understanding that SQL will excalate to 2PC in this situation?

SzymonPobiega:
@johnsimons it will escalate unless you are using outbox.

johnsimons:
@SzymonPobiega so how should we document this ? Because the user could be using:
```c#
configuration.Transactions()
                .DisableDistributedTransactions()
```

And event have MSDTC turned off.
This was actually one of the great benefits of the SQL transport where we "kind of" avoid DTC.

SimonCropp:
@SzymonPobiega is this a WIP? r u still working on it? anything blocking?

SzymonPobiega:
@SimonCropp applied your suggestion. Ready to merge.

PR #896, opened 2015-10-29
Outbox caveat no longer applicable

DavidBoike:
This caveat never made sense. You have to have your Outbox data in the same database as your business data, but if you have multiple endpoints, you have to have it in a *separate* data store?

More to the point, this refers to the bug in NHibernate that I fixed, and to NHibernate in general, which at the time was the only Outbox storage in town. Now it makes no sense.

Related to Particular/NServiceBus.NHibernate#117

/cc @indualagarsamy @SzymonPobiega @andreasohlund @SeanFeldman @johnsimons (Participants on bug issue)

### DIFF
--- a/nservicebus/outbox/index.md
+++ b/nservicebus/outbox/index.md
@@ -38,8 +38,7 @@ Here is a diagram how it all works:
 
 ## Caveats
 
-- Both the business data and deduplication data need to share the same database
-- If you're forwarding processed messages to other endpoints they need to use a different datastore since outbox records are keyed on the `MessageId`
+- Both the business data and deduplication data need to share the same database.
 - The Outbox is bypassed when sending messages "from outside" via the `IBus` interface (not from a message handler). The reason for this is lack of a driving force for repeated dispatching of the Outbox messages (which currently is the retry mechanism that applies only when handling messages).
 
 ## Using outbox with NHibernate persistence


SimonCropp:
@DavidBoike u refernece NH in the PR but that caveat is for all outbox implementations. Does it no longer applie to any other implmentation? Eg ravendb?

DavidBoike:
Oh, hold the phone.

I missed the word **forwarding** so now I think that actually does still make sense.

I opened this because I was looking for places in doco that the NH fix would necessitate changing, and this is all I found. Any others you can think of? If not, just close this PR.

SimonCropp:
@DavidBoike i am the wrong person to ask. Perhaps @johnsimons can chime in

johnsimons:
Sorry guys, I have no idea what
> If you're forwarding processed messages to other endpoints they need to use a different datastore since outbox records are keyed on the `MessageId`

means!
Is this talking about wanting to forward a message to another queue regardless of the outcome ?

SimonCropp:
ok looks like it is over to @andreasohlund  https://github.com/Particular/docs.particular.net/commit/26b69629596af6225ba3df81a46ae33de41f8046

andreasohlund:
>Is this talking about wanting to forward a message to another queue regardless of the outcome ?

Forward only happens when the message is successfully processed. But since the message ids are the same this will cause issues if both endpoints share the same storage since the outbox on the downstream endpoint will treat the forwarded message as a duplicate.

Does this make sense?

If yes this should be fixed now right? (since we include the endpoint name in the "key")

DavidBoike:
With NH this would be fixed. With Raven it still [could be broken](https://github.com/Particular/PlatformDevelopment/issues/484) if you provided the same DocumentStore for everything, but not if you use the default DB-per-endpoint model.

Does that meet the bar for removing this (somewhat confusing in the first place) bullet point?

andreasohlund:
>Does that meet the bar for removing this (somewhat confusing in the first place) bullet point?

Not sure but I'd say we need to fix RavenDB as well

SimonCropp:
@DavidBoike @andreasohlund been a week and no WIP on this. is this still worth doing? should it be deleted?

DavidBoike:
@bording is working on the Raven fix, which is barely broken to start with. NH is fixed. I say merge? Or wait for Brandon to finish then merge? @andreasohlund?

SimonCropp:
@DavidBoike a little confused. the PR makes no assertions about versions so there should be no logical reason to hold off. if the change is specific to a specific version and higher shouldnt that be called out in the content?

DavidBoike:
Well, I back ported the fix, so every *supported* version of NHibernate has no issue sharing outbox storage.

SimonCropp:
@DavidBoike ok understood. thanks. in that case @bording let us  know when you have finished your side of it.

bording:
The 3.0.6 and 2.2.3 Raven fixes were released today, so that should cover all of it now.

SimonCropp:
@bording thanks

PR #897, opened 2015-10-29
How NServiceBus can work with SQL Server Always-On

ramonsmits:


SimonCropp:
Content in this PR seems to have nothing to do with SQL always on

ramonsmits:
@SimonCropp You are right :-) weird... probably committed the wrong file.

SimonCropp:
this seems like a bad commit and hence a bogus PR. closing and deleting. please re-submit

PR #901, opened 2015-11-02
[WIP] Message Handler upgrade section

colin-higgins:
Added the upgrade guide section for Message Handlers.

I noticed that the signature also changed for Tim's implementation of the context.

Could I have some review on my text and snippets?

cc / @SeanFeldman @timbussmann @danielmarbach

### DIFF
--- a/Snippets/Snippets_6/UpgradeGuides/5to6/UpgradeMessageHandler.cs
+++ b/Snippets/Snippets_6/UpgradeGuides/5to6/UpgradeMessageHandler.cs
@@ -1,0 +1,36 @@
+﻿namespace Snippets6.Handlers
+{
+    using System.Threading.Tasks;
+    using NServiceBus;
+
+    #region 5to6-messagehandler
+    public class UpgradeMyAsynchronousHandler : IHandleMessages<MyMessage>
+    {
+        public async Task Handle(MyMessage message, IMessageHandlerContext context)
+        {
+            await SomeLibrary.SomeMethodAsync(message);
+        }
+    }
+
+    public class UpgradeMySynchronousHandler : IHandleMessages<MyMessage>
+    {
+        public Task Handle(MyMessage message, IMessageHandlerContext context)
+        {
+            // when no asynchronous code is executed in a handler Task.FromResult(0) can be returned
+            SomeLibrary.SomeMethod(message.Data);
+            return Task.FromResult(0);
+        }
+    }
+    #endregion
+
+    #region 5to6-bus-send-publish
+    public class SendAndPublishHandler : IHandleMessages<MyMessage>
+    {
+        public async Task Handle(MyMessage message, IMessageHandlerContext context)
+        {
+            await context.SendAsync(new MyOtherMessage());
+            await context.PublishAsync(new MyEvent());
+        }
+    }
+    #endregion
+}
--- a/nservicebus/upgrades/5to6.md
+++ b/nservicebus/upgrades/5to6.md
@@ -178,9 +178,10 @@ WARNING: Do not `return null` from your message handlers. We use the task result
 
 <!-- import 5to6-messagehandler -->
 
+
 ### Bus Send and Receive
 
-There is also a change in the parameters, giving you access to the `IMessageHandlerContext`, which wraps any methods that used to be called from `IBus`. Use the `IMessageHandlerContext` to send and publish messages.
+There is also a change in the parameters, giving you access to the `IMessageHandlerContext`, which provides the methods that used to be called from `IBus`. Use the `IMessageHandlerContext` to send and publish messages.
 
 <!-- import 5to6-bus-send-publish -->
 
--- a/nservicebus/upgrades/5to6.md
+++ b/nservicebus/upgrades/5to6.md
@@ -170,6 +170,20 @@ We've separated the callback support from the NServiceBus core. This means that 
 
 The differences in the API are fully covered in [handling responses on the client side](/nservicebus/messaging/handling-responses-on-the-client-side.md).
 
+## Message handlers
+
+The handler method on `IHandleMessages<T>` now returns a Task. In order to leverage async code, add the async keyword to your handler method and use `await` for async methods. In order to convert your synchronous code add `return Task.FromResult(0);` to the end of your handler methods. 
+
+WARNING: Do not `return null` from your message handlers. We use the task result internally and `null` will result in errors. You may mark your `Handle` methods as `async` to enforce this, which would result in `await Task.FromResult(0);`.
+
+<!-- import 5to6-messagehandler -->
+
+### Bus Send and Receive
+
+There is also a change in the parameters, giving you access to the `IMessageHandlerContext`, which wraps any methods that used to be called from `IBus`. Use the `IMessageHandlerContext` to send and publish messages.
+
+<!-- import 5to6-bus-send-publish -->
+
 
 ## Message handler ordering
 
--- a/Snippets/Snippets_5/Snippets_5.csproj
+++ b/Snippets/Snippets_5/Snippets_5.csproj
@@ -603,6 +603,7 @@
     <Compile Include="UpgradeGuides\4to5\Scheduler.cs" />
     <Compile Include="UpgradeGuides\4to5\Upgrade.cs" />
     <Compile Include="UpgradeGuides\5to6\Upgrade.cs" />
+    <Compile Include="UpgradeGuides\5to6\UpgradeMessageHandler.cs" />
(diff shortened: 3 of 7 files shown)


nservicebus/upgrades/5to6.md
```diff
@@ -166,6 +166,13 @@ We've separated the callback support from the NServiceBus core. This means that
 
 The differences in the API are fully covered in [handling responses on the client side](/nservicebus/messaging/handling-responses-on-the-client-side.md).
 
+## Message handlers
+
+The handler method on `IHandleMessages<T>` now returns a Task. In order to convert your synchronous code simply add `return Task.FromResult(0);` to the end of your handler methods. In order to leverage async code, add the async keyword to your handler method and use await for async methods.
```
danielmarbach: `await` in quotes?

Might also mention `return Task.CompletedTask`?

nservicebus/upgrades/5to6.md
```diff
@@ -166,6 +166,13 @@ We've separated the callback support from the NServiceBus core. This means that
...
+
+There is also a change in the parameters, giving you access to the `IMessageHandlerContext`, which wraps your `IBus` and various metadata.
```
timbussmann: I think this is the more important part of the upgrade (which isn't represented in the sample code. Moving all bus operations to the context will be the biggest "code change". In most cases, just adding await should do the trick. Why not using the following structure:
1. new handler context
2. using await
3. for synchronous code, use FromResult(0)/CompletedTask

timbussmann:
after a discussion with @hmemcpy and @danielmarbach I think we should add a "warning" statement about never returning null for methods returning Tasks (as a general guideline, not specific to handlers).

SeanFeldman:
@timbussmann should we also mention an option of marking `Task` returning methods as async to prevent null returning as @danielmarbach suggested?

danielmarbach:
I still think returning null doesn't need to be documented since null is an
invalid value by design for futures (although the language allows it, but
that is a sep. issue ;) )

-- 
*#NSBcon 2015 • The "All About NServiceBus" conference is back *
*• Learn more <http://go.particular.net/l/27442/2015-08-21/5d4wzp> →*

hmemcpy:
Warning: going to quote some blogs!

I believe that our ultimate goal should be to help our users [fall into the pit of success](https://blog.codinghorror.com/falling-into-the-pit-of-success/), by making it very difficult for them to make a mistake.

Unfortunately, it's still very easy to make such mistakes, neither the language nor the tooling prevents them from happening.

In my particular case, I implemented `IMessageHandler` one linked file between 4 projects, the fourth being NSB6, so it required a return value for Task. I typed `return null` just to shut the compiler up, and forgot about it.
When I ran it, I got tons of exceptions deep in NSB, looking like this:

```
2015-11-09 15:39:31.424 INFO  NServiceBus.FirstLevelRetriesBehavior First Level Retry is going to retry message 'e043d673-4953-4c1e-b07e-a54b01020a5b' because of an exception:
System.NullReferenceException: Object reference not set to an instance of an object.
   at NServiceBus.InvokeHandlerTerminator.<Terminate>d__0.MoveNext()
--- End of stack trace from previous location where exception was thrown ---
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.ConfiguredTaskAwaitable.ConfiguredTaskAwaiter.GetResult()
   at NServiceBus.BehaviorChain.<InvokeNext>d__3.MoveNext()
   ...
```

Now, obviously this was my own [boneheaded exception](`http://blogs.msdn.com/b/ericlippert/archive/2008/09/10/vexing-exceptions.aspx), but still, this is an innocent mistake that the user can make. C# allows returning a null for task, and nothing prevented me from doing so. Now, I don't think there's a good solution for checking for `null` being returned instead of a Task, and I agree with @danielmarbach - we shouldn't (mustn't!) check for nulls everywhere, however we should make the best effort to avoid the users making this mistake in the first place.

Until we can have better tooling-assisted migration (either some conversion tool and/or Roslyn analyzers to detect those things), the least we can do is specify this in the documentation. Not everyone is thoroughly familiar with `async/await` and how it behaves, and for many people, the "upgrade" process is going to be trying to "fix" all the compiler errors, which will cause those types of mistake.

I also love @colin-higgins's suggestion in the chat, that we can provide a sync wrapper for people who want to use sync. THIS can be a sep. issue :)

hmemcpy:
Also, you cannot force the `await` keyword on the interface, so there's really nothing that can be done (unless providing an abstract class)

nservicebus/upgrades/5to6.md
hmemcpy: Also, please avoid using the word **simply**.

Snippets/Snippets_6/UpgradeGuides/5to6/UpgradeMessageHandler.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    using System.Threading.Tasks;
+    using NServiceBus;
+
+    #region 5to6-messagehandler
+    public class UpgradeMySynchronousHandler : IHandleMessages<MyMessage>
+    {
+        public Task Handle(MyMessage message, IMessageHandlerContext context)
+        {
+            //use this example for synchronous code
```
SimonCropp: what does this comment convey?

SimonCropp: perhaps

    // when no asynchronous code is executed in a handler Task.FromResult(0) can be returned

nservicebus/upgrades/5to6.md
SimonCropp: agreed remove `simply`. 

also 

> simply add `return Task.FromResult(0);` to the end of your handler methods

is misleading since they need to add a context parameter.

I would prefer to see this broken down into a few sub headings. eg "new signature", "now async" and "using context instead of IBus"

timbussmann:
> I still think returning null doesn't need to be documented since null is an
invalid value by design for futures

I still think it's worth reminding readers of this fact since it comes with no costs.

danielmarbach:
So let's add that warning since this is the lower friction than adding null checks in the codebase at *every*  extension point

colin-higgins:
Still need to add proper story for the context usage. (This is a note for myself)

colin-higgins:
@SimonCropp @hmemcpy @timbussmann @danielmarbach @SeanFeldman 

Up for review again?

nservicebus/upgrades/5to6.md
```diff
@@ -166,6 +166,19 @@ We've separated the callback support from the NServiceBus core. This means that
 
 The differences in the API are fully covered in [handling responses on the client side](/nservicebus/messaging/handling-responses-on-the-client-side.md).
 
+## Message handlers
+
+The handler method on `IHandleMessages<T>` now returns a Task. In order to convert your synchronous code add `return Task.FromResult(0);` to the end of your handler methods. In order to leverage async code, add the async keyword to your handler method and use `await` for async methods.
```
SeanFeldman: Would it make sense to invert the order `async` and `return...` so that it goes top down?

nservicebus/upgrades/5to6.md
```diff
@@ -166,6 +166,19 @@ We've separated the callback support from the NServiceBus core. This means that
...
+
+WARNING: Do not `return null` from your message handlers. We use the task result internally and `null` will result in errors.
+
+<!-- import 5to6-messagehandler -->
+
+### Bus Send and Receive
+
+There is also a change in the parameters, giving you access to the `IMessageHandlerContext`, which wraps your `IBus` and various metadata. Use the `IMessageHandlerContext` to send and publish messages.
```
SeanFeldman: "wraps your `IBus`" or "allows operations on the bus"?

SimonCropp:
@colin-higgins  can you rebase this

nservicebus/upgrades/5to6.md
colin-higgins: Will more people use async or will more people just drop in their sync code? @SeanFeldman

SimonCropp: more people should be using async. web call, disk io, db calls. so i think have the async version first

nservicebus/upgrades/5to6.md
SimonCropp: ibus doesn really exist any more

colin-higgins:
@SimonCropp This is rebased and I made some more changes

PR #911, opened 2015-11-04
Clarification about heartbeat interval and grace period settings

SeanFeldman:
There are 2 settings that can affect each other if not configured appropriately. Since settings are not in the same doco/configuration file, adding that information in doco.

1. SC `ServiceControl/HeartbeatGracePeriod`
1. Heartbeat plugin `heartbeat/interval`

@Particular/servicecontrol please review

### DIFF
--- a/servicecontrol/creating-config-file.md
+++ b/servicecontrol/creating-config-file.md
@@ -140,6 +140,8 @@ Default: `00:00:40` (40 secs)
 
 When configuring heartbeat grace period, make sure it is synchronized with heartbeat interval defined by plugin or overwritten by [`heartbeat/interval`](/servicecontrol/plugins/heartbeat.md#configuration-heartbeat-interval) setting.
 
+Note: When monitoring multiple endpoints, ensure that heartbeat grace period is larger than any individual heartbeat interval set by the endpoints.
+
 
 #### ServiceControl/MaximumMessageThroughputPerSecond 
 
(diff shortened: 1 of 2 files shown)


johnsimons:
:+1:  to what @gbiellem said.

Except I would actually say "it must be greater" otherwise it can trigger the heartbeat missing flag, and since we are build on top of queueing it will take time to get these heartbeats messages.
I would probably also stress to not make these setting too short otherwise we fill up SC with heartbeats.

Finally it should also be pointed out that if they are not actively monitoring, there is no point of actually using heartbeats!

SeanFeldman:
Interesting scenario is when multiple endpoints have different intervals defined. In that case, Grace period should (or must, according to @johnsimons) be greater then the greatest of them all. 

So are we good to merge?

SimonCropp:
> So are we good to merge?  

but it still says `synchronized`

gbiellem:
> Interesting scenario is when multiple endpoints have different intervals defined. In that case, Grace period should (or must, according to @johnsimons) be greater then the greatest of them all. 

Yep I agree

SeanFeldman:
@gbiellem I know this is common sense, but should I add it as well?

gbiellem:
@SeanFeldman  I'd say add it.  It's would be easy to overlook

SeanFeldman:
Better safe than sorry, I hear you @gbiellem 
If no other comments, let's merge this.

SimonCropp:
can we remove `synchronized`

PR #918, opened 2015-11-11
Doco on the max 24 h retry limit

andreasohlund:
@SimonCropp as promised


SimonCropp:
@andreasohlund is it worth doing the math and if it is possible, based on counts of SLR and FLR and time increase, that we may exceed 24 hours throw an exception at startup?

andreasohlund:
@SimonCropp makes sense, raise an issue on the core?

PR #921, opened 2015-11-13
Added ServicePulse commandline help

ramonsmits:


SimonCropp:
my concern is keeping it up-to-date as the versions evolve. it is not like the nuget based apis where we can update/version and evolve

gbiellem:
> my concern is keeping it up-to-date as the versions evolve.

:+1: That's where i was coming from as well.  We've had the same issue with SC and NSB Powershell doco.

johnsimons:
cmd line help is available via the popular -? or --help, no need to create a static version of it in my opinion

SimonCropp:
ok based on the comments i am closing this

ramonsmits:
@gbiellem Yes, mentioning that they can use `-?` is fine, but that isn't indexed by google. Yes, it can get out of sync but is that a reason to not put it in the doco? Isn't that an issue for all doco?

ramonsmits:
@gbiellem @SimonCropp Reason for adding this is that a customer wanted to run multiple versions of ServiceControl on a single machine.

And yes, if you run it non elevated you get errors. That also is not really strange as that is expected.

ramonsmits:
@gbiellem What you are addressing is that basically the commandline argument doco itself is broken. If that is the case than that is different from this PR.

ramonsmits:
@gbiellem @SimonCropp Another reason for this doco is that:

1. Arguments as case sensitive
2. Unknown arguments are not resulting in commandline validation errors

I don't really have an issue with 1 is we would have validation.

Due to case sensitivity the customer was not able to manually install the instance as a different windows service.

gbiellem:
@ramonsmits case sensitivity sounds like a bug to me. Agree the command
line doco is broken and so are some of options in my opinion.

ramonsmits:
So @gbiellem should I create two issues in SP:

1. Validate commandline arguments.
2. Fix conflicting argument with constants.

ramonsmits:
Btw, I still see value in just having the output in the doco.

Its also because we don't have a KB. If we had that I could have a KB article.

gbiellem:
@ramonsmits i'd prefer to write an article on installing a second instance
which sounds like what the customer wanted. Much like we have for
servicecontrol

gbiellem:
And I m offering to write it :)

ramonsmits:
@gbiellem You have my vote :-) still, what to do about argument validation?

gbiellem:
Raise a bug - i' ll fix it

ramonsmits:
See https://github.com/Particular/ServicePulse/issues/228

gbiellem:
Thanks @ramonsmits

Issue #914, opened 2015-11-06
move home page content over to a bootstrap layout


SimonCropp:
to reduce the friction in making modifications to the html

SimonCropp:
@petersgiles @colin-higgins tracking issue for home page changes

udidahan:
is this urgent?

SimonCropp:
@udidahan urgent, no. necessary, yes.

@colin-higgins @petersgiles can one of u give an estimate on the effort?

petersgiles:
It shouldn't cost a lot engineering-wise. We will have to tidy up any inline css but it's already BS3 so there is no difficult migration to consider. 

The only thing is its a little harder to iterate over because of the fragmented way it gets glued together at build time. But @SimonCropp has already suggested a way to make that a little easier. It's a small effort with a medium impact and it fits in with our philosophy of continuous improvement. Optimistically I'd say 1 full day, pessimistically no more than 2-3 days with the odd weird bug on mobile devices and old browsers.

If it turns there are serious design changes required to make it more mobile friendly I think we might have to start a new issue.

SimonCropp:
>  pessimistically no more than 2-3 days with the odd weird bug on mobile devices and old browsers. If it turns there are serious design changes required to make it more mobile friendly I think we might have to start a new issue.

Just to be clear we are not mobile friendly ATM. so  2-3days "making it mobile friendly" would be considered outside the scope of this issue. even though we should do it

petersgiles:
> "making it mobile friendly" would be considered outside the scope of this issue. 

Ok well if we're not testing in that space my answer would be it could be done pretty quick then

colin-higgins:
Yeah, agreed with @petersgiles. Tidying things up would take about a day. We could make a fair bit of progress toward being mobile friendly as well. 

The mobile friendliness could be accomplished with a bunch of incremental changes which would be fairly low effort as well, though as you said it is out of scope.

Low effort / low risk / lowish impact

SimonCropp:
@colin-higgins 

> mobile friendliness could be accomplished with a bunch of incremental changes

yeah that was the conversation pete and I had last week.

udidahan:
OK.

petersgiles:
Is it the page that appears at http://docs.particular.net/ or is it the layout for all the pages that are created by DocsWeb?

SimonCropp:
@petersgiles the contents of this https://github.com/Particular/docs.particular.net/blob/master/home.md

[...]

PR #924, opened 2015-11-16
Update home.md

petersgiles:
Work in Progress

relates to : https://github.com/Particular/docs.particular.net/issues/914

I've refactored a few things, mostly inserted a bunch of rows and removed some divs that were resetting floating.

There's a problem with row background-color when the screen shrinks that I need to fix and the bottom bits should really be cards in BS4 but that's a way off yet.

### DIFF
--- a/home.md
+++ b/home.md
@@ -29,7 +29,6 @@
 .block.middle a,
 .productcolumn a {
   color: inherit;
-  white-space: nowrap;
 }
 
 .block {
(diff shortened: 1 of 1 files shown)


SimonCropp:
merged manually here https://github.com/Particular/docs.particular.net/commit/47d6d92118e89b7a056d597c0999eb32be429ae0

PR #929, opened 2015-11-16
Removing Obsolete EndPoints

petersgiles:
relates to https://github.com/Particular/ServicePulse/issues/197
relates to https://github.com/Particular/ServicePulse/issues/162

Just a quick summary how to remove obsolete endpoints not sure if the
tone is right, but the info is there

Might want to link to this from
http://docs.particular.net/servicepulse/intro-endpoints-heartbeats


petersgiles:
@andreasohlund as per your request.
@SimonCropp is this ok?

petersgiles:
This helps customers today. Happy to remove once the feature to manage this via Monitoring is delivered.

@johnsimons & @gbiellem  If we need to include a bigger bolder warning label please advise. Do you have objections to this being included on the docs site NOW? If so what are they?

johnsimons:
@petersgiles yes I am opposed to documenting this.
I think exposing an internal database is a risky move.
For now to mitigate this issue I think we would be better to instead provide a script for them to run.
Eg `Remove-ServiceControlExternalIntegrationSubscription http://localhost:33333 mysubscriptionname`

gbiellem:
@johnsimons :+1:

PR #937, opened 2015-11-19
[WIP] Upgrade to v6 latest unstable

danielmarbach:
@SzymonPobiega @janovesk can you guys fix the routing samples in this PR?

@SzymonPobiega there is also the issue of the `MsmqPersistence` not being available.


SzymonPobiega:
@danielmarbach why 1096 and not the latest?

danielmarbach:
Because when I did it that was the latest on myget

SzymonPobiega:
Rebased, updated to latest and cleaned most of the errors. Only request-reply and missing msmq storage are left.

PR #942, opened 2015-11-25
Sequence Diagram,  No data available

mikeminutillo:
Documentation Page for the "No data" message in ServiceInsight https://github.com/Particular/ServiceInsight/issues/515


johnsimons:
Sorry @mikeminutillo not following the 2nd paragraph

mikeminutillo:
@johnsimons which paragraph? The one starting with "Secondly"? or the one starting with "ServiceInsight Visualization are based on"?

johnsimons:
The one starting with "secondly..."

mikeminutillo:
I'll have another go at it. It boils down "there's no ConversationId header in NSB v3 or below"

mikeminutillo:
@johnsimons better?

johnsimons:
"The messages still appearing in the list are being read from an in-memory ServiceInsight cache.", not really, I think is just easier to tell them to refresh the data.

Also, are we going to mention the whole thing about errors are kept forever hence there are conversations that really the only thing they show is the error message and everything else is gone becuase it has expired ?

SimonCropp:
@johnsimons did u intend for this to appear in the menu somewhere?

johnsimons:
not sure it needs to be discoverable from the menu

PR #943, opened 2015-11-25
How to shift SqlTransport audit log to long term storage

petersgiles:
Instructs users on how to shift SqlTransport audit logs data into long term storage 
/cc @SimonCropp

### DIFF
--- a/nservicebus/sqlserver/operations-scripting.md
+++ b/nservicebus/sqlserver/operations-scripting.md
@@ -92,13 +92,14 @@ snippet:sqlserver-return-to-source-queue
 snippet:sqlserver-return-to-source-queue-usage
 
 
-### How to shift SqlTransport audit log to long term storage
+### Archiving SqlTransport audit log to long term storage
 
-There are several ways to achieve this. If you have access to a Database Administrator talk to them about techniques like [Table Partitioning](https://technet.microsoft.com/en-us/library/ms188730.aspx) and [Snapshot Replication](https://technet.microsoft.com/en-us/library/ms151832.aspx).
+There are several ways to achieve this including using techniques like [Table Partitioning](https://technet.microsoft.com/en-us/library/ms188730.aspx) and [Snapshot Replication](https://technet.microsoft.com/en-us/library/ms151832.aspx). In this example [BCP utility](https://msdn.microsoft.com/en-AU/library/ms162802.aspx) will be used.
 
-If you're just looking for some simple techniques to clear down your audit queue try this
 
-In SQL Management Studio create an audit_archive table with this SQL script.
+#### Create helper "archive" table
+
+Create an audit_archive table with this SQL script.
 
 ```
 CREATE TABLE [dbo].[audit_archive](
@@ -111,10 +112,12 @@ CREATE TABLE [dbo].[audit_archive](
 	[Body] [varbinary](max) NULL,
 	[RowVersion] [bigint] NOT NULL
 ) ON [PRIMARY] TEXTIMAGE_ON [PRIMARY]
-
 ```
 
-Next run this script. 
+
+#### Move records to archive table
+
+This script moves the contents of the audit table into `audit_archive` table.
 
 ```
 DELETE FROM [dbo].[audit] 
@@ -122,12 +125,18 @@ OUTPUT [deleted].*
 INTO [dbo].[audit_archive] 
 ```
 
-You could run this with a scheduled job if you want to clear down the archive regularly. The script moves the contents of the audit table into audit_archive table you created earlier. You also could wrap this in a transaction.
+This can be run with a scheduled job to clear down the archive regularly. 
 
-Once that query completes you are ready to archive to disk. In a command prompt use the [bcp utility](https://msdn.microsoft.com/en-AU/library/ms162802.aspx) to create an archive on disk.
+
+#### Execute BCP
+
+Once that query completes the records can be archived to disk. In a command prompt use the BCP to create an archive on disk.
 
 `bcp samples.dbo.audit_archive out archive.csv -c -q -T -S .\SQLExpress`
 
-You will still have to clear the audit records you just archived so go back to SQL Management Studio and in the samples database.
+
+#### Truncate the archive table
+
+You will still have to clear the audit records using the following script.
 
 `TRUNCATE TABLE  [dbo].[audit_archive] ;`
--- a/nservicebus/sqlserver/operations-scripting.md
+++ b/nservicebus/sqlserver/operations-scripting.md
@@ -90,3 +90,44 @@ snippet:sqlserver-return-to-source-queue
 ### Using the retry helper methods
 
 snippet:sqlserver-return-to-source-queue-usage
+
+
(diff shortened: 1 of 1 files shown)


nservicebus/sqlserver/operations-scripting.md
```diff
@@ -90,3 +90,44 @@ snippet:sqlserver-return-to-source-queue
...
+
+### How to shift SqlTransport audit log to long term storage
+
+There are several ways to achieve this. If you have access to a Database Administrator talk to them about techniques like [Table Partitioning](https://technet.microsoft.com/en-us/library/ms188730.aspx) and [Snapshot Replication](https://technet.microsoft.com/en-us/library/ms151832.aspx).
+
+If you're just looking for some simple techniques to clear down your audit queue try this
+
+In SQL Management Studio create an audit_archive table with this SQL script.
+
+```
```
andreasohlund: @SimonCropp do we create snippets for this or yagni?

nservicebus/sqlserver/operations-scripting.md
```diff
@@ -90,3 +90,44 @@ snippet:sqlserver-return-to-source-queue
...
+
+Next run this script. 
+
+```
+DELETE FROM [dbo].[audit] 
+OUTPUT [deleted].*
+INTO [dbo].[audit_archive] 
+```
+
+You could run this with a scheduled job if you want to clear down the archive regularly. The script moves the contents of the audit table into audit_archive table you created earlier. You also could wrap this in a transaction.
```
andreasohlund: Flesh out why a transaction would be needed? And what risk there would be to not have one?

nservicebus/sqlserver/operations-scripting.md
SimonCropp: @andreasohlund is it the same for every version of the sql transport? and will it continue to be so?

andreasohlund: No we have plans to change the storage format in the future cc @SzymonPobiega

SimonCropp: @andreasohlund if that is the case then a snippet is not necessary

SzymonPobiega: @andreasohlund everything is hands of @tmasternak and @weralabaj :-)

nservicebus/sqlserver/operations-scripting.md
weralabaj: Would it also make sense to mention that if you have large number of audit messages then it's better to do it in batches, not DELETE *? ;)

nservicebus/sqlserver/operations-scripting.md
weralabaj: @SzymonPobiega As of now, I don't see any upcoming changes regarding storage format (if we talk about table schema here).

SimonCropp:
Why do we need an archive table to get it to disk. This is not optimal since the disk space required for all the audits could be significant, and sql disk space is not necessarily cheap

nservicebus/sqlserver/operations-scripting.md
tmasternak: Just to clarify what @weralabaj pointed out. We don't think that changes to the format are necessary for v3. There is however a plan to switch to stored procedures interface in the future but I don't think we are in position to make any good estimations on when that will happen.

nservicebus/sqlserver/operations-scripting.md
```diff
@@ -90,3 +90,44 @@ snippet:sqlserver-return-to-source-queue
 ### Using the retry helper methods
 
 snippet:sqlserver-return-to-source-queue-usage
+
+
+### How to shift SqlTransport audit log to long term storage
```
weralabaj: Why not just "archiving audit log"?

nservicebus/sqlserver/operations-scripting.md
petersgiles: > Flesh out why a transaction would be needed

I thought about that but I didn't want the information to turn into a tutorial on how to use SQL. I think if people have got their hands grubby at this level there is enough info here to guide them but not so much that we prescribe a solution. Remember I do suggest that they talk to their DBA before doing all this.

> better to do it in batches, not DELETE

The SQL command here is essentially a move command if you wanted to batch by only shifting the oldest 1000 records you could do so by adjusting the DELETE clause but then you would introduce more table scanning etc.

The idea with including this information is to let people know it could be done but not prescribe a full solution. Like you suggest each system is different with different issues around disk management and scale.

petersgiles:
> Why do we need an archive table to get it to disk. 

This method allows you to archive a subset of your audits if you wanted or gives you the freedom to batch the process if that what you want. Doing subsets in BCP is way more complicated. 

Also that audit_archive table lets you review what you're taking without any NET gain to disk space, remember the instructions have you deleting from audit at the same time. I instruct the user to clear down the table which you would need to do anyway when using BCP, This way it is way safer that's all.

nservicebus/sqlserver/operations-scripting.md
petersgiles: This instruction is particular to SQL...there's another issue for archiving MSMQ audits so there would need to be a distinction in the title. The 'long term storage' is suggesting that you are not really ever going to put them back in your production database.

You are right though, the title is a little clumsy sounding.

petersgiles:
I'm just going to add our slack conversation here for future context

> 
simon [09:57] 
the copy isnt the problem. the problem is the approach recommended. people are likely to have a large amount of data in the audit log. forcing them to duplicate tha in another table just to get it to disk is not really great
>
pete [09:57] 
its not a duplicate…read my comments on the pull
>
​[09:58] 
…it simply staging to remove complexity with a only BCP approach
>
simon [10:00] 
ok so we move to the temp table?
>
pete [10:04] 
yep pretty much, its all done nice and cleanly…no net gain of data. when you archive with bcp you need to go back to sql and clean up anyway so there’s no difference. the method i show makes that so much easer
>
simon [10:06] 
ok will give it one mroe once over and merge
>
pete [10:08] 
thanks
>
simon [10:13] 
no TRUNCATE is needed on the audit table after the move?
>
pete [10:14] 
no
>
​[10:15] 
truncate kills all the data…the dba might want to put in a shrink database in their maintenance plan
>
simon [10:15] 
will bcp always be in the path?
>
pete [10:16] 
it is in the path of the SQL console and is typically available in other consoles after installing of SQL Management studio
>
pete [10:22] 
DBCC SHRINKDATABASE can take a long time to run. often dbas won’t bother unless they’re planning to physically move the database. you basically count on it growing back into the space again…the management of the database sizing is something that is particular to each situation so its difficult to give specific advice. A regular dev would probably truncate and shrink without really considering the overall performance impact because their measure of success is having plenty of free space. but if you leave it ‘roomy’ then some other jerk isn’t going to reclaim your disk space without telling you. horses for courses really
>
​[10:22] 
also the empty space will not slow down performance but having to grow the file again over time will.
>
​[10:23] 
and shrinking can fragment indexes…so you’ll need to reindex stuff, another performance hit

Issue #809, opened 2015-08-31
Inappropriate content in Messages, Events and Commands


adamralph:
http://docs.particular.net/nservicebus/messaging/messages-events-commands

> It is important to include the .Namespace != null; otherwise a null reference exception will occur during the type scanning.

This sentence doesn't seem to relate to anything in the page.

PR #944, opened 2015-11-25
Inappropriate content in Messages, Events and Commands

petersgiles:
fixes: https://github.com/Particular/docs.particular.net/issues/809

/cc @SimonCropp

### DIFF
--- a/nservicebus/messaging/messages-events-commands.md
+++ b/nservicebus/messaging/messages-events-commands.md
@@ -78,6 +78,5 @@ We currently have conventions that can identity:
 - [Express messages](/nservicebus/messaging/non-durable-messaging.md)
 - [TimeToBeReceived](/nservicebus/messaging/discard-old-messages.md)
 
-Note: It is important to include the `.Namespace != null`; otherwise a null reference exception will occur during the type scanning.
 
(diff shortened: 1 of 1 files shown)


nservicebus/messaging/messages-events-commands.md
```diff
@@ -78,6 +78,5 @@ We currently have conventions that can identity:
 - [Express messages](/nservicebus/messaging/non-durable-messaging.md)
 - [TimeToBeReceived](/nservicebus/messaging/discard-old-messages.md)
 
-Note: It is important to include the `.Namespace != null`; otherwise a null reference exception will occur during the type scanning.
```
andreasohlund: "if any of the scanned assemblies contains types without a namespace"

andreasohlund: Should we also note that using conventions like "Messages" could include types from other unrelated assemblies (eg. the rabbitmq.client has such types) and the fix is the include something like `t.Assembly.Name.StartsWith("MyCorp")`?

SimonCropp:
fixed in https://github.com/Particular/docs.particular.net/commit/de29ab733607800c60954429e32f86d55c0f3cf4

PR #947, opened 2015-11-26
Updated send-a-message page to V6

SzymonPobiega:
@Particular/tf-asyncawait plase review

### DIFF
--- a/Snippets/Snippets_3/BasicUsageOfIBus.cs
+++ b/Snippets/Snippets_3/BasicUsageOfIBus.cs
@@ -6,12 +6,30 @@
     {
         void Send()
         {
-            IBus bus = null;
-
             #region BasicSend
+            IBus bus = CreateAndStartTheBus();
+
             bus.Send(new MyMessage());
             #endregion
         }
+
+        IBus CreateAndStartTheBus()
+        {
+            throw new System.NotImplementedException();
+        }
+
+        #region SendFromHandler
+
+        public class MyMessageHandler : IHandleMessages<MyMessage>
+        {
+            public IBus Bus { get; set; }
+
+            public void Handle(MyMessage message)
+            {
+                Bus.Send(new OtherMessage());
+            }
+        }
+        #endregion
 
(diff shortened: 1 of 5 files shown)


Snippets/Snippets_3/BasicUsageOfIBus.cs
```diff
@@ -6,13 +6,31 @@ public class BasicUsageOfIBus
     {
         void Send()
         {
-            IBus bus = null;
-
             #region BasicSend
+            IBus bus = CreateAndStartTheBus();
```
timbussmann: in the v5 example you used the real config api to create the IBus instance which I like much better. Any reason to not do the same in v3/4?

nservicebus/messaging/send-a-message.md
```diff
@@ -6,12 +6,17 @@ redirects:
 - nservicebus/how-do-i-send-a-message
 ---
 
-To send a message, use the `Send` method on the `IBus` interface, passing as the argument the instance of the message to deliver:
+Prior to V6 sending a message involved using the `Send` method on the `IBus` interface, passing as the argument the instance of the message to deliver. In V6 the equivalent of the `IBus` is the `IBusContext` family of interfaces.
```
timbussmann: I somehow don't like that we start of with version difference. I don't think the renaming of interfaces changes much in the scope of sending a basic message. So could we put the interface name and version a little bit in the background? Something like "[...] using the `Send` method on the `IBus`(pre v6)/`IBusContext`(v6) interface[...]". Maybe skip the version altogether an let the samples speak for themselves?

SeanFeldman: @timbussmann like this: "Message sending involves using `Send` operation that take an argument of the message to be delivered. Refer to samples for details." ?

SzymonPobiega:
@timbussmann @SeanFeldman I've updated according to your suggestions.

SimonCropp:
thanks @SzymonPobiega

PR #952, opened 2015-11-27
[WIP] Set headers in v6

timbussmann:
needs to update nsb core reference to latest unstable as soon the header changes are merged.


andreasohlund:
:+1:

PR #960, opened 2015-12-01
Update Reply doco page

timbussmann:
updating snippets on this page: http://docs.particular.net/nservicebus/messaging/reply-to-a-message

### DIFF
--- a/Snippets/Snippets_6/Handlers/ReplyingHandler.cs
+++ b/Snippets/Snippets_6/Handlers/ReplyingHandler.cs
@@ -1,0 +1,38 @@
+﻿namespace Snippets6.Handlers
+{
+    using System.Threading.Tasks;
+    using NServiceBus;
+
+    public class ReplyingHandler : IHandleMessages<RequestDataMessage>
+    {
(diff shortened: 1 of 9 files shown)


Snippets/Snippets_6/Handlers/ReplyingHandler.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            //Create a response message:
+            var response = new DataResponseMessage
+            {
+                DataId = message.DataId,
+                String = message.String
+            };
+
+            //Underneath the covers, Reply sends a new message to the return address on the message being handled.
+            await context.Reply(response);
```
SzymonPobiega: Why no 'Send` equivalent?

timbussmann: I had one first, the I checked the implementation and realized that Reply is a completely different implementation so I removed the "equivalent".

SzymonPobiega: :+1:

PR #966, opened 2015-12-04
Revised controlling subscriptions page for V6

SzymonPobiega:


SimonCropp:
Why is publishers a method?

SzymonPobiega:
Same as `UnicastRouting` - an extension defined by a feature, not directly on the `BusConfiguration`.

johnsimons:
@SzymonPobiega so this api is to subscribe to events programatically ?
I'm asking because the all `.Publishers()` is throwing me out.

Also if it is, then what is the diference between this and `bus.Subscribe(...)` ?

SzymonPobiega:
@johnsimons no. The `Publishers()` API is for programatically define the publishers of events. In the message-endpoint mappings you used the same section to configure both send destinations and publishers. In the code API we wanted to make it explicit. The destinations of `Subscribe()` calls are defined in the `Publishers` collection, rather than in the `UnicastRoutingTable`.

johnsimons:
So @SzymonPobiega are you saying that publishers have to programatically sefine what they are publshing ?

SzymonPobiega:
@johnsimons no. If you don't want to use MEMs, in the subscriber you need to somehow define who the publisher of each event is. And you do it via the `Publishers()` collection.

johnsimons:
What is MEM ?

SzymonPobiega:
MessageEndpointMappings

johnsimons:
Ok, I know what MEM is

johnsimons:
In that case my original question stands, what is the difference between `bus.Subsciber(...)` vs what u proposing ?

johnsimons:
the way I see it, it makes more sense to me to have an api like:
```c#
busConfiguration.Subscribe(type, EndpointName);
```

SzymonPobiega:
In order to do `bus.Subscriber` the endpoint needs to know where the publisher is. So you need to first do `busConfig.Publishers().AddStatic(new EndpointName("sales"), typeof(MyEvent)` and then `bus.Subscribe()`.

johnsimons:
the reason is because majority of our user are already used to the old style:
`bus.Subscribe(type)`

SzymonPobiega:
@johnsimons I am not touching the `bus.Subscriber`. I only replace the source of information it uses.

johnsimons:
but again we are mixing concerns here, that to me are infrastructure concerns that should not be exposed out of busConfig

SzymonPobiega:
a mapping of event type to a *logical* endpoint is not an infra concern IMO

johnsimons:
agree, hence the reason I suggested:

`busConfiguration.Subscribe(type, new EndpointName("Sales"));`

johnsimons:
but the code that tells the framework where the "Sales" physical address is, is IMO a infrastructure concern

SzymonPobiega:
how would that relate to *auto subscribe*? Adding info to the collection of publishes does not mean we actually subscribe. This is controlled elsewhere.

johnsimons:
ok, I am confused again, maybe another face to face meeting would help here.
I can't do it right now but maybe Wednesday?

SzymonPobiega:
ok, let's try on Wednesday.

johnsimons:
@SzymonPobiega are we still discussing this API design on wednesday ?

SzymonPobiega:
Yes. I'll be back around 8:15-8:20 my time so I'll ping you.

PR #967, opened 2015-12-07
notifications section in v6 upgrade guide

timbussmann:
sadly, the example code requires all the rx packages, should we just omit it?


SimonCropp:
@timbussmann the snippets project was always going to need many nugets. so i think this is worthwhile. merging

PR #970, opened 2015-12-09
Routing rework

SzymonPobiega:
* replaced ASCII art with proper diagram
 * broken down routing doco into secrions based on roles & responsibilities (architect, ops, transport)


SzymonPobiega:
@SimonCropp @johnsimons please take a look.

johnsimons:
@SzymonPobiega I still don't understand how we planning to deal with ServiceControl ?
Is the DEV/OPS suppose to deploy an assembly into SC ?

SC is on v5, we currently do not have plans to upgrade to v6, So if a user starts using all this new features, how do they implement the same routing conventions in SC ?

SimonCropp:
not saying "this isnt a problem" but we already have many extension points that SC doesn't support

For example

- custom encryption via mutators
- custom transports

johnsimons:
@SimonCropp regarding
> custom encryption via mutators

That is not important to SC, SC does not need to decrypt the payload, the only issue is that we skip indexing of the body of the message (so SI searching may suffer because of it), but the headers are still indexed.

> custom transports

In theory they can work, we just need to provide some guidance if it comes to it.

Anything else u can think of ?

SimonCropp:
> we just need to provide some guidance if it comes to it.

but wouldnt that guidance be very similar to how someone would support routing changes @SzymonPobiega is proposing?

johnsimons:
No idea @SimonCropp 
The way I see it is that we are exposing a code only api that is only available in v6, so how does OPS configure SC to do the same ?

SimonCropp:
i dont think the intent is to provide a "code only (and not ops friendly) api". just "code first" so that more opinionated ops friendly approaches can be build on top of that API. yes those ops friendly approachs have not been build yet but thats doesnt mean "they wont". sorry @SzymonPobiega if i am mis-interpreting your plans

johnsimons:
Agree, hence the reason to raise it to make sure we don't forget

SzymonPobiega:
@SimonCropp @johnsimons I don't see it relates directly to SC. SC uses the `processing endpoint` header to figure out where to return the messages and that is not affected by the routing. That said, the dynamic scale-out thing might cause friction since endpoint instances are easier added/removed so that header might potentially refer to an instance that no longer exists.

I think the general solution here should be tapping SC to the same routing mechanism as the endpoints and re-route each message when retrying. That would solve the problems of topology changing under our feet but that requires a significant rework of SC.

andreasohlund:
>That said, the dynamic scale-out thing might cause friction since endpoint
instances are easier added/removed so that header might potentially refer
to an instance that no longer exists.

Agreed, but that's a problem we already have with the current distributor
right?

PR #977, opened 2015-12-14
Update docs to describe current behaviour - better to submit an issue to be fixed instead?

timiles:
Got stuck where my assembly wasn't reliably being picked up for subscription; tracked down the issue to be an assembly with version 0.0.1.1 apparently not being considered compliant with assembly with version 0.0.128.0.  If I force the pair to be 0.0.1.0 & 0.0.128.0, or 0.0.1.1 & 0.0.128.1, then we have success.  I can work with this for now, but do you agree this is a potential issue?

### DIFF
--- a/nservicebus/messaging/publish-subscribe/index.md
+++ b/nservicebus/messaging/publish-subscribe/index.md
@@ -153,7 +153,7 @@ Transport->Subscriber2: Send Message1
 
 ## Versioning subscriptions
 
-In NServiceBus version 3.0 and onwards subscriptions for types with the same Major version are considered compliant. This means that a subscription for MyEvent 1.1.0 will be considered valid for MyEvent 1.X.Y as well.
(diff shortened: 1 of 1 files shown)


andreasohlund:
The current version should only look at the major version, what version of NServiceBus are U on?

That said we do plan to get rid of the version check but no concrete timeline yet

timiles:
I am using Version 5.  It does respect the major version for compliance, but I believe it's a bug that it also gets upset if there's a fourth build number which does not match, i.e. 0.0.1.0 vs 0.0.1.1

andreasohlund:
Got it, I'll make sure we includes this as well in the fix

andreasohlund:
Here's the issue to track this https://github.com/Particular/NServiceBus/issues/2778

timiles:
Thanks Andreas

PR #983, opened 2015-12-15
Using BusSession instead of BusContext

danielmarbach:

Attention: Had to go. No time left to look at the upgrade guide. If someone can I'd be happy.


danielmarbach:
Since there is no text involved here I'm going to self-merge

PR #984, opened 2015-12-15
Update transaction enum name

colin-higgins:
@andreasohlund


SimonCropp:
@colin-higgins can u convert those to be snippets in a the same way as the other snippets in the upgrade guide. that way they will be verified on each nuget update

SimonCropp:
merged manually here https://github.com/Particular/docs.particular.net/commit/949318fac888c3b08b8e4df51530d71ae301c2ce

PR #986, opened 2015-12-17
Updated routing to cover changes in `EndpointInstance`

SzymonPobiega:


johnsimons:
> the next step would be to make the discriminator mandatory as we discussed earlier today.

@SzymonPobiega regarding that, should we change the api so the `config.EndpointName(name, discriminator)` ?

/cc @andreasohlund @tmasternak

PR #993, opened 2015-12-28
WIP - Adding an include file and header for SM discontinued message

indualagarsamy:
Relates to https://github.com/Particular/PlatformDevelopment/issues/454

### DIFF
--- a/platform/getting-started-with-particular-service-platform.md
+++ b/platform/getting-started-with-particular-service-platform.md
@@ -6,8 +6,9 @@ tags:
 - ServiceMatrix
 - Send and Receive
 - Visual Studio
-include: sm-discontinued
 ---
+
+include: sm-discontinued
 
 This step-by-step guide to the Particular Service Platform walks you through the creation of a send-and-receive NServiceBus distributed application using ServiceMatrix v2.0 for Visual Studio 2012, using the following steps:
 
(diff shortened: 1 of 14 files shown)


platform/getting-started-with-particular-service-platform.md
```diff
@@ -8,6 +8,8 @@ tags:
 - Visual Studio
 ---
 
+include: sm-discontinued
```
SimonCropp: I think we should move this file under the am directory. And also rename it to "getting started with am". And also move it in the menu

servicematrix/sm-discontinued.include.md
```diff
@@ -0,0 +1,3 @@
+## ServiceMatrix is no longer available. For more details, [please read this message](http://www.particular.net/servicematrix)
```
SimonCropp: Why the comma?

indualagarsamy: Introductory clauses and long phrases must always be followed by a comma, while single words and short phrases may or may not. Introductory phrases and clauses usually begin with a subordinate conjunction (such as after, because, though, or when).

I now use a tool to check my often misuse of commas. The usage of comma in this instance is correct.

platform/getting-started-with-particular-service-platform.md
indualagarsamy: Ah, you meant `SM directory`. I was wondering about the am! Yeah, I agree. It's just that I looked for usage of Matrix and disabled it. But will moving it cause a 404?

PR #994, opened 2015-12-29
Improving MessagingBestPractices/Validations exception messages

WojcikMike:
Adding information about exceptions being thrown while violating the best practices

Relates to https://github.com/Particular/NServiceBus/issues/3267

### DIFF
--- a/nservicebus/messaging/messages-events-commands.md
+++ b/nservicebus/messaging/messages-events-commands.md
@@ -37,7 +37,18 @@ Used to communicate that some action has taken place. An *Event* should be _publ
  * cannot implement `ICommand`.
  * cannot be sent using the gateway, i.e., `bus.SendToSites()`.
 
-Note: For reply messages in a request and response pattern, you may want to use `IMessage` since these replies are neither a Command nor an Event.
+Note: For reply messages in a request and response pattern, you may want to use `IMessage` since these replies are neither a Command nor an Event. 
+
+### Validation Messages (version 3 - version 6)
+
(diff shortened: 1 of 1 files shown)


SimonCropp:
is the exception message consistent for all current majors of NSB?

WojcikMike:
The above description match NSB v6.

For NSB v5 the messages are the same however the exception is of type `InvalidOperationException`

NSB v4 and v3 - as for NSB v5 the exception is `InvalidOperationException` and the following message:
"Reply is neither supported for Commands nor Events. Commands should be sent to their logical owner using bus.Send and bus. Events should be Published with bus.Publish."
Depends on the method used you either get the same message or 2 separate ones:
"Reply is not supported for Commands. Commands should be sent to their logical owner using bus.Send and bus."
and
"Reply is not supported for Events. Events should be Published with bus.Publish."

@SimonCropp should we extract the section with exceptions and do it in a way that match versions?

WojcikMike:
Hey @andreasohlund , can you chime in?

SimonCropp:
@WojcikMike i dont care about the exception type. people should never be catching it so it is irrelevant.   i would like to have the exact text for each major version captured in the doco

andreasohlund:
> i dont care about the exception type. people should never be catching it so it is irrelevant. i would like to have the exact text for each major version captured in the doco

:+1:

WojcikMike:
Ok, in that case this is the direction that I will taken. Will include exact exception messages without mentioning exception types.

PR #995, opened 2016-01-04
Updating host profiles doco to include snippets for different versions

johnsimons:
Also removed `Behavior requiring initialization to be complete` section, can't figure out the value in it

### DIFF
--- a/Snippets/Snippets_3/Profiles/behavior.cs
+++ b/Snippets/Snippets_3/Profiles/behavior.cs
@@ -1,0 +1,26 @@
+﻿using NServiceBus.Hosting.Profiles;
+#region profile_behavior [3,4]
+class LiteEmailBehavior : IHandleProfile<NServiceBus.Lite>
+{
+    public void ProfileActivated()
+    {
+        // set the NullEmailSender in the container
+    }
+}
+
+class IntegrationEmailBehavior : IHandleProfile<NServiceBus.Integration>
+{
+    public void ProfileActivated()
+    {
+        // set the FileEmailSender in the container
+    }
+}
+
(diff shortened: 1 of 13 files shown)


SimonCropp:
thanks

johnsimons:
I don't really understand the value in http://docs.particular.net/nservicebus/hosting/nservicebus-host/profiles#dependent-profile-behaviors either, I would also remove that, not sure what exactly the point it ?

SimonCropp:
@johnsimons that doco pre-dates me. i dont mind removing it, but is there anyone who is more of an expert on the host who can weigh in?

johnsimons:
Only @udidahan and @andreasohlund

udidahan:
@johnsimons do you not understand the value in the docs, or of the feature itself?

johnsimons:
The feature @udidahan

udidahan:
@johnsimons and is the value of Profiles clear?

SimonCropp:
@udidahan to me no. short term appearance of value for long term complexity and overly rigid/opinionated api

udidahan:
@SimonCropp then an issue should be opened in PlatDev to discuss that.

johnsimons:
@udidahan the one I am strugling with is http://docs.particular.net/nservicebus/hosting/nservicebus-host/profiles#dependent_profile-v3_x-4_x

udidahan:
@johnsimons the scenario is that a user may want to create their own types "decorators" for the IConfigureThisEndpoint class like AsA_Publisher - for example, ISendEmails. Then, they would want to have a profile handler which checks to see whether this endpoint sends emails, and if so, customize the way in which that's done.

PR #1001, opened 2016-01-05
removed duplicate text

dbelcham:
@Particular/documentation @SimonCropp 

First change so went conservative and created a PR. There appeared to be a duplicate section at the end of this document so I just deleted it.


SimonCropp:
Is the snippet no longer required?

dbelcham:
The snippet is still being used. Both the text and the snippet were duplicated.

SimonCropp:
Looking at the diff that is not the case.

dbelcham:
For some reason the Source Diff is showing line 63 as being deleted when it's not (Rich Diff shows correctly).

Lines 33 -> 37 are the ones that were duplicated with what I removed at lines 66->70

PR #1005, opened 2016-01-06
ConfigureAwait in behaviors

danielmarbach:
@Particular/documentation This PR adds `ConfigureAwait(false)` to the behaviors. Inside the pipeline it matters.

I know this is a slight contradiction to http://docs.particular.net/samples/#technology-choices-async-await

I also aligned Snippet 4 and Snippet 4 behavior comment to Snippets 6 behavior


SimonCropp:
@danielmarbach i thought we had previously agreed to skip the `ConfigureAwait(false)` http://docs.particular.net/samples/#technology-choices-async-await  ?

PR #1026, opened 2016-01-08
Document transaction scope wrapper UoW and MSMQ & SQL transaction scope opts

MarcinHoppe:
First stab at documenting the transaction API cleanup.

### Still to do:
- [x] MSMQ transport transaction scope options
- [x] UoW transaction scope options
- [x] SQL transport transaction scope options (when implemented)
- [x] Upgrade guide

### DIFF
--- a/nservicebus/messaging/transactions.md
+++ b/nservicebus/messaging/transactions.md
@@ -114,11 +114,14 @@ NOTE: This requires the selected storage to support enlisting in transaction sco
 
 WARNING: This might escalate to a distributed transaction if data in different databases are updated.
 
+WARNING: This API must not be used in combination with transports running in *transaction scope* mode. Starting from version 6, wrapping handlers in a `TransactionScope` in such a situation throws an exception.
+
 
 ## Controlling transaction scope options
 
 The following options for transaction scopes used during message processing can be configured.
 
+NOTE: Starting from version 6 isolation level and timeout for transaction scopes are also configured at the transport level.
 
 ### Isolation level
 
--- a/nservicebus/pipeline/unit-of-work.md
+++ b/nservicebus/pipeline/unit-of-work.md
@@ -6,6 +6,46 @@ redirects:
 - nservicebus/unit-of-work-in-nservicebus
 ---
 
+## Using a transaction scope
+
+If your business transaction is spread across multiple handlers there is always a risk of partial updates since one handler might succeed in updating the data while other won't. To avoid this use a unit of work that wraps all handlers in a `TransactionScope` and makes sure that there are no partial updates. Use following code to enable a wrapping scope:
+
+snippet:UnitOfWorkWrapHandlersInATransactionScope
+
+NOTE: This requires the selected storage to support enlisting in transaction scopes.
+
+WARNING: This might escalate to a distributed transaction if data in different databases are updated.
+
+WARNING: This API must not be used in combination with transports running in *transaction scope* mode. Starting from version 6, wrapping handlers in a `TransactionScope` in such a situation throws an exception.
+
+
+### Controlling transaction scope options
+
+The following options for transaction scopes used to wrap all handlers can be configured.
+
+
+### Isolation level
+
+NServiceBus will by default use the `ReadCommitted` [isolation level](https://msdn.microsoft.com/en-us/library/system.transactions.isolationlevel).
+
+Change the isolation level using
+
+snippet:UnitOfWorkCustomTransactionIsolationLevel
+
+
+### Transaction timeout
+
+NServiceBus will use the [default transaction timeout](https://msdn.microsoft.com/en-us/library/system.transactions.transactionmanager.defaulttimeout) of the machine the endpoint is running on.
+
+Change the transaction timeout using
+
+snippet:UnitOfWorkCustomTransactionTimeout
+
+Or via .config file using a [example DefaultSettingsSection](https://msdn.microsoft.com/en-us/library/system.transactions.configuration.defaultsettingssection.aspx#Anchor_5).
+
+
+## Implementing custom units of work
+
 When using a framework like NServiceBus you usually need to create your own units of work to avoid having to repeat code in your message handlers. For example, committing NHibernate transactions, or calling `SaveChanges` on the RavenDB session.
 
 ### IManageUnitsOfWork
--- a/nservicebus/upgrades/5to6.md
+++ b/nservicebus/upgrades/5to6.md
@@ -440,6 +440,19 @@ Version 6 provide a configuration API that is more aligned with the transaction 
 * `config.Transactions().DisableDistributedTransactions()` - replaced with `.UseTransport<MyTransport>().Transactions(TransportTransactionMode.ReceiveOnly)` or if supported by the transport `.UseTransport<MyTransport>().Transactions(TransportTransactionMode.AtomicSendsWithReceive)`
 * `config.Transactions().EnableDistributedTransactions()` - is the default mode for transactions with DTC support but can be enabled explicitly using `.UseTransport<MyTransport>().Transactions(TransportTransactionMode.TransactionScope)`
 
+### Controlling transaction scope options
+
+Version 6 allows transaction scope options to be configured at the transport level. Setting isolation level and timeout can now be done with the following:
+
+snippet:5to6TransportTransactionScopeOptions
+
+
+### Wrapping handlers execution in a transaction scope
+
+Version 6 comes with a unit of work that wraps execution of handlers in a transaction scope, which can now be done with this API:
+
+snippet:5to6WrapHandlersExecutionInATransactionScope
+
 
 ### Suppressing the ambient transaction
 
--- a/Snippets/Snippets_5/UpgradeGuides/5to6/Upgrade.cs
+++ b/Snippets/Snippets_5/UpgradeGuides/5to6/Upgrade.cs
@@ -1,6 +1,7 @@
 ﻿namespace Snippets5.UpgradeGuides._5to6
 {
     using System;
+    using System.Transactions;
     using NServiceBus;
     using NServiceBus.Unicast.Transport;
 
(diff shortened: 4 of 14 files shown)


MarcinHoppe:
@andreasohlund can you take a look and check if this is going in the right direction?

Snippets/Snippets_6/Transactions.cs
```diff
@@ -32,17 +50,17 @@ public void TransportTransactionsWithScope()
...
+            busConfiguration.UnitOfWork()
+                .WrapHandlersInATransactionScope();
             #endregion
         }
 
-        public void CustomTransactionTimeout()
+        public void UnitOfWorkWrapHandlersInATransactionScope()
         {
-            #region CustomTransactionTimeout
+            #region UnitOfWorkWrapHandlersInATransactionScope
```
andreasohlund: You need to rename the corresponding region in the v5, v4 and v3 snippets as well (if you want them to show in the same place)

nservicebus/pipeline/unit-of-work.md
```diff
@@ -6,6 +6,19 @@ redirects:
 - nservicebus/unit-of-work-in-nservicebus
 ---
 
+## Avoiding partial updates
```
andreasohlund: Perhaps the title should be more like "Using a transaction scope" and the partial updates can be moved into the text instead as a use case?

Snippets/Snippets_6/Transactions.cs
MarcinHoppe: The `CustomTransactionTimeout` is there below. It seems I out-smarted the differ by shuffling the code around a little bit too much.

I think we need both `CustomTransactionTimeout` snippet to show in the transport transactions doco and `UnitOfWorkWrapHandlersInATransactionScope` to show in the UoW doco.

MarcinHoppe: Agree?

nservicebus/pipeline/unit-of-work.md
MarcinHoppe: I reworded that a little bit. Looks better now?

Snippets/Snippets_6/Transactions.cs
SimonCropp: @MarcinHoppe so does This snippet exist in earlier versions?

nservicebus/messaging/transactions.md
```diff
@@ -119,6 +119,7 @@ WARNING: This might escalate to a distributed transaction if data in different d
 
 The following options for transaction scopes used during message processing can be configured.
 
+NOTE: In Version 6 isolation level and timeout for transaction scopes are also configured at the transport level.
```
SimonCropp: This is a fragile statement. What about version 5 and lower? When we release v7 does this apply? You seem to imply it is only v6. Perhaps "in version 6 and higher..."

MarcinHoppe: Great commend. Reworded this statement.

Snippets/Snippets_6/Transactions.cs
MarcinHoppe: The custom transaction timeout snippet exists in all versions: 3.x, 4.x, 5.x, and 6-pre. For 6-pre it is defined here:

https://github.com/Particular/docs.particular.net/pull/1026/files#diff-e43503f5ec8194d91e9e461683ad48a4R77

Does this make sense?

andreasohlund: Should we pull the UOW snippet out into a UnitOfWork.cs to make the split even more obvious?

And then perhaps rename `Transactions.cs` => `TransportTransactions.cs` ?

MarcinHoppe: Done.

MarcinHoppe:
@Particular/nservicebus @Particular/documentation this is ready for review and merge.

nservicebus/pipeline/unit-of-work.md
```diff
@@ -6,6 +6,43 @@ redirects:
...
 
+## Using a transaction scope
+
+If your business transaction is spread across multiple handlers there is always a risk of partial updates since one handler might succeed in updating the data while other won't. To avoid this use a unit of work that wraps all handlers in a `TransactionScope` and makes sure that there are no partial updates. Use following code to enable a wrapping scope:
+
+snippet:UnitOfWorkWrapHandlersInATransactionScope
+
+NOTE: This requires the selected storage to support enlisting in transaction scopes.
+
+WARNING: This might escalate to a distributed transaction if data in different databases are updated.
```
ramonsmits: When using different connection strings, not when using different databases. It is perfectly fine to use multiple database on a single connection.

nservicebus/pipeline/unit-of-work.md
```diff
@@ -6,6 +6,43 @@ redirects:
 - nservicebus/unit-of-work-in-nservicebus
 ---
 
+## Using a transaction scope
+
+If your business transaction is spread across multiple handlers there is always a risk of partial updates since one handler might succeed in updating the data while other won't. To avoid this use a unit of work that wraps all handlers in a `TransactionScope` and makes sure that there are no partial updates. Use following code to enable a wrapping scope:
+
+snippet:UnitOfWorkWrapHandlersInATransactionScope
+
+NOTE: This requires the selected storage to support enlisting in transaction scopes.
```
ramonsmits: Which selected storage? Storage from NServiceBus and/or the user scope?

nservicebus/pipeline/unit-of-work.md
```diff
@@ -6,6 +6,43 @@ redirects:
...
+
+WARNING: This might escalate to a distributed transaction if data in different databases are updated.
+
+### Controlling transaction scope options
```
ramonsmits: Shouldnt this also have the option to control the kind of scope? Like Required, RequiresNew or Suppress?

ramonsmits:
What about the following?

```c#
busConfiguration.UseTransport<MsmqTransport>()
  .Transactions(TransportTransactionMode.TransactionScope)
  .TransactionScopeOptions(isolationLevel: IsolationLevel.RepeatableRead);

busConfiguration.UnitOfWork()
  .WrapHandlersInATransactionScope(timeout: TimeSpan.FromSeconds(30));
```

Is the UoW scope not required? It is not required when it is created with *Required* and the transport defines a transaction scope.

Isn't there a case where you would like to have a transaction scope for the transport and a transaction scope RequiresNew for the UoW? This looks similar to the native transport transaction option but in a different implementation.

nservicebus/upgrades/5to6.md
```diff
@@ -440,6 +440,17 @@ Version 6 provide a configuration API that is more aligned with the transaction
...
+### Controlling transaction scope options
+
+Version 6 allows transaction scope options to be configured at the transport level. Setting isolation level and timeout can now be done with the following:
+
+snippet:5to6TransportTransactionScopeOptions
+
+
+### Wrapping handlers execution in a transaction scope
+
+`config.Transactions().WrapHandlersExecutionInATransactionScope()` was removed and wrapping handler execution hand is now done via `config.UnitOfWork().WrapHandlersInATransactionScope()`.
```
SimonCropp: I would prefer these were imported snippets

nservicebus/pipeline/unit-of-work.md
MarcinHoppe: How about:

"WARNING: This might escalate to a distributed transaction if data in more than one durable resource is updated. For example, using two distinct database connections inside a `TransactionScope` will cause an escalation."

nservicebus/pipeline/unit-of-work.md
MarcinHoppe: I think this particular sentence refers to the storage in which the user's business data is stored.

I have a vague memory of @andreasohlund saying that writes to NSB storage might (with the exclusion of Outbox) happen outside of this `TransactionScope`.

Of course, there's also fat chance I might be wrong here, so experts should feel free to correct me.

From doco perspective, should we be explicit about what storage is meant here?

nservicebus/pipeline/unit-of-work.md
MarcinHoppe: `Required` is used here. We can certainly document it but would this knowledge be beneficial to the user?

This UoW always creates a new transaction scope and throws if an ongoing transaction is detected. In such setting `Suppress` does not seem to bring any benefit, and `RequiresNew` seems to be equivalent to `Requires`.

Does this make sense?

MarcinHoppe:
@ramonsmits This snippet would throw an exception. The UoW detects if a transport transaction scope is present and only creates a new transaction scope if it can't find one (throws otherwise).

Should we document this behavior and add guidance not to use this UoW when a transport transaction scope is already present?

nservicebus/upgrades/5to6.md
MarcinHoppe: Good point. Done.

andreasohlund:
@MarcinHoppe yes we'll throw if the scope UOW is used together with a transport scope so we should document that

MarcinHoppe:
I agree. What bugs me is that we only throw at runtime. Can we detect this situation at configuration time to give user earlier feedback that their configuration is not legal?

MarcinHoppe:
@andreasohlund Proper comment added.

SimonCropp:
@MarcinHoppe can u stop adding the inprogress label. that is inferred by the fact that it is a PR

MarcinHoppe:
I think this was added when I was moving item in Waffle. I made a note to myself not to do this again on a PR.

andreasohlund:
Waffle auto moves PR to the in-progress column, should we turn that off?

SimonCropp:
@andreasohlund i dont think waffle adds much value in terms of docs

andreasohlund:
So that means that doco work won't show up on the aggregated board, this feels off to me? (why wouldn't we like to make doco work visible as well)

andreasohlund:
>Isn't there a case where you would like to have a transaction scope for the transport and a transaction scope RequiresNew for the UoW? This looks similar to the native transport transaction option but in a different implementation.

I can't see a use case for it since that means that the recive is not in sync with the storage ops so the user can just lower the transport tx level to < `TransactionScope`

PR #1052, opened 2016-01-15
Core v6 version of the ASB native integration sample

SeanFeldman:
Second ASB sample converted to use core v6 and transport v7 (`StandardTopology`)


SeanFeldman:
@SimonCropp this is another sample. Previous one was send/reply, this one is native that demonstrates how to set transport encoding that is different from the default.

SimonCropp:
I got nothing.. This all looks good, merge if u r also happy

PR #1055, opened 2016-01-20
Add info for message type fallback

dvdstelt:
I added a single paragraph. But although it might be clear or correct what it states (I have you to verify that ;-) ) I'm not sure of readers of the document will understand.

I've thought about a code snippet of a single line, which seems ridiculous. Also because it's not NSB version related, it's .NET related.
I've also thought about putting a triple ` addition, like in the step-by-step page (http://docs.particular.net/samples/step-by-step/)
But I'm not sure if I'm allowed to use that.

So I'd like some feedback

### DIFF
--- a/nservicebus/handlers/index.md
+++ b/nservicebus/handlers/index.md
@@ -23,14 +23,16 @@ If you are using the Request-Response or Full Duplex pattern, your handler will 
 
 If you are handling a message in a publish-and-subscribe scenario, see [How to Publish/Subscribe to a Message](/nservicebus/messaging/publish-subscribe/).
 
-### Mapping to FQN with fallback to type
 
-NServiceBus will try to map incoming messages to a type using its Fully Qualified Name (FQN). This is default behavior for sharing assemblies amongst endpoints and sendonly clients. When a message cannot be mapped based on FQN, NServiceBus will try to map to just the type. The following is an example of how NServiceBus gets the type information.
+### Mapping to name
+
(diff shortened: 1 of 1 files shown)


dvdstelt:
This is the line I might add, to show how to obtain the FQN

```
MyCustomMessage message = new MyCustomMessage();
string fqn = message.GetType().AssemblyQualifiedName;
```

SimonCropp:
@dvdstelt perhaps `typeof(MyCustomMessage).AssemblyQualifiedName` etc?

SimonCropp:
@andreasohlund can u please review this

dvdstelt:
@SimonCropp I thought that with the message coming into a handler, the `message.GetType()` would be clearer to the readers. To make it more clear, I added the `MyCustomMessage message = new MyCustomMessage();` so they'll know it's the message coming into a handler.

Might also change it to the following. But thought that was overkill, as you also need to add a async NSB6 version for people to understand it in the near future.

```
public class MyHandler : IHandleMessages<MyMessage>
{
    public void Handle(MyMessage message)
    {
        string fqn = message.GetType().AssemblyQualifiedName;
        string fallback = message.GetType().FullName;
    }
}
```

SimonCropp:
@dvdstelt there is actually a way to document that kind of stuff in a cross version way https://github.com/Particular/docs.particular.net#defining-snippets

dvdstelt:
I know. And with the last snippet it might make sense, because the IHandleMessages<T> code is in there. But initially it was just C# code which doesn't change over NServiceBus versions (yet) and I thought the snippets were a bit overkill.

Normally I write much sentences to explain something. But as NServiceBus docos normally don't use a lot of words/sentences, I thought I'd comply (?) to this as well and made up just this paragraph.

But that's also why I asked to have a look at what I wrote. Should I add the code? Should I use more words to explain? I know to me personally, the paragraph would probably not make a lot of sense and I'd need to study it. Especially without a code example.

SimonCropp:
wouldnt this be better documented here http://docs.particular.net/nservicebus/messaging/headers#serialization-headers

dvdstelt:
It's not where I would look for information on how messages are mapped to handlers. It could be a reference from the handlers doc to the headers doc, where it's explained further. But I've looked at the headers document as a reference for information, not where something is explained about how a mapping occurs.

PR #1093, opened 2016-01-26
Removing  MaximumMessageThroughputPerSecond from snippets

tmasternak:


nservicebus/operations/tuning.md
```diff
@@ -31,6 +31,8 @@ You can define a maximum value for the number of messages per second that the en
 NServiceBus will not enforce any throughput restrictions by default.
 
 
+NOTE: Currently version 6 of NServiceBus does not support `MaximumMessageThroughputPerSecond` setting.
```
SimonCropp: "Currently"? so it will in the future? what about when someone is reading this in the v7 timeframe?  is there any more info we need to give on top of "does not support"?

SimonCropp: since this is the new standard moving forward shouldn't this be at the top of this section?

tmasternak: I think that the throttling will likely be dropped Particular/NServiceBus#2859. I'll remove 'currently' to avoid confusion.

tmasternak:
Changed the wording and moved 'Notes' boxes to top of the sections

PR #1123, opened 2016-02-01
Update db-compaction.md

rikedje:
The commandline to run ServiceControl must be run as Administrator


gbiellem:
@rikedje  Thanks for the update

PR #1126, opened 2016-02-01
Throughput throttling sample

hmemcpy:
ping @timbussmann


hmemcpy:
Need to write sample.md

samples/throttling/Version_6/Sample/EndpointConfig.cs
```diff
@@ -0,0 +1,20 @@
+namespace ThroughputThrottlingDemo
```
SimonCropp: remove the namespace

samples/throttling/Version_6/Sample/packages.config
```diff
@@ -0,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="NServiceBus" version="6.0.0-unstable1448" targetFramework="net46" />
+  <package id="NServiceBus.Host" version="7.0.0-unstable0053" targetFramework="net46" />
```
SimonCropp: use self hosting instead

samples/throttling/Version_6/Sample/EndpointConfig.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    using NServiceBus;
+
+    public class EndpointConfig : IConfigureThisEndpoint
+    {
+        public void Customize(BusConfiguration configuration)
+        {
+            configuration.UsePersistence<InMemoryPersistence>();
+            configuration.SendFailedMessagesTo("errorq");
```
SimonCropp: the error q for most samples is "error"

SimonCropp:
why no v5 version?

samples/throttling/Version_6/Sample/GitHubSearchHandler.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using System.Threading.Tasks;
+using NServiceBus;
+using Octokit;
+
+namespace ThroughputThrottlingDemo
```
SimonCropp: remove the namespace

samples/throttling/Version_6/Sample/GitHubSearchHandler.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    public class GitHubSearchHandler : IHandleMessages<SearchGitHub>
+    {
+        // use anonymous access which has very strict rate limitations
+        private static readonly GitHubClient GitHubClient = new GitHubClient(new ProductHeaderValue("ThroughputThrottlingDemo"));
```
SimonCropp: remove the private and readonly

samples/throttling/Version_6/Sample/Sample.csproj
```diff
@@ -0,0 +1,80 @@
...
+  </ItemGroup>
+  <Import Project="$(MSBuildToolsPath)\Microsoft.CSharp.targets" />
+  <!-- To modify your build process, add your task inside one of the targets below and uncomment it. 
+       Other similar extension points exist, see Microsoft.Common.targets.
+  <Target Name="BeforeBuild">
+  </Target>
+  <Target Name="AfterBuild">
+  </Target>
+  -->
+  <PropertyGroup>
```
SimonCropp: u forgot to remove the noise from the host

samples/throttling/Version_6/Sample/Startup.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using System.Threading.Tasks;
+using NServiceBus;
+
+public class Startup : IWantToRunWhenBusStartsAndStops
```
SimonCropp: move the to the program.cs

samples/throttling/Version_6/Sample/Sample.csproj
hmemcpy: @SimonCropp sigh, you're right :|

timbussmann:
@SimonCropp we removed throttling configuration in v6 and decided to provide some guidance on how to implement a custom throttling instead. V5 still has the throttling configuration option as described in the docs.

samples/throttling/Version_6/Sample/Program.cs
```diff
@@ -0,0 +1,49 @@
...
+        AsyncMain().GetAwaiter().GetResult();
+    }
+
+    static async Task AsyncMain()
+    {
+        BusConfiguration configuration = new BusConfiguration();
+        configuration.EndpointName("Samples.Throttling");
+        configuration.UsePersistence<InMemoryPersistence>();
+        configuration.SendFailedMessagesTo("error");
+        configuration.LimitMessageProcessingConcurrencyTo(1);
```
timbussmann: we could actually remove the concurrency limitation

samples/throttling/Version_6/Sample/Sample.csproj.DotSettings
```diff
@@ -0,0 +1,2 @@
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
+	<s:String x:Key="/Default/CodeInspection/CSharpLanguageProject/LanguageLevel/@EntryValue">CSharp60</s:String></wpf:ResourceDictionary>
```
timbussmann: do we need to include this file?

SimonCropp: @timbussmann  it is too early to convert all the old samples to c# 6. so this is a short term hack. will move over to all c#6 and hence VS2015 in prob 6 months

SimonCropp:
@timbussmann but the argument to remove the throttling  was because it wasnt actually possible to use it properly with real world requirements. as per @andreasohlund logic. so based on that the same logic should make a v5 sample just a valuable

timbussmann:
we didn't receive an issues from users regarding a better way to do throughput throttling in v5 yet. So, while I agree that a v5 sample could add value, I don't see that on my priority list tbh. According to Ramon, there are customers using the v5 mechanisms and this sample is intended to show a way how to move forward.

SimonCropp:
> we didn't receive an issues from users regarding a better way to do throughput throttling in v5 yet

That doesnt mean they dont have bug they are not aware of.

timbussmann:
@hmemcpy let's also update the tuning.md page as part of this PR?

SimonCropp:
@timbussmann 

> let's also update the tuning.md page as part of this PR

prefer u didnt. this is a chunk of work that can be finished and then move onto the next. no need to add more to it

timbussmann:
I'm ok with another PR after the sample is merged. But I think it makes sense to link to this sample from the docs page.

samples/throttling/Version_6/Sample/Program.cs
hmemcpy: @timbussmann removing this causes the console output to go to hell... for the sample we could just limit to 1

timbussmann: hm should we add a comment about this?

hmemcpy:
@Particular/documentation please review.
@SimonCropp I'm not sure if I've done everything correctly for the sample to appear in /samples, please note what's missing.

PR #1133, opened 2016-02-03
Distributor cluster profiles

ramonsmits:
Based on feedback by @SimonCropp I now mention the distributor profiles for both versions earlier than V4.3 and after. Also changed the title as this document is about the distributor.


nservicebus/scalability-and-ha/deploying-to-a-windows-failover-cluster.md
```diff
@@ -137,6 +145,9 @@ The first thing you should do is to make sure that the worker servers have [uniq
 
 Set up your worker processes on all worker servers (not the cluster nodes!) as windows services, but instead of using `NServiceBus.MSMQDistributor`, use `NServiceBus.MSMQWorker` profile instead.
 
+NOTE: Please use profiles **NServiceBus.Distributor** and **NServiceBus.Worker** when using NServiceBus earlier then V4.3
```
SimonCropp: i would prefer two headings here. "version x and higher" and "version x and lower". and explain the profiles in each

SimonCropp:
@ramonsmits if this page is specific to the distributor can u move it to exist under http://docs.particular.net/nservicebus/scalability-and-ha/distributor/

SimonCropp:
@ramonsmits r u still working on this?

SzymonPobiega:
@ramonsmits Shoiuld we have a similar page describing how to make a single non-distributor endpoint HA? Or maybe extract the *clustered service* part. Because I think it sometimes makes sense to get the HA without the need for throughput scale out.

SimonCropp:
@ramonsmits if u no longer think this is required i will delete the PR.

PR #1158, opened 2016-02-09
Flipping the Raven configuration options order

DavidBoike:
Starting with the most esoteric option (because it's what the code looks for first) is just wrong. It's like giving your down-the-street neighbor directions to your house via the North Pole.


DavidBoike:
@Particular/documentation please review.

Tried to get rid of me/you type language along the way as well, but couldn't cover whole document in time I had available.

PR #1170, opened 2016-02-12
Documented the new SendOptions and ReplyOptions APIs.

SzymonPobiega:


SimonCropp:
@SzymonPobiega any reason why we would not add equivalents snippets for earlier versions where supported?

SzymonPobiega:
oversight. Let me correct that.

SimonCropp:
@SzymonPobiega thanks. other than that this is good to merge.

SzymonPobiega:
@SimonCropp I've added missing snippets but could not find a way to override the `reply to` for V4 and previous.

Issue #1177, opened 2016-02-15
Documenting 'TransportDefinition` and its extensions


hmemcpy:
Connects to https://github.com/Particular/PlatformDevelopment/issues/682.

With https://github.com/Particular/NServiceBus/pull/3025, `ConfigureTransport` was renamed to `TransportDefinition`. In terms of doco, there exists a [sample](http://docs.particular.net/samples/custom-transport/) that shows how to build a custom transport using the new API.

Question(s):
1. Should `TransportDefinition` be explicitly documented?
2. Should this change appear in the upgrade guide? If so, where?

weralabaj:
I think we should create a page similar to what we have for persistence http://docs.particular.net/nservicebus/persistence/authoring-custom and add all the details there.
In upgrade guide we can mention both links, but IMHO that content is not relevant for most users, so no point in including details in the upgrade guide for everybody.

SzymonPobiega:
OK, since we don't have currently doco on writing a custom transport, should we
 * document new V6 APIs
 * document how to upgrade a transport from V5 to V6
 * both?

indualagarsamy:
@SimonCropp @hmemcpy @SzymonPobiega - What do we want to do here? I see that we've merged a PR on ConfigureTransport where we're pointing user to a sample. Is that good enough?

SzymonPobiega:
I'd say we should at least document the V6 APIs. Ideally there should be a migration guide but I am not sure if it is doable given all the other high priority tasks. I tried to create such a guide while upgrading the EventStore transport from 5 to 6 and the biggest issue I found was `IDequeueMessages` to `IPushMessages` transition. These interfaces are very different and I ended up copying MSMQ implementation of message pump and removing the MSMQ bits rather than adjusting the existing EventStore impl. 

Maybe way to document this should be to create an in-memory transport that shows the mechanics of a message pump?

weralabaj:
IMO the sample is not enough, so either we extend sample description or create another doco with important considerations. We have some initial guidance for authoring persistence http://docs.particular.net/nservicebus/persistence/authoring-custom, but that could be improved too.

Considering priority - would it make sense to get in touch with people who maintain community transports and persistences first? Wouldn't they wait for at least beta or first release before updating their projects?

weralabaj:
Issue related to https://github.com/Particular/PlatformDevelopment/issues/682.

weralabaj:
[...]

PR #1179, opened 2016-02-15
Adding note about deprecating ConfigureTransport

hmemcpy:
Added this in the upgrade guide. Please see #1177 if anything else should be added.


SimonCropp:
That sample does not show how to transition from one to the other so it is not sufficient as it stands

SimonCropp:
@hmemcpy any thoughts on this?

hmemcpy:
@SimonCropp well, we didn't have any doco/sample on the previous API, I thought to add this to the upgrade guide in case any of the downstreams use the old class. Do you think we should provide a sample for v5 as well?

SimonCropp:
i am calling doco bankruptcy on the old transport doco and merging this

PR #1202, opened 2016-02-19
updated connecting to SC section

SeanFeldman:
For a long time the examples and guidance was MSMQ specific.
Changed it to be MSMQ agnostic and in examples state that it's for MSMQ.

### DIFF
--- a/servicecontrol/plugins/index.md
+++ b/servicecontrol/plugins/index.md
@@ -42,6 +42,6 @@ If an audit queue is configured, for example `audit@MachineName` with MSMQ trans
 
 WARNING: Endpoint with plugins installed that cannot communicate to ServiceControl will shut down.
 
-The ServiceControl queue (and all other ServiceControl related sub-queues) are created during the installation phase of ServiceControl. 
+The ServiceControl queue (and all other ServiceControl related sub-queues) are created during the installation phase of ServiceControl.   The ServiceControl queue name is based on the Window Service name. The examples above assume that ServiceControl is installed with the default Windows service name of `particular.servicecontrol`.  If the instance of ServiceControl is installed with a different Windows Service name then the endpoint configuration file must be set.
 
 NOTE: If you're using MSMQ and have configured ServiceControl to import audit and error queues those queues must be present on the same machine as ServiceControl is installed on.
--- a/servicecontrol/plugins/index.md
+++ b/servicecontrol/plugins/index.md
@@ -42,6 +42,6 @@ If an audit queue is configured, for example `audit@MachineName` with MSMQ trans
 
 WARNING: Endpoint with plugins installed that cannot communicate to ServiceControl will shut down.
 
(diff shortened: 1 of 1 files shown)


SimonCropp:
@gbiellem or @mikeminutillo can u review and merge

gbiellem:
I think we need to clarify that the ServiceControl queue name is based on the service name as well.  
That not obvious in this doco but that's unrelated to what @SeanFeldman did.  I'll add it to this PR then i think we can merge

servicecontrol/plugins/index.md
```diff
@@ -34,11 +34,11 @@ NOTE: For NServiceBus version-dependent requirements for each plugin, review the
 Once deployed on an active endpoint, the endpoint sends plugin-specific information to ServiceControl. Plugins send messages using the defined endpoint transport to the ServiceControl queue. Location of ServiceControl queue is determined by the following:
 
 1. **Endpoint`s configuration file** 
-Check for an `appSetting` named `ServiceControl/Queue` e.g. `<add key="ServiceControl/Queue" value="particular.servicecontrol@machine"/>`.
+Check for an `appSetting` named `ServiceControl/Queue` e.g. `<add key="ServiceControl/Queue" value="particular.servicecontrol"/>`.
 1. **Convention based on the configured Error Queue machine** 
-If an error queue is configured, for example `error@MachineName`, then the queue `Particular.ServiceControl@MachineName` will be used.
+If an error queue is configured, for example `error@MachineName` with MSMQ transport, then the queue `Particular.ServiceControl@MachineName` will be used.
```
mikeminutillo: This sentence is confusing. Does this mean that if we have a configured error queue that includes a machine name then we will assume the Service Control queue is on the same machine? What if the error queue is just a queue name (i.e. current machine?). What if we are using a transport that does not include machine name (i.e. RabbitMQ)?

servicecontrol/plugins/index.md
```diff
@@ -34,11 +34,11 @@ NOTE: For NServiceBus version-dependent requirements for each plugin, review the
...
 1. **Convention based on the configured Audit Queue machine** 
-If an audit queue is configured, for example `audit@MachineName`, then the queue `Particular.ServiceControl@MachineName` will be used.
+If an audit queue is configured, for example `audit@MachineName` with MSMQ transport, then the queue `Particular.ServiceControl@MachineName` will be used.
```
mikeminutillo: Are these conventions applied in this order? If I have audit and error going to different machines, where do the ServiceControl messages go?

gbiellem: Yes they are applied in that order -  https://github.com/Particular/ServiceControl.Plugin.Nsb4.Heartbeat/blob/master/src/ServiceControl.Plugin.Nsb4.Heartbeat/ServiceControlBackend.cs#L86

servicecontrol/plugins/index.md
gbiellem: @mikeminutillo I've seen this in other docs too.
@SimonCropp What's the convention for queue names generally within docs? Is this right or is there something more generic?

SimonCropp: we r not consistent with that yet. so i would not block this PR for that

Issue #1195, opened 2016-02-17
SQL Server transport v3: Upgrade guide


weralabaj:
Task force: @tmasternak @weralabaj @justabitofcode @mauroservienti

Plan of attack:
- [x] Clean up multi-db doco and snippets
- [x] Configuration api changes for multi-schema
- [x] Configuration api chagnes for providing custom connection string
- [x] ~~Changes in multi-db support~~ (will be handled by https://github.com/Particular/docs.particular.net/issues/1103)
- [ ]  ~~Document new shutdown timeout behavior for SQL Particular/NServiceBus.Host#51~~ (that was moved to https://github.com/Particular/docs.particular.net/issues/1167)
- [ ]  ~~Document new shutdown timeout behavior for MSMQ~~ (that was moved to https://github.com/Particular/docs.particular.net/issues/1167)
- [x] Mention that transaction settings are configured on the transport level.
- [x] How new transaction modes translate to previous options
- [x] Circuit Breaker - removed methods.
Related to https://github.com/Particular/NServiceBus.SqlServer/issues/100

http://docs.particular.net/nservicebus/upgrades/5to6

MarcinHoppe:
You might also want to mention that transaction settings (mode, isolation level, and timeout) are configured on the transport level and this also includes the SQL transport.

tmasternak:
We should also document:
 * methods around CircuitBreaker that have been removed
 * new transaction modes and how they translate to previous options

weralabaj:
I'm starting work on this

weralabaj:
BTW what does " Document new shutdown timeout behavior for MSMQ" do here?!

weralabaj:
I'll need some help with the shutdown behaviour. @tmasternak you know what are the changes in both MSMQ and Sql?

weralabaj:
@mikeminutillo or @SimonCropp Don't want to self-merge, another review from native would also be useful. Have a look :)

indualagarsamy:
@weralabaj - I've provided my suggestions to the doco changes.

mikeminutillo:
As the associated PR has been merged, is this issue done as well? Or is there more to do?

tmasternak:
Yep. This one is done.

Issue #1103, opened 2016-01-28
SQL Transport multi instance guidance


mauroservienti:
Task force: @mauroservienti @tmasternak @justabitofcode 

Create guidance that describes why we don't want to support multi-instance in the SQL Transport and options for customers that may benefit of such a deployment topology.

As discussed with @andreasohlund we decided to ship the multi-instance support as a sample and provide guidance around it.

- [ ] Cover changes in upgrade guide for v5 to v6

indualagarsamy:
@mauroservienti @tmasternak @justabitofcode - since the upgrade guide for SqlServer transport is already completed, and has this section http://docs.particular.net/nservicebus/upgrades/sqlserver-2to3#sql-server-transport-multi-schema-support
can this issue be closed? 
cc / @weralabaj

weralabaj:
@indualagarsamy this is actually a different thing that @justabitofcode is working on at the moment. We're tracking this in PlatformDev https://github.com/Particular/NServiceBus.SqlServer/issues/139, so if we're ok with PR with no issue in docs I"m +1 for closing.

weralabaj:
This issue is covered by https://github.com/Particular/docs.particular.net/pull/1354 and https://github.com/Particular/docs.particular.net/pull/1353. No need to track it here.

PR #1206, opened 2016-02-23
Sql Server transport upgrade guide from v5 to v6

weralabaj:
Plus related changes in other Sql Server transport docos.

Connects to #1195.


Snippets/Snippets_6/Transports/SqlServer/ConnectionStrings.config
```diff
@@ -1,5 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8" ?>
 <!--startcode sqlserver-connection-string-xml 3 --><configuration>
+<configuration>
```
SimonCropp: extra configuration node

Snippets/Snippets_6/Transports/SqlServer/MultiSchema.config.xml
```diff
@@ -0,0 +1,13 @@
...
+<configuration>
+  <!--startcode sqlserver-singledb-multischema-config 3 -->
+  <UnicastBusConfig>
+  <MessageEndpointMappings>
+    <add Assembly="Billing.Contract"
+         Endpoint="billing@schema_name"/>
+    <add Assembly="Sales.Contract"
+         Endpoint="sales@another_schema_name"/>
+  </MessageEndpointMappings>
+</UnicastBusConfig>
```
SimonCropp: can the indent be fixed?

Snippets/Snippets_6/Transports/SqlServer/MultiSchema.cs
```diff
@@ -0,0 +1,33 @@
+﻿namespace Snippets5.Transports.SqlServer
+{
+    using NServiceBus;
+    using NServiceBus.Transports.SQLServer;
+
+    public class SingleDbMultiSchema
+    {
+        void CurrentEndpointSchema()
+        {
+            BusConfiguration busConfiguration = new BusConfiguration();
```
SimonCropp: this is v6 so it should be endpointconfig

SimonCropp:
i think u forgot to save the csproj before pushing?

weralabaj:
@tmasternak can you check if I didn't write anything obviously wrong?
Also I need some help with the changes in shutdown behaviour.
Should we cover multi-db here? I think that should be done as part of providing guidance for multi-db and multi-instance, etc.

Snippets/Snippets_5/Snippets_5.csproj
```diff
@@ -368,9 +369,9 @@
     <Content Include="SCPluginsConfig\heartbeatsIntervalConfig.xml" />
     <Content Include="Transports\Mqmq\MessageQueue.config.xml" />
     <Content Include="Transports\RabbitMQ_[2.0,3.0]\rabbitmqconnection.config.xml" />
-    <Content Include="Transports\SqlServer\MultiDb.config.xml">
```
tmasternak: Why did we change the file type?

tmasternak:
`c:\git\p\docs\snippets\Snippets_6\Transports\SqlServer\ConnectionStringsNamed.config` snippet in `configuration.md` is not rendering properly.

nservicebus/sqlserver/multiple-databases.md
```diff
@@ -14,24 +14,50 @@ The SQL Server transport allows you to select, on per-endpoint basis, where the
...
+By default the SQL Server transport uses a single instance of the SQL Server to maintain queues for all endpoints in the system. In order to send a message, an endpoint needs to connect to the (usually remote) database server and execute a SQL command. The message is delivered directly to the destination queue without any store-and-forward mechanism.
```
tmasternak: Change `execute a Sql command` to `execute SQL statement`?

nservicebus/sqlserver/multiple-databases.md
```diff
@@ -14,24 +14,50 @@ The SQL Server transport allows you to select, on per-endpoint basis, where the
...
 
+Using a single database doesn't require Distributed Transaction Coordinator (MS DTC). Another advantage is the ability to take a snapshot of entire system state (all the queues) by backing up a database. This is most useful when the business data is also stored in the same database.
 
-## Database-per-endpoint
+## Single database with multiple schemas
 
-In a more complex scenario endpoints use separate databases and DTC is involved. Due to lack of store-and-forward mechanism, if a remote endpoint's database or DTC infrastructure is down, the endpoint cannot send messages to it. This potentially renders the endpoint (and all endpoints transitively depending on it) unavailable.
+The default schema used by SQL Server transport is `dbo`. A different schema might be specified using API:
 
+snippet:sqlserver-singledb-multischema
```
tmasternak: This is not rendering correctly. V2 an V3 snippets are the same.

nservicebus/sqlserver/multiple-databases.md
```diff
@@ -14,24 +14,50 @@ The SQL Server transport allows you to select, on per-endpoint basis, where the
...
 
-## Database-per-endpoint with store-and-forward
+or using a configuration file:
+
+snippet:sqlserver-singledb-multischema-config
+
+If two endpoints use different schemas then additional configuration is required. The sender needs to know the schema of the receiver, and subscriber needs the schema of the publisher. 
+
+The schema for another endpoint can be specified at compile time:
```
tmasternak: I would not call it compile time. Previously it was called `push-model` - settings provided once during configuration or `pull-model` user code will be called each time information is needed.

nservicebus/sqlserver/multiple-databases.md
```diff
@@ -14,24 +14,50 @@ The SQL Server transport allows you to select, on per-endpoint basis, where the
...
+
+snippet:sqlserver-singledb-multidb-push
```
tmasternak: This snippet is not valid for version v3

nservicebus/sqlserver/multiple-databases.md
```diff
@@ -14,24 +14,50 @@ The SQL Server transport allows you to select, on per-endpoint basis, where the
...
+
+snippet:sqlserver-singledb-multischema-connString
```
tmasternak: This snippet is not valid for version v3

Snippets/Snippets_5/Snippets_5.csproj
weralabaj: I added it before and never intended to make it '.xml' file ;) Not sure if that matters? @SimonCropp ?

nservicebus/upgrades/5to6.md
```diff
@@ -610,4 +610,38 @@ The BSON serializer build into the core has been removed. Use the [Newtonsoft se
...
+
+The custom connection factory method is now expected to be `async` and no parameters are passed to it by the framework:
+
+snippet:sqlserver-custom-connection-factory
+
+### Multi-schema support
+ 
+The API and config structure for multi-schema configuration [were changed](/nservicebus/sqlserver/multiple-databases.md#single-database-with-multiple schemas). The `Queue Schema` parameter is no longer supported neither in the config file nor the code configuration API. 
+
+The schema for the current endpoint can be specified at compile time, using `DefaultSchema` method:
```
tmasternak: The same as above. I would not call it compile time.

nservicebus/sqlserver/multiple-databases.md
weralabaj: You mean they should be? Not really, it's "EndpointConfiguration" vs "BusConfiguration". I'll include the line creating the type in the snippet.

nservicebus/sqlserver/multiple-databases.md
weralabaj: I left it there, it says exactly this "Push mode: In the push mode the whole collection of endpoint connection information objects is passed during configuration time"

I'm not sure about this, Push vs Pull means nothing to the user without context. If we want to change that then I think we should explain earlier what we mean by Push and Pull and I didn't see a good way to do it in that context. It makes sense when we have two options of code configuration, but I find confusing calling config file option pull and code option push.

nservicebus/sqlserver/multiple-databases.md
weralabaj: Good catch, I need to figure out how to change it.

nservicebus/upgrades/5to6.md
weralabaj: I think as a compromise I just won't use any of those terms :) Less confusion.

nservicebus/sqlserver/multiple-databases.md
weralabaj: > c:\git\p\docs\snippets\Snippets_6\Transports\SqlServer\ConnectionStringsNamed.config snippet in configuration.md is not rendering properly.

Can you attach a screenshot? Seems ok on my computer.

tmasternak: image

For v3 the connection string should not have `Queue Schema`. However that might not be rendering problem.

nservicebus/sqlserver/multiple-databases.md
tmasternak: :+1:

nservicebus/sqlserver/multiple-databases.md
tmasternak: I like the idea of not using the `compile time` mentioned below.

nservicebus/upgrades/5to6.md
tmasternak: :+1:

nservicebus/upgrades/5to6.md
```diff
@@ -610,4 +610,38 @@ The BSON serializer build into the core has been removed. Use the [Newtonsoft se
 
 [MSMQ Subscription Authorization](/nservicebus/msmq/subscription-authorisation.md) is now done by the `SubscriptionAuthorizer` delegate at configuration time and not the `IAuthorizeSubscriptions` interface.
 
-snippet: 5to6-MsmqSubscriptionAuthorizer
\ No newline at end of file
+snippet: 5to6-MsmqSubscriptionAuthorizer
+
+## SQL Server Transport
```
SimonCropp: @weralabaj this should be split into a specific sql upgrade guide

SimonCropp:
this needs a rebase

nservicebus/upgrades/5to6.md
weralabaj: @SimonCropp what's the idea? Have a separate page in the upgrades directory for it and link to the "main" guide as related ?

SimonCropp: yeah. we can finesse the details later. but since other nugets can have multiple releases inside on major of the core, the only way is to have upgrade guides at the package level

weralabaj:
Rebased and moved to the separate article.

nservicebus/sqlserver/configuration.md
```diff
@@ -37,6 +37,11 @@ Combined with a named connection in the `connectionStrings` node of the `app.con
 
 snippet:sqlserver-named-connection-string-xml
 
+### Via a custom connection factory
+
+It is also possible to pass to transport a custom factory method, which will provide connection strings at runtime:
```
mikeminutillo: This doesn't flow well with the document above. All of the preceding sections have the form

> Via something
> by some process

Should this just be:

> By passing the transport a custom factory method which will provide connection strings at runtime:

nservicebus/sqlserver/multiple-databases.md
```diff
@@ -14,24 +14,48 @@ The SQL Server transport allows you to select, on per-endpoint basis, where the
 
 The transport will route messages to destination endpoints based on the configuration. If no specific configuration has been provided for a particular destination endpoint, the transport assumes the destination has the same configuration (schema, database and instance name/address) as the sending endpoint. If this assumption turns out to be false (the transport cannot connect to destination queue), an exception is thrown immediately. There is no store-and-forward mechanism on the transport level (and hence -- no dead-letter queue).
 
-NOTE: If the destination endpoint uses different database or server instance, sending a message to it might cause the transaction to escalate to a distributed transaction. Usually it is not a desired effect and one can use NServiceBus Outbox to avoid it.
+NOTE: If the destination endpoint uses different database or server instance, sending a message to it might cause the transaction to escalate to a distributed transaction. Usually it is not a desired effect. Use NServiceBus [Outbox](/nservicebus/outbox/) to avoid it.
```
mikeminutillo: NOTE: Sending a message to an endpoint in a different database or server instance might cause the transaction to escalate to a distributed transaction. To avoid this, use the NServiceBus Outbox.

Out of interest, wouldn't the transaction between the servers still need to escalate when it pulls from the outbox and pushes to the receiver queue?

nservicebus/sqlserver/multiple-databases.md
```diff
@@ -14,24 +14,48 @@ The SQL Server transport allows you to select, on per-endpoint basis, where the
...
+By default the SQL Server transport uses a single instance of the SQL Server to maintain queues for all endpoints in the system. In order to send a message, an endpoint needs to connect to the (usually remote) database server and execute a SQL statement. The message is delivered directly to the destination queue without any store-and-forward mechanism. 
 
+Using a single database doesn't require Distributed Transaction Coordinator (MS DTC). Another advantage is the ability to take a snapshot of entire system state (all the queues) by backing up a database. This is most useful when the business data is also stored in the same database.
 
-## Database-per-endpoint
+## Single database with multiple schemas
 
-In a more complex scenario endpoints use separate databases and DTC is involved. Due to lack of store-and-forward mechanism, if a remote endpoint's database or DTC infrastructure is down, the endpoint cannot send messages to it. This potentially renders the endpoint (and all endpoints transitively depending on it) unavailable.
+The default schema used by SQL Server transport is `dbo`. A different schema might be specified using API:
```
mikeminutillo: > A different schema might be specified using API:

Sounds like a warning. What about:

> To specify a different schema use the following API:

nservicebus/sqlserver/multiple-databases.md
```diff
@@ -14,24 +14,48 @@ The SQL Server transport allows you to select, on per-endpoint basis, where the
...
  * Each endpoint has its own database where it stores both the queues and the user data
  * Messages are not sent immediately when calling `Bus.Send()` but are added to the *outbox* that is stored in the endpoint's own database. After completing the handling logic the messages in the *outbox* are forwarded to their destination databases
- * Should one of the forward operations fail, it will be retried by means of [standard NServiceBus retry mechanism](/nservicebus/errors/automatic-retries.md). This might result in some messages being sent more than once but it is not a problem because the outbox automatically handles the deduplication of incoming messages based on their ID.
+ * Should one of the forward operations fail, it will be retried by means of [standard NServiceBus retry mechanism](/nservicebus/errors/automatic-retries.md). This might result in some messages being sent more than once but Outbox automatically handles the deduplication of incoming messages based on their ID, providing `exactly-once` message delivery guarantee.
```
mikeminutillo: means of *the* standard NServiceBus retry mechansim

Outbox is lowercased and italicized above but capital cased here.

nservicebus/upgrades/sqlserver_2to3.md
```diff
@@ -0,0 +1,44 @@
...
+
+### Connection factory
+
+The custom connection factory method is now expected to be `async` and no parameters are passed to it by the framework:
+
+snippet:sqlserver-custom-connection-factory
+
+### Multi-schema support
+ 
+The API and config structure for multi-schema configuration [were changed](/nservicebus/sqlserver/multiple-databases.md#single-database-with-multiple schemas). The `Queue Schema` parameter is no longer supported neither in the config file nor the code configuration API.
```
mikeminutillo: I'd get rid of the neither-nor

> The `Queue Schema` parameter is no longer supported in the config file or the code configuration API.

nservicebus/upgrades/sqlserver_2to3.md
```diff
@@ -0,0 +1,44 @@
...
+snippet:sqlserver-singledb-multidb-pull 
+
+The latter is also used for specifying schemas for other endpoints that the configured endpoint communicates with.
+When using configuration file to specify schemas for other endpoints, their schemas should be placed in the `MessageEndpointMappings` section and follow `endpoint_name@schema_name` convention: 
+
+snippet:sqlserver-singledb-multischema-config
+
+### Circuit breaker
+
+The parameter `PauseAfterReceiveFailure(TimeSpan)` is no longer supported. The current pause value is 1 second.
```
mikeminutillo: > The current pause value is 1 second.

Implies that it can change but presumably not by the user. Can we be more explicit about this?

> In Version 3, the pause value is hard-coded at 1 second.

nservicebus/sqlserver/multiple-databases.md
weralabaj: My understanding is that we don't need that, because Outbox gives us at least once guarantee, not exactly once. But if you have some specific edge case in mind, might be good to consult with Szymon?

weralabaj:
Thanks a lot @mikeminutillo !

weralabaj:
@SimonCropp outside of scope of this issue, but related - should we add direct links to upgrade guides from the menu? I think right now the only way to get to them is via search, direct or related links, but not menu. Given that we'll have multiple pages now per each core version that will get messy.

nservicebus/sqlserver/multiple-databases.md
indualagarsamy: @mikeminutillo - Outbox has constraints. It needs to be in the same database as that of the business db. See http://docs.particular.net/nservicebus/outbox/#caveats.
What I am not sure is what happens when the outbox is trying to send it to different queues located in different databases. Is this scenario documented? For example, what happens if the handler transaction is ok but sending to one of these database didn't succeed due to datbase connection issues? @SzymonPobiega @mauroservienti ? But these are things that should in the outbox doco. 

@weralabaj - minor corrections on your note. I think we should call out the caveats as well. See below:

NOTE: If the destination endpoint uses a different database or server instance, sending a message to it might cause the transaction to escalate to a distributed transaction which may not be desirable. Using the [Outbox](/nservicebus/outbox/) feature, DTC escalations can be avoided as long as the both the endpoint's Outbox and Business data share the same database.

SzymonPobiega: I thought we don't support multi-db with outbox in V6? In V5 indeed you could use outbox with multi-DB to avoid DTC escalation.

tmasternak: @SzymonPobiega this part relates to V5

SzymonPobiega: Then :+1:

nservicebus/sqlserver/multiple-databases.md
```diff
@@ -14,24 +14,48 @@ The SQL Server transport allows you to select, on per-endpoint basis, where the
...
+By default the SQL Server transport uses a single instance of the SQL Server to maintain queues for all endpoints in the system. In order to send a message, an endpoint needs to connect to the (usually remote) database server and execute a SQL statement. The message is delivered directly to the destination queue without any store-and-forward mechanism. 
 
+Using a single database doesn't require Distributed Transaction Coordinator (MS DTC). Another advantage is the ability to take a snapshot of entire system state (all the queues) by backing up a database. This is most useful when the business data is also stored in the same database.
```
indualagarsamy: Looks like the disadvantage mentioned in the original statement has been removed. Is that intentional?
Also `By default` IMO is incorrect. The user configures the transport. They end up setting it up / using the same configuration for all endpoints. We don't dictate the default. 

Here are my suggestions:

Typically when using SQL Server transport, the endpoints are set up to use the same database for the storing the queues. Sending a message involves executing a SQL statement that results in delivering the message directly to the destination queue. The message is delivered directly without a store-and-forward mechanism. 

## Advantages
- As all the transactions involve the same database, the need for requiring Distributed Transaction Coordinator (MS DTC) can be eliminated. 
- A simple database backup can provide a complete snapshot of the entire system state, which includes business data and all of the messages in the queues. 

## Disadvantages
- Using a single database to store all of the queues may not always be possible especially on large systems. 
- Using schemas to distinguish logical data stores might be a good compromise for mid-size systems.

nservicebus/sqlserver/multiple-databases.md
```diff
@@ -14,24 +14,48 @@ The SQL Server transport allows you to select, on per-endpoint basis, where the
...
 
-## Database-per-endpoint
+## Single database with multiple schemas
 
-In a more complex scenario endpoints use separate databases and DTC is involved. Due to lack of store-and-forward mechanism, if a remote endpoint's database or DTC infrastructure is down, the endpoint cannot send messages to it. This potentially renders the endpoint (and all endpoints transitively depending on it) unavailable.
```
indualagarsamy: Why was this deleted? Is this no longer a concern?

nservicebus/sqlserver/multiple-databases.md
```diff
@@ -14,24 +14,48 @@ The SQL Server transport allows you to select, on per-endpoint basis, where the
...
+
+## Multiple databases with store-and-forward
 
 In order to overcome this limitation a higher level store-and-forward mechanism needs to be used. The Outbox feature can be used to effectively implement a distributed decoupled architecture where:
  * Each endpoint has its own database where it stores both the queues and the user data
- * Messages are not sent immediately when calling `Bus.Send()` but are added to the *outbox* that is stored in the endpoint's own database. After completing the handling logic the messages in the *outbox* are forwarded to their destination databases
- * Should one of the forward operations fail, it will be retried by means of [standard NServiceBus retry mechanism](/nservicebus/errors/automatic-retries.md). This might result in some messages being sent more than once but it is not a problem because the outbox automatically handles the deduplication of incoming messages based on their ID.
+ * Messages are not sent immediately when calling `Bus.Send()` but are added to the Outbox table that is stored in the endpoint's own database. After completing the handling logic the messages stored in the Outbox table are forwarded to their destination databases
```
indualagarsamy: mention the expected action first. Here's my suggestion:

When calling `Bus.Send()`, messages are stored in the Outbox table rather than getting dispatched immediately. The Outbox table is a database table that resides on the same database as that of the endpoint. After successful execution of the message handler logic, the messages stored in the Outbox table are forwarded to their destinations.

indualagarsamy: Since we are talking about how the outbox actually works, I am wondering if we also should link it to the correct section of the outbox doco here.  Thoughts?

nservicebus/sqlserver/multiple-databases.md
```diff
@@ -14,24 +14,48 @@ The SQL Server transport allows you to select, on per-endpoint basis, where the
...
+ * Should one of the forward operations fail, it will be retried by means of the [standard NServiceBus retry mechanism](/nservicebus/errors/automatic-retries.md). This might result in some messages being sent more than once but Outbox automatically handles the deduplication of incoming messages based on their ID, providing `exactly-once` message delivery guarantee.
```
indualagarsamy: I think rather than one of the forward operations, it should be any of the forward operations.

Here's a suggestion:

Should any of the forward operations were to fail, it will be retried using the standard [retry mechanism](/nservicebus/errors/automatic-retries.md). However, this may result in some messages to be sent multiple times. To mitigate this and to provide `exactly-once` message delivery guarantee, the Outbox feature automatically handles the de-duplication of incoming messages based on their ID.

nservicebus/upgrades/sqlserver_2to3.md
```diff
@@ -0,0 +1,44 @@
...
+ - migration
+related:
+- nservicebus/sagas/concurrency
+---
+
+
+## SQL Server Transport
+
+### Transactions
+Transaction settings are now handled in the transport level configuration. For details refer to [Transaction configuraton API](/nservicebus/upgrades/5to6.md#transaction-configuration-API) section.
```
indualagarsamy: > Transaction settings are now handled in the transport level configuration.

What does this mean? Its not too clear.

nservicebus/upgrades/sqlserver_2to3.md
```diff
@@ -0,0 +1,44 @@
...
+
+The native transaction support has been split into two different levels: `ReceiveOnly` and `SendAtomicWithReceive`. SQL Server Transport supports both of them. `SendAtomicWithReceive` is equivalent to disabling distributed transactions in Version 5. For more details refer to [Transaction configuraton API](/nservicebus/upgrades/5to6.md#transaction-configuration-API) section and [Transaction support](/nservicebus/messaging/transactions.md) page.
```
indualagarsamy: Seems like the below should be in SqlTransport doco?
> The native transaction support has been split into two different levels: `ReceiveOnly` and `SendAtomicWithReceive`. SQL Server Transport supports both of them.

How about, having a section called, 

#### Disabling Distributed Transactions
Use `SendAtomicWithReceive` option and show usage of that?

To me that's relevant. For example I was using the previous version and was disabling DTC, how do I do that in the new version. How would I have to change my code. To me, that info belongs in the upgrade guide. The rest can be a doco link.

nservicebus/upgrades/sqlserver_2to3.md
```diff
@@ -0,0 +1,44 @@
...
+
+### Connection factory
+
+The custom connection factory method is now expected to be `async` and no parameters are passed to it by the framework:
```
indualagarsamy: >The custom connection factory method is now expected to be `async`

is there a link to any doco that talks about Custom Connection Factory for the existing version? can we link it to that?

nservicebus/upgrades/sqlserver_2to3.md
```diff
@@ -0,0 +1,44 @@
...
+snippet:sqlserver-singledb-multidb-pull 
+
+The latter is also used for specifying schemas for other endpoints that the configured endpoint communicates with.
+When using configuration file to specify schemas for other endpoints, their schemas should be placed in the `MessageEndpointMappings` section and follow `endpoint-name@schema-name` convention: 
+
+snippet:sqlserver-singledb-multischema-config
+
+### Circuit breaker
+
+The parameter `PauseAfterReceiveFailure(TimeSpan)` is no longer supported. In Version 3, the pause value is hard-coded at 1 second.
```
indualagarsamy: I am wondering if we should mention the 
> In Version 3, the pause value is hard-coded at 1 second.

Can we link to any doco that talks about the reasons as to why we discontinued this?.

nservicebus/upgrades/sqlserver_2to3.md
```diff
@@ -0,0 +1,44 @@
...
+
+The custom connection factory method is now expected to be `async` and no parameters are passed to it by the framework:
+
+snippet:sqlserver-custom-connection-factory
+
+### Multi-schema support
+ 
+The API and config structure for multi-schema configuration [were changed](/nservicebus/sqlserver/multiple-databases.md#single-database-with-multiple schemas). The `Queue Schema` parameter is no longer supported in the config file and the code configuration API. 
+
+The schema for the current endpoint can be specified using `DefaultSchema` method:
```
indualagarsamy: > The schema for the current endpoint 
Does this mean current as in version?

nservicebus/upgrades/sqlserver_2to3.md
```diff
@@ -0,0 +1,44 @@
...
+
+snippet:sqlserver-singledb-multischema
+
+or by defining a custom connection factory:
+
+snippet:sqlserver-singledb-multidb-pull 
+
+The latter is also used for specifying schemas for other endpoints that the configured endpoint communicates with.
```
indualagarsamy: What's the latter in this context?

nservicebus/upgrades/sqlserver_2to3.md
```diff
@@ -0,0 +1,44 @@
...
+
+### Connection factory
+
+The custom connection factory method is now expected to be `async` and no parameters are passed to it by the framework:
+
+snippet:sqlserver-custom-connection-factory
+
+### Multi-schema support
+ 
+The API and config structure for multi-schema configuration [were changed](/nservicebus/sqlserver/multiple-databases.md#single-database-with-multiple schemas). The `Queue Schema` parameter is no longer supported in the config file and the code configuration API.
```
indualagarsamy: Instead of:
> The API and config structure for multi-schema configuration [were changed](/nservicebus/sqlserver/multiple-databases.md#single-database-with-multiple schemas). 

how about:
The configuration API for enabling [multi-schema support](/nservicebus/sqlserver/multiple-databases.md#single-database-with-multiple schemas) has now changed.

nservicebus/upgrades/sqlserver_2to3.md
```diff
@@ -0,0 +1,44 @@
...
+summary: Instructions on how to upgrade from NServiceBus Versions 5 to 6
+tags:
+ - upgrade
+ - migration
+related:
+- nservicebus/sagas/concurrency
+---
+
+
+## SQL Server Transport
```
indualagarsamy: Should this receive a mention?
https://github.com/Particular/docs.particular.net/pull/1206/files#r54628162

nservicebus/sqlserver/multiple-databases.md
weralabaj: @indualagarsamy if you think we need more clarification regarding Outbox, I think that should be moved to outbox doco. I don't know it that well and this doesn't seem to be right place to discuss. I use your suggestion for text.

nservicebus/sqlserver/multiple-databases.md
weralabaj: > We don't dictate the default.

I don't agree with that, especially given that we're getting rid of multi-instance setup in v3. If you want to use multi-instance you need to have a good reason to do so, because that causes a lot of complications. We tried to hide it from the user in v2 and now are not so sure it was the right approach. But the multi-db doco will be refined anyway as part of the work on multi-instance samples and guidance.

The advantages/disadvantages of each setup should be part of the guidance content, and I wasn't clear on why and in what circumstances we'd recommend each setup. E.g. you can also eliminate the need for DTC if you use Outbox, so it's not the only option.

I'd rather leave it out of scope for now and cover when we have multi-instance part ready.

weralabaj: > Looks like the disadvantage mentioned in the original statement has been removed. Is that intentional?

Which one?

nservicebus/sqlserver/multiple-databases.md
weralabaj: It's still there, starts in line 51. It's just limits of GH changes rendering.

nservicebus/sqlserver/multiple-databases.md
weralabaj: Done.

nservicebus/upgrades/sqlserver_2to3.md
weralabaj: I didn't want to repeat here the content from the other doco, so only linking to more details. I think once they go the referred article, it should be clear what's the change. Not sure how to explain it here without copying content though :(

nservicebus/upgrades/sqlserver_2to3.md
weralabaj: Again, I wanted to avoid repetition. Do you think it's worth duplicating snippets in that case?

> I was using the previous version and was disabling DTC, how do I do that in the new version. How would I have to change my code.

> `SendAtomicWithReceive` is equivalent to disabling distributed transactions in Version 5

Is the problem that the statement above is not precise enough or that the formatting could be better?

nservicebus/sqlserver/multiple-databases.md
tmasternak: @indualagarsamy I think the note should say:
 * If input queue, error queue, outbox and business data are on the same db instance user can avoid escalating to DTC
 * If handler produces at least two messages which are destined to queues on different db instances endpoint will escalate to DTC

nservicebus/upgrades/sqlserver_2to3.md
weralabaj: Not that I know of, but we have this issue still open https://github.com/Particular/NServiceBus.SqlServer/issues/136.

Apart from snippet, what would you expect to find in that doco?

nservicebus/upgrades/sqlserver_2to3.md
weralabaj: > I am wondering if we should mention the

I assumed that if somebody was using this option before then it might be useful information, but no hard feelings about it, can be removed.

> Can we link to any doco that talks about the reasons as to why we discontinued this?

I asked around but didn't get the straight answer as to why it was removed, nor why it was there in the first place and whether that was useful. I'd really appreciate if you have more info on the latter at least.

nservicebus/upgrades/sqlserver_2to3.md
weralabaj: After adding more comments it's difficult to find out what you're referring to. Could you copy that fragment here?

weralabaj:
@indualagarsamy I've changed some of the things, as you suggested. Where I wasn't sure I commented, have another look :)

nservicebus/sqlserver/multiple-databases.md
```diff
@@ -14,24 +14,48 @@ The SQL Server transport allows you to select, on per-endpoint basis, where the
...
+By default the SQL Server transport uses a single instance of the SQL Server to maintain queues for all endpoints in the system. In order to send a message, an endpoint needs to connect to the (usually remote) database server and execute a SQL statement. The message is delivered directly to the destination queue without any store-and-forward mechanism.
```
indualagarsamy: @weralabaj - I don't agree with the `By default` as mentioned here. The Sql server transport doesn't do that by default. The connection string for the transport is provided by the user as part of the endpoint configuration. The users often provide the same connection string for transports for all endpoints. Just wanted to make that more clear. 

>By default the SQL Server transport uses a single instance of the SQL Server to maintain queues for all endpoints in the system. In order to send a message, an endpoint needs to connect to the (usually remote) database server and execute a SQL statement. The message is delivered directly to the destination queue without any store-and-forward mechanism. 

How about:
Typically when using SQL Server transport, the endpoints are set up to use the same database for the storing the queues. Sending a message involves executing a SQL statement that results in delivering the message directly to the destination queue. The message is delivered directly without a store-and-forward mechanism.

nservicebus/sqlserver/multiple-databases.md
indualagarsamy: @weralabaj - I am good with the advantages / disadvantages stuff being part of #1103.

nservicebus/upgrades/sqlserver_2to3.md
indualagarsamy: @weralabaj - have a look here: http://docs.particular.net/nservicebus/upgrades/5to6#istartablebus-and-bus-static-class-are-obsolete
The snippets for both version show the user what the API was in the older version and how its changing. IMO that info is valuable in the upgrade guide. And its much better looking at the usage.

nservicebus/upgrades/sqlserver_2to3.md
indualagarsamy: @weralabaj - Not saying we should repeat content. You can always link to more details, but the first statement seems unclear. What are we trying to highlight to the user? Maybe we can come up with something better.  

Are we trying to say that with NServiceBus Versions 6 and above, the transaction API has a changed. And how it impacts the new version of SQL Server transport and the transaction scope settings?

nservicebus/upgrades/sqlserver_2to3.md
indualagarsamy: @weralabaj - I am good with that.

nservicebus/upgrades/sqlserver_2to3.md
indualagarsamy: @tmasternak - who's the best person to answer @weralabaj and my question above as to why `PauseAfterReceiveFailure(Timespan)` is no longer supported?Can you point me to the right person or any documentation around this that we could point to in the upgrade guide? cc / @andreasohlund

indualagarsamy:
@weralabaj - awesome work. I've answered your questions I believe :).

nservicebus/sqlserver/multiple-databases.md
weralabaj: OK, I'll adjust :)

nservicebus/upgrades/sqlserver_2to3.md
weralabaj: Ok, see if it's better now?

nservicebus/upgrades/sqlserver_2to3.md
weralabaj: I'll raise it as a question in core. That should be documented also in the "main upgrade guide". In SQL we removed it because we copied code from core.

weralabaj:
Can that be merged?

weralabaj:
@indualagarsamy can that be merged?

indualagarsamy:
@weralabaj - looks good. Feel free to merge.

SimonCropp:
ok i merge manually here https://github.com/Particular/docs.particular.net/commit/e35fc37d98f5ee80e23d365ebfc127bab9e8b928

@indualagarsamy @weralabaj thanks

PR #1211, opened 2016-02-24
Changing from MessageToSend to MessageTypeToSend to clarify the argum…

mauroservienti:
…ent meaning

### DIFF
--- a/Snippets/Operations/SqlServer/NativeSend.cs
+++ b/Snippets/Operations/SqlServer/NativeSend.cs
@@ -20,7 +20,7 @@
                 messageBody: "{\"Property\":\"PropertyValue\"}",
                 headers: new Dictionary<string, string>
                 {
-                    {"NServiceBus.EnclosedMessageTypes", "MessageToSend"}
+                    {"NServiceBus.EnclosedMessageTypes", "MessageTypeToSend"}
                 }
                 );
 


SimonCropp:
I think expand it in the markdown

indualagarsamy:
@SimonCropp - I still think the snippet is incorrect. IMO, if you're providing a snippet, it ought to work. Users shouldn't have to try and dig into why the snippet is wrong.

mauroservienti:
The snippet is just text that does nothing on it's own. I'll propose a change in the doco page to address the meaning of the message type and the implications when using, or not, a full type name.

indualagarsamy:
>The snippet is just text that does nothing on it's own. 

Then why have compilable snippets in the first place. When a user takes the snippet we provide it should work. Would the snippet work, if the user pulled the code from our doco page and used it or would the code result in exceptions are somethings to think about. 

>I'll propose a change in the doco page to address the meaning of the message type and the implications when using, or not, a full type name.

I agree on the detailed explanations on the snippet themselves should belong in the correct documentation. So, if the documentation doesn't address that we ought to fix it. I believe the related doco page for this is http://docs.particular.net/nservicebus/messaging/message-owner

mauroservienti:
> Then why have compilable snippets in the first place. When a user takes the snippet we provide it should work. Would the snippet work, if the user pulled the code from our doco page and used it or would the code result in exceptions are some things to think about.

As far as I can tell there is nothing in that page that mentions a message type so copying that snippet cannot work in any way, if we want to provide a meaningful `MessageTypeToSend` parameter than we should provide the message class as well, otherwise providing something like `MyNamespace.Message, MyAssembly` is as useless as `MessageTypeToSend`.

Or am I missing something?

Issue #1168, opened 2016-02-12
add v6 store-and-forward


SimonCropp:
Connects to https://github.com/Particular/NServiceBus.SqlServer/issues/139

## Plan of Attack
 - [x] Update v5 and v6 samples [PR](https://github.com/Particular/docs.particular.net/pull/1362)

SimonCropp:
@SzymonPobiega can u have a look at this one?

SzymonPobiega:
@tmasternak the purpose of this sample was to show how to add store-and-forward feature to a transport that does not have it. The main target was SQL Server transport in multi-DB mode. With this sample, instead of sending messages directly to the other database, we would send it to ourselves and use outbox to push it to the other DB.

Since we are no longer supporting multi-DB/outbox combo in V6 (at least for now), does this sample still make sense? It does not make much sense in context of other transports (in this form).

SzymonPobiega:
The alternative we can provide is a sample where we enable the outbox for the messages sent from `IMessageSession` but since we have not yet decided the fate of the outbox, I'd leave it for now.

johnsimons:
Started the v6 version in #1216, if this does not go anywhere feel free to delete PR

tmasternak:
@SzymonPobiega correct we decided to drop multi-DB/outbox combo in v6 so the sample does not really make sense.

timbussmann:
@SzymonPobiega @tmasternak so can we close this since it wouldn't make sense on v6 as far as I understood you?

SzymonPobiega:
@timbussmann :+1: for closing. I am not sure if we manage to provide a replacement for it in the 6.0 time frame but the proper replacement (which might come a bit later) would be allowing the outbox to be used with the message session unit of work. We discussed that with @andreasohlund recently. You would pass the dedup ID to the method that opens the session.

SimonCropp:
@timbussmann or @SzymonPobiega can u please add some text that explains why there is no v6 version. or perhaps link to some upgrade guide or other page that handles that explanation?

tmasternak:
@SimonCropp I think that we don't have a good place to link to yet. This should change this week. I've added this PR to PoA for Multi-instance in v6 issue and will track it from there.

SimonCropp:
@tmasternak thanks

weralabaj:
@SzymonPobiega @tmasternak Since I'm missing some context - do we now have the info we need?

tmasternak:
[...]

PR #1216, opened 2016-02-25
[WIP] V6 store and forward

johnsimons:
This deals with #1168 

Stuck on how to access the send options in v6, have no idea how to do it

### DIFF
--- a/samples/outbox/store-and-forward/Version_6/Receiver/OrderSubmittedHandler.cs
+++ b/samples/outbox/store-and-forward/Version_6/Receiver/OrderSubmittedHandler.cs
@@ -1,0 +1,15 @@
+using System;
+using System.Threading.Tasks;
+using NServiceBus;
+
+public class OrderSubmittedHandler : IHandleMessages<OrderSubmitted>
+{
+    public Task Handle(OrderSubmitted message, IMessageHandlerContext context)
+    {
+        Console.WriteLine("Order {0} worth {1} submitted", message.OrderId, message.Value);
+        return context.Reply(new OrderAccepted
+        {
+            OrderId = message.OrderId,
+        });
+    }
+}
--- a/samples/outbox/store-and-forward/Version_6/Receiver/Program.cs
+++ b/samples/outbox/store-and-forward/Version_6/Receiver/Program.cs
@@ -1,0 +1,35 @@
+﻿using System;
+using System.Threading.Tasks;
+using NServiceBus;
+using NServiceBus.Features;
+
+class Program
+{
+    static void Main()
+    {
+        AsyncRun().GetAwaiter().GetResult();
+    }
+
+    static async Task AsyncRun()
+    {
+        #region ReceiverConfiguration
+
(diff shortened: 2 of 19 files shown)


SimonCropp:
@timbussmann can u comment on this

timbussmann:
`SendOptions` are not available in the pipeline. All settings configured on the `SendOptions` are merged into the pipeline context (e.g. see here: https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Pipeline/Outgoing/OutgoingSendContext.cs#L17).

johnsimons:
@timbussmann I need access to the send type, so I can distinguish whether is a publish or a send ?
This is how we used to do it in v5:
```c#
SendOptions sendOptions = context.DeliveryOptions as SendOptions;
 +        if (sendOptions != null)
 +        {
 +            context.OutgoingLogicalMessage.Headers["$.store-and-forward.destination"] =
 +                sendOptions.Destination.ToString();
 +            sendOptions.Destination = configure.LocalAddress;
 +            //We could as well store other properties of the SendOptions to handle things like delayed delivery
 +        }
 +        else
 +        {
 +            PublishOptions publishOptions = context.DeliveryOptions as PublishOptions;
 +            if (publishOptions != null)
 +            {
 +                //Technically we don't need to store tha actual type, just a marker that this is a Publish operation
 +                context.OutgoingLogicalMessage.Headers["$.store-and-forward.eventtype"] =
 +                    publishOptions.EventType.AssemblyQualifiedName;
 +            }
 +            else
 +            {
 +                //We should never get here as is makes no sense to reply from outside of a handler
 +                throw new NotSupportedException("Not supported delivery option: " + context.DeliveryOptions.GetType().Name);
 +            }
 +        }
```

How can I achieve the same in v6 ?

timbussmann:
can you use `IOutgoingLogicalMessageContext.Message.MessageType`?

johnsimons:
@timbussmann 

> can you use IOutgoingLogicalMessageContext.Message.MessageType

Not sure, u tell me ?

timbussmann:
@johnsimons it should work. Since it's your PR, you should be able to test it yourself?

johnsimons:
@timbussmann I have tested it and still stuck

I assume this is what u mean:
```
if (conventions.IsEventType(context.Message.MessageType))
```
Is the `context.Message.MessageType` the eventtype or the proxy type ?

Also I need to get access to the destination, how do I get that ?

johnsimons:
@timbussmann I really think u should take over this PR, it is not much left to do.

timbussmann:
@johnsimons what about this comment in the linked issue here:

> correct we decided to drop multi-DB/outbox combo in v6 so the sample does not really make sense.

do we still need this sample at all then?

johnsimons:
@timbussmann not my call, as I said in https://github.com/Particular/docs.particular.net/issues/1168#issuecomment-188621568

timbussmann:
closing this as #1168 was closed to

PR #1098, opened 2016-01-28
Section on things to think about when using ServiceControl

petersgiles:
related to https://github.com/Particular/docs.particular.net/issues/1097

@dvdstelt @johnsimons check out the section at the bottom of this...needs more work but I think it worth getting out there as I have been seeing a few cases where this advice might have prevented some angst for our customers

PR #1217, opened 2016-02-25
[WIP] Mike's take on ServiceControl in Practice

mikeminutillo:
This is a modified version of https://github.com/Particular/docs.particular.net/pull/1098#event-564935177

Created a new branch to allow side-by-side comparison and discussion.

I moved from "Features" to "Capabilities and Plugins". I also modified the tone to try and remove opinion-sounding things. Structurally I tried to present the problem, propose an approach and finish with specific tips.

@Particular/docs-maintainers @petersgiles - take a look and see if you agree with this approach?

### DIFF
--- a/servicecontrol/servicecontrol-in-practice.md
+++ b/servicecontrol/servicecontrol-in-practice.md
@@ -19,7 +19,9 @@ Here are some important considerations when installing and deploying ServiceCont
 
 - Read the [Capacity Planning](/servicecontrol/capacity-and-planning.md) and [Troubleshooting](/servicecontrol/troubleshooting.md) guides for ServiceControl.
 - Each environment should have a dedicated ServiceControl instance.
-- In production, ServiceControl should run on a dedicated machine.
+- Other applications or NServiceBus endpoints running on the same machine as ServiceControl can compete for hardware resources and therefore negatively impact ServiceControl's performance. For optimal performance run SC on a dedicated machine.
+- If ServiceControl is installed on a Virtual Machine, ensure the machine is capable of high levels of I/O traffic. The amount of I/O required depends on the system being monitored. Run performance 
+tests with realistic message loads to baseline hardware requirements.
 
 
 ## Message Throughput Considerations
@@ -60,7 +62,7 @@ Once this baseline has been established, follow these steps:
 
 ## Anti-Virus Checks
 
-Exclude the ServiceControl [database directory](/servicecontrol/configure-ravendb-location.md) from anti-virus checks. ServiceControl uses an embedded database and produces a lot of I/O traffic. Some anti-virus software can add overhead to I/O operations causing a significant performance impact.
+Exclude the ServiceControl [database directory](/servicecontrol/configure-ravendb-location.md) from anti-virus checks. ServiceControl uses an embedded database and produces a high level of I/O traffic. Some anti-virus software can add overhead to I/O operations causing a significant performance impact.
 
 
 ## Version Downgrades
--- a/servicecontrol/servicecontrol-in-practice.md
+++ b/servicecontrol/servicecontrol-in-practice.md
@@ -17,17 +17,17 @@ Here are some important considerations when installing and deploying ServiceCont
 
 ## Hardware & Installation Considerations:
 
-- Read the [Capacity Planning](/servicecontrol/capacity-and-planning.md) and [Troubleshooting](/servicecontrol/troubleshooting.md) guides for ServiceControl.
-- Each environment should have a dedicated ServiceControl instance.
-- Other applications or NServiceBus endpoints running on the same machine as ServiceControl can compete for hardware resources and therefore negatively impact ServiceControl's performance. For optimal performance run SC on a dedicated machine.
-- If ServiceControl is installed on a Virtual Machine, ensure the machine is capable of high levels of network and disk I/O traffic. The amount of I/O required depends on the system being monitored, the number of messages being processed, the transport being used, etc. Run performance tests with realistic message loads to baseline hardware requirements.
+ * Read the [Capacity Planning](/servicecontrol/capacity-and-planning.md) and [Troubleshooting](/servicecontrol/troubleshooting.md) guides for ServiceControl.
+ * Each environment should have a dedicated ServiceControl instance.
+ * Other applications or NServiceBus endpoints running on the same machine as ServiceControl can compete for hardware resources and therefore negatively impact ServiceControl's performance. For optimal performance run SC on a dedicated machine.
+ * If ServiceControl is installed on a Virtual Machine, ensure the machine is capable of high levels of network and disk I/O traffic. The amount of I/O required depends on the system being monitored, the number of messages being processed, the transport being used, etc. Run performance tests with realistic message loads to baseline hardware requirements.
 
 
 ## Message Throughput Considerations
 
-Depending on the number of endpoints and message volume, audit messages can have a large impact on performance. Turn off [Message Auditing](/nservicebus/operations/auditing.md#turning-off-auditing) if it is not needed. The primary reason for the audit ingestion capability of ServiceControl is to support system analysis with ServiceInsight. If ServiceInsight is not in use then turn off messaging auditing for each endpoint. Message auditing may be important for some endpoints but not others. 
+Depending on the number of endpoints and message volume, audit messages can have a large impact on performance. Turn off [Message Auditing](/nservicebus/operations/auditing.md#turning-off-auditing) if it is not needed. The primary reason for the audit ingestion capability of ServiceControl is to support system analysis with ServiceInsight. If ServiceInsight is not in use then turn off messaging auditing for each endpoint. Message auditing may be important for some endpoints but not others.
 
-NOTE: If message auditing is required without the use of ServiceInsight, configure endpoints and ServiceControl to use different audit queues. Audit messages going to an audit queue that is not managed by ServiceControl will need to be cleaned up manually. 
+NOTE: If message auditing is required without the use of ServiceInsight, configure endpoints and ServiceControl to use different audit queues. Audit messages going to an audit queue that is not managed by ServiceControl will need to be cleaned up manually.
 
 Turn off [Audit Forwarding](/servicecontrol/errorlog-auditlog-behavior.md) if it is not needed. ServiceControl sends a copy of each audited message to configured Audit Forwarding queue. If these messages are not being used, turn this feature off.
 
@@ -39,14 +39,14 @@ Turn off [Audit Forwarding](/servicecontrol/errorlog-auditlog-behavior.md) if it
 
 ### Heartbeats
 
-Not all endpoints are mission critical and need to be monitored with [heartbeats](/servicepulse/intro-endpoints-heartbeats.md) using the same SLA. For endpoints that are less critical to business, [adjust the Heartbeat Interval](/servicecontrol/plugins/heartbeat#configuration-heartbeat-interval). Increasing the interval ensures that ServiceControl is able to process heartbeats in a timely manner. Increasing the heartbeat interval for endpoints requires a corresponding increase in the [heartbeat grace period](/servicecontrol/creating-config-file#plugin-specific-servicecontrol-heartbeatgraceperiod) in ServiceControl. 
+Not all endpoints are mission critical and need to be monitored with [heartbeats](/servicepulse/intro-endpoints-heartbeats.md) using the same SLA. For endpoints that are less critical to business, [adjust the Heartbeat Interval](/servicecontrol/plugins/heartbeat.md#configuration-heartbeat-interval). Increasing the interval ensures that ServiceControl is able to process heartbeats in a timely manner. Increasing the heartbeat interval for endpoints requires a corresponding increase in the [heartbeat grace period](/servicecontrol/creating-config-file.md#plugin-specific-servicecontrol-heartbeatgraceperiod) in ServiceControl.
 
-Heartbeat messages tend to be frequent, and a large backlog can occur if ServiceControl is offline for an extended period. When this happens, it can take ServiceControl some time to process old heartbeats when it restarts. 
+Heartbeat messages tend to be frequent, and a large backlog can occur if ServiceControl is offline for an extended period. When this happens, it can take ServiceControl some time to process old heartbeats when it restarts.
 
 
 ### Saga Audit
 
-The [Saga Audit](/servicecontrol/plugins/saga-audit.md) plugin produces a lot of data. It's use outside of a development environment is not recommended. If the saga audit plugin is already deployed in production, [follow the instructions outlined](/servicecontrol/plugins/saga-audit#removing-the-plugin-from-production) to remove it.
+The [Saga Audit](/servicecontrol/plugins/saga-audit.md) plugin produces a lot of data. It's use outside of a development environment is not recommended. If the saga audit plugin is already deployed in production, [follow the instructions outlined](/servicecontrol/plugins/saga-audit.md#removing-the-plugin-from-production) to remove it.
 
 
 ## Performance Considerations
@@ -54,9 +54,10 @@ The [Saga Audit](/servicecontrol/plugins/saga-audit.md) plugin produces a lot of
 Run a performance test using the expected peak and average message throughput for the environment to baseline the system. The baseline test should not include audit ingestion or any ServiceControl plugins. 
 
 Once this baseline has been established, follow these steps:
-- Install and configure the Heartbeat plugin in each endpoint where it is needed. Re-run the performance test suite and monitor ServiceControl to ensure that it can effectively monitor the system under load. This may require adjustments to the Heartbeat interval. Re-run the performance tests after each adjustment.
-- Turn on auditing for each endpoint that needs it and re-run the performance tests to assess impact.
-- For each endpoint, turn on any required Custom Checks and re-run the performance tests to assess impact.
+
+ * Install and configure the Heartbeat plugin in each endpoint where it is needed. Re-run the performance test suite and monitor ServiceControl to ensure that it can effectively monitor the system under load. This may require adjustments to the Heartbeat interval. Re-run the performance tests after each adjustment.
+ * Turn on auditing for each endpoint that needs it and re-run the performance tests to assess impact.
+ * For each endpoint, turn on any required Custom Checks and re-run the performance tests to assess impact.
 
 When an infrastructure outage occurs in a production environment it's possible that every message processed every endpoint may end up in the error queue. It can take ServiceControl some time to ingest all of these messages. Once ingested, a bulk retry operation will consume additional network and disk I/O above the usual requirements. It is important to simulate these conditions as a part of performance testing to ensure that these times and resources are catered for in recovery plans.
  
--- a/servicecontrol/servicecontrol-in-practice.md
+++ b/servicecontrol/servicecontrol-in-practice.md
@@ -1,40 +1,51 @@
 ---
 title: ServiceControl in Practice
-summary: Considerations for configuring ServiceControl for use in production environments
+summary: Optimizing ServiceControl for use in different environments
 tags:
 - ServiceControl
 ---
 
 
-ServiceControl has many capabilities which can be extended by adding optional plugins into the endpoints of the system being monitored. Each capability and plugin can provide valuable information but also comes with a cost in terms of resource usage and performance. There are other factors that can influence the resource and performance profile of ServiceControl including hardware, message throughput, and number of endpoints. These factors can vary greatly between environments. Capabilities and plugins that provide value in one environment may have a negative impact if included in another environment.
+ServiceControl provides many capabilities such as endpoint monitoring, advanced debgugging, and failed message management. These capabilities can be extended by adding optional [plugins](http://docs.particular.net/servicecontrol/plugins/) into the endpoints being monitored. Each capability and plugin provides valuable information, but they have certain resource and performance costs. Harware, peak and average message throughput, and number of endpoints in the system all have a large impact on the performance of ServiceControl. These factors can vary greatly between environments. Capabilities and plugins that provide value in one environment may have a negative impact if included in another environment.
 
-For example, the [Saga Audit](/servicecontrol/plugins/saga-audit.md) plugin provides additional information to support a development environment where message load is low. In a production environment, where are there are many more saga instances to audit, the increased overhead is magnified and can have a significant performance impact.  
+For example, the [Saga Audit](/servicecontrol/plugins/saga-audit.md) plugin provides additional information to support a development environment where message load is low. In a production environment, where there are many more saga instances to audit, the increased overhead is magnified and can have a significant performance impact.  
 
 Each capability and plugin needs to be considered for each environment to determine if the value that it provides in that environment outweighs the costs that it imposes. 
 
+Here are some important considerations when installing and deploying ServiceControl for better performance.
 
-## General Approach
 
-When configuring ServiceControl for a new environment follow these steps:
+## Hardware & Installation Considerations:
 
 - Read the [Capacity Planning](/servicecontrol/capacity-and-planning.md) and [Troubleshooting](/servicecontrol/troubleshooting.md) guides for ServiceControl.
-- Turn off Auditing and remove ServiceControl plugins from Endpoints in the new environment.
-- Run a performance test using the expected peak and average message throughput for the environment to baseline the system. This test suite should include simulating infrastructure outages that will cause large numbers of message processing failures. 
-- For each Endpoint, install and configure the Heartbeat plugin (if needed). Re-run the performance test suite and monitor ServiceControl to ensure that it is able to effectively monitor the system under load. This may require adjustments to the Heartbeat Interval. Re-run the performance tests after each adjustment.
-- For each endpoint, turn on Auditing (if needed) and re-run the performance tests to determine impact.
-- For each endpoint, turn on any required Custom Checks and re-run the performance tests to determine impact.
-
-NOTE: There are some factors that influence ServiceControl resource usage and performance that are less likely to change between environments but are specific to the system being monitored such as choice of transport and average message size. 
+- Each enviornment should have a dedicated ServiceControl instance.
+- In production, ServiceControl should run on a dedicated machine.
 
 
-## Specific Considerations
+## Message Throughput Considerations
 
-- Each environment should have a dedicated ServiceControl instance.
-- In production, ServiceControl should run on a **dedicated machine**.
-- Turn off [Message Auditing](/nservicebus/operations/auditing.md) if it is not needed. The audit ingestion capability of ServiceControl is primarily to support system analysis with ServiceInsight. If ServiceInsight is not deployed in an environment then configure ServiceControl to read audit messages from an empty queue or turn off messaging auditing for each endpoint. Message auditing may be important for some endpoints but not other others. 
+- Depending on the number of endpoints and message volume, audit messages can have a large impact on performance. Turn off [Message Auditing](/nservicebus/operations/auditing.md) if it is not needed. The primary reason for the audit ingestion capability of ServiceControl is to support system analysis with ServiceInsight. If ServiceInsight is not deployed in an environment then configure ServiceControl to read audit messages from an empty queue or turn off messaging auditing for each endpoint. Message auditing may be important for some endpoints but not others. 
 - Turn off [Audit Forwarding](/servicecontrol/errorlog-auditlog-behavior.md) if it is not needed. ServiceControl sends a copy of each audited message to configured Audit Forwarding queue. If these messages are not being used, turn this feature off.
-- Plugins communicate between the Endpoint they installed in to ServiceControl using messaging and the configured transport of your system. Each instance of a plugin adds more messages to the ServiceControl queue which can delay the processing of each message and make ServiceControl less responsive.
-  - [Heartbeats](/servicepulse/intro-endpoints-heartbeats.md) - Not all endpoints are mission critical and need to be monitored with heartbeats. For Endpoints that do need to be monitored, adjust the Heartbeat Interval. Increasing the interval ensures that ServiceControl is able to process heartbeats in a timely manner but also increases the window that an Endpoint can be unavailable for before being detected by ServiceControl. Heartbeat messages tend to be frequent and a large backlog can occur if ServiceControl is offline for an extended period. When this happens, it can take ServiceControl some time to process old heartbeats when it restarts. 
-  - [Saga Audit](/servicecontrol/plugins/saga-audit.md) - the saga audit plugin produces a lot of data. It's use outside of a development environment is not recommended.
-- Exclude the ServiceControl [database directory](/servicecontrol/configure-ravendb-location.md) from anti-virus checks.
-- Do not downgrade releases of ServiceControl. ServiceControl uses an embedded database and changes to the internal data structures can occur between releases. Rolling back may cause index corruption or data loss. Perform testing in a lower environment before upgrading in production environments.
+- [Plugins](http://docs.particular.net/servicecontrol/plugins/) are installed in an Endpoint and  sned data to ServiceControl. This communication uses messaging over the configured transport of the endpoint. Each instance of a plugin adds more messages to the ServiceControl queue which can delay the processing of each message and make ServiceControl less responsive.
+  - [Heartbeats](/servicepulse/intro-endpoints-heartbeats.md) - Not all endpoints are mission critical and need to be monitored with heartbeats using the same SLA. For endpoints that are less critical to business, [adjust the Heartbeat Interval](http://docs.particular.net/servicecontrol/plugins/heartbeat#configuration-heartbeat-interval). Increasing the interval ensures that ServiceControl is able to process heartbeats in a timely manner but also increases the window that an Endpoint can be unavailable for before being detected by ServiceControl. Heartbeat messages tend to be frequent, and a large backlog can occur if ServiceControl is offline for an extended period. When this happens, it can take ServiceControl some time to process old heartbeats when it restarts. 
+  - [Saga Audit](/servicecontrol/plugins/saga-audit.md) - the saga audit plugin produces a lot of data. It's use outside of a development environment is not recommended. If the saga audit plugin is already deployed in production, [follow the instructions outlined](http://docs.particular.net/servicecontrol/plugins/saga-audit#removing-the-plugin-from-production) to remove it.
+
+
+## Performance Considerations
+
+Run a performance test using the expected peak and average message throughput for the environment to baseline the system. This test suite should include simulating infrastructure outages that will cause large numbers of message processing failures. The baseline test should not include audit ingestion or any ServiceControl plugins. 
+
+Once this baseline has been established, follow these steps:
+- For each endpoint, install and configure the Heartbeat plugin (if needed). Re-run the performance test suite and monitor ServiceControl to ensure that it can effectively monitor the system under load. This may require adjustments to the Heartbeat interval. Re-run the performance tests after each adjustment.
+- For each endpoint, turn on auditing (if needed) and re-run the performance tests to assess impact.
+- For each endpoint, turn on any required Custom Checks and re-run the performance tests to assess impact.
+
+
+## Anti-Virus Checks
+
+Exclude the ServiceControl [database directory](/servicecontrol/configure-ravendb-location.md) from anti-virus checks. ServiceControl uses an embedded database and produces a lot of I/O traffic. Some anti-virus software can add overhead to I/O operations causing a significant performance impact.
+
+
+## Version Downgrades
+
+Do not downgrade releases of ServiceControl. ServiceControl uses an embedded database and changes to the internal data structures can occur between releases. Rolling back may cause index corruption or data loss. Perform testing in a lower environment before upgrading in production environments.
--- a/servicecontrol/servicecontrol-in-practice.md
+++ b/servicecontrol/servicecontrol-in-practice.md
@@ -6,9 +6,9 @@ tags:
 ---
 
 
-ServiceControl provides many capabilities such as endpoint monitoring, advanced debgugging, and failed message management. These capabilities can be extended by adding optional [plugins](http://docs.particular.net/servicecontrol/plugins/) into the endpoints being monitored. Each capability and plugin provides valuable information, but they have certain resource and performance costs. Harware, peak and average message throughput, and number of endpoints in the system all have a large impact on the performance of ServiceControl. These factors can vary greatly between environments. Capabilities and plugins that provide value in one environment may have a negative impact if included in another environment.
+ServiceControl provides many capabilities such as endpoint monitoring, advanced debgugging, and failed message management. These capabilities can be extended by adding optional [plugins](/servicecontrol/plugins/) into the endpoints being monitored. Each capability and plugin provides valuable information, but they have certain resource and performance costs. 
 
-For example, the [Saga Audit](/servicecontrol/plugins/saga-audit.md) plugin provides additional information to support a development environment where message load is low. In a production environment, where there are many more saga instances to audit, the increased overhead is magnified and can have a significant performance impact.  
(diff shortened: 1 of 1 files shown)


servicecontrol/servicecontrol-in-practice.md
```diff
@@ -1,35 +1,40 @@
 ---
 title: ServiceControl in Practice
-summary: Details how to think about the use of ServiceControl in production environments
+summary: Considerations for configuring ServiceControl for use in production environments
```
SimonCropp: configuring ? perhaps optimising? not sure?

servicecontrol/servicecontrol-in-practice.md
```diff
@@ -1,35 +1,40 @@
...
+  - [Saga Audit](/servicecontrol/plugins/saga-audit.md) - the saga audit plugin produces a lot of data. It's use outside of a development environment is not recommended.
+- Exclude the ServiceControl [database directory](/servicecontrol/configure-ravendb-location.md) from anti-virus checks.
+- Do not downgrade releases of ServiceControl. ServiceControl uses an embedded database and changes to the internal data structures can occur between releases. Rolling back may cause index corruption or data loss. Perform testing in a lower environment before upgrading in production environments.
```
SimonCropp: i think this is a good point. Is it called out elsewhere as well?

servicecontrol/servicecontrol-in-practice.md
```diff
@@ -1,35 +1,40 @@
...
+
+
+## Specific Considerations
+
+- Each environment should have a dedicated ServiceControl instance.
+- In production, ServiceControl should run on a **dedicated machine**.
+- Turn off [Message Auditing](/nservicebus/operations/auditing.md) if it is not needed. The audit ingestion capability of ServiceControl is primarily to support system analysis with ServiceInsight. If ServiceInsight is not deployed in an environment then configure ServiceControl to read audit messages from an empty queue or turn off messaging auditing for each endpoint. Message auditing may be important for some endpoints but not other others. 
+- Turn off [Audit Forwarding](/servicecontrol/errorlog-auditlog-behavior.md) if it is not needed. ServiceControl sends a copy of each audited message to configured Audit Forwarding queue. If these messages are not being used, turn this feature off.
+- Plugins communicate between the Endpoint they installed in to ServiceControl using messaging and the configured transport of your system. Each instance of a plugin adds more messages to the ServiceControl queue which can delay the processing of each message and make ServiceControl less responsive.
```
SimonCropp: "Plugins communicate between the Endpoint they installed in to ServiceControl using messaging and the configured transport of your system. "   i think that needs to be split up

servicecontrol/servicecontrol-in-practice.md
weralabaj: why not just "Using ServiceControl in production"?

servicecontrol/servicecontrol-in-practice.md
```diff
@@ -1,35 +1,40 @@
...
 
-For example, in a development environment the [Debug Session](/servicecontrol/plugins/debug-session.md) and [Saga Audit](/servicecontrol/plugins/saga-audit.md) plugins provide detailed information to support problem analysis. Particular advise that customers not use these plugins outside of a development environment.
+ServiceControl has many capabilities which can be extended by adding optional plugins into the endpoints of the system being monitored. Each capability and plugin can provide valuable information but also comes with a cost in terms of resource usage and performance. There are other factors that can influence the resource and performance profile of ServiceControl including hardware, message throughput, and number of endpoints. These factors can vary greatly between environments. Capabilities and plugins that provide value in one environment may have a negative impact if included in another environment.
```
weralabaj: Link to plugins http://docs.particular.net/servicecontrol/plugins/.

weralabaj: Would it make sense to list what capabilities SC has out of the box, without plugins?

weralabaj: > Each capability and plugin can provide valuable information but also comes with a cost in terms of resource usage and performance.

Each capability and plugin provides valuable information, but they have certain storage and performance costs.

> There are other factors that can influence the resource and performance profile of ServiceControl including hardware, message throughput, and number of endpoints. These factors can vary greatly between environments. Capabilities and plugins that provide value in one environment may have a negative impact if included in another environment.

How about: "Hardware, (expected/real?) message throughput (load?), and number of endpoints in the system all have a large impact on the performance of ServiceControl. Since these factors vary greatly between systems and environments, the decision about using a specific feature in production should be made carefully."

servicecontrol/servicecontrol-in-practice.md
```diff
@@ -1,35 +1,40 @@
...
 
-Making decisions about the use of the other plugins and features requires a little more thought. The temptation to 'just use them all' as a catch-all insurance policy is not a good choice. 
+For example, the [Saga Audit](/servicecontrol/plugins/saga-audit.md) plugin provides additional information to support a development environment where message load is low. In a production environment, where are there are many more saga instances to audit, the increased overhead is magnified and can have a significant performance impact.
```
weralabaj: > , where are there are many

, where ~~are~~ there are many

servicecontrol/servicecontrol-in-practice.md
```diff
@@ -1,35 +1,40 @@
...
\ No newline at end of file
+NOTE: There are some factors that influence ServiceControl resource usage and performance that are less likely to change between environments but are specific to the system being monitored such as choice of transport and average message size. 
+
+
+## Specific Considerations
+
+- Each environment should have a dedicated ServiceControl instance.
+- In production, ServiceControl should run on a **dedicated machine**.
+- Turn off [Message Auditing](/nservicebus/operations/auditing.md) if it is not needed. The audit ingestion capability of ServiceControl is primarily to support system analysis with ServiceInsight. If ServiceInsight is not deployed in an environment then configure ServiceControl to read audit messages from an empty queue or turn off messaging auditing for each endpoint. Message auditing may be important for some endpoints but not other others.
```
weralabaj: > but not other others. 

but not ~~other~~ others.

servicecontrol/servicecontrol-in-practice.md
```diff
@@ -1,35 +1,40 @@
...
-- [Heartbeats](/servicepulse/intro-endpoints-heartbeats.md) and [Custom Checks](/servicepulse/intro-endpoints-custom-checks.md) can enhance a system's monitoring capability, but they add extra noise. Often, not all endpoints are mission critical. Therefore it is worth making sure that the risks to the system if an endpoint is not available are examined and understood. All communication from endpoints with ServiceControl is performed via messaging. Adding a message to the queue every second may have little impact on when the notification shows up in ServicePulse. If ServiceControl is shutdown without pausing system Endpoints, heartbeats will continue to be added to the queue. ServiceControl will attempt to process all the messages in the queue when it restarts.
+Each capability and plugin needs to be considered for each environment to determine if the value that it provides in that environment outweighs the costs that it imposes. 
 
-##### Less can be more
 
-Moving forward, take a thoughtful approach in the adoption of plugins and features:
+## General Approach
```
weralabaj: I'd switch "Specific Considerations" with "General Approach" (when it comes to order in which they appear).

servicecontrol/servicecontrol-in-practice.md
```diff
@@ -1,35 +1,40 @@
...
+- For each endpoint, turn on Auditing (if needed) and re-run the performance tests to determine impact.
+- For each endpoint, turn on any required Custom Checks and re-run the performance tests to determine impact.
 
-- If running anti-virus software exclude the ServiceControl [database directory](/servicecontrol/configure-ravendb-location.md) from virus checks.
-- The recommended approach is that customers don't downgrade major and minor releases of ServiceControl. ServiceControl uses an embedded database and changes to the internal data structures can occur between releases. Rolling back may cause index corruption or data loss. The recommended approach is that customers perform testing in a lower environment before upgrading in production environments.
\ No newline at end of file
+NOTE: There are some factors that influence ServiceControl resource usage and performance that are less likely to change between environments but are specific to the system being monitored such as choice of transport and average message size. 
+
+
+## Specific Considerations
```
weralabaj: "Recommended (basic) setup/deployment"?

servicecontrol/servicecontrol-in-practice.md
weralabaj: "Measuring/Verifying ServiceControl setup in production"?

servicecontrol/servicecontrol-in-practice.md
weralabaj: I think this looks like we're trying very hard not to say anything specific ;)
Rather than say that audit ingestion is to support system analysis with ServiceInsight would it make sense to say it's primary usage scenario is verifying implementation in development environment? IMO "system analysis" sounds like something I'd potentially want to do in production, to have an idea how all the pieces work together.

What if all my endpoints are important and need auditing? How do I know when I need or not need it? What other options do I have?

I know those are hard questions, but can we at least give some examples, e.g. if your system was thoroughly tested then the important audits may be 3rd party integration or less reliable parts of the system?

weralabaj: Another thing - if I turn off auditing, what impact does it have for my ability to investigate failures? Can I do that without auditing? Will I miss some information in the critical time for my system?

servicecontrol/servicecontrol-in-practice.md
```diff
@@ -1,35 +1,40 @@
...
+- Plugins communicate between the Endpoint they installed in to ServiceControl using messaging and the configured transport of your system. Each instance of a plugin adds more messages to the ServiceControl queue which can delay the processing of each message and make ServiceControl less responsive.
+  - [Heartbeats](/servicepulse/intro-endpoints-heartbeats.md) - Not all endpoints are mission critical and need to be monitored with heartbeats. For Endpoints that do need to be monitored, adjust the Heartbeat Interval. Increasing the interval ensures that ServiceControl is able to process heartbeats in a timely manner but also increases the window that an Endpoint can be unavailable for before being detected by ServiceControl. Heartbeat messages tend to be frequent and a large backlog can occur if ServiceControl is offline for an extended period. When this happens, it can take ServiceControl some time to process old heartbeats when it restarts.
```
weralabaj: What does make one endpoint mission critical, and others not?

Isn't it about how fast this endpoint needs to process the messages and if it's a big deal if you had say 10 min. delay?
Would be good also to ensure the user here that even though the endpoint is down, the messages won't be lost. So once it goes up again it'll pick them up. It's not a loss, just delay. Not "what messages can you lose (because the endpoint is not mission critical)", but "which you can process with delay without losing your job".

I think it's important to make that clear here.

servicecontrol/servicecontrol-in-practice.md
indualagarsamy: "Optimizing ServiceControl for use in Production" sounds better to me. 
Although you are technically configuring SC, you're still providing optimization advice.

indualagarsamy:
@mikeminutillo - In general the changes look good.  

However, I think it makes sense to categorize the information into separate considerations rather than General Approach and Special Considerations. 
For example, grouping the information into 
- Hardware & Installation Considerations (where you talk about installing on a separate dedicated machine, look at capacity planning etc)
- Message Throughput Considerations (Where you talk about how the messages impact SC, including the plugins and the endpoint message volume itself. 
- Testing Considerations (Where you talk about the baseline test setup and the testing itself)
- Antivirus Software and its impact (Talk about a little bit more in detail as to why we need to exclude the directory)
- Version Upgrade/Downgrade Considerations
- Troubleshooting SC - with the all else fails catch and provide the link. 

I think if the information is organized as such, it's easier to read and follow along or focus in the specific area.

servicecontrol/servicecontrol-in-practice.md
petersgiles: > Optimizing ServiceControl for use in Production

Maybe this is part of what made original attempt at this fail. If we focus on Production only it simplifies the advice.

servicecontrol/servicecontrol-in-practice.md
petersgiles: > Would it make sense to list what capabilities SC has out of the box, without plugins?

Interesting...if there are no plugins installed on endpoints does it do anything?

> Hardware, (expected/real?) message throughput (load?)

When I was initially writing this I was trying to encourage people to gather their own metrics through performance testing. I don't think we should provide numbers, customers should seek their own and do their own analysis.

petersgiles: @mikeminutillo I like this opening paragraph

servicecontrol/servicecontrol-in-practice.md
```diff
@@ -1,35 +1,40 @@
...
+- For each endpoint, turn on Auditing (if needed) and re-run the performance tests to determine impact.
+- For each endpoint, turn on any required Custom Checks and re-run the performance tests to determine impact.
 
-- If running anti-virus software exclude the ServiceControl [database directory](/servicecontrol/configure-ravendb-location.md) from virus checks.
-- The recommended approach is that customers don't downgrade major and minor releases of ServiceControl. ServiceControl uses an embedded database and changes to the internal data structures can occur between releases. Rolling back may cause index corruption or data loss. The recommended approach is that customers perform testing in a lower environment before upgrading in production environments.
\ No newline at end of file
+NOTE: There are some factors that influence ServiceControl resource usage and performance that are less likely to change between environments but are specific to the system being monitored such as choice of transport and average message size.
```
petersgiles: @mikeminutillo I like this NOTE ...its a valuable point that I had trouble articulating

servicecontrol/servicecontrol-in-practice.md
petersgiles: @weralabaj I don't think we are trying to make a case for the value of auditing. We are filling a gap in the documentation around performance considerations and the very abrupt limits that you might hit with SC if its deployed without care. It's advice akin to 'keep your tires properly inflated and you will improve your car's fuel consumption' rather than 'everyone needs a Ferrari'

servicecontrol/servicecontrol-in-practice.md
petersgiles: > What does make one endpoint mission critical, and others not?

My original statement started `Often, not all endpoints are mission critical` the word 'often' changes the tone to put the onus on the customer, not us.

@weralabaj I do understand with your point but I think this is the same worry as the comment about the value of auditing. The purpose of this document is to establish in our customer's mind that everything you enable has a consequence. We are asking that they weight the cost benefit of that consequence when they activate a feature.

petersgiles:
Looking at this now and reviewing the comments with less personal attachment to the content I think it's worth reminding everyone that I was considering our licensing page in my thoughts http://particular.net/licensing when I was writing this.

One could argue that if you're on the Advanced license (SP only) then you would never use audits because you should never use SI. The licensing model is the same reason that I didn't want to restrict the advice to Production only. We want people to use SI and therefore be on the Enterprise license.

Overall the tone of this is a better match for docs than mine. I say it is worth releasing and collecting reactions to it.

servicecontrol/servicecontrol-in-practice.md
mikeminutillo: @petersgiles What about this

> Optimizing ServiceControl for use in different environments

petersgiles: yep

servicecontrol/servicecontrol-in-practice.md
mikeminutillo: I don't know. It was a part of @petersgiles content and I didn't want to lose it.

petersgiles: I put it in here after a discussion with @mikeminutillo

servicecontrol/servicecontrol-in-practice.md
mikeminutillo: > Interesting...if there are no plugins installed on endpoints does it do anything?

Yes. SC can collect audits and errors with no plugins installed. It enables ServiceInsight and the recoverability capabilities of Pulse as well with no plugins installed.

mikeminutillo:
Ready for another round of reviews.

servicecontrol/servicecontrol-in-practice.md
```diff
@@ -1,35 +1,51 @@
...
-- Try increasing the [heartbeat interval](/servicecontrol/plugins/heartbeat.md). Ideally heartbeat updates should not occur more frequently than ServiceControl can process them or more than Operations staff are prepared to monitor them with ServicePulse.
-- With each additional change perform a load test again adjusting the heartbeat interval satisfied with the result.
-- Repeat the process of considering the business relevance, system impact and load testing with each [Custom Check](/servicecontrol/plugins/custom-checks.md).
-- Repeat the process of considering the business relevance, system impact and load testing with each Endpoint [Audit](/nservicebus/operations/auditing.md).
+## Hardware & Installation Considerations:
 
-##### Tips and Tricks
+- Read the [Capacity Planning](/servicecontrol/capacity-and-planning.md) and [Troubleshooting](/servicecontrol/troubleshooting.md) guides for ServiceControl.
+- Each enviornment should have a dedicated ServiceControl instance.
+- In production, ServiceControl should run on a dedicated machine.
```
johnsimons: I wonder if we should either copy or provide a link to https://ravendb.net/docs/article-page/2.5/csharp/server/deployment/recommended-configuration, we may as well be prescriptive about hardware specs, we know SC does not work otherwise

servicecontrol/servicecontrol-in-practice.md
```diff
@@ -1,35 +1,51 @@
...
-- When comfortable with the performance of the system try adding [Heartbeats](/servicepulse/intro-endpoints-heartbeats.md) to monitor the system again.
-- Try increasing the [heartbeat interval](/servicecontrol/plugins/heartbeat.md). Ideally heartbeat updates should not occur more frequently than ServiceControl can process them or more than Operations staff are prepared to monitor them with ServicePulse.
-- With each additional change perform a load test again adjusting the heartbeat interval satisfied with the result.
-- Repeat the process of considering the business relevance, system impact and load testing with each [Custom Check](/servicecontrol/plugins/custom-checks.md).
-- Repeat the process of considering the business relevance, system impact and load testing with each Endpoint [Audit](/nservicebus/operations/auditing.md).
+## Hardware & Installation Considerations:
 
-##### Tips and Tricks
+- Read the [Capacity Planning](/servicecontrol/capacity-and-planning.md) and [Troubleshooting](/servicecontrol/troubleshooting.md) guides for ServiceControl.
+- Each enviornment should have a dedicated ServiceControl instance.
```
johnsimons: Apart from the spelling mistake in environment, I actually don't think SC works if they don't use a dedicated instance, so the word "should" maybe change it to "must"

servicecontrol/servicecontrol-in-practice.md
```diff
@@ -1,35 +1,51 @@
...
 
-For example, in a development environment the [Debug Session](/servicecontrol/plugins/debug-session.md) and [Saga Audit](/servicecontrol/plugins/saga-audit.md) plugins provide detailed information to support problem analysis. Particular advise that customers not use these plugins outside of a development environment.
+ServiceControl provides many capabilities such as endpoint monitoring, advanced debgugging, and failed message management. These capabilities can be extended by adding optional [plugins](http://docs.particular.net/servicecontrol/plugins/) into the endpoints being monitored. Each capability and plugin provides valuable information, but they have certain resource and performance costs. Harware, peak and average message throughput, and number of endpoints in the system all have a large impact on the performance of ServiceControl. These factors can vary greatly between environments. Capabilities and plugins that provide value in one environment may have a negative impact if included in another environment.
```
johnsimons: spelling in "hardware"

indualagarsamy: 1. You have a typo. harware -> hardware.
I think you can break this into 2 paragraphs. 

ServiceControl provides many capabilities such as endpoint monitoring, advanced debgugging, and failed message management. These capabilities can be extended by adding optional [plugins](http://docs.particular.net/servicecontrol/plugins/) into the endpoints being monitored. Each capability and plugin provides valuable information, but they have certain resource and performance costs. 

Hardware, peak and average message throughput, and number of endpoints in the system all have a large impact on the performance of ServiceControl. These factors can vary greatly between environments. Capabilities and plugins that provide value in one environment may have a negative impact if included in another environment. For example, the ....

Combine the For example paragraph into the above, so it reads better.

servicecontrol/servicecontrol-in-practice.md
```diff
@@ -1,35 +1,51 @@
...
+
+- Depending on the number of endpoints and message volume, audit messages can have a large impact on performance. Turn off [Message Auditing](/nservicebus/operations/auditing.md) if it is not needed. The primary reason for the audit ingestion capability of ServiceControl is to support system analysis with ServiceInsight. If ServiceInsight is not deployed in an environment then configure ServiceControl to read audit messages from an empty queue or turn off messaging auditing for each endpoint. Message auditing may be important for some endpoints but not others. 
+- Turn off [Audit Forwarding](/servicecontrol/errorlog-auditlog-behavior.md) if it is not needed. ServiceControl sends a copy of each audited message to configured Audit Forwarding queue. If these messages are not being used, turn this feature off.
+- [Plugins](http://docs.particular.net/servicecontrol/plugins/) are installed in an Endpoint and  sned data to ServiceControl. This communication uses messaging over the configured transport of the endpoint. Each instance of a plugin adds more messages to the ServiceControl queue which can delay the processing of each message and make ServiceControl less responsive.
```
johnsimons: spelling 'sends"

servicecontrol/servicecontrol-in-practice.md
```diff
@@ -1,35 +1,51 @@
...
 
-Making decisions about the use of the other plugins and features requires a little more thought. The temptation to 'just use them all' as a catch-all insurance policy is not a good choice. 
+For example, the [Saga Audit](/servicecontrol/plugins/saga-audit.md) plugin provides additional information to support a development environment where message load is low. In a production environment, where there are many more saga instances to audit, the increased overhead is magnified and can have a significant performance impact.
```
indualagarsamy: combine this with the previous paragraph so the info flows better.

servicecontrol/servicecontrol-in-practice.md
```diff
@@ -1,35 +1,51 @@
...
 
-- If running anti-virus software exclude the ServiceControl [database directory](/servicecontrol/configure-ravendb-location.md) from virus checks.
-- The recommended approach is that customers don't downgrade major and minor releases of ServiceControl. ServiceControl uses an embedded database and changes to the internal data structures can occur between releases. Rolling back may cause index corruption or data loss. The recommended approach is that customers perform testing in a lower environment before upgrading in production environments.
\ No newline at end of file
+
+## Message Throughput Considerations
+
+- Depending on the number of endpoints and message volume, audit messages can have a large impact on performance. Turn off [Message Auditing](/nservicebus/operations/auditing.md) if it is not needed. The primary reason for the audit ingestion capability of ServiceControl is to support system analysis with ServiceInsight. If ServiceInsight is not deployed in an environment then configure ServiceControl to read audit messages from an empty queue or turn off messaging auditing for each endpoint. Message auditing may be important for some endpoints but not others.
```
indualagarsamy: I think you should link it to a specific section in auditing, in particular to 
https://github.com/Particular/docs.particular.net/blob/servicecontrol-in-practice2/nservicebus/operations/auditing.md#turning-off-auditing

petersgiles: @SimonCropp has an opinion on that @indualagarsamy

SimonCropp:
note that links cant start with `http://docs.particular.net`

servicecontrol/servicecontrol-in-practice.md
indualagarsamy: >If ServiceInsight is not deployed in an environment then configure ServiceControl to read audit messages from an empty queue or turn off messaging auditing for each endpoint.

SC is deployed in "what" environment? Not sure if it reads well. 

Also, this ought to be explained in a bit more detail. For example, what is configure SC to read audit messages from an empty queue mean? 

Highlight that if you do that, then the regular audit queue might be piling up and the Ops folks might need to have something in place to purge the message build up in the audit queue.

servicecontrol/servicecontrol-in-practice.md
```diff
@@ -1,35 +1,51 @@
...
+- [Plugins](http://docs.particular.net/servicecontrol/plugins/) are installed in an Endpoint and  sned data to ServiceControl. This communication uses messaging over the configured transport of the endpoint. Each instance of a plugin adds more messages to the ServiceControl queue which can delay the processing of each message and make ServiceControl less responsive.
+  - [Heartbeats](/servicepulse/intro-endpoints-heartbeats.md) - Not all endpoints are mission critical and need to be monitored with heartbeats using the same SLA. For endpoints that are less critical to business, [adjust the Heartbeat Interval](http://docs.particular.net/servicecontrol/plugins/heartbeat#configuration-heartbeat-interval). Increasing the interval ensures that ServiceControl is able to process heartbeats in a timely manner but also increases the window that an Endpoint can be unavailable for before being detected by ServiceControl. Heartbeat messages tend to be frequent, and a large backlog can occur if ServiceControl is offline for an extended period. When this happens, it can take ServiceControl some time to process old heartbeats when it restarts.
```
indualagarsamy: > but also increases the window that an Endpoint can be unavailable for before being detected by ServiceControl.

Is it relevant? If so, can you reword this sentence?

servicecontrol/servicecontrol-in-practice.md
mikeminutillo: I was wondering if we want to give some guidance about the machine. I'm not sure that the guidance given in that link is sufficient though. You will definitely need AT LEAST that.

So much depends on the behavior and characteristics of the actual system being monitored.

servicecontrol/servicecontrol-in-practice.md
indualagarsamy: Wondering if we can format this as sub headings instead of bullets and sub bullets.

servicecontrol/servicecontrol-in-practice.md
mikeminutillo: @johnsimons There are two issues here.

1. You should not connect Dev and Prod endpoints to the same ServiceControl
2. You should not have multiple ServiceControl instances in Prod (or any environment)

This sentence was designed to address the first concern. You can do that, especially if you modify endpoint name based on environment and we want to advise against it. 

Do we need to address the second issue as well?

mikeminutillo:
> This is excellent, I like the way you've broken up the sections and your titles are good.

I like it too. Kudos to @indualagarsamy on that

servicecontrol/servicecontrol-in-practice.md
mikeminutillo: > Wondering if we can format this as sub headings instead of bullets and sub bullets.

Awesome idea. Implemented.

mikeminutillo:
Round 3. FIGHT!

servicecontrol/servicecontrol-in-practice.md
```diff
@@ -1,35 +1,68 @@
...
+
+Once this baseline has been established, follow these steps:
+- For each endpoint, install and configure the Heartbeat plugin (if needed). Re-run the performance test suite and monitor ServiceControl to ensure that it can effectively monitor the system under load. This may require adjustments to the Heartbeat interval. Re-run the performance tests after each adjustment.
+- For each endpoint, turn on auditing (if needed) and re-run the performance tests to assess impact.
+- For each endpoint, turn on any required Custom Checks and re-run the performance tests to assess impact.
+
+
+## Anti-Virus Checks
+
+Exclude the ServiceControl [database directory](/servicecontrol/configure-ravendb-location.md) from anti-virus checks. ServiceControl uses an embedded database and produces a lot of I/O traffic. Some anti-virus software can add overhead to I/O operations causing a significant performance impact.
```
SimonCropp: "a lot of I/O traffic" > "a high level of I/O traffic"

required picture of an alot http://thewritepractice.com/wp-content/uploads/2012/05/Alot-vs-a-lot1-600x450.png

servicecontrol/servicecontrol-in-practice.md
```diff
@@ -1,35 +1,68 @@
...
+### Saga Audit
+
+The [Saga Audit](/servicecontrol/plugins/saga-audit.md) plugin produces a lot of data. It's use outside of a development environment is not recommended. If the saga audit plugin is already deployed in production, [follow the instructions outlined](/servicecontrol/plugins/saga-audit#removing-the-plugin-from-production) to remove it.
+
+
+## Performance Considerations
+
+Run a performance test using the expected peak and average message throughput for the environment to baseline the system. This test suite should include simulating infrastructure outages that will cause large numbers of message processing failures. The baseline test should not include audit ingestion or any ServiceControl plugins. 
+
+Once this baseline has been established, follow these steps:
```
SimonCropp: perhaps "follow these steps for each endpoint:" so u dont need it on every point?

servicecontrol/servicecontrol-in-practice.md
```diff
@@ -1,35 +1,68 @@
...
-- Try increasing the [heartbeat interval](/servicecontrol/plugins/heartbeat.md). Ideally heartbeat updates should not occur more frequently than ServiceControl can process them or more than Operations staff are prepared to monitor them with ServicePulse.
-- With each additional change perform a load test again adjusting the heartbeat interval satisfied with the result.
-- Repeat the process of considering the business relevance, system impact and load testing with each [Custom Check](/servicecontrol/plugins/custom-checks.md).
-- Repeat the process of considering the business relevance, system impact and load testing with each Endpoint [Audit](/nservicebus/operations/auditing.md).
+## Hardware & Installation Considerations:
 
-##### Tips and Tricks
+- Read the [Capacity Planning](/servicecontrol/capacity-and-planning.md) and [Troubleshooting](/servicecontrol/troubleshooting.md) guides for ServiceControl.
+- Each environment should have a dedicated ServiceControl instance.
+- In production, ServiceControl should run on a dedicated machine.
```
weralabaj: By "dedicated machine" we mean it should be the only thing running or different machine than NServiceBus endpoints or ...?

mikeminutillo: It should be the only thing running. I want to say something like:

> This machine should be configure with at least X RAM and Y I/O. 

I don't think we can provide those numbers though. Can we say something along the lines of high I/O and RAM usage is expected?

servicecontrol/servicecontrol-in-practice.md
mikeminutillo: The problem is that subtly alters the meaning. This is 3 steps where each step address every endpoint.

The change proposed here suggests that the user fully configure each endpoint and run performance tests for each one individually.

servicecontrol/servicecontrol-in-practice.md
mikeminutillo: Ah yes, [the Alot](http://hyperboleandahalf.blogspot.com.au/2010/04/alot-is-better-than-you-at-everything.html)

Thanks, I struggled with how to phrase this.

servicecontrol/servicecontrol-in-practice.md
weralabaj: I think numbers are tricky, so maybe "Note that other applications running on the machine can negatively impact ServiceControl's performance. The recommended scenario is to run SC on a dedicated machine."

indualagarsamy: To go along with what @weralabaj said, but tweaking it slightly:
Other applications or NServiceBus endpoints running on the same machine as ServiceControl can compete for hardware resources and therefore negatively impact ServiceControl's performance. Hence, it is recommended to run SC on a dedicated machine.

indualagarsamy:
@SimonCropp -My intent was to point to the documentation and the correct section on the bookmark. I understand that links can't start with http:// - Perhaps next time, I'll truncate all that to be more clear.

servicecontrol/servicecontrol-in-practice.md
HEskandari: Should we also comment on running on the cloud environments? Might be stating the obvious but it is easy to forget VMs running on the clouds are more often than not, competing for IO, degrading the performance.

mikeminutillo: @HEskandari I want to say something like this

> If ServiceControl is installed on a Virtual Machine, ensure the machine is capable of high levels of I/O traffic. The amount of I/O required depends on the system being monitored. Run [performance 
tests]() with realistic message loads to baseline hardware requirements.

Where the Performance Tests link off to the section below describing the tests. Does that work?

HEskandari: @mikeminutillo :+1: Sounds good to me.

servicecontrol/servicecontrol-in-practice.md
```diff
@@ -1,35 +1,70 @@
...
-- With each additional change perform a load test again adjusting the heartbeat interval satisfied with the result.
-- Repeat the process of considering the business relevance, system impact and load testing with each [Custom Check](/servicecontrol/plugins/custom-checks.md).
-- Repeat the process of considering the business relevance, system impact and load testing with each Endpoint [Audit](/nservicebus/operations/auditing.md).
+## Hardware & Installation Considerations:
 
-##### Tips and Tricks
+- Read the [Capacity Planning](/servicecontrol/capacity-and-planning.md) and [Troubleshooting](/servicecontrol/troubleshooting.md) guides for ServiceControl.
+- Each environment should have a dedicated ServiceControl instance.
+- Other applications or NServiceBus endpoints running on the same machine as ServiceControl can compete for hardware resources and therefore negatively impact ServiceControl's performance. For optimal performance run SC on a dedicated machine.
+- If ServiceControl is installed on a Virtual Machine, ensure the machine is capable of high levels of I/O traffic. The amount of I/O required depends on the system being monitored. Run performance
```
ramonsmits: I would be a bit more specific, network IO and disk IO.

servicecontrol/servicecontrol-in-practice.md
```diff
@@ -1,35 +1,70 @@
...
\ No newline at end of file
+
+## Message Throughput Considerations
+
+Depending on the number of endpoints and message volume, audit messages can have a large impact on performance. Turn off [Message Auditing](/nservicebus/operations/auditing.md#turning-off-auditing) if it is not needed. The primary reason for the audit ingestion capability of ServiceControl is to support system analysis with ServiceInsight. If ServiceInsight is not in use then turn off messaging auditing for each endpoint. Message auditing may be important for some endpoints but not others. 
+
+NOTE: If message auditing is required without the use of ServiceInsight, configure endpoints and ServiceControl to use different audit queues. Audit messages going to an audit queue that is not managed by ServiceControl will need to be cleaned up manually.
```
ramonsmits: I don't think this is entirely true. Based on the audit stream we know that a retried message is now succeeded or isn't their such correlation happening?

servicecontrol/servicecontrol-in-practice.md
```diff
@@ -1,35 +1,70 @@
...
+- Other applications or NServiceBus endpoints running on the same machine as ServiceControl can compete for hardware resources and therefore negatively impact ServiceControl's performance. For optimal performance run SC on a dedicated machine.
+- If ServiceControl is installed on a Virtual Machine, ensure the machine is capable of high levels of I/O traffic. The amount of I/O required depends on the system being monitored. Run performance 
+tests with realistic message loads to baseline hardware requirements.
 
-- If running anti-virus software exclude the ServiceControl [database directory](/servicecontrol/configure-ravendb-location.md) from virus checks.
-- The recommended approach is that customers don't downgrade major and minor releases of ServiceControl. ServiceControl uses an embedded database and changes to the internal data structures can occur between releases. Rolling back may cause index corruption or data loss. The recommended approach is that customers perform testing in a lower environment before upgrading in production environments.
\ No newline at end of file
+
+## Message Throughput Considerations
```
ramonsmits: Message throughput also included how many messages you want to be able to retry. We should mention that we need to reserve capacity for the IO involved when doing large groups of retries.

servicecontrol/servicecontrol-in-practice.md
```diff
@@ -1,35 +1,70 @@
...
+
+
+## Plugin Considerations
+
+[Plugins](/servicecontrol/plugins/) are installed in an Endpoint and send data to ServiceControl. This communication uses messaging over the configured transport of the endpoint. Each instance of a plugin adds more messages to the ServiceControl queue which can delay the processing of each message and make ServiceControl less responsive.
+
+
+### Heartbeats
+
+Not all endpoints are mission critical and need to be monitored with [heartbeats](/servicepulse/intro-endpoints-heartbeats.md) using the same SLA. For endpoints that are less critical to business, [adjust the Heartbeat Interval](/servicecontrol/plugins/heartbeat#configuration-heartbeat-interval). Increasing the interval ensures that ServiceControl is able to process heartbeats in a timely manner.
```
ramonsmits: Increasing the heartbeat interval in the plugin also requires to increase the interval in SC:

- http://docs.particular.net/servicecontrol/creating-config-file#plugin-specific-servicecontrol-heartbeatgraceperiod

servicecontrol/servicecontrol-in-practice.md
```diff
@@ -1,35 +1,70 @@
...
+
+
+### Saga Audit
+
+The [Saga Audit](/servicecontrol/plugins/saga-audit.md) plugin produces a lot of data. It's use outside of a development environment is not recommended. If the saga audit plugin is already deployed in production, [follow the instructions outlined](/servicecontrol/plugins/saga-audit#removing-the-plugin-from-production) to remove it.
+
+
+## Performance Considerations
+
+Run a performance test using the expected peak and average message throughput for the environment to baseline the system. This test suite should include simulating infrastructure outages that will cause large numbers of message processing failures. The baseline test should not include audit ingestion or any ServiceControl plugins.
```
ramonsmits: Simulating outages, should that be a section on its own? Basically, they should plan for maintenance windows where ServiceControl is able to 'recover' the buffered data in a certain timeframe and reserve capacity for that.

servicecontrol/servicecontrol-in-practice.md
```diff
@@ -1,35 +1,70 @@
...
+
+
+## Anti-Virus Checks
+
+Exclude the ServiceControl [database directory](/servicecontrol/configure-ravendb-location.md) from anti-virus checks. ServiceControl uses an embedded database and produces a high level of I/O traffic. Some anti-virus software can add overhead to I/O operations causing a significant performance impact.
+
+
+## Version Downgrades
+
+Do not downgrade releases of ServiceControl. ServiceControl uses an embedded database and changes to the internal data structures can occur between releases. Rolling back may cause index corruption or data loss. Perform testing in a lower environment before upgrading in production environments.
```
ramonsmits: I've added an issue for this: https://github.com/Particular/ServiceControl/issues/686

servicecontrol/servicecontrol-in-practice.md
mikeminutillo: @ramonsmits that is true. However, all it does is move a failed message from `RetryIssued` to `RetrySuccessful`. Failed messages of both types are excluded from groups, and from the UI so they will never be seen in Pulse. The only place this matters is Insight.

servicecontrol/servicecontrol-in-practice.md
mikeminutillo: I wonder if this is best to put in message throughput or the performance test section. i.e. Your perf tests should include retrying a group of a non-trivial size

servicecontrol/servicecontrol-in-practice.md
mikeminutillo: Good catch

mikeminutillo:
Alrighty. Let's get this merged (to the original PR and then to master).

servicecontrol/servicecontrol-in-practice.md
ramonsmits: :+1:

ramonsmits:
Shouldn't we add a TL;DR to it? I always say to customer treat SC the same as your SQL Server in regards to cpu, memory, network and disk resources, monitoring, backups and availability.

ramonsmits:
@mikeminutillo @weralabaj  Good stuff, merge it and see if you want to address the comments in an additional revision of this PR. No need to block this PR for that.

PR #1221, opened 2016-02-26
[WIP] NHibernate timeout cleanup configuration

tmasternak:


tmasternak:
This option will not be supported.

PR #1231, opened 2016-02-29
Removed statement about `await Task.FromResult` from upgrade guide

danielmarbach:
The statement with `await Task.FromResult` was weird. Removed it for better clarity. It was also not clear to what "this" was referring to.

### DIFF
--- a/nservicebus/upgrades/5to6.md
+++ b/nservicebus/upgrades/5to6.md
@@ -32,9 +32,9 @@ In previous versions of NServiceBus, to send messages or publish messages within
 
 ## Message handlers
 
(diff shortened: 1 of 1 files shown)


danielmarbach:
The question is:

Should we still add

> You may mark your handle method `async`

? I think it is redundant since it is already stated above. Or maybe we can structure it better with subtitles about async vs sync handlers. Thoughts?

indualagarsamy:
@danielmarbach - I like your idea of providing more info as Async vs Sync handlers.

weralabaj:
Not sure how much you want to cover in "Async vs Sync handlers", would it make sense to make it a separate article (guidance content) and only link from here?
If that's just 2-3 sentences though then let's do it here.

SimonCropp:
@indualagarsamy @weralabaj @danielmarbach happy with more info being provided elsewhere. but this PR is incrementally better and should not be blocked by a more significant effort in terms of documenting async

PR #1242, opened 2016-03-02
Clarifying time sent header

andreasohlund:


nservicebus/operations/performance-counters.md
```diff
@@ -36,6 +36,7 @@ Since all performance counters in Windows are exposed via Windows Management Ins
 
 Monitors the age of the oldest message in the queue. This takes into account the whole chain, from the message being sent from the client machine until successfully processed by the server. Define an SLA for each of your endpoints and use the `CriticalTime` counter to make sure you adhere to it.
 
+Note: TimeSent is captured when the call to emit the message is performed even though the [batching](/nservicebus/messaging/batched-dispatch.md) introduced in Version 6 means that the message will be handed over to the tranport infrastructure at a later time. This means that slow calls in handlers happening **after** the outgoing operation will be included in "TimeSent".
```
indualagarsamy: @andreasohlund - rather than a note, would a sub heading as Versions 6 and above actually work better?
Also, what do you think about the below? I made some tweaks. 

For Versions 6 and above: 
The value recorded in the `TimeSent` header is the time when the call to send the message is executed, not the actual time when the message was dispatched to the destination queue. Since the outgoing messages in the handlers are [dispatched as a batched operation](/nservicebus/messaging/batched-dispatch.md), depending on how long the message handler takes to complete, the actual sends might happen at a time later than the time that was recorded in the `TimeSent` header.

indualagarsamy: Here's a better one:

For Versions 6 and above: 
The value recorded in the `TimeSent` header is the time when the call to send the message is executed, not the actual time when the message was dispatched to the destination queue. Since the outgoing messages in the handlers are [sent as a batched operation](/nservicebus/messaging/batched-dispatch.md), depending on how long the message handler takes to complete, the actual dispatch to the transport infrastructure may happen at a later time than the time recorded in the `TimeSent` header.

timbussmann: > This means that slow calls in handlers happening **after** the outgoing operation will be included in "TimeSent"

seems to be wrong with the current implementation? Routing happens immediately when calling context.Send?

andreasohlund:
@indualagarsamy great, updated with some minor tweaks https://github.com/Particular/docs.particular.net/pull/1242/files#diff-08fa4e3c37ef15090701b9074a8530fa

andreasohlund:
@timbussmann what I meant is that the sent time will include slow calls in the handlers and not reflect the time that the message was actually "sent" via the infrastructure (including the outbox)

Does the new wording read better?

timbussmann:
> Does the new wording read better?

:+1: :+1:

Issue #1237, opened 2016-03-02
Do a search and QA on usages of `IBus`


SimonCropp:

WojcikMike:
I will first identify places where IBus is used and then decide colelctively decision will be made which sentences fix in what way

WojcikMike:
After I searched the doco for IBus reference I found following files (I search only through *.md files):

* [NServiceBus/Containers/Index IBus Resoultion Section] (https://github.com/Particular/docs.particular.net/blob/master/nservicebus/containers/index.md#ibus-resolution)
* [NServiceBus/Containers/Index Cleanup Section] (https://github.com/Particular/docs.particular.net/blob/master/nservicebus/containers/index.md#cleanup)
* [NServiceBux/Containers/Injecting IBus] (https://github.com/Particular/docs.particular.net/blob/master/nservicebus/containers/injecting-ibus.md) - this one contains note saying that in v6 IBus was removed but I would add a link to a page showing sample with IMessageHandlerContext 
* [NServiceBus/gateway/Index Using the gateway first line] (https://github.com/Particular/docs.particular.net/blob/master/nservicebus/gateway/index.md#using-the-gateway)
* [NServiceBus/histing/publishing from web applciation - Storage Driven Transport Topology section] (https://github.com/Particular/docs.particular.net/blob/master/nservicebus/hosting/publishing-from-web-applications.md#storage-driven-transport-topology)
* [NServiceBus/messaging/memory-transport How to Raise In memory... section] (https://github.com/Particular/docs.particular.net/blob/master/nservicebus/messaging/memory-transport.md#how-to-raise-an-in-memory-event)
* [NServiceBus/messaging/reply-to-a-message ] (https://github.com/Particular/docs.particular.net/blob/master/nservicebus/messaging/reply-to-a-message.md) - summary field
* [NServiceBus/sagas/reply-replaytooriginator-differences] (https://github.com/Particular/docs.particular.net/blob/master/nservicebus/sagas/reply-replaytooriginator-differences.md) - contains a note that IBus is deprecated but summary still contains IBus
* [samples/hosting/multi-hosting/sample] (https://github.com/Particular/docs.particular.net/blob/master/samples/hosting/multi-hosting/sample.md)
* [samples/hosting/windows-service/sample] (https://github.com/Particular/docs.particular.net/blob/master/samples/hosting/windows-service/sample.md#code-walk-through)
[...]

PR #1278, opened 2016-03-07
Do a search and QA on usages of `IBus` in Samples

WojcikMike:
Connects to #1237

### DIFF
--- a/samples/hosting/windows-service/sample.md
+++ b/samples/hosting/windows-service/sample.md
@@ -19,7 +19,7 @@ The sample is a console application whose `Main` entry point detects if the appl
 
 snippet:windowsservice-hosting-main
 
-The `OnStart` method, manually called when running in interactive mode and automatically called by the Windows Service when running as service, configures the IBus instance:
+The `OnStart` method, manually called when running in interactive mode and automatically called by the Windows Service when running as service, configures the Endpoint Instance:
 
 snippet:windowsservice-hosting-onstart
 
--- a/samples/web/asp-mvc-application/sample.md
+++ b/samples/web/asp-mvc-application/sample.md
@@ -45,13 +45,7 @@ In `AsyncPagesMvc`, open `Global.asax.cs` and see the code for the `ApplicationS
 
 snippet:ApplicationStart
 
-By calling `With()`, the code indicates to NServiceBus to scan the directory where the web application is deployed (different from non-web applications).
-
-The `.ForMvc` extension method injects `IBus` into the controllers by implementing the MVC interfaces `IDependencyResolver` and `IControllerActivator`.
-
-The NServiceBus builder registers and instantiates `IControllerActivator` so that when the controllers are requested, the NServiceBus builder has the opportunity to inject the `IBus` implementation into their `IBus` public property.
-
-Read [how the IBus is injected into the controllers](/samples/web/asp-mvc-injecting-bus/).
+For more details on how to inject NServiceBus classes into the controllers, check the [ASP.NET Controller Injection Sample](/samples/web/asp-mvc-injecting-bus/).
 
 
 ## Sending a message
@@ -64,7 +58,7 @@ Using `AsyncController`:
 snippet:AsyncController
 
 
-### Synchronous message sending: SendAndBlockController controller
+### Synchronous message sending: SendAndBlockController controller (version 4-5)
 
 Open the SendAndBlockController class:
 
@@ -85,4 +79,4 @@ This class implements the NServiceBus interface `IHandleMessages<T>` where `T` i
 
 NServiceBus manages the classes that implement this interface. When a message arrives in the input queue, it is deserialized, and then, based on its type, NServiceBus instantiates the relevant classes and calls their Handle method, passing in the message object.
 
-Notice the `IBus` property of the class. This is how it gets a reference to the bus. In the method body notice it calling the `Return` method on the bus, which results in a message being returned to `WebApplication`, specifically putting a message in the input queue whose name is determined by the namespace where the bus was configured; in this case, the `global.asax`: `AsyncPagesMVC`.
+In the method body notice the response being returned to the originating endpoint. This will result in a message being added to the input queue for `AsyncPagesMVC`.
--- a/samples/hosting/multi-hosting/sample.md
+++ b/samples/hosting/multi-hosting/sample.md
@@ -10,7 +10,7 @@ related:
 
 ## Code walk-through
 
-This sample shows how to host multiple NServiceBus instances in one process. At start up the application creates two `IBus` instances, connected to different queues using different configurations:
+This sample shows how to host multiple NServiceBus instances in one process. At start up the application creates two Endpoint Instances, connected to different queues using different configurations:
 
 snippet:multi-hosting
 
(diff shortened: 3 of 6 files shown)


mikeminutillo:
Discussion in the doco channel today 

```

Simon Cropp [11:50 AM] 
@here: thoughts on this https://github.com/Particular/docs.particular.net/pull/1278/files

Mike Minutillo [12:00 PM] 
@simon: I hate having to qualify everything with (in Version X) but I can't think of a better way off the top of my head

[12:01] 
Could we get away with backporting the term Endpoint even for IBus?

[12:02] 
> At start up the application creates two Endpoint Instances, connected to different queues using different configurations:

[12:03] 
Ideally then the text of the docs talks about requirements while the snippets show the implementation of those requirements

Simon Cropp [12:05 PM] 
@mike: so in some of those changes in the PR can u think of a way to reword them to we can avoid both bus and endpoint?

[12:05] 
i would prefer to link to some other page that explains the diff in one location

Mike Minutillo [12:09 PM] 
@simon: I think we'll end up with an Endpoint page that describes what an endpoint is, what an Endpoint instance, and what those things looked like pre-v6

[12:09] 
Then we could use the term Endpoint Instance and link that to that page

Simon Cropp [12:09 PM] 
ok. i think that might work

[12:09] 
simpler moving forward

[12:10] 
which i always like

Mike Minutillo [12:10 PM] 
Not `IEndpointInstance` which is implementation. `Endpoint Instance` which is requirements level

Simon Cropp [12:10 PM] 
Ok. R u happy to ping that pr and suggest that?
```

With this in mind, we need to put together a doco page about Endpoints and Endpoint Instances before progressing this issue any further so that we can link to it when updating each of these instances of `IBus`.

WojcikMike:
@mikeminutillo can you help me define what should be on those pages? I found that high level definitions are described in here:
http://docs.particular.net/nservicebus/concept-overview#endpoint

My idea:
* Create a page for Endpoint and Endpoint Instance and link it from concept-overview page
* Describe implementation details (samples) of endpoint instance for v3-6

Later on change the above commits to use this concepts.

mikeminutillo:
@WojcikMike I started doing exactly this today and then got pulled away. For now just lift the content from the overview page and we'll add to it later.

I think this should be the next menu item after "Messaging" under NServiceBus as it's a pretty important concept.

WojcikMike:
@mikeminutillo I created PR #1287 to capture that work. I made you contributor so that you can help if you want to.

WojcikMike:
I adjusted some of the descriptions ot don't containt IBus nor iEndpointInstance but rather a concept. There are few places that I don't know how to change as it talkes about the code rather than concepts

weralabaj:
I think from the user perspective the `IBus vs IEndpointInstance` thing is very confusing (in the course Udi says you can't have an instance of the bus, but here it is; but wait it's not a bus, it's an endpoint instance but it was called a bus before...).

Seems to me it could be made simpler if we focused on the user - why do they need it, what exactly do they need to know, what are they trying to achieve? 

They want to send messages. Just that.

Whether that is possible from one interface or another doesn't make a big difference. So instead of having article saying "Injecting the Bus into ASP.NET MVC Controller" we could have "Sending messages (directly) from an ASP.NET MVC Controller" or something like that.

Instead of:
> Notice the `IBus` property of the class. In version 4 and 5 this is how it gets a reference to the bus. In version 6 `IMessageHandlerContext` is used instead. In the method body notice it calling the `Return` method on the bus, which results in a message being returned to `WebApplication`, specifically putting a message in the input queue whose name is determined by the namespace where the bus was configured; in this case, the `global.asax`: `AsyncPagesMVC`.

We could say:
> Notice how the message is put in (returned to?) the input queue of the `WebApplication` (endpoint?) inside the `Handle` method. The input queue name is determined by the namespace where the bus (or rather NSB? endpoint?) was configured; in this sample it's `global.asax`: `AsyncPagesMVC`.

Plus below we could add the note similar to what is in http://docs.particular.net/nservicebus/messaging/send-a-message#outside-a-message-handler.

Of course, the exact text could be tweaked to sound better.

@WojcikMike @mikeminutillo Do you think that approach could simplify the samples and other docos texts? Or did I misunderstand some part of the issue?

mikeminutillo:
@weralabaj I'm not sure. The point of the samples is to point out significant implementation details. I don't know how easily we can get away from them. In this specific case I would go with something like:

> In the method body notice the response being returned to the originating endpoint. This will result in a message being added to the input queue for `MyWebClient`.

weralabaj:
@mikeminutillo I like your suggestion better than mine. So in this specific case - doable? We could try with others too?

I really feel that this IBus vs IEndpointInstance in description in some cases is irrelevant detail that introduces complexity to the text. I didn't check for all occurrences, but I could try to come up with something with Michal today.

WojcikMike:
Ok I updated descriptions which allowed me to remove almost all `IBus` and `IEndpointInstance`.

I also removed 3 phrases from here:
http://docs.particular.net/samples/web/asp-mvc-application/#initializing-the-bus

As I talked with @DavidBoike and he pointed out that they refered to v3 which we don't have as a sample in here (we have samples from Version 4 and above). Which also removed some references to `IBus`

@Particular/docs-maintainers please review and merge if ok.

samples/hosting/windows-service/sample.md
```diff
@@ -19,7 +19,7 @@ The sample is a console application whose `Main` entry point detects if the appl
 
 snippet:windowsservice-hosting-main
 
-The `OnStart` method, manually called when running in interactive mode and automatically called by the Windows Service when running as service, configures the IBus instance:
+The `OnStart` method, manually called when running in interactive mode and automatically called by the Windows Service when running as service, configures the instance of Endpoint Instance:
```
indualagarsamy: redundant usage of instance at the end of the sentence. How about:

The `OnStart` method, manually called when running in interactive mode and automatically called by the Windows Service when running as service, configures the Endpoint Instance:

samples/web/asp-mvc-application/sample.md
```diff
@@ -45,13 +45,7 @@ In `AsyncPagesMvc`, open `Global.asax.cs` and see the code for the `ApplicationS
...
 snippet:ApplicationStart
 
-By calling `With()`, the code indicates to NServiceBus to scan the directory where the web application is deployed (different from non-web applications).
-
-The `.ForMvc` extension method injects `IBus` into the controllers by implementing the MVC interfaces `IDependencyResolver` and `IControllerActivator`.
-
-The NServiceBus builder registers and instantiates `IControllerActivator` so that when the controllers are requested, the NServiceBus builder has the opportunity to inject the `IBus` implementation into their `IBus` public property.
-
-Read [how the IBus is injected into the controllers](/samples/web/asp-mvc-injecting-bus/).
+Read [how the IEndpointInstance/IBus is injected into the controllers](/samples/web/asp-mvc-injecting-bus/).
```
indualagarsamy: > Read [how the IEndpointInstance/IBus is injected into the controllers](/samples/web/asp-mvc-injecting-bus/).

How about:

For more details on how to either inject the IBus (for Versions 5 and below) or the IEndpointInstance (for Versions 6 and above) to the controllers, check the [ASP.NET MVC Sample](/samples/web/asp-mvc-injecting-bus/).

mikeminutillo: Or even

> For more details on how to inject NServiceBus classes into the controllers, check the ASP.NET Controller Injection Sample.

WojcikMike:
Changes included, commits squashed, please merge

weralabaj:
I see comments above from Mike and Indu were addressed, so merging now.

PR #1279, opened 2016-03-07
Do a search and QA on usages of `IBus` in Documentation

WojcikMike:
Connects to #1237


WojcikMike:
I messed with commits

PR #1284, opened 2016-03-08
A sample of multi-tenant enabled endpoint with DB-level isolation

SzymonPobiega:
* MSMQ transport
 * NH persistence
 * Outbox
 * Database per tenant
 * Sagas
 * Timeouts


SimonCropp:
@SzymonPobiega i pushed my changes. do u think this is done? or do u have more finishing touches?

PR #1285, opened 2016-03-08
Do a search and QA on usages of `IBus` in Doco

WojcikMike:
Connects to #1237

Blocked by definitions of Endpoint and Endpoint Instance requested in: #1278

### DIFF
--- a/nservicebus/containers/index.md
+++ b/nservicebus/containers/index.md
@@ -34,12 +34,14 @@ Support for other containers is provided via custom integrations.
 The above pages all have examples of how to pass in an instance of an existing container. This is useful when you want to make use of the full features of the container and share the DI behavior with NServiceBus.
 
 
-### IBus resolution
+### Endpoint resolution
 
-Note that the instance of `IBus` is scoped for the lifetime of the container. Hence if you resolve `IBus` and then dispose of it the endpoint will stop processing messages. Note that all NServiceBus services, including `IBus`, will be injected into the passed in container instance. As such there is no need to register these instances at configuration time.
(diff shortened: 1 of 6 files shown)


WojcikMike:
I reviewed the changes and I have no better idea how to write it.

WojcikMike:
Currently I have following text:
>The simplest way to reply to a message is using the `Reply` method provided by the `IMessageHandlerContext` (in Version 6) or IBus (in Version 3-5), like this:

Should I rather go for below solution:
>The simplest way to reply to a message is using the `Reply` method, like this:

And allow for code to do the talking?
cc: @weralabaj @mikeminutillo @SimonCropp

WojcikMike:
The same thing goes with:
>In order to send message to other sites you need to call the `IMessageHandlerContext.SendToSites` (in Version 6) or `IBus.SendToSites` (in Version 3-5).

Should I remove IBus and IMessageHandlerContext and simply mention method name and create sample? Thoughts?

weralabaj:
Looking at other examples (e.g. http://docs.particular.net/samples/outbox/sqltransport-nhpersistence/#code-walk-through-receiver-project or http://docs.particular.net/nservicebus/sqlserver/configuration#connection-strings-via-a-named-connection-string) I'm +1 for leaving only method names.

> The simplest way to reply to a message is using the Reply method, like this: 

It could be even "The simplest way to reply to a message is using `Reply` method:

> In order to send message to other sites you need to call the IMessageHandlerContext.SendToSites (in Version 6) or IBus.SendToSites (in Version 3-5).

"In order to send message to other sites call `SendToSites` method:

WojcikMike:
Ok, I introduced the changes, please review @Particular/docs-maintainers

PR #1289, opened 2016-03-08
Alternative to "NServiceBus 5" at the top of the RavenDB article

DavidBoike:
@SimonCropp I committed a little fix to master because you had put NServiceBus.RavenDB 5 at the top of the RavenDB article which doesn't make any sense - it actually referred to Core. There is no NSB.Raven V4 yet let alone V5. But I agree that it's confusing to have NServiceBus 5 right at the top of the Raven article. What do you think of this attempt to get rid of that confusion?

### DIFF
--- a/nservicebus/ravendb/index.md
+++ b/nservicebus/ravendb/index.md
@@ -11,12 +11,7 @@ redirects:
  - nservicebus/ravendb/connecting
 ---
 
-Uses the [RavenDB document database](http://ravendb.net/) for storage.
-
-
-## NServiceBus 5 and above
-
-RavenDB is no longer merged into the core. The RavenDB-backed persistence is available as a separate [NuGet package](https://www.nuget.org/packages/NServiceBus.RavenDB). This allows NServiceBus and RavenDB to be upgraded independently.
+Uses the [RavenDB document database](http://ravendb.net/) for storage. Starting with NServiceBus 5, the RavenDB-backed persistence is no longer ILMerged into the Core and is available as a separate [NuGet package](https://www.nuget.org/packages/NServiceBus.RavenDB), which allows NServiceBus and RavenDB to be upgraded independently. See below for information specifically related to NServiceBus 3 and NServiceBus 4. 
 
 
 ### Connection options for RavenDB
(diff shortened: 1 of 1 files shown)


weralabaj:
"NServiceBus X" or "NServiceBus Version X"?

Either way I think we should consider using it as a standard in other documents too. It's an issue not only for Raven article.

weralabaj:
If we decide to add "Version" later, it's easy to do. I'm merging as it is, as it's clearer than before.

weralabaj:
> See below for information specifically related to NServiceBus 3 and NServiceBus 4. 

@DavidBoike where is the info for other versions then? Do we need to link?

DavidBoike:
We don't support < v3 and there was no Raven in < v3. Does that address your concern?

nservicebus/ravendb/index.md
```diff
@@ -11,12 +11,7 @@ redirects:
  - nservicebus/ravendb/connecting
 ---
 
-Uses the [RavenDB document database](http://ravendb.net/) for storage.
-
-
-## NServiceBus 5 and above
-
-RavenDB is no longer merged into the core. The RavenDB-backed persistence is available as a separate [NuGet package](https://www.nuget.org/packages/NServiceBus.RavenDB). This allows NServiceBus and RavenDB to be upgraded independently.
+Uses the [RavenDB document database](http://ravendb.net/) for storage. Starting with NServiceBus 5, the RavenDB-backed persistence is no longer ILMerged into the Core and is available as a separate [NuGet package](https://www.nuget.org/packages/NServiceBus.RavenDB), which allows NServiceBus and RavenDB to be upgraded independently. See below for information specifically related to NServiceBus 3 and NServiceBus 4.
```
indualagarsamy: > Starting with NServiceBus 5, 

See the version specific language here:
https://github.com/particular/docs.particular.net#version-language

So, maybe `When using NServiceBus Versions 5 and above`?

indualagarsamy: > See below for information specifically related to NServiceBus 3 and NServiceBus 4.

Does this mean that the below information does not apply to NServiceBus Versions 5 and above?

indualagarsamy:
@DavidBoike - I ended up linking the section, as the section is buried deep below.

PR #1295, opened 2016-03-10
Distributor upgrade sample

SzymonPobiega:
A bit different from other samples, it shows the steps required to upgrade the distributor setup (same as in the distributor sample) to V6.

Not sure how to represent the actual steps in the end result but maybe a collection of `patch` files users can apply themselves to see the upgrade happening would be cool.


SimonCropp:
@SzymonPobiega 

> Not sure how to represent the actual steps in the end result but maybe a collection of patch files users can apply themselves to see the upgrade happening would be cool.

Sounds interesting. but i think this PR is good enough to go for now. so i will tweak and merge and the above can happen later

PR #1298, opened 2016-03-10
Language/grammar pass on RabbitMQ native integration sample doc

bording:


SimonCropp:
@bording thanks

samples/rabbitmq/native-integration/sample.md
```diff
@@ -10,12 +11,12 @@ related:
 
 ## Code walk-through
 
-The sample consists of 2 console apps, a simple NServiceBus receiver and a native sender. These apps demonstrate how to have senders on other platforms send messages to NServiceBus endpoints. While the .NET RabbitMQ client is used in this sample, the approach demonstrated can be used with any of the [official](https://www.rabbitmq.com/download.html) or [community](https://www.rabbitmq.com/devtools.html) native RabbitMQ client libraries.
+The sample consists of two endpoints, a simple NServiceBus receiver and a native sender. These endpoints demonstrate how to have senders on other platforms send messages to NServiceBus endpoints. While the .NET RabbitMQ client is used in this sample, the approach demonstrated can be used with any of the [official](https://www.rabbitmq.com/download.html) or [community](https://www.rabbitmq.com/devtools.html) native RabbitMQ client libraries.
```
bording: Is endpoint the right word here? One of the projects doesn't use NServiceBus at all, only the native RabbitMQ client. Seems like endpoint has a specific meaning that wouldn't apply to that project.

SimonCropp: yep. will fix

samples/rabbitmq/native-integration/sample.md
```diff
@@ -39,8 +40,8 @@ snippet:DefineNSBMessage
...
+NServiceBus requires all messages to be uniquely identified in order to be able to perform retries in a safe way. Unfortunately, RabbitMQ doesn't provide a unique id for messages by automatically, so a unique id will need to be manually generated. By default, NServiceBus will look for this message id in the optional [AMQP](https://www.amqp.org/) `message-id` message header. This behavior can be modified by using a [custom message identifier strategy](/nservicebus/rabbitmq/configuration-api.md#configuring-rabbitmq-transport-to-be-used-controlling-the-message-id-strategy) to tell NServiceBus to look in a different location for the message identifier. Using this custom strategy, the id can be extracted from any message header, or even the message payload itself.
```
bording: The link to the AMQP spec that was added here should probably point to https://www.rabbitmq.com/amqp-0-9-1-reference.html instead since that is the version of the spec that RabbitMQ implements.

samples/rabbitmq/native-integration/sample.md
```diff
@@ -39,8 +40,8 @@ snippet:DefineNSBMessage
...
 
-To set this up for this sample, generate a unique id on the sender side and attach it to the `MessageId` property:
+To set this up for this sample, generate a unique identifier on the sender side and attach it to the `MessageId` property:
```
bording: I see that you changed some uses of "id" to "identifier" instead, but not all of them. What is the guideline you're following on when to use the abbreviation? Should I change the remaining uses to "identifier" also?

Issue #1281, opened 2016-03-07
No Doco for `IManageMessageFailures` extension now obsolete message in v6


indualagarsamy:
In version 5, in order to override the behavior of SLR and moving messages to error queue, the user can extend the IManageMessageFailures interface and do custom logic there. See v5 code. 

In Version 6, we now have an obsolete message:
IManageMessageFailures is no longer an extension point. If full control over what happens when a message fails (including retries) is required then override the MoveFaultsToErrorQueue behavior. To get notified when messages are being moved use BusNotifications.Errors.MessageSentToErrorQueue.Subscribe(e=>{})

However we don't have any doco on this in the upgrade guide. 

**Version 5 code**
```c#
public class CustomFaultManager : IManageMessageFailures, INeedInitialization
{
    public void SerializationFailedForMessage(TransportMessage message, Exception e)
    {
        Console.WriteLine("CustomFaultManager - SerializationFailedForMessage");
    }

    public void ProcessingAlwaysFailsForMessage(TransportMessage message, Exception e)
    {
        Console.WriteLine("CustomFaultManager - ProcessingAlwaysFailsForMessage");
    }

    public void Init(Address address)
    {
        Console.WriteLine("CustomFaultManager - Init");
    }
        
    public void Customize(BusConfiguration configuration)
    {
        configuration.RegisterComponents(c => c.ConfigureComponent<CustomFaultManager>(DependencyLifecycle.InstancePerCall));
    }
}
```

indualagarsamy:
@SeanFeldman - any reason you removed the `Project: V6 Launch` label from this issue?

SimonCropp:
@indualagarsamy it was a GH glitch. we need to go through and add them all back

indualagarsamy:
@SimonCropp - ok, thanks.

indualagarsamy:
PR merged. https://github.com/Particular/docs.particular.net/pull/1418

PR #1300, opened 2016-03-11
WIP - Added Guidance for moving away from IManageMessageFailures in V6.

petersgiles:
Relates to https://github.com/Particular/docs.particular.net/issues/1281

In version 5, in order to override the behavior of SLR and moving messages to error queue, the user can extend the IManageMessageFailures interface and do custom logic there. See v5 code. 

**Version 5 code**
```c#
public class CustomFaultManager : IManageMessageFailures, INeedInitialization
{
    public void SerializationFailedForMessage(TransportMessage message, Exception e)
    {
        Console.WriteLine("CustomFaultManager - SerializationFailedForMessage");
    }

    public void ProcessingAlwaysFailsForMessage(TransportMessage message, Exception e)
    {
        Console.WriteLine("CustomFaultManager - ProcessingAlwaysFailsForMessage");
    }

    public void Init(Address address)
    {
        Console.WriteLine("CustomFaultManager - Init");
    }
        
    public void Customize(BusConfiguration configuration)
    {
        configuration.RegisterComponents(c => c.ConfigureComponent<CustomFaultManager>(DependencyLifecycle.InstancePerCall));
    }
}
```

In Version 6, this is done by adding a new message processing behavior in the pipeline. 
User can control if the behavior should still invoke SLR or not, by turning off SLR or customizing it a the endpoint level. 
User can control if the message can be forwarded to the error queue or not in the behavior.

### DIFF
--- a/Snippets/Snippets_6/UpgradeGuides/5to6/CustomErrorHandlingBehaviour.cs
+++ b/Snippets/Snippets_6/UpgradeGuides/5to6/CustomErrorHandlingBehaviour.cs
@@ -48,12 +48,9 @@ namespace Snippets6.UpgradeGuides._5to6
     #endregion
 
     #region 5to6-registercustomerrorhandling
-    class RegisterCustomErrorHandling : INeedInitialization
+    public void Customize(EndpointConfiguration endpointConfiguration)
     {
-        public void Customize(EndpointConfiguration configuration)
-        {
-            configuration.Pipeline.Register<NewMessageProcessingPipelineStep>();
-        }
+        endpointConfiguration.Pipeline.Register<NewMessageProcessingPipelineStep>();
     }
     #endregion
 }
--- a/Snippets/Snippets_6/UpgradeGuides/5to6/CustomErrorHandlingBehaviour.cs
+++ b/Snippets/Snippets_6/UpgradeGuides/5to6/CustomErrorHandlingBehaviour.cs
@@ -1,0 +1,59 @@
+namespace Snippets6.UpgradeGuides._5to6
+{
+    using System;
+    using System.Threading.Tasks;
+    using NServiceBus;
+    using NServiceBus.Pipeline;
+
+    #region 5to6-customerrorhandlingbehaviour
+    class CustomErrorHandlingBehaviour : Behavior<ITransportReceiveContext>
+    {
+        public override async Task Invoke(ITransportReceiveContext context, Func<Task> next)
+        {
+            try
+            {
+                await next().ConfigureAwait(false);
+            }
+            catch (MessageDeserializationException)
+            {
+                Console.WriteLine("CustomFaultManager - MessageDeserializationException");
+            }
+            catch (Exception)
+            {
+                Console.WriteLine("CustomFaultManager - ProcessingAlwaysFailsForMessage");
+                // do not rethrow the message if you want to mark the message as processed.
+                // rethrow to move the message to the error queue.
+                throw;
+                // if you want to rollback the receive operation instead of mark as processed:
+                //context.AbortReceiveOperation();
+            }
+        }
+    }
+    #endregion
+
+    #region 5to6-newmessageprocessingpipelinestep
+    class NewMessageProcessingPipelineStep : RegisterStep
+    {
+        public NewMessageProcessingPipelineStep()
+            : base("CustomErrorHandlingBehaviour", typeof(CustomErrorHandlingBehaviour), "Adds custom error behavior to pipeline")
+        {
+            InsertAfter("MoveFaultsToErrorQueue");
+            // only invoke this behavior if FLR fails to handle the message
+            InsertBeforeIfExists("FirstLevelRetries");
+            // if you want to handle the message before it is moved to error queue, insert after SLR.
+            // if you want to handle the message before it is handled by SLR, insert it before SLR.
+            InsertBeforeIfExists("SecondLevelRetries");
+        }
+    }
+    #endregion
+
+    #region 5to6-registercustomerrorhandling
+    class RegisterCustomErrorHandling : INeedInitialization
+    {
+        public void Customize(EndpointConfiguration configuration)
+        {
+            configuration.Pipeline.Register<NewMessageProcessingPipelineStep>();
+        }
+    }
+    #endregion
+}
--- a/Snippets/Snippets_6/UpgradeGuides/5to6/CustomErrorHandlingBehaviour.cs
+++ b/Snippets/Snippets_6/UpgradeGuides/5to6/CustomErrorHandlingBehaviour.cs
@@ -21,11 +21,11 @@ namespace Snippets6.UpgradeGuides._5to6
             catch (Exception)
             {
                 Console.WriteLine("CustomFaultManager - ProcessingAlwaysFailsForMessage");
-                // do not rethrow the message if you want to mark the message as processed.
-                // rethrow to move the message to the error queue.
+                // To mark the message as processed DO NOT rethrow the message.
+                // Rethrow to move the message to the error queue.
                 throw;
-                // if you want to rollback the receive operation instead of mark as processed:
-                //context.AbortReceiveOperation();
+                // To rollback the receive operation instead of mark as processed:
+                // context.AbortReceiveOperation();
             }
         }
     }
@@ -38,19 +38,21 @@ namespace Snippets6.UpgradeGuides._5to6
             : base("CustomErrorHandlingBehaviour", typeof(CustomErrorHandlingBehaviour), "Adds custom error behavior to pipeline")
         {
             InsertAfter("MoveFaultsToErrorQueue");
-            // only invoke this behavior if FLR fails to handle the message
+            // Only invoke this behavior if FLR fails to handle the message
             InsertBeforeIfExists("FirstLevelRetries");
-            // if you want to handle the message before it is moved to error queue, insert after SLR.
-            // if you want to handle the message before it is handled by SLR, insert it before SLR.
+            // To handle the message before it is moved to error queue, insert after SLR.
+            // To handle the message before it is handled by SLR, insert it before SLR.
             InsertBeforeIfExists("SecondLevelRetries");
         }
     }
     #endregion
 
-    #region 5to6-registercustomerrorhandling
-    public void Customize(EndpointConfiguration endpointConfiguration)
-    {
-        endpointConfiguration.Pipeline.Register<NewMessageProcessingPipelineStep>();
+    class RegisterCustomErrorHandling : INeedInitialization {
+        public void Customize(EndpointConfiguration endpointConfiguration)
+        {
+            #region 5to6-registercustomerrorhandling
+            endpointConfiguration.Pipeline.Register<NewMessageProcessingPipelineStep>();
+            #endregion
+        }
     }
-    #endregion
 }
--- a/nservicebus/errors/managing-errors-with-the-pipeline.md
+++ b/nservicebus/errors/managing-errors-with-the-pipeline.md
@@ -1,0 +1,30 @@
+---
+title: Managing errors with the pipeline
+summary: Instructions on how to perform custom logic during FLR and SLR with the pipeline
+tags:
+ - upgrade
+ - example
+---
+
+In versions of NServiceBus lower than v6 to add logic to FLR and SLR processing NServiceBus provided the `IManageMessageFailures` interface to implement. Now developers must instead add a new behavior class which will be invoked during the message processing pipeline. In this behavior, both the deserialization exception and the general exception can be handled. Any custom action that needs to be performed when an exception happens can be specified. 
+
+While Version 5 did not allow an easier way to invoke both the SLR and the custom fault handling policy, in Version 6, SLR can either be disabled or customized as necessary. For more details, please read this article on how to [disable SLR](/nservicebus/errors/automatic-retries.md#second-level-retries-disabling-slr-through-code).
+
+
+### Create a new behavior that has access to the message at the transport message context
+
+snippet: 5to6-customerrorhandlingbehaviour
+
+As illustrated in the above example, by choosing to throw in the catch block, this custom behavior can then let NServiceBus forward the message to the error queue. If that's not the desired behavior, remove the throw to let NServiceBus know that the message handling is complete.
+
+
+### Create a new message processing step in the pipeline.
+
+Specify where and how this behavior ought to be invoked. 
+
+snippet: 5to6-newmessageprocessingpipelinestep
+
+
+### Register the new step in the pipeline
+
+snippet: 5to6-registercustomerrorhandling
--- a/nservicebus/errors/managing-errors-with-the-pipeline.md
+++ b/nservicebus/errors/managing-errors-with-the-pipeline.md
@@ -1,26 +1,26 @@
 ---
 title: Managing errors with the pipeline
-summary: Instructions on how to perform custom logic during FLR and SLR with the pipeline
+summary: Instructions on how to perform logic during FLR and SLR with the pipeline
 tags:
  - upgrade
  - example
 ---
 
-In versions of NServiceBus lower than v6 to add logic to FLR and SLR processing NServiceBus provided the `IManageMessageFailures` interface to implement. Now developers must instead add a new behavior class which will be invoked during the message processing pipeline. In this behavior, both the deserialization exception and the general exception can be handled. Any custom action that needs to be performed when an exception happens can be specified. 
+In versions of NServiceBus lower than v6 to add logic to FLR and SLR processing NServiceBus provided the `IManageMessageFailures` interface to implement. Now developers must instead add a new behavior class which will be invoked during the message processing pipeline. In this behavior, both the deserialization exception and the general exception can be handled as well as any custom action that needs to be performed when an exception happens. 
 
-While Version 5 did not allow an easier way to invoke both the SLR and the custom fault handling policy, in Version 6, SLR can either be disabled or customized as necessary. For more details, please read this article on how to [disable SLR](/nservicebus/errors/automatic-retries.md#second-level-retries-disabling-slr-through-code).
+While Version 5 did not allow an easier way to invoke both the SLR and the custom fault handling policy, in Version 6, SLR can either be disabled or customized as necessary. For more details, read this article on how to [disable SLR](/nservicebus/errors/automatic-retries.md#second-level-retries-disabling-slr-through-code).
 
 
 ### Create a new behavior that has access to the message at the transport message context
 
 snippet: 5to6-customerrorhandlingbehaviour
 
-As illustrated in the above example, by choosing to throw in the catch block, this custom behavior can then let NServiceBus forward the message to the error queue. If that's not the desired behavior, remove the throw to let NServiceBus know that the message handling is complete.
+As illustrated in the above example, by choosing to throw in the catch block, this behavior can then let NServiceBus forward the message to the error queue. If that's not the desired behavior, remove the throw to let NServiceBus know that the message handling is complete.
 
 
 ### Create a new message processing step in the pipeline.
 
-Specify where and how this behavior ought to be invoked. 
+Specify where and how this behavior is invoked. 
 
 snippet: 5to6-newmessageprocessingpipelinestep
 
--- a/nservicebus/errors/managing-errors-with-the-pipeline.md
+++ b/nservicebus/errors/managing-errors-with-the-pipeline.md
@@ -6,10 +6,9 @@ tags:
  - example
 ---
 
-In versions of NServiceBus lower than v6 to add logic to FLR and SLR processing NServiceBus provided the `IManageMessageFailures` interface to implement. Now developers must instead add a new behavior class which will be invoked during the message processing pipeline. In this behavior, both the deserialization exception and the general exception can be handled as well as any custom action that needs to be performed when an exception happens. 
+In Versions 5 and below to add logic to FLR and SLR processing required implementing the `IManageMessageFailures` interface. Instead add a new behavior which is invoked during the message processing pipeline. In this behavior, both the deserialization exception and the general exception can be handled as well as any custom action that needs to be performed when an exception occurs. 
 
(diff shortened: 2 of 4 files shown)


indualagarsamy:
@Particular/docs-maintainers - can you please review?

Snippets/Snippets_6/UpgradeGuides/5to6/CustomErrorHandlingBehaviour.cs
```diff
@@ -0,0 +1,59 @@
...
+                Console.WriteLine("CustomFaultManager - MessageDeserializationException");
+            }
+            catch (Exception)
+            {
+                Console.WriteLine("CustomFaultManager - ProcessingAlwaysFailsForMessage");
+                // do not rethrow the message if you want to mark the message as processed.
+                // rethrow to move the message to the error queue.
+                throw;
+                // if you want to rollback the receive operation instead of mark as processed:
+                //context.AbortReceiveOperation();
```
SimonCropp: would prefer this as a different smaller snippet so any change to AbortReceiveOperation in the future is captured

indualagarsamy:
@andreasohlund @timbussmann - can you review the guidance for IManageMessageFailures?

SimonCropp:
i think the comments in snippets here would be better explained in md content

Snippets/Snippets_6/UpgradeGuides/5to6/CustomErrorHandlingBehaviour.cs
```diff
@@ -0,0 +1,59 @@
...
+            InsertBeforeIfExists("FirstLevelRetries");
+            // if you want to handle the message before it is moved to error queue, insert after SLR.
+            // if you want to handle the message before it is handled by SLR, insert it before SLR.
+            InsertBeforeIfExists("SecondLevelRetries");
+        }
+    }
+    #endregion
+
+    #region 5to6-registercustomerrorhandling
+    class RegisterCustomErrorHandling : INeedInitialization
```
SimonCropp: INeedInitialization seems to have noting to do with this snippet. perhaps just show the `Pipeline.Register`

Snippets/Snippets_6/UpgradeGuides/5to6/CustomErrorHandlingBehaviour.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        public void Customize(EndpointConfiguration configuration)
+        {
+            configuration.Pipeline.Register<NewMessageProcessingPipelineStep>();
```
SimonCropp: `configuration` => `endpointConfiguration`

nservicebus/upgrades/5to6.md
```diff
@@ -600,3 +600,34 @@ In Version 6 the callback routing is based on user-provided unique endpoint inst
...
+### Create a new behavior that has access to the message at the transport message context
+
+snippet: 5to6-customerrorhandlingbehaviour
+
+As illustrated in the above example, by choosing to throw in the catch block, this custom behavior can then let NServiceBus forward the message to the error queue. If that's not the desired behavior, remove the throw to let NServiceBus know that the message handling is complete.  
+
+
+### Create a new message processing step in the pipeline.
+
+Specify where and how this behavior ought to be invoked.
```
SimonCropp: ought seems strange phrasing for doco. perhaps just "should"

nservicebus/upgrades/5to6.md
```diff
@@ -600,3 +600,34 @@ In Version 6 the callback routing is based on user-provided unique endpoint inst
...
+To convert existing code that used `IManageMessageFailures`, add a new behavior class which will be invoked during the message processing pipeline. In this behavior, both the deserialization exception and the general exception can be handled. Any custom action that needs to be performed when an exception happens can be specified. 
+
+While Version 5 did not allow an easier way to invoke both the SLR and the custom fault handling policy, in Version 6, SLR can either be disabled or customized as necessary. For more details, please read this article on how to [disable SLR](/nservicebus/errors/automatic-retries.md#second-level-retries-disabling-slr-through-code).
+
+
+### Create a new behavior that has access to the message at the transport message context
+
+snippet: 5to6-customerrorhandlingbehaviour
+
+As illustrated in the above example, by choosing to throw in the catch block, this custom behavior can then let NServiceBus forward the message to the error queue. If that's not the desired behavior, remove the throw to let NServiceBus know that the message handling is complete.
```
SimonCropp: custom is redundant

nservicebus/upgrades/5to6.md
```diff
@@ -600,3 +600,34 @@ In Version 6 the callback routing is based on user-provided unique endpoint inst
...
+
+## IManageMessageFailures is now Obsolete
+
+In Version 5, the `IManageMessageFailures` interface served as an extension point to customize the process of handling the action when a message fails the first level retries before the message is forwarded to the error queue.  
+
+Version 6 allows more control of this behavior with the [more advanced message processing pipeline](/nservicebus/pipeline/customizing-v6.md). 
+ 
+To convert existing code that used `IManageMessageFailures`, add a new behavior class which will be invoked during the message processing pipeline. In this behavior, both the deserialization exception and the general exception can be handled. Any custom action that needs to be performed when an exception happens can be specified. 
+
+While Version 5 did not allow an easier way to invoke both the SLR and the custom fault handling policy, in Version 6, SLR can either be disabled or customized as necessary. For more details, please read this article on how to [disable SLR](/nservicebus/errors/automatic-retries.md#second-level-retries-disabling-slr-through-code).
```
SimonCropp: remove please

Snippets/Snippets_6/UpgradeGuides/5to6/CustomErrorHandlingBehaviour.cs
```diff
@@ -0,0 +1,59 @@
...
+    #region 5to6-customerrorhandlingbehaviour
+    class CustomErrorHandlingBehaviour : Behavior<ITransportReceiveContext>
+    {
+        public override async Task Invoke(ITransportReceiveContext context, Func<Task> next)
+        {
+            try
+            {
+                await next().ConfigureAwait(false);
+            }
+            catch (MessageDeserializationException)
```
mikeminutillo: Why does the snippet code swallow this exception?

petersgiles: It's an example of how to swallow an exception

Snippets/Snippets_6/UpgradeGuides/5to6/CustomErrorHandlingBehaviour.cs
mikeminutillo: I'd break this into 3 different snippets. One that throws, one that doesn't, and one that aborts. Then explain the difference in the text between. You could then ditch the Console.WriteLine statements as well.

petersgiles: Fair call

Snippets/Snippets_6/UpgradeGuides/5to6/CustomErrorHandlingBehaviour.cs
indualagarsamy: @mikeminutillo - The IManageMessageFailures had a method that'd get invoked on Serialization failures. This exception is there just to show that the code from that function needs to go in that catch block. 
Need to add something there or when the snippets get broken apart - show this as how to handle serialization failures in v6

indualagarsamy: I updated the PR description with the v5 code so you can see.

mikeminutillo: Ah. Looking at the v5 code I now see why the Console.Writelines are in there as well. I'd handle each use case as a separate snippet. This becomes a 4th case added on to the list of 3 I wrote below.

indualagarsamy: @mikeminutillo - yep. Agreed. The snippets need some work.

Snippets/Snippets_6/UpgradeGuides/5to6/CustomErrorHandlingBehaviour.cs
```diff
@@ -0,0 +1,59 @@
...
+    }
+    #endregion
+
+    #region 5to6-newmessageprocessingpipelinestep
+    class NewMessageProcessingPipelineStep : RegisterStep
+    {
+        public NewMessageProcessingPipelineStep()
+            : base("CustomErrorHandlingBehaviour", typeof(CustomErrorHandlingBehaviour), "Adds custom error behavior to pipeline")
+        {
+            InsertAfter("MoveFaultsToErrorQueue");
```
andreasohlund: Something feels off, in the old solution we where completely replacing the "recoverability" parts of the pipeline but here we're combining?

Note: There are plans to merge all 3 recoverability behaviors into one which would break this.

My proposal is to make this a plain "PhysicalReceiveStage" behavior instead. The flaw with that is that there is no way to request a FLR since the "AbortReceiveOperation" method is only available on the TransportReceive stage. Perhaps we should change that?

timbussmann: > There are plans to merge all 3 recoverability behaviors into one which would break this.

I'm still thinking we should not merge retries with the errorqueue. It seems we need to think more carefully about retries "extensibility" support instead of just following design ideas?

> My proposal is to make this a plain "PhysicalReceiveStage" behavior instead. The flaw with that is that there is no way to request a FLR since the "AbortReceiveOperation" method is only available on the TransportReceive stage. Perhaps we should change that?

This shouldn't be a big deal to implement. But the error queue behavior swallows all exceptions before.

andreasohlund: >But the error queue behavior swallows all exceptions before.

The physical receive is wrapped by the transport receive so a that shouldn't happen?

timbussmann: yeah sorry misread that. I think we clarified everything on the call?

indualagarsamy: @timbussmann @andreasohlund - Can you summarize what you finalized in your call?

Snippets/Snippets_6/UpgradeGuides/5to6/CustomErrorHandlingBehaviour.cs
```diff
@@ -0,0 +1,56 @@
...
+            // only invoke this behavior if FLR fails to handle the message
+            InsertBeforeIfExists("FirstLevelRetries");
+            // if you want to handle the message before it is moved to error queue, insert after SLR.
```
SimonCropp: remove "you"

nservicebus/errors/managing-errors-with-the-pipeline.md
```diff
@@ -0,0 +1,30 @@
+---
+title: Managing errors with the pipeline
+summary: Instructions on how to perform logic during FLR and SLR with the pipeline
+tags:
+ - upgrade
+ - example
+---
+
+In versions of NServiceBus lower than v6 to add logic to FLR and SLR processing NServiceBus provided the `IManageMessageFailures` interface to implement. Now developers must instead add a new behavior class which will be invoked during the message processing pipeline. In this behavior, both the deserialization exception and the general exception can be handled as well as any custom action that needs to be performed when an exception happens.
```
SimonCropp: "In Versions 5 and below..."

Snippets/Snippets_6/UpgradeGuides/5to6/CustomErrorHandlingBehaviour.cs
```diff
@@ -0,0 +1,56 @@
...
+    using NServiceBus.Pipeline;
+
+    #region 5to6-customerrorhandlingbehaviour
+    class CustomErrorHandlingBehaviour : Behavior<ITransportReceiveContext>
+    {
+        public override async Task Invoke(ITransportReceiveContext context, Func<Task> next)
+        {
+            try
+            {
+                await next().ConfigureAwait(false);
```
SimonCropp: we dont ConfigureAwait in doco http://docs.particular.net/samples/#technology-choices-async-await

SimonCropp:
@petersgiles did u forget to push the csproj?

Snippets/Snippets_6/UpgradeGuides/5to6/CustomErrorHandlingBehaviour.cs
```diff
@@ -0,0 +1,56 @@
...
+            InsertBeforeIfExists("FirstLevelRetries");
+            // if you want to handle the message before it is moved to error queue, insert after SLR.
+            // if you want to handle the message before it is handled by SLR, insert it before SLR.
+            InsertBeforeIfExists("SecondLevelRetries");
+        }
+    }
+    #endregion
+
+    #region 5to6-registercustomerrorhandling
+    public void Customize(EndpointConfiguration endpointConfiguration)
```
SimonCropp: this needs to be in a class. also shouldnt the region be only around the `endpointConfiguration.Pipeline.Register<NewMessageProcessingPipelineStep`  ?

nservicebus/upgrades/5to6.md
```diff
@@ -600,3 +600,11 @@ In Version 6 the callback routing is based on user-provided unique endpoint inst
 snippet: 5to6-Callbacks-InstanceId
 
 NOTE: This identifier needs to be stable and should never be hardcoded, e.g. it can be read from the configuration file or from the environment (e.g. role ID in Azure).
+
+
+## IManageMessageFailures is now Obsolete
+
+In Version 5, the `IManageMessageFailures` interface served as an extension point to customize the process of handling the action when a message fails the first level retries before the message is forwarded to the error queue.  
+
+Version 6 allows more control of this behavior with the [more advanced message processing pipeline](/nservicebus/pipeline/customizing-v6.md) and there is also Version 6 specific guidance for [managing errors with the pipeline](/nservicebus/errors/managing-errors-with-the-pipeline.md).
```
SimonCropp: In versions 6 and below

nservicebus/upgrades/5to6.md
```diff
@@ -600,3 +600,11 @@ In Version 6 the callback routing is based on user-provided unique endpoint inst
 snippet: 5to6-Callbacks-InstanceId
 
 NOTE: This identifier needs to be stable and should never be hardcoded, e.g. it can be read from the configuration file or from the environment (e.g. role ID in Azure).
+
+
+## IManageMessageFailures is now Obsolete
+
+In Version 5, the `IManageMessageFailures` interface served as an extension point to customize the process of handling the action when a message fails the first level retries before the message is forwarded to the error queue.
```
SimonCropp: In Versions 5 and below

nservicebus/upgrades/5to6.md
SimonCropp: i think it is a bit generous to say "Version 6 allows more control of this behavior with the". i dont think v6 is more customizable than v5 in terms of the pipeline. at least not measurable from the consuming point of view

nservicebus/errors/managing-errors-with-the-pipeline.md
```diff
@@ -0,0 +1,30 @@
...
+title: Managing errors with the pipeline
+summary: Instructions on how to perform logic during FLR and SLR with the pipeline
+tags:
+ - upgrade
+ - example
+---
+
+In versions of NServiceBus lower than v6 to add logic to FLR and SLR processing NServiceBus provided the `IManageMessageFailures` interface to implement. Now developers must instead add a new behavior class which will be invoked during the message processing pipeline. In this behavior, both the deserialization exception and the general exception can be handled as well as any custom action that needs to be performed when an exception happens. 
+
+While Version 5 did not allow an easier way to invoke both the SLR and the custom fault handling policy, in Version 6, SLR can either be disabled or customized as necessary. For more details, read this article on how to [disable SLR](/nservicebus/errors/automatic-retries.md#second-level-retries-disabling-slr-through-code).
```
SimonCropp: > did not allow an easier way to invoke both the SLR and the custom fault handling policy

why call that out? just say "now v6 supports..." or "in v6 x was added"

SimonCropp: > For more details, read this article on how to [disable SLR](/nservicebus/errors/automatic-retries.md#second-level-retries-disabling-slr-through-code).

> See also [disabling SLR](/nservicebus/errors/automatic-retries.md#second-level-retries-disabling-slr-through-code).

nservicebus/errors/managing-errors-with-the-pipeline.md
SimonCropp: > Now developers must instead add a new behavior class which will be invoked during the message processing pipeline.

how about

> Instead add a new behavior which is invoked during the message processing pipeline.

SimonCropp: > when an exception occurs

Snippets/Snippets_6/UpgradeGuides/5to6/CustomErrorHandlingBehaviour.cs
timbussmann: we were mainly talking about the idea of removing InsertBefore & After and how we currently need to that to emulate `IManageMessageFailures`. We both agreed that the current workaround isn't pretty and we need to find a better extensibility mechanisms. For that we need to have a better understanding on what you actually want to achieve with `IManageMessageFailures`. We don't want to reimplement IManageMessageFailures capabilities 1:1 (since that interface also seems weird) but we'd like to gather some usage scenarios where you want to modify the existing "failed message handling mechanisms" and come up with a better API.

Snippets/Snippets_6/UpgradeGuides/5to6/CustomErrorHandlingBehaviour.cs
petersgiles: didn't you leave a note to get rid of it

SimonCropp: Yes the class added no value inside the snippet. But u removed the class so the code will now no longer compile

SimonCropp: Only the register needs to be in the region

Snippets/Snippets_6/UpgradeGuides/5to6/CustomErrorHandlingBehaviour.cs
andreasohlund: Scenarios that I can think of (please add more)

* To get notified when message is retried or sent to error - Handled by the notifications
* Control the number of FLR's per message type/ex - Can this be done with a custom SLR policy?
* Control the number and delay of SLR's per message type/ex - Can this be done with a custom SLR policy?
* Send a given message to a different "error queue" - Add a behavior to the logical receive pipeline that try/catch and bus.Sends to the give queue

More?

Snippets/Snippets_6/UpgradeGuides/5to6/CustomErrorHandlingBehaviour.cs
indualagarsamy: @SimonCropp - I tried to take a look at the some of the Pipeline samples in v6, they all use this. I am not sure how you want to change this.
See: https://github.com/Particular/docs.particular.net/blob/master/samples/pipeline/stream-properties/Version_6/Shared/StreamSendBehavior.cs#L62

SimonCropp: unless u can think of a good exception to the rule i say we remove it from all the pipeline samples

indualagarsamy: @SimonCropp - I am not sure, that it improves readability by removing ConfigureAwait(false). 
It's just one line of code and it follows best practice. I actually see no harm in leaving it the way it currently is. Anyone wanting to add stuff to the pipeline will be looking at the samples and I think it serves best to point them to the right way of doing things, in this case, by using ConfigureAwait(false) - thoughts?

indualagarsamy: @SimonCropp - thoughts?

>@SimonCropp - I am not sure, that it improves readability by removing ConfigureAwait(false). 
It's just one line of code and it follows best practice. I actually see no harm in leaving it the way it currently is. Anyone wanting to add stuff to the pipeline will be looking at the samples and I think it serves best to point them to the right way of doing things, in this case, by using ConfigureAwait(false) - thoughts?

SimonCropp: Yeah leave as is

petersgiles:
@andreasohlund & @timbussmann It was my understanding that you guys were going to finish this off because you weren't happy with how this had to be done with the API as it is. Have you progressed it? Seems like nothing is happening. How do we finish it and get it done?

timbussmann:
> @andreasohlund & @timbussmann It was my understanding that you guys were going to finish this off 

that's new to me :) 

> you weren't happy with how this had to be done with the API as it is

I think we ruled this out as v6 release related work.

petersgiles:
> I think we ruled this out as v6 release related work.

that's news to me ;)

---

We should probably make a decision on what to do with it and I don't feel qualified to make that decision.

- Do we close it?
- Do we remove v6 label?
- Do we fix the bits of the API that make this tricky?

indualagarsamy:
@petersgiles - This is the PR I was telling you that I was going to work on. cc / @andreasohlund @timbussmann 
What I was going to do was:
1. Create a new doco on how to customize error message handling using the new pipeline. 
2. And in that doco, add some scenarios of what we already have
3. Link that up to the upgrade guide. 
Going to do this as soon as I wrap up with the satellite doco I have going right now.

SimonCropp:
Ok. No movement for a week. Do i kill this?

indualagarsamy:
@petersgiles - I've taken the work we've put in and moved it to here: https://github.com/Particular/docs.particular.net/pull/1418

PR #1303, opened 2016-03-11
Update concurrency.md

dvdstelt:
Small updates about locking and rowversion, with link to our own documentation instead of Ayendes

### DIFF
--- a/nservicebus/sagas/concurrency.md
+++ b/nservicebus/sagas/concurrency.md
@@ -32,19 +32,15 @@ NOTE: In NServiceBus Version 2 you had to create the constraint yourself in the 
 
 This works predictably due to reliance on the underlying database providing optimistic concurrency support. When more than one thread tries to update the same saga instance, the database detects it and only allows one of them to commit. If this happens the retries will occur and the race condition be solved.
 
+### SQL Server
+When running using NHibernate persistance, the NServiceBus framework enables pessimistic locking, resulting in saga data being locked for a single thread. Adding ["Version" property to the saga data](http://docs.particular.net/nservicebus/nhibernate/saga-concurrency) gives NHibernate the option to work its magic and gain additional performance, especially in high volume messaging environments.
+
+Another option is to use a [transaction isolation level](https://msdn.microsoft.com/en-us/library/system.transactions.isolationlevel.aspx) of serializable but that can cause [excessive locking](https://msdn.microsoft.com/en-us/library/ms173763.aspx) with considerable performance degradation, if not handled with care. Since NServiceBus 4 the default isolation level is "ReadCommitted".
+
+NOTE: "Serializable" is the default isolation level for distributed transactions.
+
+### RavenDb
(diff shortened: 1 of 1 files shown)


nservicebus/sagas/concurrency.md
```diff
@@ -32,19 +32,15 @@ NOTE: In NServiceBus Version 2 you had to create the constraint yourself in the
...
+### SQL Server
+When running using NHibernate persistance, the NServiceBus framework enables pessimistic locking, resulting in saga data being locked for a single thread. Adding ["Version" property to the saga data](http://docs.particular.net/nservicebus/nhibernate/saga-concurrency) gives NHibernate the option to work its magic and gain additional performance, especially in high volume messaging environments.
 
-Another option is to use a [transaction isolation level](https://msdn.microsoft.com/en-us/library/system.transactions.isolationlevel.aspx) of serializable but that causes [excessive locking](https://msdn.microsoft.com/en-us/library/ms173763.aspx) with considerable performance degradation.
+Another option is to use a [transaction isolation level](https://msdn.microsoft.com/en-us/library/system.transactions.isolationlevel.aspx) of serializable but that can cause [excessive locking](https://msdn.microsoft.com/en-us/library/ms173763.aspx) with considerable performance degradation, if not handled with care. Since NServiceBus 4 the default isolation level is "ReadCommitted".
```
dvdstelt: Not sure if it's "ReadCommitted" for Sagas. There's a hint (updlock, rowlock) added, which perhaps should be described here as well?

nservicebus/sagas/concurrency.md
```diff
@@ -32,19 +32,15 @@ NOTE: In NServiceBus Version 2 you had to create the constraint yourself in the
...
 
-NOTE: "Serializable" is the default isolation level for TransactionScopes.
-
-In NServiceBus Version 4 the default isolation level is
-"ReadCommitted", which is a more sensible default.
+NOTE: "Serializable" is the default isolation level for distributed transactions.
```
dvdstelt: Not default for TransactionScope, but distributed transactions are promoted to serializable

ramonsmits: @dvdstelt TransactionScope is something different than a DTC transaction. A TransactionScope can potentially be escalated to a fully distributed transaction when needed.

The default isolation level for a TransactionScope is also Serializable but we set that to ReadCommitted.

It is true that the default distributed transaction isolation level is serializable but that is not used due to it copying the selected isolation level from the scope. 

I don't see the behavior with a DTC transaction to be escalated to Serializable. It uses Read Committed as I get optimistic concurrency errors which could not be happening if the isolation level would be serializable as serializable should result in either sequential processing or deadlocking.

dvdstelt: I know what a scope vs dtc is, but what I did not know is that default is serializable for scope. My bad. Will see if I should fix it after all. Thanks for the comment.

SimonCropp:
@dvdstelt did u forget to delete the branch?

PR #1305, opened 2016-03-14
Updating unit testing snippets to v6

hmemcpy:
Connects to https://github.com/Particular/NServiceBus.Testing/issues/29

This adds testing snippets, converted to v6.
Also adding a few tweaks in the testing docs (mainly, removing the word "unit" whenever possible)

### DIFF
--- a/Snippets/Snippets_6/UnitTesting/AdditionalServices/MyMessage.cs
+++ b/Snippets/Snippets_6/UnitTesting/AdditionalServices/MyMessage.cs
@@ -1,0 +1,6 @@
+namespace Snippets6.UnitTesting.AdditionalServices
+{
+    class MyMessage
+    {
+    }
+}
--- a/Snippets/Snippets_6/UnitTesting/AdditionalServices/MyMessage.cs
+++ b/Snippets/Snippets_6/UnitTesting/AdditionalServices/MyMessage.cs
@@ -1,6 +1,0 @@
-namespace Snippets6.UnitTesting.AdditionalServices
-{
-    class MyMessage
-    {
-    }
-}
--- a/nservicebus/testing/index.md
+++ b/nservicebus/testing/index.md
@@ -7,7 +7,7 @@ redirects:
  - nservicebus/testing/unit-testing
 ---
 
-Developing enterprise-scale distributed systems is hard and testing them is just as challenging a task. The architectural approach supported by NServiceBus makes these challenges more manageable. And the testing facilities provided actually make unit testing endpoints and workflows easy. You can now develop service layers and long-running processes using test-driven development.
+Developing enterprise-scale distributed systems is hard and testing them is just as challenging a task. The architectural approach supported by NServiceBus makes these challenges more manageable. The testing facilities provided make unit testing endpoints and workflows easy, allowing developing service layers and long-running processes using Test-Driven Development.
 
 
 ## Getting started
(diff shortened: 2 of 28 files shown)


hmemcpy:
@Particular/docs-maintainers @timbussmann please review

Snippets/Snippets_6/UnitTesting/AdditionalServices/MyMessage.cs
```diff
@@ -0,0 +1,6 @@
+namespace Snippets6.UnitTesting.AdditionalServices
+{
+    class MyMessage
+    {
+    }
+}
```
timbussmann: any reason to not inline the messages into the snippet file?

nservicebus/testing/index.md
```diff
@@ -18,12 +18,12 @@ To install this package:
 ```
 Install-Package NServiceBus.Testing
 ```
-Once the package is installed you need ensure that you call `Test.Initialize()` (or any of its overloads) before executing any test method.
+Once the package is installed you need ensure that you call `Test.Initialize()` (or any of its overloads) before executing any test method. This is **not needed **for Version 6.
```
indualagarsamy: Avoid language like 'you' which has a personal tone. 
Please see: https://github.com/particular/docs.particular.net#language-preferences

Also, it should be Versions 6 and above:
See: https://github.com/particular/docs.particular.net#version-language

nservicebus/testing/index.md
```diff
@@ -18,12 +18,12 @@ To install this package:
...
 
 NOTE: To limit the assemblies and types scanned by the test framework it is possible to use the `Initialize()` overload that accepts a delegate you can use to customize the `ConfigurationBuilder`.
```
indualagarsamy: Change the usage of You. See above.

nservicebus/testing/index.md
```diff
@@ -39,7 +39,7 @@ snippet:TestingSaga
 
 ## Configuring unobtrusive message conventions
 
-If you are using [unobtrusive mode](/nservicebus/messaging/unobtrusive-mode.md) you need to configure the unit test support with those conventions as shown below.
+Prior to Version 6, if you are using [unobtrusive mode](/nservicebus/messaging/unobtrusive-mode.md) you need to configure the unit test support with those conventions as shown below.
```
indualagarsamy: See the proper version specific language, see https://github.com/particular/docs.particular.net#version-language
It should be something like 
For Versions 5 and below.

Also remove the usage of you.

nservicebus/testing/index.md
```diff
@@ -66,4 +66,6 @@ snippet:TestingAdditionalDependencies
 
 ## Constructor injected bus
 
+NOTE: The `IBus` interface was deprecated in Version 6, and replaced with the contextual `IMessageHandlerContext` parameter on the `IHandleMessages<T>.Handle()` methods.
```
indualagarsamy: Is there a documentation link we can point to? 
We talk about this in the upgrade guide. But a doco link would be best.

nservicebus/testing/index.md
SimonCropp: calling out that something is not needed in future versions at the end is kind of annoying. call it out before do people who keep up to date can skip the following text

Snippets/Snippets_6/UnitTesting/AdditionalServices/MyMessage.cs
hmemcpy: @timbussmann I could, but they are empty and provide no additional value, so I think it would just create noise

nservicebus/testing/index.md
hmemcpy: @indualagarsamy when previously adding deprecation notice, I used to link to the upgrade guide, but we realized this wasn't a good idea. Now, the deprecation notice suggests what to do instead, e.g. http://docs.particular.net/nservicebus/containers/injecting-ibus

We don't (currently) have a dedicated doc for handler context methods. Maybe we should add it? @timbussmann

Snippets/Snippets_6/UnitTesting/AdditionalServices/MyMessage.cs
timbussmann: well the still can be outside the region. but I think the additional files add more noise?

hmemcpy: Ah, you mean in the actual file, rather than rendered in docs? If so, sure, no problem!

PR #1321, opened 2016-03-17
Added suggestions about replacing the bus

SzymonPobiega:


SzymonPobiega:
cc @SimonCropp @mikeminutillo 
cc @andreasohlund we had a discussion on `centralized` vs `decentralized` and `federated`. Do you have anything to add/change here?

README.md
```diff
@@ -662,6 +662,14 @@ Avoid ambiguity.
 
 **Versions X** and NOT **VX** or **version X**.
 
+## Terminology
+
+### Bus
+
+We agreed that we the word `Bus` should be avoided. Some replacements incldue:
+ * When referring to the topology, use decentralized or federated (as opposed to centralized e.g. RabbitMQ)
```
mikeminutillo: Can we find a sentence for this use? I'm not sure under what circumstances I would use decentralized vs bus. 

Also, the bracketed bit makes it sound as if we are replacing the term "centralized" but I guess you mean if the terms are Broker vs Bus or Centralized vs Bus. Is that right?

andreasohlund: The terms I've been using is 

Bus vs Broker

Federated vs Centralized

SzymonPobiega: Here's the page that needs to be updated: http://docs.particular.net/nservicebus/transports/. I'll do it.

SzymonPobiega: I've clarified the federated part.

indualagarsamy:
@SzymonPobiega - What's the B-Word? That sounds wrong on so many levels. Call a bus the bus.

README.md
```diff
@@ -662,6 +662,14 @@ Avoid ambiguity.
 
 **Versions X** and NOT **VX** or **version X**.
 
+## Terminology
+
+### Bus
+
+We agreed that we the word `Bus` should be avoided. Some replacements incldue:
```
indualagarsamy: > We agreed that we the word `Bus` should be avoided. Some replacements incldue:

has typos. 

Also is this advice only for Versions 6 and above?

SzymonPobiega:
I am sorry @indualagarsamy . I was not aware of the meaning of it and I did not mean anything wrong. Thanks for correcting it.

README.md
SzymonPobiega: I fixed the typo and added a note that the `Bus` is allowed in pages specific to older versions. Bus most of the pages should be version-agnostic and should use the new terminology.

SimonCropp:
@SzymonPobiega thanks. i think we will refine this a bit more moving forward. but it is a good start

PR #1324, opened 2016-03-17
Update docs for SQL Server multi-X

weralabaj:
Connects to https://github.com/Particular/NServiceBus.SqlServer/issues/139


nservicebus/sqlserver/configuration.md
```diff
@@ -46,6 +46,27 @@ By passing the transport a custom factory method which will provide connection s
 
 snippet:sqlserver-custom-connection-factory
 
+## Custom database schemas
+
+The default schema in SQL Server transport is `dbo'. If no custom schema is provided in configuration, then transport will use the default value.
```
SimonCropp: incorrect tick? after dbo

SimonCropp: > If no custom schema is provided in configuration, then transport will use the default value.

isnt that kind of implied by the "default" in the previous statement?

Snippets/Snippets_5/Transports/SqlServer_2/MultiSchema.cs
```diff
@@ -0,0 +1,69 @@
...
+                .DefaultSchema("myschema");
+
+            #endregion
+        }
+
+        void NonStandardSchemaInConnectionString()
+        {
+            BusConfiguration busConfiguration = new BusConfiguration();
+
+            #region sqlserver-non-standard-schema-connString [2.1,2.0]
```
SimonCropp: 2.1 to 2.0?

Snippets/Snippets_5/Transports/SqlServer_2/MultiSchema.cs
```diff
@@ -0,0 +1,69 @@
...
+            #region sqlserver-multischema-config-pull 2.1
+
+            busConfiguration.UseTransport<SqlServerTransport>()
+                .UseSpecificConnectionInformation(endpointName =>
+                {
+                    if (endpointName == "sales")
+                        return ConnectionInfo.Create()
+                                             .UseSchema("salesSchema");
+                    if (endpointName == "billing")
+                        return ConnectionInfo.Create()
```
SimonCropp: i know this is minor but for consistency can u add the `{}` inside the ifs

Snippets/Snippets_5/Transports/SqlServer_2/MultiSchema.cs
```diff
@@ -0,0 +1,69 @@
...
+                                             .UseSchema("billingSchema");
+                    return null;
```
SimonCropp: shouldnt this throw?

SimonCropp:
@weralabaj as part of this r u doing a full review on that md file?

Snippets/Snippets_5/Transports/SqlServer_2/MultiSchema.cs
weralabaj: I want to have 2.1-2.x displayed. What's another way?

weralabaj:
@SimonCropp yes, that's coming.

Snippets/Snippets_5/Transports/SqlServer_2/MultiSchema.cs
SimonCropp: it is nuget syntax so `[2.1,3.0)`

SimonCropp:
ok awesome

Snippets/Snippets_5/Transports/SqlServer_2/MultiSchema.cs
weralabaj: not really, if no value is provided then it assumes defaults will be used (AFIR)

weralabaj:
@tmasternak Could you have a quick look? I'll continue tomorrow morning.

One more thing. I think that in terms of SQL transport docs one thing that we are missing is a section on distributed transactions, i.e. in what scenarios transaction will be escalated and how to prevent it (also covering the impact of persistence and business data storage choices). We have bits and pieces here and there, but it's difficult to put it all together and understand fully.

Then we could also move the [Outbox](http://docs.particular.net/nservicebus/sqlserver/configuration#sql-server-transport-the-outbox-and-user-data-disabling-the-dtc) and [Entity Framework caveats](http://docs.particular.net/nservicebus/sqlserver/configuration#sql-server-transport-the-outbox-and-user-data-disabling-the-dtc-entity-framework-caveats) over there, they don't fit very well in configuration IMO.

There are a few options related to where to place it:
- Extend multi-database.md document
- Extend design.md section on transactions (http://docs.particular.net/nservicebus/sqlserver/design#transactions-and-delivery-guarantees)
- Extract that transactions section to a separate page and add this info there

To me personally the latest makes most sense, thoughts on that?

Snippets/Snippets_6/Transports/SqlServer_3/MultiSchema.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        void OtherEndpointConnectionParamsPull()
+        {
+            EndpointConfiguration endpointConfiguration = new EndpointConfiguration();
+            #region sqlserver-multischema-config-pull
+
+            endpointConfiguration.UseTransport<SqlServerTransport>()
+                .UseSpecificSchema(tn =>
+                {
+                    if (tn == "sales")
```
tmasternak: I think it would be nice to show that tn here is not only endpoint name but also `error` and we can define here a custom schema for `error` as well.

nservicebus/sqlserver/multiple-databases.md
```diff
@@ -1,21 +1,66 @@
 ---
-title: Multi-database support
-summary: How to configure SQL Server transport to use multiple instances of the database and route messages between them.
+title: SQL Server transport setup - default, multi-schema, multi-catalog and multi-instance
```
tmasternak: should we change the name of md file as well and keep redirects for old `multiple-database.md`

tmasternak: Maybe chagne to `SQL Server transport deployment options`?

nservicebus/sqlserver/multiple-databases.md
```diff
@@ -1,21 +1,66 @@
...
+summary: How to configure and manage SQL Server transport in default, multi-schema, multi-catalog and multi-instance modes
 tags:
 - SQL Server
 - Transports
 ---
 
-The SQL Server transport supports selecting, on per-endpoint basis, where the table queues should be created. The selection can be done on multiple levels:
+In the default SQL Server transport setup, the messages and other data are stored in multiple tables in a single database. The default schema is `dbo`, but it is possible to [specify a custom schema](/nservicebus/sqlserver/configuration.md#custom-database-schemas).
```
tmasternak: What is other data?

nservicebus/sqlserver/multiple-databases.md
```diff
@@ -1,21 +1,66 @@
...
 
- * different schemas in a single database
- * different databases in a single SQL Server instance
- * different SQL Server instances
+In some environments it might be necessary to use a more advanced setup. The supported options are:
+ * ***default***: single schemas in a single database (Versions 1.x and higher)
```
tmasternak: No "s" at the end

nservicebus/sqlserver/multiple-databases.md
```diff
@@ -1,21 +1,66 @@
...
+ * ***multi-schema***: multiple schemas in a single database (Versions 2.1 and higher)
+ * ***multi-catalog***: multiple databases in a single SQL Server instance (Versions 2.1 and higher)
```
tmasternak: Should we change "database" to "catalogy" or is it going to be separate PR.

nservicebus/sqlserver/multiple-databases.md
```diff
@@ -1,21 +1,66 @@
...
+ * ***multi-instance***: multiple databases in various SQL Server instances (Versions 2.1 to 2.x).
 
-The transport will route messages to destination endpoints based on the configuration. If no specific configuration has been provided for a particular destination endpoint, the transport assumes the destination has the same configuration (schema, database and instance name/address) as the sending endpoint. If this assumption turns out to be false (the transport cannot connect to destination queue), an exception is thrown immediately. There is no store-and-forward mechanism on the transport level (and hence -- no dead-letter queue).
+The transport will route messages to destination endpoints based on the configuration. If no specific configuration has been provided for a particular destination endpoint, the transport assumes the destination has the same configuration (schema, database and instance name) as the sending endpoint. If the destination endpoint has a different configuration and it hasn't been provided, then exception will be thrown immediately (since the transport cannot connect to the destination queue).
```
tmasternak: Do we want to say that this not apply to ReplyTo or Publish in which case non-default configuration will be taken from header or subscription store?

nservicebus/sqlserver/multiple-databases.md
```diff
@@ -1,21 +1,66 @@
...
 
-NOTE: If the destination endpoint uses a different database or server instance, sending a message to it might cause the transaction to escalate to a distributed transaction which may not be desirable. Using the [Outbox](/nservicebus/outbox/ feature, DTC escalations can be avoided as long as the both the endpoint's Outbox and business data share the same database.
+WARNING: SQL Server transport does not support store-and-forward mechanism on a transport level, so there is no dead-letter queue. It means that in *multi-catalog* and *multi-database* modes if a remote endpoint's infrastructure (e.g. DTC or SQL Server instance) is down, then messages can't be delivered to it. That makes this particular endpoint, and all endpoints depending on it, unavailable. That problem can be addressed by using [Outbox](/nservicebus/outbox/) feature, as explained in [*Multi-catalog* and *multi-instance* with store-and-forward](#multi-catalog-and-multi-instance-with-store-and-forward) section.
```
tmasternak: Do we need to mention DLQ here. For me it's a bit misleading but it might be only me.

nservicebus/sqlserver/multiple-databases.md
tmasternak: It should be (Versions 2.1 to 2.x) - we do not support it yet in V3 and wont' in 3.0

nservicebus/sqlserver/multiple-databases.md
```diff
@@ -1,21 +1,66 @@
...
+- Supported in Versions 2.1 and higher
+- Simple configuration and setup
+- Doesn't require Distributed Transaction Coordinator (MS DTC)
+- The snapshot (backup) of the entire system state is done by backing up a single database (especially useful if business data is also stored in the same database)
+- A fine-grained security control on a database schema level
+- Can be monitored with ServiceControl
+
+## Multi-catalog
+
+- Supported in Versions 2.1 and higher
```
tmasternak: We do not support this in 3.0 - probably will in 3.x

nservicebus/sqlserver/multiple-databases.md
```diff
@@ -1,21 +1,66 @@
...
+- Requires Distributed Transaction Coordinator by default
+- Escalation to distributed transactions can be avoided by using Outbox and storing business data in the same database as Outbox data
```
tmasternak: Do we want to add that careful configuration `error` queue is required as well. If `error` queue is on different instance than input queue the escalation will happen as well.

nservicebus/sqlserver/multiple-databases.md
```diff
@@ -1,21 +1,66 @@
...
+
+In order to overcome the limitation caused by the lack of store-and-forward mechanism on a transport level, a higher level store-and-forward mechanism needs to be used. That functionality is essentially provided by [Outbox](/nservicebus/outbox/) feature, because when Outbox is enabled:
+ * Each endpoint has its own database in which it stores both the queues and the user data
+ * Messages are not dispatched immediately after the `Send()` method is called. Instead they are first stored in the Outbox table in the same database that endpoint's persistence is using. After the handler logic completes successfully, the messages stored in the Outbox table are forwarded to their destinations.
+ * If any of the forward operations fails, the message sending will be retried using the standard [retry mechanism](/nservicebus/errors/automatic-retries.md). This might result in sending some messages multiple times, which is known as `at-least-once` delivery. The Outbox feature performs automatically de-duplication of incoming messages based on their IDs, effectively providing `exactly-once` message delivery.
+
+
+=====> OLD => move to configuration.md <=========
```
SimonCropp: i think a merge conflict got pushed

nservicebus/sqlserver/multiple-databases.md
```diff
@@ -1,21 +1,66 @@
...
+
+- Supported in Versions 2.1 and higher
+- Requires Distributed Transaction Coordinator by default
+- Escalation to distributed transactions can be avoided by using Outbox and storing business data in the same database as Outbox data
+- Can be monitored with ServiceControl
+
+## Multi-instance
+
+- Supported in Versions 2.1 to 3.x, won't be supported in Versions 4.0 and higher
+- Requires Distributed Transaction Coordinator, even when using Outbox
```
tmasternak: This is the same as for multi-catalog. We can avoid this if `errors` are on the same instance.

tmasternak:
@weralabaj I think that multi-catalog and multi-instance are pretty much the same from the SqlServer perspective for now. There are same rules for using it as store-and-forward and when transactions escalate to DTC.
In current form both are considered legacy in V3 and will be removed in V4. It is possible however that in early 3.x we will provide multi-catalog but working only inside single sql instance and with different configuration.

I would propose to merge the multi-catalog and multi-instance descriptions for now. Thoughts? /cc: @weralabaj @mauroservienti

nservicebus/sqlserver/multiple-databases.md
weralabaj: Let's discuss.

nservicebus/sqlserver/multiple-databases.md
weralabaj: no, this is my message to future self

SimonCropp: lol fair enough

weralabaj:
@tmasternak Given this https://github.com/Particular/NServiceBus.SqlServer/issues/156, I think we should remove the section describing how to configure SC with SQL Server transport from here http://docs.particular.net/nservicebus/sqlserver/multiple-databases#other-endpoints-servicecontrol and add TODO to the issue saying that after fix is complete the doco needs to be updated.

weralabaj:
@tmasternak please, review. if I made some content mistakes (i.e. something is incorrect) then feel free to commit changes without asking.

then I'll ask for editor's review.

@mauroservienti could you help me with defining guidance around when to use each of the available options? as far I have only "A fine-grained security control on a database schema level" for multi-schema.

weralabaj:
@mauroservienti @ramonsmits please, could you review from the content perspective?

nservicebus/sqlserver/configuration.md
```diff
@@ -46,6 +43,69 @@ By passing the transport a custom factory method which will provide connection s
 
 snippet:sqlserver-custom-connection-factory
 
+NOTE: The connection string configuration API in NServiceBus core favors code over configuration for the current endpoint. If the same connection string is passed both in `app.config` and via the `ConnectionString()` method, the latter will be used.
```
ramonsmits: Does the word 'core' add value? I misread it a couple of times, reading code.

nservicebus/sqlserver/configuration.md
```diff
@@ -46,6 +43,69 @@ By passing the transport a custom factory method which will provide connection s
...
+
+## Multiple connection strings
+
+In [*multi-catalog* and *multi-instance* modes](/nservicebus/sqlserver/deployment-options.md) it is necessary to include additional information in configuration.  The sender needs to know the connection string of the receiver, and subscriber needs the connection string of the publisher.
```
ramonsmits: >, and subscriber

, and **the** subscriber

nservicebus/sqlserver/configuration.md
```diff
@@ -46,6 +43,69 @@ By passing the transport a custom factory method which will provide connection s
...
+
+Connection string for the remote endpoint can be configured in several ways:
```
ramonsmits: Might plural be more appropriate? 

Connection string**s** for the remote endpoint**s** can be configured in several ways:

nservicebus/sqlserver/configuration.md
```diff
@@ -46,6 +43,69 @@ By passing the transport a custom factory method which will provide connection s
...
+
+### Via the App.Config
+
+In Versions 2.1 to 2.x the endpoint-specific connection information is discovered by reading the connection strings from the configuration file with `NServiceBus/Transport/{name of the endpoint in the message mappings}` naming convention. If such a connection string is found, it is used for a given endpoint and this setting has precedence over the code-provided connection information.
```
ramonsmits: Line 46 mentions that code has preference over configuration. Now I'm confused...

nservicebus/sqlserver/configuration.md
```diff
@@ -46,6 +43,69 @@ By passing the transport a custom factory method which will provide connection s
...
 
+NOTE: The connection string configuration API in NServiceBus core favors code over configuration for the current endpoint. If the same connection string is passed both in `app.config` and via the `ConnectionString()` method, the latter will be used.
+
+## Multiple connection strings
+
+In [*multi-catalog* and *multi-instance* modes](/nservicebus/sqlserver/deployment-options.md) it is necessary to include additional information in configuration.  The sender needs to know the connection string of the receiver, and subscriber needs the connection string of the publisher.
+
+Connection string for the remote endpoint can be configured in several ways:
+
+### Via the App.Config
```
ramonsmits: Isn't is better to have each section have a header mentioning the version instead of hiding the version in the paragraph text?

nservicebus/sqlserver/configuration.md
```diff
@@ -46,6 +43,69 @@ By passing the transport a custom factory method which will provide connection s
...
+snippet:sqlserver-multidb-other-endpoint-connection-push
+
+
+### Via the configuration API - Pull mode
+
+The pull mode can be used when specific information is not available at configuration time. One can pass a `Func<String, ConnectionInfo>` that will be used by the SQL Server transport to resolve connection information at runtime.
+
+snippet: sqlserver-multidb-other-endpoint-connection-pull
+
+### Example
```
ramonsmits: I would swear I have seen the push/pull configuration somewhere before. Cannot find a sample:

I found: https://github.com/Particular/docs.particular.net/blob/master/samples/sqltransport-nhpersistence/Version_3/Receiver/Program.cs#L49

Its at a weird location...

nservicebus/sqlserver/deployment-options.md
```diff
@@ -0,0 +1,58 @@
...
+- nservicebus/sqlserver/multiple-databases
+---
+
+In default configuration SQL Server transport assumes that all tables used for storing messages are located in a single catalog and within a single schema. The configuration can be changed to partition message storage between different schemas and catalogs. The decision can me made on per queue basis (including error and audit queues).
+
+The supported deployment options are:
+ * ***default***: all queues are stored in a single catalog and a single schema (Versions 1.x and higher)
+ * ***multi-schema***: queues are stored in a single catalog but in more than one schema (Versions 2.1 and higher)
+ * ***multi-catalog***: queues are stored in multiple catalogs but on a single SQL Server instance. This mode is not explicitly supported by SQL Server transport, but can be achieved  by using multi-instance option. Therefore in this document both options are covered under *multi-instance* term.
+ * ***multi-instance***: queues are stored in multiple catalogs on more than one SQL Server instance (Versions 2.1 to 2.x).
```
ramonsmits: Wasn't multi instance already there in an earlier version? I suspect it is supported since 1.2

nservicebus/sqlserver/deployment-options.md
```diff
@@ -0,0 +1,58 @@
...
+redirects:
+- nservicebus/sqlserver/multiple-databases
+---
+
+In default configuration SQL Server transport assumes that all tables used for storing messages are located in a single catalog and within a single schema. The configuration can be changed to partition message storage between different schemas and catalogs. The decision can me made on per queue basis (including error and audit queues).
+
+The supported deployment options are:
+ * ***default***: all queues are stored in a single catalog and a single schema (Versions 1.x and higher)
+ * ***multi-schema***: queues are stored in a single catalog but in more than one schema (Versions 2.1 and higher)
+ * ***multi-catalog***: queues are stored in multiple catalogs but on a single SQL Server instance. This mode is not explicitly supported by SQL Server transport, but can be achieved  by using multi-instance option. Therefore in this document both options are covered under *multi-instance* term.
```
ramonsmits: multi catalog, single instance => indirectly support by using multi catalog but requires DTC.

nservicebus/sqlserver/deployment-options.md
```diff
@@ -0,0 +1,58 @@
...
+- Supported in Versions 2.1 and higher.
+- Has simple configuration and setup.
+- Doesn't require Distributed Transaction Coordinator (MS DTC).
+- The snapshot (backup) of the entire system state is done by backing up a single database. It is especially useful if business data is also stored in the same database.
+- Enables security configuration on a schema level.
+- Can't be monitored with ServiceControl.
+
+## Multi-instance (same applies for multi-catalog)
+
+- Supported in Versions 2.1 to 2.x, legacy support is offered in Versions 3.x, won't be supported directly in Versions 4.0 and higher.
```
ramonsmits: Multi-instance supported I think since 1.2

nservicebus/sqlserver/deployment-options.md
```diff
@@ -0,0 +1,58 @@
...
+- Can be monitored with ServiceControl.
+
+## Multi-schema
+
+- Supported in Versions 2.1 and higher.
+- Has simple configuration and setup.
+- Doesn't require Distributed Transaction Coordinator (MS DTC).
+- The snapshot (backup) of the entire system state is done by backing up a single database. It is especially useful if business data is also stored in the same database.
+- Enables security configuration on a schema level.
+- Can't be monitored with ServiceControl.
```
ramonsmits: Is this true @mauroservienti? I thought the schema support was backported?

ramonsmits:
@weralabaj I've made some changes but have to read it as a whole tomorrow.

nservicebus/sqlserver/configuration.md
weralabaj: Sorry @ramonsmits I don't understand that comment. Could you clarify what's the problem?

nservicebus/sqlserver/deployment-options.md
weralabaj: I'll check that, this is what the existing doco said so I assumed it's true.

weralabaj:
Ok @ramonsmits thanks, and ping me when it's ready to make my changes :)

nservicebus/sqlserver/deployment-options.md
tmasternak: @ramonsmits it's a thing to be done. https://github.com/Particular/NServiceBus.SqlServer/issues/156

nservicebus/sqlserver/configuration.md
weralabaj: I think that would require further reorganization of the content. At the moment I dont' have that information for all options.

nservicebus/sqlserver/configuration.md
weralabaj: So was I ;) That's why there we say "Core" and here it concerns transport. Don't ask me why it's inconsistent though, no idea.

weralabaj:
@Particular/docs-maintainers please, review

nservicebus/sqlserver/configuration.md
```diff
@@ -7,14 +7,12 @@ redirects:
 - nservicebus/sqlserver/concurrency
 ---
 
-
 ## Connection strings
 
-The SQL Server transport is built on top of ADO.NET and uses its connection pooling mechanism. This may result in connection pool being shared by transport and other parts of endpoint process. Depending on the situation it might be necessary to adjust the default connection pool size. See also [SQL Server Connection Pooling and Configuration](https://msdn.microsoft.com/en-us/library/8xx3tyca.aspx).
+NOTE: The SQL Server transport is built on top of ADO.NET and will use connection pooling. This may result in the connection pool being shared by the transport and other parts of the endpoint process. Depending on the situation it might be necessary to adjust the default connection pool size. For more details refer to the official [SQL Server Connection Pooling and Configuration](https://msdn.microsoft.com/en-us/library/8xx3tyca.aspx) document.
```
indualagarsamy: Why is this a NOTE? I think I there was a previous PR, where I had removed the NOTE.

>For more details refer to the official [SQL Server Connection Pooling and Configuration](https://msdn.microsoft.com/en-us/library/8xx3tyca.aspx) document.

I think it should be either `documentation` or `article`. Maybe:

nservicebus/sqlserver/configuration.md
```diff
@@ -7,14 +7,12 @@ redirects:
...
 
 Connection string can be configured in several ways:
 
-
```
indualagarsamy: 2 empty lines before any heading. See https://github.com/particular/docs.particular.net#spaces

nservicebus/sqlserver/configuration.md
weralabaj: Hm so that explains the merge conflicts I have?

nservicebus/sqlserver/configuration.md
```diff
@@ -28,7 +26,6 @@ By adding a connection named `NServiceBus/Transport` in the `connectionStrings`
 
 snippet:sqlserver-connection-string-xml
 
-
```
indualagarsamy: 2 empty lines before any heading. See https://github.com/particular/docs.particular.net#spaces

nservicebus/sqlserver/configuration.md
weralabaj: I thought it should be made more visible, because it's important information. But I don't really care whether it's note or not, so I'll remove it.

nservicebus/sqlserver/configuration.md
```diff
@@ -46,6 +43,68 @@ By passing the transport a custom factory method which will provide connection s
 
 snippet:sqlserver-custom-connection-factory
 
+## Multiple connection strings
+
+In [*multi-catalog* and *multi-instance* modes](/nservicebus/sqlserver/deployment-options.md) it is necessary to include additional information in configuration.  The sender needs to know the connection string of the receiver, and the subscriber needs the connection string of the publisher.
```
indualagarsamy: > In [*multi-catalog* and *multi-instance* modes](/nservicebus/sqlserver/deployment-options.md) it is necessary to include additional information in configuration. 

would be nice to add why. 
In order to be able to route messages to endpoints that are in different databases?

nservicebus/sqlserver/configuration.md
```diff
@@ -46,6 +43,68 @@ By passing the transport a custom factory method which will provide connection s
...
+
+Connection strings for the remote endpoint can be configured in several ways:
+
+### Via the App.Config
```
indualagarsamy: I'd move the app.config way as the last section, since we prefer the code first configuration approach.

nservicebus/sqlserver/configuration.md
```diff
@@ -46,6 +43,68 @@ By passing the transport a custom factory method which will provide connection s
...
+
+In Versions 2.1 to 2.x the endpoint-specific connection information is discovered by reading the connection strings from the configuration file with `NServiceBus/Transport/{name of the endpoint in the message mappings}` naming convention. If such a connection string is found, it is used for a given endpoint and this setting has precedence over the code-provided connection information.
```
indualagarsamy: > If such a connection string is found, it is used for a given endpoint and this setting has precedence over the code-provided connection information.

I thought code has precedence over app.config? Hmmm... I am not sure. I thought in the core, code config overrides app config. If the above statement is true and we have a conflict in behavior between core and transport, I think we need to raise an issue. Ping @andreasohlund ?

nservicebus/sqlserver/configuration.md
```diff
@@ -46,6 +43,68 @@ By passing the transport a custom factory method which will provide connection s
...
+
+Starting from Version 3.0, the connection string for a remote endpoint has to be provided using configuration API.
```
indualagarsamy: Should be Versions 3 and above.
See version naming convention here: https://github.com/Particular/docs.particular.net#version-language

nservicebus/sqlserver/configuration.md
indualagarsamy: Is there an explanation of push and pull modes?

nservicebus/sqlserver/configuration.md
```diff
@@ -46,6 +43,68 @@ By passing the transport a custom factory method which will provide connection s
...
+
+Given the following mappings:
+
+snippet:sqlserver-multidb-messagemapping
+
+and the following connection strings:
+
+snippet:sqlserver-multidb-connectionstrings
+
+The messages sent to the endpoint called `billing` will be sent to the database catalog `Billing` on the server instance `DbServerB`. The messages to the endpoint called `sales` will be sent to the default database catalog and database server instance, because the endpoint specific configuration wasn't passed, i.e. `MyDefaultDB` on server `DbServerA`.
```
indualagarsamy: sent is repeated too many times. 

How about:
The messages sent to the endpoint called `billing` will be dispatched to the database catalog `Billing` on server instance `DbServerB`. Because the endpoint configuration isn't specified for `sales`, any messages sent to the `sales` endpoint will be dispatched to the default database catalog and database server instance, in this example to `MyDefaultDB` on server `DbServerA`

nservicebus/sqlserver/configuration.md
```diff
@@ -46,6 +43,68 @@ By passing the transport a custom factory method which will provide connection s
...
+
+## Custom database schemas
+
+The default schema in SQL Server transport is `dbo`.
+The schema for the specific endpoint can be configured using `DefaultSchema` method:
+
+snippet:sqlserver-non-standard-schema
+
+In Versions 1.2.3 to 2.x it was also possible to pass custom schema in the connection string, using `Queue Schema` parameter:
```
indualagarsamy: > In Versions 1.2.3 to 2.x
Should this be in Versions 2.x and below ?

nservicebus/sqlserver/configuration.md
andreasohlund: Yea this is funky, another reason to just layer config on top of code first apis => then it's last one wins

> If the above statement is true and we have a conflict in behavior between core and transport

What specific part of the core are you referring to?

indualagarsamy: @andreasohlund - I was referring to specifying configuration in the core. My understanding was code config wins over config specified in app.config.

nservicebus/sqlserver/configuration.md
```diff
@@ -46,6 +43,68 @@ By passing the transport a custom factory method which will provide connection s
...
+ 
+snippet:sqlserver-non-standard-schema-connString
+snippet:sqlserver-non-standard-schema-connString-xml
+
+## Multiple custom schemas
```
indualagarsamy: Can we split this into different sub headings for different behaviors for versions rather than notes and sentences with Notice that? For example:
### For Versions x-y

### For Versions z and above

nservicebus/sqlserver/configuration.md
```diff
@@ -46,6 +43,68 @@ By passing the transport a custom factory method which will provide connection s
...
+
+If two endpoints use different schemas then additional configuration is required. The sender needs to know the schema of the receiver, and subscriber needs the schema of the publisher.
```
indualagarsamy: How about something like:

If the endpoints are configured to use a different schema, then additional configuration is required for proper message routing. The sending endpoint needs to know the schema information of the receiving endpoint. The same applies when the publishing endpoint and subscribing endpoint use different schemas. The subscriber will need the schema information of the publisher in order to send subscription requests and to receive events from the publisher.

nservicebus/sqlserver/configuration.md
```diff
@@ -53,11 +112,10 @@ In an environment where DTC is disabled and [Outbox](/nservicebus/outbox/) is en
 
 The following conditions need to be met:
 
- * the business specific data and the `Outbox` storage must be in the same database;
+ * the business specific data and the `Outbox` storage must be in the same database instance;
  * the user code accessing business related data must use the same `connection string` as the `Outbox` storage.
 
-
```
indualagarsamy: Keep the newline. I think the spacing guidelines is 2 before and 1 after.

nservicebus/sqlserver/deployment-options.md
```diff
@@ -0,0 +1,61 @@
...
+title: SQL Server transport deployment options
+summary: Describes available SQL Server transport deployment options
+tags:
+- SQL Server
+- Transports
+redirects:
+- nservicebus/sqlserver/multiple-databases
+---
+
+In default configuration SQL Server transport assumes that all tables used for storing messages are located in a single catalog and within a single schema. The configuration can be changed to partition message storage between different schemas and catalogs. The decision can me made on per queue basis (including error and audit queues).
```
indualagarsamy: > In default configuration SQL Server transport assumes that all tables used for storing messages are located in a single catalog and within a single schema.

I think this is better as:
When using the default configuration, SQL Server transport assumes that all tables used for storing messages for endpoints are located in a single catalog and within a single schema. 

>The decision can me made on per queue basis (including error and audit queues).

typo `can me` -> `can be`. 
How about: 
The schemas and catalogs can also be specified at a queue level. For example, the error and the audit queues can be configured to use a different schema and a different database catalog.

nservicebus/sqlserver/deployment-options.md
```diff
@@ -0,0 +1,61 @@
...
+
+The supported deployment options are:
+ * ***default***: all queues are stored in a single catalog and a single schema (Versions 1.x and higher)
+ * ***multi-schema***: queues are stored in a single catalog but in more than one schema (Versions 2.1 and higher)
```
indualagarsamy: > queues are stored in a single catalog but in more than one schema (Versions 2.1 and higher)

how about:
queues are stored in a single catalog but in more than one schema. This option is only available on Versions 2.1 and above.

nservicebus/sqlserver/deployment-options.md
```diff
@@ -0,0 +1,61 @@
...
+tags:
+- SQL Server
+- Transports
+redirects:
+- nservicebus/sqlserver/multiple-databases
+---
+
+In default configuration SQL Server transport assumes that all tables used for storing messages are located in a single catalog and within a single schema. The configuration can be changed to partition message storage between different schemas and catalogs. The decision can me made on per queue basis (including error and audit queues).
+
+The supported deployment options are:
```
indualagarsamy: Should we separate this based on version? It's confusing to know what is and what isn't for what versions.

nservicebus/sqlserver/deployment-options.md
```diff
@@ -0,0 +1,61 @@
...
+ * ***multi-catalog***: queues are stored in multiple catalogs but on a single SQL Server instance. This mode is indirectly supported by using *multi-instance* option, and requires using DTC. In this document both options are covered under *multi-instance* term.
+ * ***multi-instance***: queues are stored in multiple catalogs on more than one SQL Server instance (Versions 2.1 to 2.x).
+
+NOTE: To properly identify the chosen deployment option all queues that the endpoint interacts with need to be taken into consideration, including error and audit queues. If either of them are stored in a separate SQL Server instance then the deployment is a *multi-instance* one.
+
+The transport will route messages to destination based on the configuration. If no specific configuration has been provided for a particular destination, the transport assumes the destination has the same configuration as the sending endpoint (i.e. identical schema, catalog and instance name). If the destination has a different configuration and it hasn't been provided, then exception will be thrown immediately, because the transport wouldn't be able to connect to the destination queue.
```
indualagarsamy: >  then exception will be thrown immediately, because the transport wouldn't be able to connect to the destination queue.

do you mean, an exception will be thrown when the endpoint starts?

nservicebus/sqlserver/deployment-options.md
```diff
@@ -0,0 +1,61 @@
...
+redirects:
+- nservicebus/sqlserver/multiple-databases
+---
+
+In default configuration SQL Server transport assumes that all tables used for storing messages are located in a single catalog and within a single schema. The configuration can be changed to partition message storage between different schemas and catalogs. The decision can me made on per queue basis (including error and audit queues).
+
+The supported deployment options are:
+ * ***default***: all queues are stored in a single catalog and a single schema (Versions 1.x and higher)
+ * ***multi-schema***: queues are stored in a single catalog but in more than one schema (Versions 2.1 and higher)
+ * ***multi-catalog***: queues are stored in multiple catalogs but on a single SQL Server instance. This mode is indirectly supported by using *multi-instance* option, and requires using DTC. In this document both options are covered under *multi-instance* term.
```
indualagarsamy: > This mode is indirectly supported by using *multi-instance* option, and requires using DTC. In this document both options are covered under *multi-instance* term.

Since we also have a multi-instance option, why not specify this later?

nservicebus/sqlserver/deployment-options.md
```diff
@@ -0,0 +1,61 @@
...
+
+NOTE: To properly identify the chosen deployment option all queues that the endpoint interacts with need to be taken into consideration, including error and audit queues. If either of them are stored in a separate SQL Server instance then the deployment is a *multi-instance* one.
+
+The transport will route messages to destination based on the configuration. If no specific configuration has been provided for a particular destination, the transport assumes the destination has the same configuration as the sending endpoint (i.e. identical schema, catalog and instance name). If the destination has a different configuration and it hasn't been provided, then exception will be thrown immediately, because the transport wouldn't be able to connect to the destination queue.
+
+# Modes overview
+
+## Default (single schema, single catalog, single SQL Server instance)
+
+- Supported in Versions 1.0 and higher.
```
indualagarsamy: Since this is supported in all versions, and its default, can we remove this?

nservicebus/sqlserver/deployment-options.md
```diff
@@ -0,0 +1,61 @@
...
+- Supported in Versions 2.1 and higher.
+- Has simple configuration and setup.
+- Doesn't require Distributed Transaction Coordinator (MS DTC).
+- The snapshot (backup) of the entire system state is done by backing up a single database. It is especially useful if business data is also stored in the same database.
+- Enables security configuration on a schema level.
+- Can't be monitored with ServiceControl.
+
+## Multi-instance (and multi-catalog)
+
+- Supported in Versions 2.1 to 2.x, legacy support is offered in Versions 3.x, won't be supported directly in Versions 4.0 and higher.
```
indualagarsamy: I think we should add:
WARNING: This option will not be supported in Versions 4 and above.

nservicebus/sqlserver/deployment-options.md
```diff
@@ -0,0 +1,61 @@
...
+- Requires Distributed Transaction Coordinator, or using Outbox and storing business data in the same database as Outbox data.
+- Can't be monitored with ServiceControl.
+
+## Multi-instance with store-and-forward
+
+SQL Server transport does not support store-and-forward mechanism natively. It means that in *multi-instance* mode if a remote endpoint's infrastructure (e.g. DTC or SQL Server instance) is down, then messages can't be delivered. That makes this particular endpoint, and all endpoints depending on it, unavailable.
```
indualagarsamy: > That makes this particular endpoint, and all endpoints depending on it, unavailable. 

What's the context for this particular endpoint? Which endpoint are we talking about here?

nservicebus/sqlserver/configuration.md
andreasohlund: Correct, eg. the error q setting will use code config if present https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Recoverability/Faults/ErrorQueueSettings.cs#L14

nservicebus/sqlserver/configuration.md
weralabaj: But we have "The sender needs to know the connection string of the receiver, and the subscriber needs the connection string of the publisher." in the next sentence, isn't that clear then that they need that info to communicate?

nservicebus/sqlserver/configuration.md
weralabaj: Is that true for every version then?
I'm not sure either, this phrase has been here unmodified for a loooong time. I'll have to triple check it.

nservicebus/sqlserver/configuration.md
weralabaj: But then wouldn't it be from version 1.0? That also predates me, so I'm assuming here the original author put 1.2.3 on purpose. I didn't check that though.

nservicebus/sqlserver/configuration.md
weralabaj: Then we'll need to repeat some of the content. Not sure option which one is better, but I'll give it a go and see how much repetition that will result in.

nservicebus/sqlserver/deployment-options.md
weralabaj: Upon sending a message if I recall correctly.

nservicebus/sqlserver/deployment-options.md
weralabaj: Remote endpoint from the previous sentence. If I said "that remote endpoint" instead would it be clearer?

nservicebus/sqlserver/configuration.md
indualagarsamy: Thanks @andreasohlund, I raised an issue here. https://github.com/Particular/NServiceBus.SqlServer/issues/187
@weralabaj - thanks! As far as the original statement goes, we can leave it as how you have it and change it once the issue is addressed.

nservicebus/sqlserver/deployment-options.md
indualagarsamy: Can you make that more explicit, not sure what immediately meant, in what context.

nservicebus/sqlserver/configuration.md
indualagarsamy: It's not very clear.. the current statements say do this, provide this, but doesn't tell you why its needed. Somewhere down below, i had the following suggestion. You can decide if that makes more sense or not. 

If the endpoints are configured to use a different schema, then additional configuration is required for proper message routing. The sending endpoint needs to know the schema information of the receiving endpoint. The same applies when the publishing endpoint and subscribing endpoint use different schemas. The subscriber will need the schema information of the publisher to send subscription requests and to receive events from the publisher.

weralabaj: OK :)

nservicebus/sqlserver/deployment-options.md
indualagarsamy: @weralabaj - How about:

SQL Server transport does not support store-and-forward mechanism natively.  Therefore, if the receiving endpoint's infrastructure e.g. DTC or SQL Server instance is unavailable especially in a *multi-instance* mode, messages to the endpoint can't be delivered. The sending endpoint and all the other endpoints that depend on it will also be unavailable.

nservicebus/sqlserver/deployment-options.md
```diff
@@ -0,0 +1,61 @@
...
+
+- Supported in Versions 2.1 to 2.x, legacy support is offered in Versions 3.x, won't be supported directly in Versions 4.0 and higher.
+- Requires Distributed Transaction Coordinator, or using Outbox and storing business data in the same database as Outbox data.
+- Can't be monitored with ServiceControl.
+
+## Multi-instance with store-and-forward
+
+SQL Server transport does not support store-and-forward mechanism natively. It means that in *multi-instance* mode if a remote endpoint's infrastructure (e.g. DTC or SQL Server instance) is down, then messages can't be delivered. That makes this particular endpoint, and all endpoints depending on it, unavailable. 
+
+The problem can be addressed using [Outbox](/nservicebus/outbox/) feature. In addition Outbox can be used to avoid escalating transactions to DTC, given that each endpoint has its own database in which it stores input queue, error and audit queues, and the user data. That means it's not possible to avoid distributed transactions in case when any of the queues (including error and audit) is on a different SQL Server instance or catalog.
```
indualagarsamy: This shouldn't start as a new paragraph. I think this should be a continuation of the previous paragraph, so it reads more continuously. 

>The problem can be addressed using [Outbox](/nservicebus/outbox/) feature.

should be 
This problem can be addressed by using the [Outbox](/nservicebus/outbox/) feature.

>In addition Outbox can be used to avoid escalating transactions to DTC, given that each endpoint has its own database in which it stores input queue, error and audit queues, and the user data. That means it's not possible to avoid distributed transactions in case when any of the queues (including error and audit) is on a different SQL Server instance or catalog. 

The difference between when its possible to avoid DTC using Outbox and when it's not isn't very clear. How about the below? Did I understand this correctly? 

Also, the Outbox feature can be used to avoid escalating transactions to DTC, when each endpoint has a separate database for storing queues and business data on the same SQL Server instance. However, it's not possible to avoid distributed transactions when any of the queues (including error and audit) are on a different SQL Server instance or catalog.

nservicebus/sqlserver/deployment-options.md
```diff
@@ -0,0 +1,61 @@
...
+
+When using Outbox:
+
+ * Messages are not dispatched immediately after the `Send()` method is called. Instead, they are first stored in the Outbox table in the same database that the endpoint's persistence is using. After the handler logic completes successfully, the messages stored in the Outbox table are forwarded to their final destinations.
+ * If any of the forward operations fails, the message sending will be retried using the standard [retry mechanism](/nservicebus/errors/automatic-retries.md). This might result in sending some messages multiple times, which is known as `at-least-once` delivery. The Outbox feature performs automatically de-duplication of incoming messages based on their IDs, effectively providing `exactly-once` message delivery. The receiving endpoint should also be configured to use the outbox.
```
indualagarsamy: >If any of the forward operations fails, the message sending will be retried using the standard [retry mechanism](/nservicebus/errors/automatic-retries.md). This might result in sending some messages multiple times, which is known as `at-least-once` delivery. The Outbox feature performs automatically de-duplication of incoming messages based on their IDs, effectively providing `exactly-once` message delivery. The receiving endpoint should also be configured to use the outbox.

How about:

If any of the forward operations fails, the message sending will be retried using the standard [retry mechanism](/nservicebus/errors/automatic-retries.md). Attempting to retry the forward operation may result in dispatching some messages multiple times. However, the Outbox feature automatically de-duplicates the incoming messages based on their IDs, therefore providing `exactly-once` message delivery. For this to work, the receiving endpoint should also be configured to use the outbox.

nservicebus/sqlserver/deployment-options.md
```diff
@@ -0,0 +1,61 @@
...
+
+### Error and audit queues per instance
+
+As mentioned before, store-and-forward requires `error` and `audit` queues per endpoint. As a result system using such a deployment scenario cannot be monitored by a single ServiceControl instance.
```
indualagarsamy: Should this be in the SqlServer transport documentation? This seems like a limitation of ServiceControl and should be in relevant SC doco.

Snippets/Snippets_5/Snippets_5.csproj
```diff
@@ -378,7 +378,9 @@
     <None Include="Transports\SqlServer_2\ConnectionStrings.config" />
     <None Include="Transports\SqlServer_2\ConnectionStringsNamed.config" />
     <None Include="Transports\SqlServer_2\MultiDB.SC.config" />
-    <None Include="Transports\SqlServer_2\MultiSchema.config" />
+    <None Include="Transports\SqlServer_2\MultiSchema.config">
+      <SubType>Designer</SubType>
```
indualagarsamy: Do we need this subtype?

SimonCropp: This is a bug in vs where it randomly adds and removes subtype

nservicebus/sqlserver/configuration.md
weralabaj: Nothing beyond "In the push mode the whole collection of endpoint connection information objects is passed during configuration time." and "The pull mode can be used when specific information is not available at configuration time.".
I don't like it either but don't have a good idea of what we could replace it with.

weralabaj:
@indualagarsamy I've got so many issues with rebase that it seemed simpler to copy the content here https://github.com/Particular/docs.particular.net/pull/1346. Please, review the new one and merge if it's ok.

PR #1335, opened 2016-03-18
[WIP] Databus removing strategy

dvdstelt:
The change contains hints on removing strategy for Databus feature.
This was added after multiple questions in Desk on how/when to remove files that were saved using the Databus feature.


nservicebus/messaging/databus.md
```diff
@@ -68,6 +68,21 @@ snippet:MessageWithLargePayloadUsingConvention
...
+
+Looking at it from another perspective, the file system also isn’t transactional. Transactions spawned by NServiceBus might fail, but the file system cannot participate in them, resulting in a deleted file even when the transaction is aborted. Or if the outbox feature in NServiceBus is enabled, the message will be removed from the queuing storage, but it has not actually been processed yet. 
+
+There are a large number of scenarios where removing the file can cause a problem. The only person who actually can tell when which file could be removed, is the developer, who should come up with a strategy to remove files. An option is a strategy like removing the file after (Max(SLA) + x-days), so that it is unlikely that the file still hasn’t been processed.
+
+In the [databus sample](http://docs.particular.net/samples/databus/) you can see how you can obtain the file location:
+
+    public void Handle(MessageWithLargePayload message)
```
dvdstelt: The code should probably be updated in the sample. Is it okay to alter the current sample referred to?

nservicebus/messaging/databus.md
```diff
@@ -68,6 +68,21 @@ snippet:MessageWithLargePayloadUsingConvention
 
 NServiceBus DataBus implementations currently behave differently with regard to cleanup of physical attachments used to transfer data properties. `FileShareDataBus` **does not** remove physical attachments once the message is gone. `AzureDataBus` **does** remove Azure storage blobs used for physical attachments once the message is gone.
 
+### Custom cleanup strategy
+
+Theoretically NServiceBus cannot tell when a file should be deleted. Perhaps the message is (functionally) deferred for the file to be processed later, or it is passed along to another handler. When the file is removed upon committing the transaction, the deferred or passed on message cannot process the file anymore. Removing the file depends on your functional needs.
```
SimonCropp: Remove your

Theoretically is redundant 

I think listing the scenarios in point for would be better

nservicebus/messaging/databus.md
```diff
@@ -68,6 +68,21 @@ snippet:MessageWithLargePayloadUsingConvention
...
+
+Looking at it from another perspective, the file system also isn’t transactional. Transactions spawned by NServiceBus might fail, but the file system cannot participate in them, resulting in a deleted file even when the transaction is aborted. Or if the outbox feature in NServiceBus is enabled, the message will be removed from the queuing storage, but it has not actually been processed yet. 
+
+There are a large number of scenarios where removing the file can cause a problem. The only person who actually can tell when which file could be removed, is the developer, who should come up with a strategy to remove files. An option is a strategy like removing the file after (Max(SLA) + x-days), so that it is unlikely that the file still hasn’t been processed.
```
SimonCropp: I would re phrase the "person" "developer" to talk about "business" and "requirements"

nservicebus/messaging/databus.md
```diff
@@ -68,6 +68,21 @@ snippet:MessageWithLargePayloadUsingConvention
...
+
+In the [databus sample](http://docs.particular.net/samples/databus/) you can see how you can obtain the file location:
```
SimonCropp: I dont think people should need to look through the whole sample to get the location. Perhaps a snippet

nservicebus/messaging/databus.md
SimonCropp: This is lacking too much context to be useful

nservicebus/messaging/databus.md
```diff
@@ -68,6 +68,21 @@ snippet:MessageWithLargePayloadUsingConvention
 
 NServiceBus DataBus implementations currently behave differently with regard to cleanup of physical attachments used to transfer data properties. `FileShareDataBus` **does not** remove physical attachments once the message is gone. `AzureDataBus` **does** remove Azure storage blobs used for physical attachments once the message is gone.
 
+### Custom cleanup strategy
+
+Theoretically NServiceBus cannot tell when a file should be deleted. Perhaps the message is (functionally) deferred for the file to be processed later, or it is passed along to another handler. When the file is removed upon committing the transaction, the deferred or passed on message cannot process the file anymore. Removing the file depends on your functional needs.
+
+Looking at it from another perspective, the file system also isn’t transactional. Transactions spawned by NServiceBus might fail, but the file system cannot participate in them, resulting in a deleted file even when the transaction is aborted. Or if the outbox feature in NServiceBus is enabled, the message will be removed from the queuing storage, but it has not actually been processed yet.
```
SimonCropp: did u copy this from word? since `isn’t` has a dodgy single quote

SimonCropp: > Looking at it from another perspective

this seems a little awkward. perhaps just list the scenarios with a sub heading?

nservicebus/messaging/databus.md
SimonCropp: also this will be a broken link

nservicebus/messaging/databus.md
```diff
@@ -68,7 +68,22 @@ snippet:MessageWithLargePayloadUsingConvention
 
 NServiceBus DataBus implementations currently behave differently with regard to cleanup of physical attachments used to transfer data properties. `FileShareDataBus` **does not** remove physical attachments once the message is gone. `AzureDataBus` **does** remove Azure storage blobs used for physical attachments once the message is gone.
 
+### Custom cleanup strategy
```
indualagarsamy: Is the custom cleanup strategy only apply to FileShareDataBus? 
Should we change the first paragraph into two sections, i.e. FileShareDataBus and AzureDataBus and spell out the behavior differences?

nservicebus/messaging/databus.md
```diff
@@ -68,7 +68,22 @@ snippet:MessageWithLargePayloadUsingConvention
...
 
+NServiceBus cannot tell when a file should be deleted, as removing the file depends on your functional needs. There are a large number of scenarios where removing the file can cause a problem.
```
indualagarsamy: Avoid language with personal tone. See https://github.com/particular/docs.particular.net#language-preferences

How about:
NServiceBus does not remove the transferred message attachments once a message is delivered. Automatically removing these attachments can cause problems in many situations. For example:

nservicebus/messaging/databus.md
```diff
@@ -68,7 +68,22 @@ snippet:MessageWithLargePayloadUsingConvention
 
 NServiceBus DataBus implementations currently behave differently with regard to cleanup of physical attachments used to transfer data properties. `FileShareDataBus` **does not** remove physical attachments once the message is gone. `AzureDataBus` **does** remove Azure storage blobs used for physical attachments once the message is gone.
 
+### Custom cleanup strategy
 
+NServiceBus cannot tell when a file should be deleted, as removing the file depends on your functional needs. There are a large number of scenarios where removing the file can cause a problem. 
+
+- The message can be (functionally) deferred so that the file will be processed later, or it is passed along to another handler. When the file is removed upon committing the transaction, the deferred or passed on message cannot process the file anymore.
```
indualagarsamy: Can you split these into two cases? i.e. when a message is deferred and the other case? 
Also what does passed along to another handler mean?

nservicebus/messaging/databus.md
```diff
@@ -68,7 +68,22 @@ snippet:MessageWithLargePayloadUsingConvention
...
+- The file system isn’t transactional. Transactions spawned by NServiceBus might fail, but the file system cannot participate in them, resulting in a deleted file even when the transaction is aborted.
```
indualagarsamy: How about:
- The file system does not participate in distributed transactions. If for some reason, the message handler throws an exception and the transaction rolls back, the file delete operation on the attachment cannot be rolled back. Therefore, when the message is retried, the attachment will no longer be present causing additional problems.

nservicebus/messaging/databus.md
```diff
@@ -68,7 +68,22 @@ snippet:MessageWithLargePayloadUsingConvention
...
+- If the outbox feature in NServiceBus is enabled, the message will be removed from the incoming queue, but it might not actually have been processed yet.
```
indualagarsamy: `actually` is redundant and can be removed. 

Can we expand the outbox behavior a little more?

nservicebus/messaging/databus.md
```diff
@@ -68,7 +68,22 @@ snippet:MessageWithLargePayloadUsingConvention
...
 
+### Custom cleanup strategy
 
+NServiceBus cannot tell when a file should be deleted, as removing the file depends on your functional needs. There are a large number of scenarios where removing the file can cause a problem. 
+
+- The message can be (functionally) deferred so that the file will be processed later, or it is passed along to another handler. When the file is removed upon committing the transaction, the deferred or passed on message cannot process the file anymore.
+- The file system isn’t transactional. Transactions spawned by NServiceBus might fail, but the file system cannot participate in them, resulting in a deleted file even when the transaction is aborted.
+- If the outbox feature in NServiceBus is enabled, the message will be removed from the incoming queue, but it might not actually have been processed yet. 
+
+The business requirements should specify when the file should be processed and when it can be removed. Therefor the business should come up with a strategy to remove the files. An option is a strategy like removing the file after (Max(SLA) + x-days), so that it is unlikely that the file still hasn’t been processed.
```
indualagarsamy: typo on Therefore. 

Suggestion:
Evaluating the business requirements can indicate how long the processing of a message should take and when the attachments can safely be removed. One strategy to deal with these attachments is to set up a cleanup policy which removes any files after a certain number of days has passed based on business SLA.

nservicebus/messaging/databus.md
```diff
@@ -68,7 +68,22 @@ snippet:MessageWithLargePayloadUsingConvention
...
+
+To obtain the file location use the base path set up during confiration and the incoming DataBus properties
```
indualagarsamy: typo on `configuration`

Snippets/Snippets_3/DataBus/CleanupStrategy/Define.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace Snippets5.DataBus.CleanupStrategy
```
indualagarsamy: this looks wrong. Copy paste error?

Snippets/Snippets_3/DataBus/CleanupStrategy/Usage.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace Snippets5.DataBus.CleanupStrategy
```
indualagarsamy: this looks wrong. Copy paste error? Shouldn't this be Snippets_3 and not Snippets5?

Snippets/Snippets_4/DataBus/CleanupStrategy/Define.cs
```diff
@@ -0,0 +1,18 @@
+﻿namespace Snippets5.DataBus.CleanupStrategy
```
indualagarsamy: this looks wrong. Copy paste error? Shouldn't this be Snippets_4 and not Snippets5?

Snippets/Snippets_4/DataBus/CleanupStrategy/Usage.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace Snippets5.DataBus.CleanupStrategy
```
indualagarsamy: this looks wrong. Copy paste error? Shouldn't this be Snippets_4 and not Snippets5?

Snippets/Snippets_6/DataBus/CleanupStrategy/Define.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace Snippets5.DataBus.CleanupStrategy
```
indualagarsamy: this looks wrong. Copy paste error? Shouldn't this be Snippets_6 and not Snippets5?

Snippets/Snippets_6/DataBus/CleanupStrategy/Usage.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace Snippets5.DataBus.CleanupStrategy
```
indualagarsamy: this looks wrong. Copy paste error? Shouldn't this be Snippets_6 and not Snippets5?

nservicebus/messaging/databus.md
dvdstelt: Thanks @indu, used

nservicebus/messaging/databus.md
dvdstelt: Well, you could say the strategy applies to both. When first picking up support cases asking about FileShare not deleting files, I was told "it's not deleting, because you can't know when it should be deleted" and I thought that was logical and came up with some scenarios myself. Then I read that AzureDataBus actually does remove the files. Which seems kind of strange, especially considering the current PR. :)

And also considering this : https://github.com/Particular/NServiceBus/issues/846

I would say both need a cleaning strategy based on x-amount of days, which we can provide, but customers should also be able to provide their own?

Snippets/Snippets_6/DataBus/CleanupStrategy/Define.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    using NServiceBus;
+
+    public class Define
+    {
+        public Define()
+        {
+            #region DefineFileLocationForDatabusFiles
+            EndpointConfiguration endpointConfiguration = new EndpointConfiguration("endpointName");
+            endpointConfiguration.UseDataBus<FileShareDataBus>().BasePath(@"c:\database_files\");
```
indualagarsamy: Should this be databus_files or databus_attachments? instead of `\database_files\`?

nservicebus/messaging/databus.md
```diff
@@ -68,7 +68,25 @@ snippet:MessageWithLargePayloadUsingConvention
...
+
+- The message can be (functionally) deferred so that the file will be processed later. Removing the file after deferring the message, results in a message without the corresponding file.
+- Functional requirements might dictate the message to be available for a longer duration.
+- The file system does not participate in distributed transactions. If for some reason, the message handler throws an exception and the transaction rolls back, the file delete operation on the attachment cannot be rolled back. Therefore, when the message is retried, the attachment will no longer be present causing additional problems.
+- If the outbox feature in NServiceBus is enabled, the message will be removed from the incoming queue, but it might not have been processed yet. 
+
+The business requirements should specify when the file should be processed and when it can be removed. Therefore the business should come up with a strategy to remove the files. An option is a strategy like removing the file after (Max(SLA) + x-days), so that it is unlikely that the file still hasn’t been processed.
```
indualagarsamy: >The business requirements should specify when the file should be processed and when it can be removed. Therefore the business should come up with a strategy to remove the files. An option is a strategy like removing the file after (Max(SLA) + x-days), so that it is unlikely that the file still hasn’t been processed.

This paragraph is redundant as the following paragraph has the updated information.

Snippets/Snippets_3/DataBus/CleanupStrategy/Usage.cs
```diff
@@ -0,0 +1,17 @@
...
+    using System;
+    using System.IO;
+    using Snippets3.DataBus.DataBusProperty;
+
+    public class Usage
+    {
+        #region FileLocationForDatabusFiles
+        public void Handle(MessageWithLargePayload message)
+        {
+            string filename = Path.Combine(@"c:\databus_files\", message.LargeBlob.Key);
```
SimonCropp: how would they do this for a databus property that uses conventions

Snippets/Snippets_3/DataBus/CleanupStrategy/Define.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace Snippets3.DataBus.CleanupStrategy
+{
+    using NServiceBus;
+
+    public class Define
+    {
+        public Define()
+        {
+            #region DefineFileLocationForDatabusFiles
+            Configure.With()
```
SimonCropp: wouldnt this snippet suffice https://github.com/Particular/docs.particular.net/blob/master/Snippets/Snippets_4/DataBus/FileShare/DataBus.cs

nservicebus/messaging/databus.md
```diff
@@ -66,9 +66,45 @@ snippet:MessageWithLargePayloadUsingConvention
...
+
+* The supported data bus implementations do not participate in distributed transactions. If for some reason, the message handler throws an exception and the transaction rolls back, the delete operation on the attachment cannot be rolled back. Therefore, when the message is retried, the attachment will no longer be present causing additional problems.
+* The message can be deferred so that the file will be processed later. Removing the file after deferring the message, results in a message without the corresponding file.
+* Functional requirements might dictate the message to be available for a longer duration.
+* If the outbox feature in NServiceBus is enabled, the message will be removed from the incoming queue, but it might not have been processed yet. 
+
+
+### AzureDataBus Implementation
+
+AzureDataBus will **remove** the Azure storage blobs used for physical attachments after the message is processed if the `TimeToBeReceived` attribute is specified. When this attribute isn’t provided, the physical attachments will not be removed.
```
SimonCropp: dodgy quote in `isn’t`

SimonCropp: an attribute isnt the only way to define TTBR. link to this instead  http://docs.particular.net/nservicebus/messaging/discard-old-messages

nservicebus/messaging/databus.md
```diff
@@ -66,9 +66,45 @@ snippet:MessageWithLargePayloadUsingConvention
...
+
+
+### AzureDataBus Implementation
+
+AzureDataBus will **remove** the Azure storage blobs used for physical attachments after the message is processed if the `TimeToBeReceived` attribute is specified. When this attribute isn’t provided, the physical attachments will not be removed.  
+
+
+#### Cleanup Strategy for AzureDataBus
+
+Specify the `TimeToBeReceived` attribute. For more details, read this article on [discarding old messages](/nservicebus/messaging/discard-old-messages.md).
```
SimonCropp: not "just an attribute". see above

nservicebus/messaging/databus.md
indualagarsamy: :) Copy paste from TextEdit in Mac to MarkdownPad :)

Snippets/Snippets_3/DataBus/CleanupStrategy/Usage.cs
indualagarsamy: Discussed this @SimonCropp 
Also why do we need the Handle? What are we illustrating in a message handler? 
Shouldn't the cleanup policy be part of a timer, as opposed to a message handler?

Removing files as part of the handler is not a good idea because file deletes aren't part of DTC and can transactions can rollback.

indualagarsamy:
@dvdstelt - I was trying to make minor changes and have this merged. But ended up with more questions. After discussing with @SimonCropp here are the outstanding questions:
1. The snippets, i.e.,the define.cs is the same as the one we already have as @SimonCropp mentioned in an earlier comment. Also it's better not to have C:\ as the path and instead use a networkFileSharePath as a variable. 
2. The usage.cs only shows one variation, doesn't address the convention approach. 
3. Cleanup strategy needs a little more work in how the users can remove those files, i.e. use of a timer, etc. 
Perhaps we can modify the databus sample to implement a cleanup strategy (doesn't have to be part of this PR, we can do that in a separate PR). 
@SimonCropp - please add your comments in case I've missed anything.

nservicebus/messaging/databus.md
indualagarsamy: I am linking to that article in the cleanup stragegy. Will remove the word attribute. Replacing it with value instead to be generic.

nservicebus/messaging/databus.md
indualagarsamy: will fix this.

Snippets/Snippets_3/DataBus/CleanupStrategy/Usage.cs
dvdstelt: It was a question by a customer how to obtain the file location. I'm not surprised they can't figure it out by themselves, that's why I added it. And the fact that it's in "handle" method, is so they understand that it's the receiving side.

And hopefully they understand that removing files in the handler isn't smart, due to the story they just read. But I could add something like "// Execute cleanup strategy, like scheduling the delete after x-days"

SimonCropp:
@dvdstelt have u had a look at @indualagarsamy feedback?

Snippets/Snippets_6/DataBus/CleanupStrategy/Define.cs
dvdstelt: Definitely databus instead of database! Good catch! And I see the current doco mentioned attachments, I'll rename that as well.

Snippets/Snippets_3/DataBus/CleanupStrategy/Usage.cs
indualagarsamy: @dvdstelt - Often times whether we like it or not, users pick stuff from samples. So, I am all for scheduling a delete after x days policy and executing the cleanup as opposed to doing it in a message handler.

indualagarsamy:
@dvdstelt - did you have a chance to read through this:

>@dvdstelt - I was trying to make minor changes and have this merged. But ended up with more questions. After discussing with @SimonCropp here are the outstanding questions:
1. The snippets, i.e.,the define.cs is the same as the one we already have as @SimonCropp mentioned in an earlier comment. Also it's better not to have C:\ as the path and instead use a networkFileSharePath as a variable. 
2. The usage.cs only shows one variation, doesn't address the convention approach. 
3. Cleanup strategy needs a little more work in how the users can remove those files, i.e. use of a timer, etc. 
Perhaps we can modify the databus sample to implement a cleanup strategy (doesn't have to be part of this PR, we can do that in a separate PR). 
@SimonCropp - please add your comments in case I've missed anything.

Snippets/Snippets_3/DataBus/CleanupStrategy/Usage.cs
dvdstelt: @indualagarsamy What would that result in, technically? Do you mean sending a deferred message that will later execute the cleanup policy, or some other NSB feature that will handle this?

dvdstelt:
Sorry, can't have missed the messages, but must've slipped into the background.

1. Definitely true, will update this
2. Should I add both?
3. Created a new issue #1402

dvdstelt:
I've updated v5 snippet, let me know if this is better, I'll update the other version snippets as well.

SimonCropp:
@dvdstelt can u rebase

dvdstelt:
Rebased

Snippets/Snippets_3/DataBus/CleanupStrategy/Usage.cs
indualagarsamy: @dvdstelt - sorry i missed your question. I was thinking more of a simple timer that executes a cleanup operation/check. @SimonCropp - thoughts?

mikeminutillo:
Doesn't this leak the internal details of the `FileShareDataBus`? Wouldn't we be better off adding a `Delete(key)` to the `IDataBus` interface and calling that with the key? The File Share one could implement what you are talking about here and we could move the functionality for deleting old databus properties after the TTBR expires into an easily installed plugin (or even make it a part of the DataBus feature). 

Even if we don't go that far, adding the `Delete(key)` behavior now would simplify this code a lot.

samples/containers/unity/Unity_3/Sample/app.config
```diff
@@ -10,6 +10,38 @@
         <assemblyIdentity name="Microsoft.Practices.ServiceLocation" publicKeyToken="31bf3856ad364e35" culture="neutral" />
         <bindingRedirect oldVersion="0.0.0.0-1.3.0.0" newVersion="1.3.0.0" />
       </dependentAssembly>
+      <dependentAssembly>
```
weralabaj: Do we need those redirects?

weralabaj:
@dvdstelt should that be a WIP? can I help to complete it?

dvdstelt:
@weralabaj I got some questions on an actual example for cleaning strategy and updated v5 snippet for this, to see if it was better.
https://github.com/Particular/docs.particular.net/pull/1335#issuecomment-207594122

Never got a response though. I'd rather have the cleanup strategy implemented like with ASB, then the extensive code snippet in doco.

Including sample:
```
        #region FileLocationForDatabusFiles
        public void Handle(MessageWithLargePayload message)
        {
            string filename = message.LargeBlob.Key;
            Bus.Defer(TimeSpan.FromDays(30), new RemoveDatabusAttachment(filename));
        }

        public void Handle(RemoveDatabusAttachment message)
        {
            CleanUp(message.Filename);
        }

        private void CleanUp(string filename)
        {
            // Code to clean up
        }
        #endregion
```
Excluding
```
        #region FileLocationForDatabusFiles
        public void Handle(MessageWithLargePayload message)
        {
            string filename = Path.Combine(@"c:\databus_files\", message.LargeBlob.Key);
            Console.WriteLine(filename);
        }
        #endregion
```
Option #3 
```
        #region FileLocationForDatabusFiles
        public void Handle(MessageWithLargePayload message)
        {
            // Example how to obtain exact file location
            string filename = Path.Combine(@"c:\databus_files\", message.LargeBlob.Key);
            // Now execute cleanup strategy, for example by deferring a new message and removing file after 30 days.
        }
        #endregion
```

dvdstelt:
It's also a my fault, as this isn't high on my prio list, although I keep coming back at it :-)

indualagarsamy:
@dvdstelt - Based on my first few reviews, I suggested the following:

> The snippets, i.e.,the define.cs is the same as the one we already have as @SimonCropp mentioned in an earlier comment. Also it's better not to have C:\ as the path and instead use a networkFileSharePath as a variable. 

The snippets still point to C:\
Also, rather than have a cleanup operation on a message handler (in a message, deferred or scheduled) a simpler approach would be to use a timer.

dvdstelt:
@indualagarsamy Hmmm, you're right. I could swear I fixed that.
And you mean something like `var timer = new System.Timers.Timer(really_large_number);`? Because that's not durable.

dvdstelt:
> Doesn't this leak the internal details of the FileShareDataBus? Wouldn't we be better off adding a Delete(key) to the IDataBus interface and calling that with the key?

@mikeminutillo You are absolutely right, but they already leak. The fact that we mention them in a snippet doesn't make them leak more than before! ;-)

Fact is that several customers are asking how to get rid of old files and nothing is mentioned in the doco on how to achieve this. So after giving the same reply for several times, I thought of writing it in doco. Now the ideal solution would be to solve one of the issues mentioning this

- https://github.com/Particular/NServiceBus/issues/3231
- https://github.com/Particular/NServiceBus/issues/846
- https://github.com/Particular/NServiceBus/issues/3205
- https://github.com/Particular/ServiceControl/issues/607 (This would be my choice!)
- https://github.com/Particular/ServiceControl/issues/608

But that's not happening for some time. So I think we need to provide _some information_ to help out customers, as they're completely in the dark at the moment.

dvdstelt:
@indualagarsamy @SimonCropp How can we get this done? Should we schedule a call to discuss and decide on final solution? I'd love to get this off of my list.
d

nservicebus/messaging/databus.md
```diff
@@ -66,9 +66,46 @@ snippet:MessageWithLargePayloadUsingConvention
...
 
-NServiceBus DataBus implementations currently behave differently with regard to cleanup of physical attachments used to transfer data properties. `FileShareDataBus` **does not** remove physical attachments once the message is gone. `AzureDataBus` **does** remove Azure storage blobs used for physical attachments once the message is gone.
+NServiceBus DataBus implementations currently behaves differently with regard to cleanup of physical attachments used to transfer data properties depending on the implementation used.
 
 
+### Reasons why attachments are not removed by default
+
+Automatically removing these attachments can cause problems in many situations. For example:
+
+* The supported data bus implementations do not participate in distributed transactions. If for some reason, the message handler throws an exception and the transaction rolls back, the delete operation on the attachment cannot be rolled back. Therefore, when the message is retried, the attachment will no longer be present causing additional problems.
```
andreasohlund: I would mention the error q/ retrying from SP since that can happen "days" after the message failed

SimonCropp:
ok this is good enough for now. I will add a cleanup strategy example to the DB sample https://github.com/Particular/docs.particular.net/issues/1462

dvdstelt:
Now the cleanup strategy idea is in the v5 snippet, but not in v3 & v6?

Second snippet under this : http://docs.particular.net/nservicebus/messaging/databus#databus-attachments-cleanup-filesharedatabus

SimonCropp:
should be fixed now

PR #1337, opened 2016-03-20
FLR does not work without transport transactions

### LINKED ISSUE #1320: Retries only work when transactions are enabled
based on the discussion here https://github.com/Particular/NServiceBus/issues/3476#issuecomment-197518418 the question came up whether we should improve the [retries doco](http://docs.particular.net/nservicebus/errors/automatic-retries) to state that retries only work when transactions are enabled?

MarcinHoppe:
Resolves #1320.


SimonCropp:
@MarcinHoppe thanks

PR #1341, opened 2016-03-22
Simplified the ServiceControl steps for compaction

johnsimons:
Changed to use esentutl


johnsimons:
@indualagarsamy @gbiellem here is PR for the compaction using esentutl.
Feel free to do any adjustments to it.

servicecontrol/db-compaction.md
```diff
@@ -7,80 +7,39 @@ tags:
 ---
 
 ## Overview
+In order to compact the embedded RavenDB database used by ServiceControl instances, the Extensible Storage Engine Utility (esentutl) needs to be run.
```
SimonCropp: can we have a link to esentutl official doco?

servicecontrol/db-compaction.md
```diff
@@ -7,80 +7,39 @@ tags:
...
 
-ServiceControl 1.4 introduced a database maintenance feature which allows ServiceControl to be run with all features except for RavenDB Studio disabled. While in this mode no messages are ingested from the queuing system.
+## Step 1: Backup ServiceControl instance
 
-Once ServiceControl is running in this mode the following procedure can be used to compact the embedded RavenDB database.
+- Follow the [backup instructions](backup-sc-database#backup).
```
SimonCropp: backup-sc-database needs a .md to not be a broken link

servicecontrol/db-compaction.md
```diff
@@ -7,80 +7,39 @@ tags:
 ---
 
 ## Overview
```
indualagarsamy: @johnsimons @gbiellem - instead of removing the manual steps, do we keep them as a secondary option?
Also, do we need to mention the cases that @boblangley mentioned, i.e. for the esent utl to work SC service needs to be shutdown properly?

servicecontrol/db-compaction.md
johnsimons: done

servicecontrol/db-compaction.md
johnsimons: done

servicecontrol/db-compaction.md
```diff
@@ -7,80 +7,39 @@ tags:
...
+In order to compact the embedded RavenDB database used by ServiceControl instances, the [Extensible Storage Engine Utility (esentutl)](https://technet.microsoft.com/en-us/library/hh875546.aspx) needs to be run.
 
-ServiceControl 1.4 introduced a database maintenance feature which allows ServiceControl to be run with all features except for RavenDB Studio disabled. While in this mode no messages are ingested from the queuing system.
+## Step 1: Backup ServiceControl instance
 
-Once ServiceControl is running in this mode the following procedure can be used to compact the embedded RavenDB database.
+- Follow the [backup instructions](backup-sc-database.md#backup).
 
+## Step 1: Stop ServiceControl
```
mikeminutillo: There are 2 Step 1s

servicecontrol/db-compaction.md
```diff
@@ -7,80 +7,39 @@ tags:
 ---
 
 ## Overview
+In order to compact the embedded RavenDB database used by ServiceControl instances, the [Extensible Storage Engine Utility (esentutl)](https://technet.microsoft.com/en-us/library/hh875546.aspx) needs to be run.
```
mikeminutillo: Is it worth adding here something about Raven (/ServiceControl) not recovering disk space once it is consumed?

servicecontrol/db-compaction.md
gbiellem: @indualagarsamy  I like the idea of retaining the secondary option although this get messy when we move to Raven 3 embedded and have to document a second UI.  But then we'll  also not be using teh system DB anymore so perhaps we'll need a whole separate doc for that.

Regarding  Clean Shutdown, yes the compact will fail if the DB isn't gracefully shutdown. That's my experience and  also what @boblangley  described.   There is a command line recovery option to get the DB into a clean state .  Based on this [thread](https://groups.google.com/forum/#!topic/ravendb/eKUYf2cHUkQ) it looks like we should also tell our users not to run repair prior to running recovery




.

servicecontrol/db-compaction.md
johnsimons: fixed

indualagarsamy:
@johnsimons - I reviewed with @gbiellem and added the old steps as a backup and added another esentutl command for recovery.

PR #1345, opened 2016-03-23
ServiceControl - Changes for SC issue 693

gbiellem:
Documents the changes in https://github.com/Particular/ServiceControl/pull/693
This should be merged when SC 1.12 ships

### DIFF
--- a/servicecontrol/installation-powershell.md
+++ b/servicecontrol/installation-powershell.md
@@ -1,6 +1,7 @@
 ---
 title: Managing ServiceControl Instances via PowerShell
-summary: Managing ServiceControl Instances via PowerShell
+summary: Managing ServiceControl Instances via PowerShell.
+reviewed: 2016-03-24
 tags:
 - ServiceControl
 - Installation
@@ -10,8 +11,7 @@ tags:
 
 ## ServiceControl PowerShell
 
-ServiceControl 1.7 introduced a new graphical management utility to add, remove, update and delete instances of the ServiceControl service. 
-These actions and some additional tools have also exposed via PowerShell module called `ServiceControlMgmt`
+ServiceControl 1.7 introduced a new graphical management utility to add, remove, update and delete instances of the ServiceControl service. These actions and some additional tools have also exposed via PowerShell module called `ServiceControlMgmt`.
 
 
 ## Prerequisites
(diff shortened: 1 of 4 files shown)


servicecontrol/installation-powershell.md
```diff
@@ -108,6 +108,15 @@ Invoke-ServiceControlInstanceUpgrade -Name <Instance To upgrade>
 The upgrade will stop the service if it is running.
 
 
+Note: Version 1.12 introduced a new mandatory app setting which allows error forwarding to be configured. 
+When upgrading instances running on Version 1.11.1 and below the `Invoke-ServiceControlInstanceUpgrade` requires a value for this setting.
+
+```bat
+Invoke-ServiceControlInstanceUpgrade -Name <Instance To upgrade> -ForwardErorrMessages [$true|$false]
+```
+he Error Forwarding Queue queue exists to allow external tools to receive error messages. If there is no process reading messages from the Error Forwarding Queue this setting should be `$false`.
```
SimonCropp: "he"

gbiellem: Damn it - "The"

SimonCropp: should there be an upgrade guide for this release?

gbiellem: @SimonCropp 

Wasn't planning on it, i was only to cover it off in release notes. I know the note regarding "Version 1.12 introduced a new mandatory app setting" caught your eye.  It doesn't tell the whole story thought.    If they are doing an interactive update the SC Management util it prompts them to supply the value for the setting:

capture

Same story for anyone using the PS cmdlet to update an instance, the PS cmdlet will force them to answer the question.

capture2

For unattended upgrades we require a new command line switch to update the instance,
If the end user doesn't supply it the MSI install will log the fact it wasn;t supplied and complete the update of the SCMU and PS modules only,  The user can then update manually via the SCMU or PS cmdlet

For unattended installations  the file they supply must contain the setting, it not it does a similar thing to the upgrade above - SCMU and PS are installed and the reason we didn't accept the file is logged. The log entry tells them to update the supplied unattended file.

Do you think it's sufficient to have this info in the release notes and docs  or do you still think a specific upgrade guide is required?

//cc @johnsimons

SimonCropp:
review done. merge when SC is released

PR #1349, opened 2016-03-23
Enabling installers in sql transport sample - native integration

WojcikMike:
@Particular/docs-maintainers please review & merge

### DIFF
--- a/samples/sqltransport/native-integration/Version_3/Receiver/PlaceOrderHandler.cs
+++ b/samples/sqltransport/native-integration/Version_3/Receiver/PlaceOrderHandler.cs
@@ -4,7 +4,7 @@ using NServiceBus.Logging;
 
 public class PlaceOrderHandler : IHandleMessages<PlaceOrder>
 {
-    static ILog log = LogManager.GetLogger<LegacyOrderDetectedHandler>();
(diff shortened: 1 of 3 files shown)


samples/sqltransport/native-integration/Version_3/Receiver/PlaceOrderHandler.cs
```diff
@@ -4,7 +4,7 @@
 
 public class PlaceOrderHandler : IHandleMessages<PlaceOrder>
 {
-    static ILog log = LogManager.GetLogger<LegacyOrderDetectedHandler>();
+    static ILog log = LogManager.GetLogger<PlaceOrderHandler>();
```
WojcikMike: I also decided to fix this copy&paste issue, as this is obviously wrong.

SimonCropp: @WojcikMike good catch

WojcikMike: After running sample first I thought that this is a bug in v3 and wrong handler is called, turned out to be something more trivial

PR #1365, opened 2016-03-30
Explaining SLR and Error behaviros with transport transactions disabled

tmasternak:
Connects to: https://github.com/Particular/NServiceBus/issues/3032

### DIFF
--- a/nservicebus/errors/automatic-retries.md
+++ b/nservicebus/errors/automatic-retries.md
@@ -31,7 +31,7 @@ Note: The configured value describes the minimum number of times a message will 
 
 ### Transport transaction requirements
 
-The FLR mechanism is implemented by rolling back the [transport transaction](/nservicebus/messaging/transactions.md). This returns the message to the input queue and the endpoint processes it again immediately. Aborting the receive operation when transactions are turned off would results in message loss. Therefore FLR cannot be used when transport transactions are disabled. 
+The FLR mechanism is implemented by rolling back the [transport transaction](/nservicebus/messaging/transactions.md). This returns the message to the input queue and the endpoint processes it again immediately. Aborting the receive operation when transactions are turned off would results in a message loss. Therefore FLR cannot be used when transport transactions are disabled. 
 
 
 ### Configuring FLR using app.config
@@ -72,7 +72,7 @@ SLR can be configured in several ways.
 
 ### Transport transaction requirements
 
-The SLR mechanism is implemented by rolling back the [transport transaction](/nservicebus/messaging/transactions.md) and scheduling message for [delayed-delivery](/nservicebus/messaging/delayed-delivery.md) next time it is picked from the input queue. Aborting the receive operation when transactions are turned off would results in message loss. Therefore SLR cannot be used when transport transactions are disabled.
+The SLR mechanism is implemented by rolling back the [transport transaction](/nservicebus/messaging/transactions.md) and scheduling message for [delayed-delivery](/nservicebus/messaging/delayed-delivery.md) the next time it is picked from the input queue. Aborting the receive operation when transactions are turned off would results in a message loss. Therefore SLR cannot be used when transport transactions are disabled.
 
 
 ### Configuring SLR using app.config
--- a/nservicebus/errors/automatic-retries.md
+++ b/nservicebus/errors/automatic-retries.md
@@ -1,37 +1,38 @@
 ---
 title: Automatic Retries
-summary: With retries, the message causing the exception is instantly retried configured number of times before forwarding to the error queue.
+summary: The message which caused the exception during processing, is automatically retried for the configured number of times before being forwarded to the error queue.
 tags:
 - Second Level Retry
 - Error Handling
 - Exceptions
 - Retry
+reviewed: 2016-03-31
 redirects:
  - nservicebus/second-level-retries
 related:
 - samples/faulttolerance
 ---
 
-Sometimes processing of a message can fail. This could be due to a transient problem like a deadlock in the database, in which case retrying the message a few times might overcome this problem. Or, if the problem is more protracted, like a third party web service going down or a database being unavailable, where it might be useful to wait a little longer before retrying the message again.
+Sometimes processing of a message fails. This could be due to a transient problem like a deadlock in the database, in which case retrying the message a few times should solve the issue. If the problem is more protracted, like a third party web service going down or a database being unavailable, solving the issue would take longer. It is therefore useful to wait longer before retrying the message again.
 
-For situations like these, NServiceBus offers two levels of retries:
+NServiceBus offers two levels of retries:
 
-- First Level Retry(FLR) is for the transient errors where quick successive retries could solve the problem.
-- Second Level Retry(SLR) is when a small delay is needed between retries.
+- First Level Retry(FLR) is for transient errors, where quick successive retries solve the problem.
+- Second Level Retry(SLR) is for errors that persist after FLR, where a small delay is needed between retries.
 
-NOTE: When a message cannot be deserialized, it will bypass all retry mechanisms both the FLR and SLR and the message will be moved directly to the error queue.
+NOTE: When a message cannot be deserialized, it will bypass all retry mechanisms and the message will be moved directly to the error queue.
 
 
 ## First Level Retries
 
-NServiceBus automatically retries the message when an exception is thrown during message processing up to five successive times by default. This value can be configured through `app.config` or code.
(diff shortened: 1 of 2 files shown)


nservicebus/errors/index.md
```diff
@@ -18,8 +18,9 @@ NServiceBus has exception catching and handling logic of its own which surrounds
 
 When a message fails NServiceBus [automatically retries](/nservicebus/errors/automatic-retries.md) the message. On repeated failure NServiceBus forwards that message to a designated error queue.
 
-Error queue can be configured in several ways.
+NOTE: When running with [transport transactions](/nservicebus/messaging/transactions.md) disabled NServiceBus bus will perform best-effort error message forwarding.
```
MarcinHoppe: The standalone "bus" word seems redundant.

MarcinHoppe: I think we should explain what best effort means, i.e. indicate that if moving to error queue fails, the message will be lost.

MarcinHoppe:
Looks good to me.

nservicebus/errors/automatic-retries.md
```diff
@@ -31,7 +31,7 @@ Note: The configured value describes the minimum number of times a message will
 
 ### Transport transaction requirements
 
-The FLR mechanism is implemented by rolling back the [transport transaction](/nservicebus/messaging/transactions.md). This returns the message to the input queue and the endpoint processes it again immediately. Therefore FLR cannot be used when transport transactions are disabled. In such cases aborting the receive operation results in message loss.
+The FLR mechanism is implemented by rolling back the [transport transaction](/nservicebus/messaging/transactions.md). This returns the message to the input queue and the endpoint processes it again immediately. Aborting the receive operation when transactions are turned off would results in message loss. Therefore FLR cannot be used when transport transactions are disabled.
```
weralabaj: > borting the receive operation when transactions are turned off would results in message loss.

would result in a message loss

weralabaj: Meaning that even if they're configured by the user they won't be used or that if you try to enable them then exception is thrown? Might be good to clarify what happens under the hoods.

nservicebus/errors/automatic-retries.md
```diff
@@ -70,6 +70,10 @@ NOTE: Retrying messages for extended periods of time would hide failures from op
 
 SLR can be configured in several ways.
 
+### Transport transaction requirements
+
+The SLR mechanism is implemented by rolling back the [transport transaction](/nservicebus/messaging/transactions.md) and scheduling message for [delayed-delivery](/nservicebus/messaging/delayed-delivery.md) next time it is picked from the input queue. Aborting the receive operation when transactions are turned off would results in message loss. Therefore SLR cannot be used when transport transactions are disabled.
```
weralabaj: > next time it is picked from the input queue

That seems unnecessary to me. Skip? Or say "the next time...".

weralabaj: > would results in message loss

as above

weralabaj: > Therefore SLR cannot be used when transport transactions are disabled.

As above, what will the user see happenning?

nservicebus/errors/index.md
```diff
@@ -18,8 +18,9 @@ NServiceBus has exception catching and handling logic of its own which surrounds
 
 When a message fails NServiceBus [automatically retries](/nservicebus/errors/automatic-retries.md) the message. On repeated failure NServiceBus forwards that message to a designated error queue.
 
-Error queue can be configured in several ways.
+NOTE: When running with [transport transactions](/nservicebus/messaging/transactions.md) disabled NServiceBus will perform best-effort error message forwarding i.e. if moving to error queue fails, the message will be lost.
```
weralabaj: Would it make sense to make linking more precise, i.e.:

http://docs.particular.net/nservicebus/messaging/transactions.md#transactions-unreliable-transactions-disabled on "[transport transactions disabled]"

weralabaj: > moving to error queue

moving to the error queue

> forwarding i.e.

forwarding, i.e.

weralabaj:
@tmasternak see comments

nservicebus/errors/automatic-retries.md
tmasternak: > That seems unnecessary to me. Skip? Or say "the next time...".

I would like to leave that because it shows the user that for SLR to work the message has to be delivered to the endpoint second time.

tmasternak:
@weralabaj I've added fixes. Please see my [comment](https://github.com/Particular/docs.particular.net/pull/1365#discussion_r57951930).

When it comes to behavior when user turns off transactions I think that it is not consistent across different versions. For V6 the FLR will be silently disabled and SLR will throw unless user explicitly disables it. For V5 I think that both will be silently disabled. I have no idea how that behaves in V4 or V3. @ramonsmits @indualagarsamy do you know what will be the behavior for those versions?

ramonsmits:
FLR/SLR has always been silently disabled.

nservicebus/errors/automatic-retries.md
weralabaj: IMO it's saying something else. The message is scheduled for delayed delivery right away, which also implies that it will be delivered again (I think that's clear from "delayed delivery"). 
"scheduling message for delayed-delivery the next time it is picked from the input queue" sounds like the scheduling will happen when the message is picked for the second time, but why would it be picked again if it wasn't scheduled for delayed delivery?

If my understanding is accurate (first paragraph) then I'd change the text. I'll do a full review of both articles anyway, because they have other issues which predate your changes, so just let me know if I misunderstood something.

nservicebus/errors/automatic-retries.md
```diff
@@ -70,6 +70,10 @@ NOTE: Retrying messages for extended periods of time would hide failures from op
 
 SLR can be configured in several ways.
 
+### Transport transaction requirements
+
+The SLR mechanism is implemented by rolling back the [transport transaction](/nservicebus/messaging/transactions.md) and scheduling message for [delayed-delivery](/nservicebus/messaging/delayed-delivery.md) the next time it is picked from the input queue. Aborting the receive operation when transactions are turned off would results in a message loss. Therefore SLR cannot be used when transport transactions are disabled.
```
ramonsmits: Shouldn't this state that SLR has a dependency on FLR? FLR does the counting semantics correct?

Saying SLR cannot work because the transaction is rolled back is more an implementation issue as we are forwarding failed messages to the error queue in case of exceptions when there are no transactions.

If we are doing a best effort in moving it to the error queue then why can't we do a best effort in processing it for SLR?

It might help here to list the steps to understand the behavior.

nservicebus/errors/index.md
```diff
@@ -18,8 +18,9 @@ NServiceBus has exception catching and handling logic of its own which surrounds
 
 When a message fails NServiceBus [automatically retries](/nservicebus/errors/automatic-retries.md) the message. On repeated failure NServiceBus forwards that message to a designated error queue.
 
-Error queue can be configured in several ways.
+NOTE: When running with [transport transactions](/nservicebus/messaging/transactions.md#transactions-unreliable-transactions-disabled) disabled NServiceBus will perform best-effort error message forwarding, i.e. if moving to the error queue fails, the message will be lost.
```
ramonsmits: Do we log an ERROR or FATAL is this would happen and which message is logged when we fail delivery to the error queue?

nservicebus/errors/automatic-retries.md
weralabaj: > FLR does the counting semantics correct?

I don't understand, what do you mean?

> It might help here to list the steps to understand the behavior.

You mean first FLR, then SLR, then EQ? Can you give an example of what you had in mind? I'm not sure if going very detailed would be useful for the user.

weralabaj: > If we are doing a best effort in moving it to the error queue then why can't we do a best effort in processing it for SLR?

I think this is an issue that should be raised in NSB repo, not addressed here.

nservicebus/errors/index.md
weralabaj: Is that what you had in mind http://docs.particular.net/nservicebus/errors/automatic-retries#retry-logging?

weralabaj:
>  For V6 the FLR will be silently disabled and SLR will throw unless user explicitly disables it. For V5 I think that both will be silently disabled.

Seems like something that should be explicitly stated in this NOTE/WARNING part.

> SLR will throw

At runtime?

nservicebus/errors/index.md
weralabaj: @ramonsmits If we want to extend the logging example I think that should be a separate issue.

PR #1366, opened 2016-03-30
Add example of initializing test with lambda

stimms:
The assemblies to scan must include the testing assembly or an exception will be thrown. http://stackoverflow.com/questions/28722542/nservicebus-test-initialize-given-key-was-not-present


SimonCropp:
@stimms thanks

PR #1367, opened 2016-03-30
Marking warning as such

SeanFeldman:
It is difficult to see the warning the way it was originally set up. Needs to be more outstanding.

@Particular/servicecontrol-maintainers please review

### DIFF
--- a/servicecontrol/installation.md
+++ b/servicecontrol/installation.md
@@ -45,7 +45,9 @@ The installation of the NServiceBus Performance counters is optional for Service
 
 #### Using the ServiceControl Management Utility to upgrade ServiceControl instances.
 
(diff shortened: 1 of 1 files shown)


boblangley:
:+1: Adding the `WARN`. Since it will now draw the eye apart from the preceding sentence we should end it with:

> The ability to add multiple instances is primarily intended to assist development or test environments and is not recommended for use in production environments.

indualagarsamy:
@SeanFeldman - For future reference, it's WARNING and not WARN. See https://github.com/particular/docs.particular.net#alerts

SeanFeldman:
Thanks @indualagarsamy I could swear it used to be `WARN`.
/cc @colin-higgins

PR #1369, opened 2016-03-30
[WIP] Guidance on configuring cleanup interval

ramonsmits:
Guidance on configuring outbox in regards to the cleanup interval and deduplication time window.

Partially based on response by @janovesk  here https://groups.google.com/forum/#!searchin/particularsoftware/TimeToKeepDeduplicationData/particularsoftware/RRig2_y6S48/A7ctrIp0AwAJ

### DIFF
--- a/nservicebus/outbox/index.md
+++ b/nservicebus/outbox/index.md
@@ -101,23 +101,23 @@ If your endpoint processes 1,000 message per second the default value would mean
 
 There is no exact calculation that yields seven days as a magic number. The things you need to consider is: how fast will my organisation notice and fix any failure that can lead to duplication of messages?
 
-Server failure*
+#### Server failure
 
 If a sender goes down between the two operations mentioned because of HW issues on the machine, how fast will you react to it? Can you spin up the, hopefully, virtualized machine, on other HW before TimeToKeepDeduplicationData runs out? Can you do it in minutes? In hours? In days? Most likely not minutes, maybe hours depending on your ops capabilities, but most likely in seven days.
 
-Network failure
+#### Network failure
 
 How about if a funky network switch drops packages that result in delayed acks and timeout for the SendMessageOverTransport operation? Detect and fix in minutes? Hours? Days?
 
-OPS Detect, asses and fix period
+#### OPS Detect, asses and fix period
 
 Minutes or hours is usually too short a period to detect, assess and fix a scenario as mentioned by your ops team. In seven days you would likely have detected that something is wrong and might be able to resolve it. If your organisations ops team has their act together, and / or you know that receiving duplicate messages will not produce costly side effects, you have every possibility to reduce the deduplication window from 7 days accordingly.
 
-Increase when there are errors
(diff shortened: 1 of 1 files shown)


mikeminutillo:
As the content presented here is based on a mailing list post, it has quite a bit of personal voice and is very conversational in nature. We have been working actively in doco to remove these traits from our content. 

For example:

Instead of 
> If your endpoint processes a high volume of message per second it might be needed to change this interval to run more frequently.
>
> If your endpoint processes 1,000 message per second the default value would mean that every minute 60,000 outbox records or documents will be deleted. In that case it probably makes more sense to set the interval to every second to delete 1,000 records per second or even more often like every 100 milli seconds to delete on average 100 records at each interval.

I would say
> An endpoint that processes a high volume of messages per second requires a shorter interval to clean up more frequently.
>
> For example, an endpoint that processes 1,000 messages per second generates 60,000 outbox records per minute. In that case, running the cleanup task every second would result in deleting 1,000 records each interval. Running the cleanup task every 100 milliseconds reduces this number to 100 records each run.

 @ramonsmits are you happy to try and remove words like "you" and "your" from this content?

ramonsmits:
Thanks @mikeminutillo , I've rewritten it. Is this better?

weralabaj:
> For example, an endpoint that processes 1,000 messages per second generates 60,000 outbox records per minute. In that case, running the cleanup task every second would result in deleting 1,000 records at each 1 second interval. Running the cleanup task every 100 milliseconds reduces this number to 100 records per each run.

Is 1000 msgs deleted at once too big number? Mentioning it in the example makes it seem so.

Snippet showing how to change this setting would make sense here.

> In such case the deduplication time window could be set to expire immediately.

That should be followed by a snippet that shows how to do that. But do we really want to encourage people to do that?

> Outbox records will only be removed when all the transport operations succeeded

How about "when all the operations succeeded" or "when message was processed successfully".

> Server failure/Network failure/OPS Detect, asses and fix period

I don't understand what you're getting at in those sections. Is there some config setting to control that? or is it more like "Things you should consider before deployment"?

What is "OPS"?

> there is always the possibility to *increase* the deduplication window on the receivers to create more time to come up with a fix.

Again, show how. That seems like a "Troubleshooting/Recovery" section?

> A large set of deduplication records might need to be deleted when reverting the deduplication time window to the old value.

Again, that's "Troubleshooting" not standard configuration topic.

weralabaj:
I think we need to move some of this content to a separate section or article, something like what we've done for ServiceControl here http://docs.particular.net/servicecontrol/servicecontrol-in-practice.

So one part could be configuration - what settings are available, what they mean, how to adjust them (snippets) and how to find good values. Then troubleshooting/recovery guide which covers the rest, but we'd need some intro to provide some link (like in the ServiceControl article), right now it's a bit confusing.

SimonCropp:
@ramonsmits r u still working on this? Should it be handed to someone else? Or should it be abandoned?

Issue #1280, opened 2016-03-07
No Doco for this obsolete message 'Address' is obsolete: 'Use the string based overloads.'	


petersgiles:
I was upgrading NServiceBus.Distributor.MSMQ and got this message `Address' is obsolete: 'Use the string based overloads. Will be removed in version 7.0.0.'	`

I looked through the [upgrade guide ](http://docs.particular.net/nservicebus/upgrades/5to6 ) and wasn't able to find examples of what to do.

I looked at the NServiceBus v6 code no hints there

``` C#
    [ObsoleteEx(
        Message = "Use the string based overloads",
        RemoveInVersion = "7.0",
        TreatAsErrorFromVersion = "6.0")]
    public class Address
    {
    }
```

How do I `Use the string based overloads`?

Here is an example of it's use in NServiceBus.Distributor.MSMQ. It seems to be used to Parse and carry address info around. 

``` C#
 public static Address GetMasterNodeAddress(ReadOnlySettings settings)
        {
            var unicastBusConfig = settings.GetConfigSection<UnicastBusConfig>();

            //allow users to override data address in config
            if (!string.IsNullOrWhiteSpace(unicastBusConfig?.DistributorDataAddress))
            {
                return Address.Parse(unicastBusConfig.DistributorDataAddress);
            }

            var masterNode = GetMasterNode(settings);

            if (string.IsNullOrWhiteSpace(masterNode))
            {
                return Address.Parse(settings.EndpointName());
            }

            ValidateHostName(masterNode);

            return new Address(settings.EndpointName(), masterNode);
        }
```

Should there there be an example of what to do in the 5to6 upgrade guide? The code example here is using it to parse a config setting how is config setting parsing done now?

DavidBoike:
@Particular/nservicebus

DavidBoike:
Grr @Particular/nservicebus-maintainers

timbussmann:
I think you just need to change `Address` o `string` here? What part of the posted code is used by the user?

petersgiles:
@timbussmann I don't have access to customer code so I can't tell you anything other than this is part of our public API and therefore customers are likely to be using this. 

It seems to me that `Address.Parse(unicastBusConfig.DistributorDataAddress)` is a richer API than `string.Parse` otherwise why would we write it and `Address(settings.EndpointName(), masterNode)` is obviously doing something special.

[...]

PR #1371, opened 2016-03-31
Added info about deprecating Address

SzymonPobiega:


SzymonPobiega:
Solves #1280

mikeminutillo:
This PR moves a bunch of 3rd level headings under this one. Was that intended?

nservicebus/upgrades/5to6.md
```diff
@@ -353,6 +353,11 @@ A new [Message Routing Feature](/nservicebus/messaging/routing.md) has been adde
 
 Previous versions of NServiceBus used `TransportMessage` as a generic container both for outgoing and incoming messages. For a better separation of concerns that class has been split into `IncomingMessage` and `OutgoingMessage`. So all code paths related to outgoing messages will use `OutgoingMessage` and all code paths related to incoming messages will use `IncomingMessages`. The class `TransportMessage` has been deprecated entirely. Here are a few common scenarios related to `TransportMessage` and how they can be addressed with either `IncomingMessage` or `OutgoingMessage`.
 
+## Deprecated Address
+
+Previous versions of NServiceBus forced usage of `Address` class as a representation of an address in all transports despite the fact that the form *queue@host* did not match most of the transports. In Version 6 the transport address is represented as an opaque `string` and NServiceBus core no longer makes any assumptions on its form.
```
mikeminutillo: Not sure about the use of "Previous versions of NServiceBus...". If this is the 5 to 6 upgrade guide I would expect this to say "Version 5...".

The current wording is consistent with the rest of the document so I'll raise another issue to discuss this.

mikeminutillo: Rather than "forced usage of `Address` class" can we just say 

> Version 5 of NServiceBus represents addresses with an `Address` class. The `Address` class maintains addresses in the *queue@host* format. This format was originally developed for the MSMQ transport but does not meet the needs of other transports. In Version 6, addresses are represented as opaque strings.

SzymonPobiega:
@mikeminutillo I've applied your suggestion about rephrasing this and moved the address section before the transport message so the 3rd level headings are back to their original place.

Issue #1184, opened 2016-02-16
Clarify MSMQ transport transaction options


MarcinHoppe:
MSMQ transport configuration doco does not mention transport transaction modes. These need to be reconciled with description on the [Transoport transactions](http://docs.particular.net/nservicebus/messaging/transactions) doco page.

/cc @weralabaj @indualagarsamy

weralabaj:
would it be possible to get more details, similar to what Yves and Sean said on ASB https://github.com/Particular/docs.particular.net/issues/1187#issuecomment-187606884?

weralabaj:
@andreasohlund in MSMQ do we differentiate between ReceiveOnly and SendAtomicWithReceive?

andreasohlund:
No, but we should => https://github.com/Particular/NServiceBus/issues/3562

weralabaj:
@andreasohlund Given that it was decided to be low-impact should I document as it is now?

andreasohlund:
Just doco that ReceiveOnly works just like SendAtomicWithReceive and make a note that its a "known issue"?

Issue #1185, opened 2016-02-16
Clarify RabbitMQ transport transaction options


MarcinHoppe:
RabbitMQ transport configuration doco does not mention anything about [transport transaction modes](http://docs.particular.net/nservicebus/messaging/transactions). These need to be reconciled.

/cc @weralabaj @indualagarsamy

indualagarsamy:
@bording - Can you please comment on this issue on what the modes are and what they mean, so we can capture them in doco? Thanks.

bording:
The highest transaction mode that RabbitMQ supports is Receive Only.

weralabaj:
@bording would it be possible to get more details, similar to what Yves and Sean said on ASB https://github.com/Particular/docs.particular.net/issues/1187#issuecomment-187606884?

bording:
After making some changes in the v5 version of the transport to backport a fix, I want to review the v6 version to make sure everything is correct. Let me do that first, and then I should be able to give you some more details.

indualagarsamy:
@bording - after all the conversations we've had regarding transaction modes, are we in a good position to update this now?

bording:
I'm still working through some changes to the core v6 version of the transport that came out of those discussions, but I'm almost done with them. Once that's done then I should be able to go into more detail.

Looking over this issue, is this scoped to just the core v6 version of the transport, or do you need details for v5 as well?

weralabaj:
@bording for now it's V6 specific and should also cover differences between V5 and V6 (what to expect after upgrade), we can extend it later

weralabaj:
ping @bording, do we have info we need now?

bording:
#1395 is merged, so this is done.

PR #1372, opened 2016-03-31
Clarify MSMQ transport transaction options

weralabaj:
Resolves https://github.com/Particular/docs.particular.net/issues/1184.

### DIFF
--- a/nservicebus/msmq/transportconfig.md
+++ b/nservicebus/msmq/transportconfig.md
@@ -80,30 +80,3 @@ The native transaction for receiving messages is shared with sending operations.
 ### Unreliable (Transactions Disabled)
 
 In this mode when message is received from an input queue it's immediately removed from it. If processing fails the message is lost, because the operation cannot be rolled back. Also any other operation performed when processing the message is executed outside of the transaction, it can't be rolled back. That might lead to undesired side effects.
-
-
-### Controlling transaction scope options
-
-The following options can be configured when the MSMQ transport is working in the transaction scope mode.
-
-
-### Isolation level
-
-NServiceBus will by default use the `ReadCommitted` [isolation level](https://msdn.microsoft.com/en-us/library/system.transactions.isolationlevel).
-
-NOTE: Version 3 and below used the default isolation level of .Net which is `Serializable`.
-
-Change the isolation level using
-
-snippet:MsmqTransactionScopeIsolationLevel
-
-
-### Transaction timeout
-
-NServiceBus will use the [default transaction timeout](https://msdn.microsoft.com/en-us/library/system.transactions.transactionmanager.defaulttimeout) of the machine the endpoint is running on.
-
-Change the transaction timeout using
-
-snippet:MsmqTransactionScopeTimeout
-
-Or via .config file using a [example DefaultSettingsSection](https://msdn.microsoft.com/en-us/library/system.transactions.configuration.defaultsettingssection.aspx#Anchor_5).
--- a/nservicebus/msmq/transportconfig.md
+++ b/nservicebus/msmq/transportconfig.md
@@ -25,7 +25,7 @@ Because of historic reasons, the configuration for MSMQ transport has been coupl
 
 ### Queue permissions
 
-By default, queues are created with `Everyone` and `Anonymous Logon` permissions to allow messages to be sent and received without additional configuration. Whilst initially convenient, this unrestricted access may be a security concern. If required, the appropriate permissions can be set on a queue after its creation.
+By default, queues are created with `Everyone` and `Anonymous Logon` permissions to allow messages to be sent and received without additional configuration. If required, the appropriate permissions can be set on a queue after its creation.
 
 NOTE: From Version 6 if the above default permissions are set, a log message will be written during the transport startup, reminding that the queue is configured with default permissions. During development, if running with an attached debugger, this message will be logged as `INFO` level, otherwise `WARN`.
 
--- a/nservicebus/msmq/transportconfig.md
+++ b/nservicebus/msmq/transportconfig.md
@@ -48,8 +48,6 @@ From Version 4 onwards these settings are configured via a transport connection 
 snippet:MessageQueueConfiguration
 
 
-
-
 ### MSMQ Label
 
 WARNING: This feature was added in Version 6 and can be used to communicate with Version 5 (and higher) endpoints. However it should **not** be used when communicating to earlier versions (2, 3 or 4) since in those versions the MSMQ Label was used to communicate certain NServiceBus implementation details.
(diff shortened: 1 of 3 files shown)


weralabaj:
@Particular/docs-maintainers I've removed "Controlling transaction scope options" from MSMQ doco, but I'm having second thoughts. The content is duplicated in http://docs.particular.net/nservicebus/messaging/transactions#controlling-transaction-scope-options, but maybe the better approach would be to extract it to a separate file and create "include" for the current location, MSMQ and SQL Server transports. Thoughts?

weralabaj:
@mikeminutillo do you have time to have a look?

mikeminutillo:
Is there no `transactions-and-delivery-guarantees` section for Rabbit? Is there a separate issue to add that content?

weralabaj:
yep, https://github.com/Particular/docs.particular.net/issues/1185

nservicebus/msmq/transportconfig.md
```diff
@@ -25,7 +25,7 @@ Because of historic reasons, the configuration for MSMQ transport has been coupl
 
 ### Queue permissions
 
-By default, queues are created with `Everyone` and `Anonymous Logon` permissions to allow messages to be sent and received without additional configuration. Whilst initially convenient, this unrestricted access may be a security concern. If your organization requires it, you should set appropriate permissions on a queue after its creation.
+By default, queues are created with `Everyone` and `Anonymous Logon` permissions to allow messages to be sent and received without additional configuration. Whilst initially convenient, this unrestricted access may be a security concern. If required, the appropriate permissions can be set on a queue after its creation.
```
mikeminutillo: >  Whilst initially convenient, this unrestricted access may be a security concern.

Does this sentence add value? If the reader believes the previous sentence to be a security concern then the following sentence helps them fix it. If the reader did not think it is of concern, saying it might be doesn't really help them make a decision.

nservicebus/msmq/transportconfig.md
```diff
@@ -58,29 +58,25 @@ Often when debugging MSMQ using [native tools](viewing-message-content-in-msmq.m
 
 snippet:ApplyLabelToMessages
 
+## Transactions and delivery guarantees
 
-## Controlling transaction scope options
+MSMQ transport supports all [transaction handling modes](/nservicebus/messaging/transactions.md), i.e. Transaction scope, Receive only, Sends atomic with Receive and No transactions.
 
-The following options can be configured when the MSMQ transport is working in the transaction scope mode.
+Refer to [Transport Transactions](/nservicebus/messaging/transactions.md) for detailed explanation of the supported transaction handling modes and available configuration options.
```
mikeminutillo: These two sentences are very similar, even linking to the same content. I would say something like

> MSMQ supports the following transaction modes:
> * Transaction scope
> * Receive only
> * Sends atomic with Receive
> * No transactions
> 
> Refer to [Transport Transactions](/nservicebus/messaging/transactions.md) for a detailed explanation of the supported transaction handling modes and available configuration options.

The fact that this is all of the transaction modes is irrelevant to the reader.

nservicebus/msmq/transportconfig.md
weralabaj: Good point, removed

nservicebus/msmq/transportconfig.md
```diff
@@ -58,29 +58,25 @@ Often when debugging MSMQ using [native tools](viewing-message-content-in-msmq.m
...
 
-The following options can be configured when the MSMQ transport is working in the transaction scope mode.
+Refer to [Transport Transactions](/nservicebus/messaging/transactions.md) for detailed explanation of the supported transaction handling modes and available configuration options. 
 
 
-### Isolation level
+### Transaction scope
 
-NServiceBus will by default use the `ReadCommitted` [isolation level](https://msdn.microsoft.com/en-us/library/system.transactions.isolationlevel).
+In this mode the ambient transaction is started before receiving the message. The transaction encompasses all stages of processing including user data access and saga data access. If all the logical data stores (transport, user data, saga data) use the same physical store there is no escalation to Distributed Transaction Coordinator (DTC).
```
mikeminutillo: Does this content add anything MSMQ specific or is it a repeat of the content from the Transport Transactions page?

weralabaj:
@mikeminutillo Maybe I should do the full review first. I've only fixed issues that were preventing rendering.

nservicebus/msmq/transportconfig.md
```diff
@@ -58,29 +58,25 @@ Often when debugging MSMQ using [native tools](viewing-message-content-in-msmq.m
...
 
-NServiceBus will use the [default transaction timeout](https://msdn.microsoft.com/en-us/library/system.transactions.transactionmanager.defaulttimeout) of the machine the endpoint is running on.
+### Unreliable (Transactions Disabled)
 
-Change the transaction timeout using
-
-snippet:MsmqTransactionScopeTimeout
-
-Or via .config file using a [example DefaultSettingsSection](https://msdn.microsoft.com/en-us/library/system.transactions.configuration.defaultsettingssection.aspx#Anchor_5).
+In this mode when message is received from an input queue it's immediately removed from it. If processing fails the message is lost, because the operation cannot be rolled back. Also any other operation performed when processing the message is executed outside of the transaction, it can't be rolled back. That might lead to undesired side effects.
```
mikeminutillo: > In this mode when message is received from an input queue it's immediately removed from it.

"it's removed from it." is awkward. How about:

> In this mode when a message is received it is immediately removed from the input queue.

mikeminutillo: > If processing fails the message is lost, because the operation cannot be rolled back.

I'm notoriously terrible at commas. Is it right to have one here?

mikeminutillo: > Also any other operation performed when processing the message is executed outside of the transaction, it can't be rolled back. That might lead to undesired side effects.

"Also" feels out of place. and the sentence is talking about "the transaction" in the section about having No Transactions which is confusing.

How about:

> Any other operation that is performed when processing the message is executed without a transaction and cannot be rolled back. This can lead to undesired side effects when message processing fails part way through.

mikeminutillo:
All of the content about isolation levels, timeouts, etc. has been lost. Is that intentional? Is it covered elsewhere?

Review complete.

weralabaj:
> All of the content about isolation levels, timeouts, etc. has been lost. Is that intentional? Is it covered elsewhere?

Yes, it's in the main transactions doco, so either we can link to the main doco (which I did) or create "include" with that fragment which will be included in the main doco and here. If you think the latter would be better I can change that.

nservicebus/sqlserver/design.md
```diff
@@ -90,8 +90,6 @@ Refer to [Transport Transactions](/nservicebus/messaging/transactions.md) for de
 
 In this mode the ambient transaction is started before receiving the message. The transaction encompasses all stages of processing including user data access and saga data access. If all the logical data stores (transport, user data, saga data) use the same physical store there is no escalation to Distributed Transaction Coordinator (DTC).
 
-snippet:OutboxSqlServerConnectionStrings
```
SimonCropp: why?

weralabaj: I think it was a left-over, didn't connect with the text.

nservicebus/msmq/transportconfig.md
weralabaj: Nothing specific, but for SQL it's also covered in the transport specific doco. Since it's only for those two I think it makes sense to leave it in those places, instead of putting in the shared article.

weralabaj:
Comments addressed in the commit

PR #1390, opened 2016-04-06
Removing internal details about Start and Stop of IWantToRunWhenEndpointStartsAndStops

SeanFeldman:
Connects to https://github.com/Particular/NServiceBus.Host/issues/68

[Startup behaviour](http://docs.particular.net/nservicebus/hosting/nservicebus-host/#startup-behavior) described in http://docs.particular.net/nservicebus/hosting/nservicebus-host/ - descriptions for `Start` and `Stop` go into implementation details:

>  Each call to Start is called in the OnStart method of a FeatureStartupTask.

and 

> Each call to Stop is called in the OnStop method of a FeatureStartupTask.

From a user perspective, this could be confusing internal implementation details .

/cc @weralabaj @bording @colin-higgins


indualagarsamy:
@SeanFeldman - I agree with your changes.

Issue #1176, opened 2016-02-15
List transports that support a given transaction mode


MarcinHoppe:
We describe different [transport transaction modes](http://docs.particular.net/nservicebus/messaging/transactions#transactions) but do not tell the reader to which of the transport can a given option be applied.

On the other side, we need to clearly document how transport transaction modes map to transport-specific behaviors:
- [x] SQL: #1175 
- [x] MSMQ: #1184 
- [x] RabbitMQ: #1185 
- [x] ASQ: #1186 
- [x] ASB: #1187 

/cc @weralabaj @indualagarsamy

andreasohlund:
:+1: 

So some kind of "Transaction support" matrix on http://docs.particular.net/nservicebus/transports/ ?

MarcinHoppe:
Looks like a good place to put and link from http://docs.particular.net/nservicebus/messaging/transactions.

SimonCropp:
@andreasohlund no not a matrix. Just different pages that explain it for each transport

MarcinHoppe:
@SimonCropp That's one thing. Another thing is that we don't have an overview of what are transactional capabilities of different transports.

indualagarsamy:
@MarcinHoppe - I added task list on the issue description.

MarcinHoppe:
@indualagarsamy Thanks.

weralabaj:
I started work on it, I suggest limiting it to v 6 for now only if possible and raising a new issue to cover others later.

SimonCropp:
@weralabaj no objection to limiting to v6

PR #1395, opened 2016-04-07
Add RabbitMQ transaction mode details

bording:
Connects to #1185 

Once this is merged, it will also wrap up #1176.

### DIFF
--- a/nservicebus/rabbitmq/configuration-api.md
+++ b/nservicebus/rabbitmq/configuration-api.md
@@ -111,21 +111,21 @@ snippet:rabbitmq-config-useroutingtopology
 
 ## Transactions and delivery guarantees
 
-### Version 6 and above
+### Versions 4 and above
 
 The RabbitMQ transport supports the following [Transport Transaction Modes](/nservicebus/messaging/transactions.md):
 
-* Receive Only
-* No transactions
+* `ReceiveOnly`
+* `None`
 
-### Receive Only
+### ReceiveOnly
 
-When running in Receive Only mode, the RabbitMQ transport consumes messages from the broker in manual acknowledgment mode. After a message is successfully processed, it is acknowledged via the AMQP [basic.ack](http://www.rabbitmq.com/amqp-0-9-1-quickref.html#basic.ack) method, which lets the broker know that the message can be removed from the queue. If a message is not successfully processed and needs to be retried, it is requeued via the AMQP [basic.reject](http://www.rabbitmq.com/amqp-0-9-1-quickref.html#basic.reject) method.
+When running in `ReceiveOnly` mode, the RabbitMQ transport consumes messages from the broker in manual acknowledgment mode. After a message is successfully processed, it is acknowledged via the AMQP [basic.ack](http://www.rabbitmq.com/amqp-0-9-1-quickref.html#basic.ack) method, which lets the broker know that the message can be removed from the queue. If a message is not successfully processed and needs to be retried, it is requeued via the AMQP [basic.reject](http://www.rabbitmq.com/amqp-0-9-1-quickref.html#basic.reject) method.
 
-If the connection to the broker is lost for any reason before a message can be acknowledged, the message will automatically be requeued by the broker, letting it be retried by the endpoint.
+WARNING: If the connection to the broker is lost for any reason before a message can be acknowledged, even if the message was successfully processed, the message will automatically be requeued by the broker. This will result in the endpoint processing the same message multiple times.
 
-### No transactions
+### None
 
-When running without transactions, the RabbitMQ transport consumes messages from the broker in manual acknowledgment mode. Regardless of whether a message is successfully processed or not, it is acknowledged via the AMQP [basic.ack](http://www.rabbitmq.com/amqp-0-9-1-quickref.html#basic.ack) method after the processing attempt. This means that a message will be attempted once, and moved to the error queue if it fails.
+When running in `None` mode, the RabbitMQ transport consumes messages from the broker in manual acknowledgment mode. Regardless of whether a message is successfully processed or not, it is acknowledged via the AMQP [basic.ack](http://www.rabbitmq.com/amqp-0-9-1-quickref.html#basic.ack) method after the processing attempt. This means that a message will be attempted once, and moved to the error queue if it fails.
 
-However, because manual acknowledgment mode is being used, if the connection to the broker is lost for any reason before a message can be acknowledged, the message will automatically be requeued by the broker. If this occurs, the message will be retried by the endpoint, despite the transaction mode setting.
+WARNING: Since manual acknowledgment mode is being used, if the connection to the broker is lost for any reason before a message can be acknowledged, the message will automatically be requeued by the broker. If this occurs, the message will be retried by the endpoint, despite the transaction mode setting.
--- a/nservicebus/rabbitmq/configuration-api.md
+++ b/nservicebus/rabbitmq/configuration-api.md
@@ -108,3 +108,24 @@ If the routing topologies mentioned above isn't flexible enough then take full c
  1. Register it with the transport calling `.UseRoutingTopology` as shown below
 
 snippet:rabbitmq-config-useroutingtopology
+
+## Transactions and delivery guarantees
+
+### Version 6 and above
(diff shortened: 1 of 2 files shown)


nservicebus/rabbitmq/configuration-api.md
```diff
@@ -107,4 +107,25 @@ If the routing topologies mentioned above isn't flexible enough then take full c
  1. Define the topology by creating a class implementing `IRoutingTopology`
  1. Register it with the transport calling `.UseRoutingTopology` as shown below
 
-snippet:rabbitmq-config-useroutingtopology
\ No newline at end of file
+snippet:rabbitmq-config-useroutingtopology
+
+## Transactions and delivery guarantees
+
+### Version 6 and above
```
indualagarsamy: Should be `Versions 6 and above`
See: https://github.com/Particular/docs.particular.net#version-language

indualagarsamy: Also, what about Versions 5 and below? Are the transactions and delivery guarantees different, documented?

bording: > Should be Versions 6 and above

I copied that from http://docs.particular.net/nservicebus/azure/azure-storage-queues-transport#transactions-and-delivery-guarantees so it looks like it's wrong on the azure pages as well.

nservicebus/rabbitmq/configuration-api.md
```diff
@@ -107,4 +107,25 @@ If the routing topologies mentioned above isn't flexible enough then take full c
...
+
+The RabbitMQ transport supports the following [Transport Transaction Modes](/nservicebus/messaging/transactions.md):
+
+* Receive Only
+* No transactions
+
+### Receive Only
+
+When running in Receive Only mode, the RabbitMQ transport consumes messages from the broker in manual acknowledgment mode. After a message is successfully processed, it is acknowledged via the AMQP [basic.ack](http://www.rabbitmq.com/amqp-0-9-1-quickref.html#basic.ack) method, which lets the broker know that the message can be removed from the queue. If a message is not successfully processed and needs to be retried, it is requeued via the AMQP [basic.reject](http://www.rabbitmq.com/amqp-0-9-1-quickref.html#basic.reject) method.
```
indualagarsamy: I think it should be formatted as `Receive Only` instead of just spelling out Receive Only

nservicebus/rabbitmq/configuration-api.md
bording: > Also, what about Versions 5 and below? Are the transactions and delivery guarantees different, documented?

#1185 and #1176 talk about adding this in relation to core v6, so that's all I did here.

bording: You mentioning versions, made me realize that I should have said "versions 4 and above" here since it should be talking about RabbitMQ transport versions here and not core, right?

nservicebus/rabbitmq/configuration-api.md
bording: I can do that, though while reviewing all the linked pages here: http://docs.particular.net/nservicebus/messaging/transactions its very inconsistent.

For example, everyone seems to be calling `TransportTransactionMode.None` something different.

nservicebus/rabbitmq/configuration-api.md
indualagarsamy: @bording - yes! It should be the transport version! Thx!

nservicebus/rabbitmq/configuration-api.md
indualagarsamy: yeah, the TransportTransactionMode enum will work better.

bording: What about the header and the list of options? How should that look?

nservicebus/rabbitmq/configuration-api.md
```diff
@@ -107,4 +107,25 @@ If the routing topologies mentioned above isn't flexible enough then take full c
...
+
+If the connection to the broker is lost for any reason before a message can be acknowledged, the message will automatically be requeued by the broker, letting it be retried by the endpoint.
```
indualagarsamy: Can we highlight the danger of this better? For example something like:

If the connection to the broker is lost for any reason before a message can be acknowledged even if the message was successfully processed, the message will still be automatically be requeued by the broker. This will result in the endpoint processing the same message multiple times.

nservicebus/rabbitmq/configuration-api.md
```diff
@@ -107,4 +107,25 @@ If the routing topologies mentioned above isn't flexible enough then take full c
...
+
+### No transactions
+
+When running without transactions, the RabbitMQ transport consumes messages from the broker in manual acknowledgment mode. Regardless of whether a message is successfully processed or not, it is acknowledged via the AMQP [basic.ack](http://www.rabbitmq.com/amqp-0-9-1-quickref.html#basic.ack) method after the processing attempt. This means that a message will be attempted once, and moved to the error queue if it fails.
```
indualagarsamy: Can you start with, When running in the `TransportTransactionMode.None` mode ..

nservicebus/rabbitmq/configuration-api.md
bording: Hmm, I wasn't really thinking of this as being dangerous for this mode, (or needing an explicit call out) since we already talk about that here: http://docs.particular.net/nservicebus/messaging/transactions#transactions-transport-transaction-receive-only

I was under the impression that these transport-specific sections were supposed to call out things that were unique to how the transport handled it.

nservicebus/rabbitmq/configuration-api.md
bording: Including the full enum name there feels a bit redundant since it has "mode" in the name.

What about "When running in `None` mode" ?

indualagarsamy: Can you clarify this sentence better to highlight the difference better?

Regardless of whether a message is successfully processed or not, it is acknowledged via the AMQP [basic.ack](http://www.rabbitmq.com/amqp-0-9-1-quickref.html#basic.ack) method after the processing attempt.

When looking at the ReceiveOnly mode, it is also manually ack-ed.  

Something like:
In this mode, the acknowledgement to the broker is done right away even before the message is processed by the endpoint's message handler. In this scenario, if the processing of the message fails, then the message will be forwarded to the error queue.

nservicebus/rabbitmq/configuration-api.md
```diff
@@ -107,4 +107,25 @@ If the routing topologies mentioned above isn't flexible enough then take full c
...
+
+However, because manual acknowledgment mode is being used, if the connection to the broker is lost for any reason before a message can be acknowledged, the message will automatically be requeued by the broker. If this occurs, the message will be retried by the endpoint, despite the transaction mode setting.
```
indualagarsamy: add this as a WARNING?

WARNING: Since manual acknowledgment mode is being used, if the connection to the broker is lost for any reason before a message can be acknowledged, the message will automatically be requeued by the broker. If this occurs, the message will be retried by the endpoint, despite the transaction mode setting.

nservicebus/rabbitmq/configuration-api.md
bording: > In this mode, the acknowledgement to the broker is done right away even before the message is processed by the endpoint's message handler. In this scenario, if the processing of the message fails, then the message will be forwarded to the error queue.

That's not how it works though. It really is identical. The only difference is that since FLR is disabled in 'None` the you always ack the message.

indualagarsamy:
Do we need to add anything in the Rabbit MQ Upgrade Guide?

bording:
> Do we need to add anything in the Rabbit MQ Upgrade Guide?

I'd need to review if anything has actually changed here in regards to previous versions. There is also a separate issue regarding finishing the Rabbit MQ Upgrade Guide (all of the v4 docs actually) so I'm going to defer that to when we can work on that.

nservicebus/rabbitmq/configuration-api.md
bording: I like the WARNING suggestion below, so I think I'll use that here as well.

PR #1398, opened 2016-04-08
Replaces the static field with AsyncLocal and simplifies the connection mgmt

SzymonPobiega:
Connects to Particular/NServiceBus.Testing#29

There were some problems with this version of multi tenant DB. I've noticed that it does not go to the correct (tenant DB), at least for some of the DB calls. I've done two changes
 * The incoming pipeline behavior now only sets the correct connection string (does not open the connection). This is because the new connection management in V7 of NH does a better job than in V5
 * The mentioned behavior has been moved forward to the transport receive stage to execute before the outbox is involved
 * The static storage used to pass context between the behavior and the NH connection provider has been changed to use .NET 4.6 `AsyncLocal<T>`


SzymonPobiega:
@SimonCropp @HEskandari please review.

samples/outbox/multi-tenant/Version_6/Receiver/Receiver.csproj
```diff
@@ -9,7 +9,7 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>Receiver</RootNamespace>
     <AssemblyName>Receiver</AssemblyName>
-    <TargetFrameworkVersion>v4.5.2</TargetFrameworkVersion>
+    <TargetFrameworkVersion>v4.6</TargetFrameworkVersion>
```
SimonCropp: why 4.6?

samples/outbox/multi-tenant/Version_6/Sender/app.config
```diff
@@ -19,4 +19,12 @@
       </providers>
     </roleManager>
   </system.web>
+  <runtime>
+    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
+      <dependentAssembly>
+        <assemblyIdentity name="NServiceBus.Core" publicKeyToken="9fc386479f8a226c" culture="neutral" />
```
SimonCropp: why is this required?

SzymonPobiega: It isn't:( package manager added it silently...

samples/outbox/multi-tenant/Version_6/Receiver/Receiver.csproj
SzymonPobiega: `AsyncLocal<T>` is 4.6. I think it is fair to require this since this is a very specific sample. Previous .NET editions did not have a good solution for async local state according to @danielmarbach

samples/outbox/multi-tenant/Version_6/Sender/app.config
SimonCropp: yeah i have started doing many of my nuget updates using sublime text search and replace

samples/outbox/multi-tenant/Version_6/Receiver/Receiver.csproj
SimonCropp: ahh ok then. this is fine. i suspect the other samples will go 4.6 within the next few months anyway

samples/outbox/multi-tenant/Version_6/Receiver/ExtractTenantConnectionStringBehavior.cs
```diff
@@ -0,0 +1,48 @@
+﻿using System;
+using System.Configuration;
+using System.Data;
```
SimonCropp: redundant usings

PR #1427, opened 2016-04-19
lucidchart update

SeanFeldman:
Capture how generated images could be maintained.

### DIFF
--- a/README.md
+++ b/README.md
@@ -628,6 +628,12 @@ This will result in the image being re-sized with the following parameters
 It will also wrap the image in a clickable lightbox so the full image can be accessed.
 
 
+### Maintaining images
+
+When creating images, strive to keep sources to for updating and re-creating images later.
+For diagrams, [LucidChart](https://www.lucidchart.com) can be used. LucidChart allows export and import of Visio (VDX) formatted documents. Visio formatted document can be used to generate images and should be committed along with the images. To generate images from LucidChar (or a Visio document), exported image as PNG, using "Crop to content" option.
+
+
 ### Sequence diagrams
 
 Sequence diagram images are generated using https://bramp.github.io/js-sequence-diagrams/ online service. Keep the source text used to generate sequence image in the document as an HTML comment to allow future modifications in case images need to be re-generated.


weralabaj:
@SeanFeldman question - since that is public readme is there anything we could ask for when people outside the company contribute? (any source files, exports...) If not then it's ok, I don't think it'll happen very often.

Could we make an announcement and add in it info on how to get a license? Do we have a shared account for this?

SeanFeldman:
@weralabaj I've never seen anyone external contributing anything other than textual content. We can deal with that when it happens.

SimonCropp:
@SeanFeldman can we deep link into lucid chart? so we dont need to copy images over?

SimonCropp:
then if we deep link we could set this up http://support.blakeschool.org/entries/23015117-Automatic-Synchronization-from-LucidChart-To-Google-Drive

SeanFeldman:
Who needs images. ASCII FTW!
[EndpointOrientedTopology.png.txt](https://github.com/Particular/docs.particular.net/files/227152/EndpointOrientedTopology.png.txt)

SeanFeldman:
@SimonCropp not sure how deep linking would help us. The images would be still generated manually. Also, having content spread across 2 sources will complicate things. I might be wrong though :)

SimonCropp:
was thinking if  lucid chart has a deep link standard to an image the update process would be just edit on their web app. no export needed

SeanFeldman:
Unfortunately, images need to be exported manually. Go ASCII :P

weralabaj:
Ok so the last question , do we need to ping Sam to get more licenses https://github.com/Particular/Collaboration/issues/392? Or could we have one shared for docs?

SeanFeldman:
@weralabaj it depends on what @SimonCropp has discovered with deep linking. I saw him hitting lucidchart URLs directly to generate images w/o exporting images explicitly. If that works and and rendering is to what we need, we might rethink the approach (and PR).

weralabaj:
@SimonCropp given deep linking investigation is there anything you'd like to change here? or can I merge as is?

SimonCropp:
merge as is. deep linking can happen later as a second step

Issue #448, opened 2014-10-19
Document that we dont support FIPS compliance


SimonCropp:
https://github.com/Particular/NServiceBus/issues/1217

PR #1439, opened 2016-04-20
ASQ SHA1-MD5

Scooletz:
Connects to https://github.com/Particular/PlatformDevelopment/issues/723
Addresses SHA1 MD5 configuration.

### DIFF
--- a/Snippets/ASQ/ASQ_7/Usage.cs
+++ b/Snippets/ASQ/ASQ_7/Usage.cs
@@ -39,9 +39,9 @@ namespace ASQ_7
             #endregion
         }
 
-        void UseSha1InsteadOfMD5(EndpointConfiguration endpointConfiguration)
+        void UseSha1(EndpointConfiguration endpointConfiguration)
         {
-            #region AzureStorageQueueUseSha1InsteadOfMD5
+            #region AzureStorageQueueUseSha1
 
             var transport = endpointConfiguration.UseTransport<AzureStorageQueueTransport>();
             transport.UseSha1ForShortening();
--- a/Snippets/ASQ/ASQ_7/Usage.cs
+++ b/Snippets/ASQ/ASQ_7/Usage.cs
@@ -39,6 +39,16 @@ namespace ASQ_7
             #endregion
         }
 
+        void UseSha1InsteadOfMD5(EndpointConfiguration endpointConfiguration)
+        {
+            #region AzureStorageQueueUseSha1InsteadOfMD5
+
+            var transport = endpointConfiguration.UseTransport<AzureStorageQueueTransport>();
+            transport.UseSha1ForShortening();
+
+            #endregion
+        }
+
         #region AzureStorageQueueTransportWithAzureHost
 
         public class EndpointConfig : IConfigureThisEndpoint
--- a/nservicebus/azure-storage-queues/configuration.md
+++ b/nservicebus/azure-storage-queues/configuration.md
@@ -80,3 +80,10 @@ This feature can be enabled by specifying `.UseAccountNamesInsteadOfConnectionSt
 snippet:AzureStorageQueueUseAccountNamesInsteadOfConnectionStrings
 
 NOTE: This feature is not available in `NServiceBus.Azure.Transports.WindowsAzureStorageQueues` Versions 6 and below.
+
+## Hashing algorithms
+If the queue is longer than [63 characters](https://msdn.microsoft.com/en-us/library/azure/dd179349.aspx), the Azure Storage Queues Transport uses a hashing algorithm to rename it. The default algorithm is `MD5`. In order to use `SHA1` instead, use the following configuration:
+
+snippet:AzureStorageQueueUseSha1InsteadOfMD5
+
+NOTE: This feature is not available in `NServiceBus.Azure.Transports.WindowsAzureStorageQueues` Versions 6 and below.
--- a/nservicebus/azure-storage-queues/configuration.md
+++ b/nservicebus/azure-storage-queues/configuration.md
@@ -79,11 +79,11 @@ This feature can be enabled by specifying `.UseAccountNamesInsteadOfConnectionSt
 
 snippet:AzureStorageQueueUseAccountNamesInsteadOfConnectionStrings
 
-NOTE: This feature is not available in `NServiceBus.Azure.Transports.WindowsAzureStorageQueues` Versions 6 and below.
+NOTE: This feature is available in `NServiceBus.Azure.Transports.WindowsAzureStorageQueues` Versions 7 and above.
 
 ## Hashing algorithms
-If the queue is longer than [63 characters](https://msdn.microsoft.com/en-us/library/azure/dd179349.aspx), the Azure Storage Queues Transport uses a hashing algorithm to rename it. The default algorithm is `MD5`. In order to use `SHA1` instead, use the following configuration:
+If a queue name is longer than [63 characters](https://msdn.microsoft.com/en-us/library/azure/dd179349.aspx), the Azure Storage Queues Transport uses a hashing algorithm to rename it. The default algorithm is `MD5`. In order to use `SHA1` instead, use the following configuration:
 
-snippet:AzureStorageQueueUseSha1InsteadOfMD5
+snippet:AzureStorageQueueUseSha1
 
-NOTE: This feature is not available in `NServiceBus.Azure.Transports.WindowsAzureStorageQueues` Versions 6 and below.
+NOTE: This feature is available in `NServiceBus.Azure.Transports.WindowsAzureStorageQueues` Versions 7 and above.
--- a/nservicebus/azure-storage-queues/configuration.md
+++ b/nservicebus/azure-storage-queues/configuration.md
@@ -81,8 +81,10 @@ snippet:AzureStorageQueueUseAccountNamesInsteadOfConnectionStrings
 
 NOTE: This feature is available in `NServiceBus.Azure.Transports.WindowsAzureStorageQueues` Versions 7 and above.
 
+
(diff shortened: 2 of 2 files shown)


Scooletz:
Ping @Particular/docs-maintainers for review & merge (when build is green).

MarcinHoppe:
I'm a little bit uneasy about even mentioning FIPS in our doco. We've been fairly deliberate about making any sort of statements about FIPS. The core is not FIPS compliant at this point so fixing this in ASQ solves nothing for the customer.

I'm not saying we should not ditch MD5 in favor of SHA1, I'm just saying we don't mention we did this to support FIPS compliance.

Thoughts?

/cc @mauroservienti @DavidBoike

DavidBoike:
This is changing a bit. It's possible to get around FIPS in the core. Only place MD5 is used there is the host identifier that identifies the endpoint uniquely to ServiceControl, and you can specify that manually.

We even have a customer working with a US government agency trying to get clearance to run FIPS but with a minor modification so that the MD5 hashing algorithm is allowed, which theoretically allows all our products (including ServiceControl) to run.

I would turn it around. I don't like the name `UseSha1ForShortening` in the API. It's too in the weeds. I would replace this with `UseFipsHashingAlgorithms`. Then the doco should say something like while NServiceBus does not officially support FIPS mode, this setting will cause ASQ specifically to use algorithms that are compliant. Should also include a yellow warning box to be sure to test everything in a FIPS-enabled environment if that's your intent, and how changing that setting with existing data will break everything.

I could see us putting together a new doco article coming out of what we learn from this existing customer soon about FIPS and what you can do to use our products in a FIPS environment, even though we don't officially support it.

nservicebus/azure-storage-queues/configuration.md
```diff
@@ -60,3 +60,10 @@ This feature can be enabled by specifying `.UseAccountNamesInsteadOfConnectionSt
 snippet:AzureStorageQueueUseAccountNamesInsteadOfConnectionStrings
 
 NOTE: This feature is not available in `NServiceBus.Azure.Transports.WindowsAzureStorageQueues` Versions 6 and below.
+
+## FIPS compliance
```
SeanFeldman: There was a discussion where-there we support FIPS or not. @DavidBoike do you remember what was the decisions? I wonder because FIPS part could be a general `include` to be used by everything that does support the compliance.

DavidBoike:
This comment is our current policy on FIPS: https://github.com/Particular/NServiceBus/issues/1217#issuecomment-141098744

We should really think about adding that to docs. Pending the results of what the customer @jdrat2000 and @johncod are working with (omitting name because public repo) that comment could serve as the kernel of a new article on that.

nservicebus/azure-storage-queues/configuration.md
indualagarsamy: I agree with @MarcinHoppe. No where do we say, FIPS or not. Starting here seems incorrect.

indualagarsamy: @SeanFeldman - See this https://github.com/Particular/docs.particular.net/issues/448

SeanFeldman:
We should get better at update issue descriptions / adding summaries rather than going through lengthy discussions :) That' what I was after. If we support FIPS: have it as an include, otherwise, no need. Thanks @indualagarsamy @DavidBoike

Scooletz:
I'm sorry but I'm a bit lost. What's the conclusion?
As far as I understand, if we don't want to mention `FIPS`, how we could event provide documentation for this change in `ASQ`. Another question is, if this feature is really needed then.

SimonCropp:
technically this feature does not have anything to do with FIPS compliance. possible compliance may be a side effect. want me to have a go at documenting it without mentioning fips?

weralabaj:
If you don't want to mention FIPS then why not just mention - hey, you can use other hashing algorithms?

BTW since we have that kind of a discussion shouldn't that be WIP?

Scooletz:
@weralabaj I've marked it as `WIP`. If that's ok, I think rephrasing this section would work the best. It's easier to replace FIPS with `other algos` than to come up with an idea how to describe it without using this word. I'm proposing a new sketch for this right now.

Scooletz:
I've replaced not wanted words.

nservicebus/azure-storage-queues/configuration.md
```diff
@@ -60,3 +60,10 @@ This feature can be enabled by specifying `.UseAccountNamesInsteadOfConnectionSt
 snippet:AzureStorageQueueUseAccountNamesInsteadOfConnectionStrings
 
 NOTE: This feature is not available in `NServiceBus.Azure.Transports.WindowsAzureStorageQueues` Versions 6 and below.
+
+## Hashing algorithms
+By default The Azure Storage Queues Transport uses `MD5` hashing algorithm for renaming queues when an entity name is too large. If your solution requires a stronger algorithm, you can enable renaming queues with `SHA1` by specifying `UseSha1ForShortening` when configuring the `AzureStorageQueueTransport`:
```
weralabaj: Sounds a bit complex and uses another forbidden word (you) ;P, how about:

If the entity name is too long, the Azure Storage Queues Transport uses hashing algorithm for renaming queues. The default algorithm is `MD5`. In order to use `SHA1` algorithm instead, use the following configuration:

weralabaj: first you say "entity names" then "renaming queues" - it's not clear whether it applies to something other than queues, can you clarify?

weralabaj: > too large/too long

How do I know how long is too long? Where do I check?

Scooletz: I'll adjust this document. I'll add the length as well.

Scooletz:

Snippets/ASQ/ASQ_7/Usage.cs
```diff
@@ -39,6 +39,16 @@ void AccountNamesInsteadOfConnectionStrings(EndpointConfiguration endpointConfig
             #endregion
         }
 
+        void UseSha1InsteadOfMD5(EndpointConfiguration endpointConfiguration)
+        {
+            #region AzureStorageQueueUseSha1InsteadOfMD5
```
SimonCropp: minor but `InsteadOfMD5` is redundant

nservicebus/azure-storage-queues/configuration.md
```diff
@@ -79,4 +79,11 @@ This feature can be enabled by specifying `.UseAccountNamesInsteadOfConnectionSt
 
 snippet:AzureStorageQueueUseAccountNamesInsteadOfConnectionStrings
 
-NOTE: This feature is not available in `NServiceBus.Azure.Transports.WindowsAzureStorageQueues` Versions 6 and below.
\ No newline at end of file
+NOTE: This feature is not available in `NServiceBus.Azure.Transports.WindowsAzureStorageQueues` Versions 6 and below.
```
SimonCropp: i would go the positive with "Available in Versions 7 and above"

nservicebus/azure-storage-queues/configuration.md
```diff
@@ -79,4 +79,11 @@ This feature can be enabled by specifying `.UseAccountNamesInsteadOfConnectionSt
...
+
+## Hashing algorithms
+If the queue is longer than [63 characters](https://msdn.microsoft.com/en-us/library/azure/dd179349.aspx), the Azure Storage Queues Transport uses a hashing algorithm to rename it. The default algorithm is `MD5`. In order to use `SHA1` instead, use the following configuration:
```
SimonCropp: i know there is a lot of legacy. but the difficulty in debugging hashed queue names makes me think we should deprecate it and throw an exception at startup for long queue names. they can always work around by hashing themselves

nservicebus/azure-storage-queues/configuration.md
```diff
@@ -79,4 +79,11 @@ This feature can be enabled by specifying `.UseAccountNamesInsteadOfConnectionSt
...
+
+snippet:AzureStorageQueueUseSha1InsteadOfMD5
+
+NOTE: This feature is not available in `NServiceBus.Azure.Transports.WindowsAzureStorageQueues` Versions 6 and below.
```
SimonCropp: again invert this statement

nservicebus/azure-storage-queues/configuration.md
MarcinHoppe: If the queue **name** is longer than...

weralabaj: @SimonCropp raise an issue for it? it'll probably get lost here.

Snippets/ASQ/ASQ_7/Usage.cs
Scooletz: Will change.

nservicebus/azure-storage-queues/configuration.md
Scooletz: Will replace.

nservicebus/azure-storage-queues/configuration.md
Scooletz: @SimonCropp The hash is added as part of the queue name at the end to match the length criterion, so there's a non-zero chance that your queue will be discoverable by it. I'll register an issue for this and link to the comment of yours.
@MarcinHoppe I'll update.

Scooletz:

nservicebus/azure-storage-queues/configuration.md
SeanFeldman: @SimonCropp this was discussed when endpoint name became a mandatory thing. While core mandates the endpoint name, transport is the one that can check validity of the name. It feels as if this should be a dependency of the core on the transports to provide that validation.

PR #1464, opened 2016-04-29
Raven Func<ReadOnlySettings, IDocumentStore> options

DavidBoike:
Related to Particular/NServiceBus.RavenDB#214

Wasn't sure what the best way was of indicating that the new options were NSB.Raven 4.0 and above. Didn't think it merited a NOTE: and so I put a line in italics above the snippets. Of course the snippets themselves will only show the 4.0-pre version. Is that enough? Looking for guidance.

It's possible we might backport this to 3.x to better support DTC customizations, but for now, this is what is true.

### DIFF
--- a/Snippets/Raven/Raven_4/Configure.cs
+++ b/Snippets/Raven/Raven_4/Configure.cs
@@ -58,6 +58,40 @@
             persistence.UseDocumentStoreForSubscriptions(myDocumentStore);
             persistence.UseDocumentStoreForSagas(myDocumentStore);
             persistence.UseDocumentStoreForTimeouts(myDocumentStore);
+            persistence.UseDocumentStoreForGatewayDeduplication(myDocumentStore);
+
+            #endregion
+        }
+
+        void CreateSpecificDocumentStoreByFunc(EndpointConfiguration endpointConfiguration)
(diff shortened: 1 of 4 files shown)


SimonCropp:
What conventions would you need to use from the settings? Ie can u clarify an example use case better?

DavidBoike:
For the PR, or in the doco?

I could do a localhost URL and get the database name from settings.EndpointName(), would that work?

Later on in the Raven Beta POA is a task where I am planning a guide on setting DTC settings based on how deployments are done, including using settings.LocalAddress() for ResourceManagerId. Then I would revise the index page to link off to that guide under these sections.

SimonCropp:
Ok the endpoint name makes sense. Would that example be useful in the docs?

DavidBoike:
It might be useful to explain "Why would I ever want to use that option?" but it will bloat the code snippet by 5 lines or so.

SimonCropp:
The snippet is fine. I think add the use case to the text

SimonCropp:
merged manually here https://github.com/Particular/docs.particular.net/commit/00b235755475a99039a9f73502b0b1300307f20f

PR #1475, opened 2016-05-03
[WIP] Clarify MSDTC and TX usage in SC

ramonsmits:
# Open questions:

- [ ] Are sends participating in the transport receive transaction?
- [ ] What is committed first, the transport transaction or the RavenDB tx?

# References

- https://particularsoftware.slack.com/archives/control/p1462265867000268

### DIFF
--- a/servicecontrol/dtc-and-transactions.md
+++ b/servicecontrol/dtc-and-transactions.md
@@ -1,0 +1,21 @@
+---
+title: ServiceControl, MSDTC and transaction support
+summary: ServiceControl uses idempotent message processing and does not rely on MSDTC for exact-once processing but does use transport transactions is supported.
+tags:
+- ServiceControl
+- MSDTC
+---
+
+ServiceControl does not rely on the Microsoft Distributed Transaction Coordinator (MSDTC) for exact-one processing of its messages. Meaning that ServiceControl can be used in environments that do not rely on MSDTC
+
+## Receiving of messages
+
+It processes its messages idempotently when storing messages in its database. Duplicate messages will not result in any issues.
+
+## Sending of messages
+
+ServiceControl can be used to retry messages but also publishes several events. These messages are send as atomic as possible. We have potentially two transactions, the transport transaction and the persistence (RavenDB) transaction. 
+
+Tranports that *do not* support transactions will result in potentially sending the same retry more then once. This should not be any issue as this is expected behavior for this transport.
+
+Transports that *do* support transactions can potentially result in not delivering the retry as the database transaction will be committted before the transport transaction.
--- a/servicecontrol/dtc-and-transactions.md
+++ b/servicecontrol/dtc-and-transactions.md
@@ -1,21 +1,29 @@
 ---
 title: ServiceControl, MSDTC and transaction support
-summary: ServiceControl uses idempotent message processing and does not rely on MSDTC for exact-once processing but does use transport transactions is supported.
+summary: Summary of ServiceControl messaging reliability
 tags:
 - ServiceControl
 - MSDTC
 ---
 
-ServiceControl does not rely on the Microsoft Distributed Transaction Coordinator (MSDTC) for exact-one processing of its messages. Meaning that ServiceControl can be used in environments that do not rely on MSDTC
+ServiceControl does not rely on the Microsoft Distributed Transaction Coordinator (MSDTC) for exact-one processing of its messages. ServiceControl can be used in environments that do not rely on MSDTC.
+
 
 ## Receiving of messages
 
-It processes its messages idempotently when storing messages in its database. Duplicate messages will not result in any issues.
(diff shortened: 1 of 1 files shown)


servicecontrol/dtc-and-transactions.md
```diff
@@ -0,0 +1,21 @@
+---
+title: ServiceControl, MSDTC and transaction support
+summary: ServiceControl uses idempotent message processing and does not rely on MSDTC for exact-once processing but does use transport transactions is supported.
```
SimonCropp: "but does use transport transactions is supported." ?

servicecontrol/dtc-and-transactions.md
```diff
@@ -0,0 +1,21 @@
...
+
+ServiceControl does not rely on the Microsoft Distributed Transaction Coordinator (MSDTC) for exact-one processing of its messages. Meaning that ServiceControl can be used in environments that do not rely on MSDTC
+
+## Receiving of messages
+
+It processes its messages idempotently when storing messages in its database. Duplicate messages will not result in any issues.
+
+## Sending of messages
+
+ServiceControl can be used to retry messages but also publishes several events. These messages are send as atomic as possible. We have potentially two transactions, the transport transaction and the persistence (RavenDB) transaction.
```
SimonCropp: "These messages are send as atomic as possible" 

what exactly is " atomic as possible"?

SimonCropp: remove "we"

SimonCropp: why potentially two? dont we always have two?

ramonsmits: Sorry @SimonCropp, I didn't mark it as WIP

SimonCropp: ok will wait till u r done

SimonCropp:
@Particular/servicecontrol-maintainers can you please weigh in here as to if this should be in our public doco or not.

johnsimons:
IMO this doco should not be included in docs website, this doco is technical and belongs in the readme file in the repo

mikeminutillo:
I don't think we want to go into any great detail here. I think we need to make a statement along the lines of:

> ServiceControl does not rely on MSDTC and all internal message processing is Idempotent. When ServiceControl retries a message, it forwards it back to the processing endpoint from an internally owned staging queue. This enables ServiceControl to rely on the transactionality of the underlying transport to ensure that each failed message is only processed once.

mikeminutillo:
That could be a NOTE on an existing page

SimonCropp:
also this failed content verification

mikeminutillo:
I updated the content to match how SC behaves with respect to receives and retries. In reviewing the External Integration Events I believe these can be delivered more than once because we go from transport->raven->transport for every external integration event with no DTC. Here is a summary of my findings (from Slack):


Mike Minutillo [10:49 AM] 
@servicecontrol-maintainers: OK. So in trying to update the doco about why we don't use/need MSDTC I came across the Integration events. I think it is very possible that these could be sent multiple times.

Mike Minutillo [10:55 AM] 
The `EventMappingHandler` subscribes to these events, converts them into a `DispatchContext` and stores them in an `ExternalIntegrationDispatchRequest` in Raven. Those don't have a deterministic Id so if there's a problem committing the receive AFTER saving the event then we'll get duplicates.

[10:59] 
Similarly, the `EventDispatcher` will load up all of the `ExternalIntegrationBatchRequests` and do a `Bus.Publish` on them. Once the whole batch has been published we delete the dispatched messages. If we publish half of the batch and then encounter a problem those integration messages will still be in the DB to be published again.

mikeminutillo:
OK. Apparently this is correct. External Integration events can be delivered multiple times

mikeminutillo:
> OK. Apparently this is correct. External Integration events can be delivered multiple times.

This point should probably be added to the integration event doco as a note.

For the rest of it.... 

I'm talking this through with @johnsimons and he rightly points out that a customer does not need to know this level of detail. When they purchase SC they are getting a product. It should be a black box. When we say "you can retry a message" we are stating that we will deliver it back to the input queue of the endpoint that tried to process it before. Whether or not DTC is used to get the message there is irrelevant isn't it?

SimonCropp:
@mikeminutillo so based on your last comment, should this be closed an a new PR started?

mikeminutillo:
@SimonCropp I think so. I wanted to sync with @ramonsmits and make sure he was on the same page before doing that.

ramonsmits:
@mikeminutillo @SimonCropp I'm fine with a note saying that even though SC is not using MSDTC but that we make use of the transport its transaction to guarantee exact-once behavior for retries if the transport supports that and internally uses idempotent processing for it to not having to rely on MSDTC.

ramonsmits:
Just read the edits by @mikeminutillo but that info about atleast-once deliver of Integration Events must be documented too.

mikeminutillo:
Replaced by https://github.com/Particular/docs.particular.net/pull/1509

PR #1476, opened 2016-05-03
[WIP] NullReferenceException in TimeoutPersisterReceiver

ramonsmits:
This documents three knows reasons for getting this exception.


SimonCropp:
NullReferenceException should never be thrown. please raise an issue in the upstream

ramonsmits:
@SimonCropp This exception is happening due to various reasons in old versions. This page needs to be there to make this discoverable even when this behavior would be resolved.

I don't understand that you just close it.

SimonCropp:
because no matter the cause, throwing a NullReferenceException  is a bug. if the version is still supported it should be patched. we should not be documenting things that are clearly bugs and in our capacity to fix. if the bug is deemed "not worth fixing" because it is so unlikely to happen then nor is it worthy of documentation

PR #1477, opened 2016-05-03
upgrade samples to raven beta1

colin-higgins:
@DavidBoike @SimonCropp 

I figured this should go in even though a fix for the upgrade sample is pending. It didn't work before and doesn't work for v3.


DavidBoike:
FWIW, would also fail for a customer on any version if they switch to connect to their own database by connection string.

SimonCropp:
this is a corrupted PR

and we are already on raven beta001

2016-05-04 08_23_34-cortana

PR #1481, opened 2016-05-04
WIP: NullReferenceException in TimeoutPersisterReceiver

ramonsmits:
This documents three knows reasons for getting this exception.

This needs to be documented to make it discoverable for our customers. This issue was a WIP in #1476 but got closed before it was even finished.

### DIFF
--- a/exceptions/null-reference-exception-in-timeoutpersisterreceiver.md
+++ b/exceptions/null-reference-exception-in-timeoutpersisterreceiver.md
@@ -1,0 +1,32 @@
+---
+title: Null Reference Exception in TimeoutPersisterReceiver.Poll
+summary: Null Reference Exception in TimeoutPersisterReceiver.Poll
+tags:
+- Exception
+---
+
+# Null Reference Exception in TimeoutPersisterReceiver.Poll
+
+This exception can occur due to:
+
+* Incomplete assembly scanning
+* Version mismatch between NServiceBus and the selected persistence
+* Selected a persistence that does not provide a specific storage feature
+
+
+## Incomplete assembly scanning
+
+When using custom [assembly scanning](/nservicebus/hosting/assembly-scanning) it is important to include all relavant assemblies that provide logic including extensions to NServiceBus like packages provided by the community or custom packages.
+
+Make sure all relevant assemblies are included.
+
+## Version mismatch between NServiceBus and the used persistence
+
+NServiceBus versions 4.4.8, 4.5.7, 4.6.10, 4.7.7, 5.0.6, 5.1.4 and 5.2.8 introduced a new interface called `IPersistTimeoutsV2` and is used by persistance implementations. If you did not update NServiceBus to atleast version 5.2.8 but are using the latest version of a transport that is using `IPersistTimeoutsV2` then this exception can occur.
+
+Please update NServiceBus to atleast the its latest patch version to resolve this issue.
+
+
+## Selected a persistence that does not provide a specific storage feature
+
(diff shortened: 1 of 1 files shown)


SimonCropp:
a nullreferenceexception is a bug. it should not be documented. it should be fixed

SimonCropp:
@Particular/nservicebus-maintainers can you please chime in here.

timbussmann:
> If you did not update NServiceBus to at least version 5.2.8 but are using the latest version of a transport that is using `IPersistTimeoutsV2` then this exception can occur.

For IPersistTimeoutsV2 we added checks in the core which won't let the endpoint start for scenarios which require the `IPersistTimeoutsV2`. So I'm not understanding why this can cause an exception?

ramonsmits:
@timbussmann The downstream package uses an interface which is not present in the NServiceBus assembly thus is cannot create that implementation.

See https://github.com/Particular/NServiceBus.NHibernate/issues/140 for details.

ramonsmits:
I agree with @SimonCropp that we must prevent NRE's from throwing. Still, this documentation is needed when customer get this error and start googling.

SimonCropp:
> Still, this documentation is needed when customer get this error and start googling.

This would be better served by raising a descriptive github issue. Then people can

 * subscribe to be notified when it is fixed
 * engage in the conversation
 * see what version  it is planned to be fixed in
 * +1 it

And it will be just as discoverable by google

mauroservienti:
I agree that this is a bug, NRE can never be considered a good answer, IMO.

ramonsmits:
Just a quick Google:

* https://groups.google.com/forum/#!topic/particularsoftware/yoHSZQjwUQE
* https://github.com/Particular/NServiceBus/issues/1042
* https://github.com/Particular/NServiceBus/issues/2243
* https://github.com/Particular/NServiceBus.NHibernate/issues/140
* http://stackoverflow.com/questions/28473442/nservicebus-5-using-azurestoragepersistence-results-in-failed-to-fetch-timeouts

So yes, that NRE can probably resolved by doing a better job at creation time to see those types can actually be created.

@SimonCropp So you are suggesting that this doco is put iin all these issues?

danielmarbach:
I'm with @SimonCropp on this one. In general we should not document `NullReferenceException`. leading down that path it would require us to document every single bug on the documentation site. Which is far too much effort to be owned by the maintainers. I vote to close this PR. The issue in the NH repo is discoverable enough.

The core already has a decent check

https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Timeout/Core/TimeoutPersistenceVersionCheck.cs#L41

It seems that NHibernate was injecting the timeout persister interface into an infrastructure before that check had the time to run. So the fix is like you did to update the nuspec range and/or analyze what part injects the interface to early. Talked to @andreasohlund and we concluded that it is not worth backporting.

ramonsmits:
@danielmarbach The NH issue was related to a non backwards compatible interface on a type that should have been backward compatible. That ~~was~~ is  an implementation bug.

What is not worth backporting?

There is still the assembly scanning issue that can cause this NRE exception ( https://github.com/Particular/NServiceBus/issues/3729 )

SimonCropp:
@ramonsmits 

> @SimonCropp So you are suggesting that this doco is put iin all these issues?

No i am suggesting that the bug should be fixed so it is not an issue. And if it is effecting enough people perhaps it should be back ported. But we should never be documenting a nullref exception as if it is expected behaviour. If you still think this issue should be pursued then please raise in platform dev as it clearly spans multiple repo

PR #1485, opened 2016-05-04
corrected NHibernate link on pub/sub

d-pemberton:
NHibernate link was pointing to "ravendb", changed it to "nhibernate"


SimonCropp:
@d-pemberton thanks for the fix :)

PR #1492, opened 2016-05-06
Original message was not clear

SeanFeldman:
Do we really need to have a warning in the upgrade guide @SimonCropp ? 
I can see a warning in the general doco, but the upgrade guide.


SimonCropp:
i think we do need it in the upgrade guide. considering most upgrades can be done in any order this needs to be called out as different

SeanFeldman:
Ok. But the wording has to change.

>as older versions cannot translate the namespace.

makes little sense.

PR #1494, opened 2016-05-06
Review ASB doco

weralabaj:
Connects to https://github.com/Particular/PlatformDevelopment/issues/718

ping @SeanFeldman, as discussed yesterday


weralabaj:
Question: is batching a new feature? If not did anything change from v6 (apart from the default value of BatchSize)?

weralabaj:
I have a little trouble with http://docs.particular.net/nservicebus/azure-service-bus/message-lock-renewal, it's not clear why would be users interested in that and how to use it. the big picture and "Why" is missing.

E.g. is increasing it used to avoid duplicates? Is it an advanced option or standard one? How do I determine the value of the settings?

Maybe because I don't understand where it fits, it's hard for me to say if it should be linked to anything or if we should include any links in the article itself.

weralabaj:
Another question - namespaces partitioning - is that a completely new thing? Was partitioning possible before (even if using hacks)? Is the "Single namespace partitioning" backwards compatible?

I think we should at least mention it in an upgrade guide, but I need more info first.

weralabaj:
For native integration http://docs.particular.net/nservicebus/azure-service-bus/native-integration there are no changes or it hasn't been updated yet?

weralabaj:
Full configuration API page http://docs.particular.net/nservicebus/azure-service-bus/configuration/configuration is only accessible via link from http://docs.particular.net/nservicebus/azure-service-bus/configuration/. It's not included in the menu. The same goes for http://docs.particular.net/nservicebus/azure-service-bus/configuration/azureservicebusqueueconfig.

Was that intentional?

weralabaj:
That's all for now, I need some help with the questions above and to verify the changes make sense.

melkio:
> Another question - namespaces partitioning - is that a completely new thing? Was partitioning possible before (even if using hacks)? 
Yes

> Is the "Single namespace partitioning" backwards compatible?
Yes. Backward compatibility tests (`EndToEnd` repo) are configured with this strategy

nservicebus/azure-service-bus/batching.md
```diff
@@ -12,16 +12,16 @@ reviewed: 2016-04-26
...
+In order to increase messages throughput, the Azure Service Bus batches operations prior to writing to its internal store. This involves delaying write operations by 20 ms. Read operations are not affected.
 
 Batching store operations should be disabled for low throughput scenarios requiring low latency.
 
-The batching is enabled by default. In order to disable it, set `EnableBatching` to `false` on [queues](/nservicebus/azure-service-bus/configuration/configuration.md#Queues), [topics](/nservicebus/azure-service-bus/configuration/configuration.md#Topics), and [subscriptions](/nservicebus/azure-service-bus/configuration/configuration.md#Subscriptions).
+Batching is enabled by default. In order to disable it, set `EnableBatching` flag to `false` on [queues](/nservicebus/azure-service-bus/configuration/configuration.md#Queues), [topics](/nservicebus/azure-service-bus/configuration/configuration.md#Topics), and [subscriptions](/nservicebus/azure-service-bus/configuration/configuration.md#Subscriptions).
```
SeanFeldman: Perhaps the paragraph should start with "Batching store access". Just batching could be ambiguous.

SeanFeldman: > `EnableBatching` flag

Not sure if I like the "flag". What do you think about this?
> In order to disable it, use `EnableBatching()` configuration API setting it to `false`...

nservicebus/azure-service-bus/batching.md
```diff
@@ -63,7 +63,7 @@ The default value for `MessageSizePaddingPercentage` is 5%. The custom percentag
 | 180K  | 247,201 | 261,149 | 6% |
 
 
-`BrokeredMessageBodyType` set to `SupportedBrokeredMessageBodyTypes.Steam`. 
+`BrokeredMessageBodyType` set to `SupportedBrokeredMessageBodyTypes.Stream`.
```
SeanFeldman: Steam :joy:

nservicebus/azure-service-bus/index.md
```diff
@@ -11,37 +11,37 @@ related:
  - samples/azure/azure-service-bus
 ---
 
-In some environments it is not possible or recommended to rely heavily on distributed transactions to ensure reliability and consistency. Therefore in environments such as very large cloud networks or hybrid networks using MSMQ is not the best idea. In those scenarios a good alternative is Azure Service Bus.
+[Azure Service Bus (ASB)](https://azure.microsoft.com/en-us/services/service-bus/)  is a messaging service hosted on the Azure platform, that allows for exchanging messages between various applications in a loosely coupled fashion. ASB Queues offer FIFO (First In, First Out) guaranteed message delivery, and support a range of standard protocols (REST, AMQP, WS*) and APIs (to put/pull messages on/off the queue). ASB Topics deliver messages to multiple subscribers and allow to use fan-out pattern to deliver messages to downstream systems.
```
SeanFeldman: I would recommend to drop FIFO from the doco. It's not possible with ASB Transport, just with raw ASB.

nservicebus/azure-service-bus/index.md
```diff
@@ -11,37 +11,37 @@ related:
...
 
-NOTE: As part of the Azure support for NServiceBus, one can choose between two transports provided by the Azure platform Azure Storage Queues and Azure Service Bus. Each of these two options has separate features, capabilities, and usage characteristics. A detailed comparison and discussion of when to select which is beyond the scope of this document. To help decide which option best suits the application's needs, review the Azure article [Azure Queues and Azure Service Bus Queues - Compared and Contrasted](https://azure.microsoft.com/en-us/documentation/articles/service-bus-azure-and-service-bus-queues-compared-contrasted/).
+NServiceBus is an abstraction over ASB. It takes advantage of ASB's built-in features, such as message deduplication and deferred messages. It also provides higher-level, convenient API for programmers on top of ASB.
```
SeanFeldman: >  message deduplication

we don't leverage that.

How about 
>such as message batching and deferred messages

nservicebus/azure-service-bus/index.md
```diff
@@ -11,37 +11,37 @@ related:
...
+ * The main disadvantage of ASB is its dependency on TCP (for low latency), which may require opening outbound ports on the firewall. Additionally, in some systems the price for the service may get high ($1 per million messages).
 
-Note: Publish/Subscribe and Timeouts (including message deferral) are supported natively by the ASB transport and do not use persistence.
+NOTE: As part of the Azure support for NServiceBus, one can choose between two transports provided by the Azure platform: Azure Storage Queues and Azure Service Bus. Each of these two options has separate features, capabilities, and usage characteristics. A detailed comparison and discussion of when to select which is beyond the scope of this document. To help decide which option best suits the application's needs, review the Azure article [Azure Queues and Azure Service Bus Queues - Compared and Contrasted](https://azure.microsoft.com/en-us/documentation/articles/service-bus-azure-and-service-bus-queues-compared-contrasted/).
```
SeanFeldman: That MSFT reference is good, I made some changes. [PR](https://github.com/Azure/azure-content/pull/6512)

nservicebus/azure-service-bus/index.md
```diff
@@ -11,37 +11,37 @@ related:
...
-Then use the Configuration API to set up NServiceBus, by specifying `.UseTransport<T>()` to override the default transport:
+Then use the Configuration API to set up NServiceBus, by specifying `.UseTransport<AzureServiceBus>()` to override the default transport:
 
 snippet:AzureServiceBusTransportWithAzure
 
 
 ## Setting the Connection String
 
-The default way to set the connection string is using the .NET provided `connectionStrings` configuration section in app.config or web.config, with the name `NServicebus\Transport`:
+The default way to set the connection string is using the .NET provided `connectionStrings` configuration section in an `app.config` or a `web.config` file, with the name `NServicebus\Transport`:
```
SeanFeldman: If we're moving away from the xml based config, should we promote app.config is the default way? Doesn't sound right.

nservicebus/azure-service-bus/performance-tuning.md
```diff
@@ -10,12 +10,15 @@ tags:
...
 
-In scenario where handlers are IO intense, it is advised to set number of threads ([`MaximumConcurrencyLevel`](/nservicebus/operations/tuning.md) in NServiceBus) to 12 threads per logical core and `BatchSize` to a number of messages that takes to process, taking into account possible/measured single message processing time and IO latency. Start with a small `BatchSize` and through adjustment and measurement bring it up, adjusting accordingly `LockDuration` and `MaxDeliveryCount`.
\ No newline at end of file
+In scenario where handlers are IO intense, it is advised to set the number of threads  to 12 threads per logical core using `MaximumConcurrencyLevel` setting, and set the `BatchSize` to a number of messages that takes to process. Take into account the expected (or measured) processing time and IO latency of a single message. Start with a small `BatchSize` and through adjustment and measurement gradually increase it, while adjusting accordingly `LockDuration` and `MaxDeliveryCount`.
```
SeanFeldman: `MaximumConcurrencyLevel` is v5 looks like. In v6 concurrency level is set via code only. Perhaps to include the link we used to have? http://docs.particular.net/nservicebus/operations/tuning

nservicebus/azure-service-bus/publisher-names-configuration.md
```diff
@@ -9,17 +9,17 @@ tags:
 reviewed: 2016-04-18
 ---
 
-When `EndpointOrientedTopology` is selected, a mapping between publisher names and message types has to be properly configured, in order to ensure that subscriber receives event messages.  
+When the [`EndpointOrientedTopology`](/nservicebus/azure-service-bus/topologies/#version-7-and-above-endpoint-oriented-topology) is selected, a mapping between publisher names and message types has to be properly configured, in order to ensure that subscriber can receive event messages.
```
SeanFeldman: :+1:

nservicebus/azure-service-bus/publisher-names-configuration.md
```diff
@@ -9,17 +9,17 @@ tags:
...
 snippet: publisher_names_mapping_by_message_type
 
 
-or for an assembly containing message types:
+or for an assembly containing multiple message types:
 
 snippet: publisher_names_mapping_by_assembly
 
-In this case, the transport analyzes all types in the assembly to identify which are messages, according to the configured [`message conventions`](/nservicebus/messaging/conventions.md). Then for each message type, the transport registers a mapping between the type and the publisher name.
+In the latter case, the transport analyzes all types in the assembly to identify which are messages, according to the configured [`message conventions`](/nservicebus/messaging/conventions.md). Then for each message type, the transport registers a mapping between the type and the publisher name.
```
SeanFeldman: What if someone doesn't use message conventions, but marker interfaces? Will it confuse them and lead to the thinking that they **have to** use message conventions instead?

nservicebus/azure-service-bus/topologies/index.md
```diff
@@ -27,7 +27,7 @@ The `ForwardingTopology` was introduced to take advantage of the broker nature o
 
 Both topologies create a single input queue per endpoint and implement [Publish-Subscribe](/nservicebus/messaging/publish-subscribe/) mechanism. However, there is a significant difference in the way the mechanism is implemented.
 
-Starting from Version 7, no default topology is set by the Azure Service Bus transport. Topology has to be configured using [configuration API](/nservicebus/azure-service-bus/configuration/configuration.md).
+Starting from Version 7, no default topology is set by the Azure Service Bus transport. Topology has to be  explicitly configured using [configuration API](/nservicebus/azure-service-bus/configuration/configuration.md).
```
SeanFeldman: :+1:

nservicebus/upgrades/asb-6to7.md
```diff
@@ -49,21 +49,27 @@ The default values of the following settings have been changed:
...
 ## [Topology](/nservicebus/azure-service-bus/topologies/) is mandatory
 
 In Versions 7 and above the topology selection is mandatory:
 
 snippet:topology-selection-upgrade-guide
 
-When the `EndpointOrientedTopology` is selected, it is also necessary to configure [publisher names](/nservicebus/azure-service-bus/publisher-names-configuration.md), in order to ensure that subscribers receive event messages:
+The [`EndpointOrientedTopology`](/nservicebus/azure-service-bus/topologies/#version-7-and-above-endpoint-oriented-topology)  is backward compatible with the Azure Service Bus transport Version 6 and below. The [`ForwardingTopology`](/nservicebus/azure-service-bus/topologies/#version-7-and-above-forwarding-topology) is the recommended option for new projects. 
+
+When selecting `EndpointOrientedTopology`, it is also necessary to configure [publisher names](/nservicebus/azure-service-bus/publisher-names-configuration.md), in order to ensure that subscribers receive event messages:
```
SeanFeldman: > in order to ensure that subscribers receive event messages:

"in case pub/sub is used"?

SeanFeldman:
Well done @weralabaj. Left a few comments.

I'm not sure about "Secure credentials". The only thing we're securing is connection strings.
Should it be named differently? /cc @melkio

yvesgoeleven:
@weralabaj Batching is new yes, it will result in significant performance improvements over V6

yvesgoeleven:
@weralabaj regarding message lock duration: there are multiple reasons why it could change, ranging from a handler implementation that just takes long, to performance problems...

yvesgoeleven:
@weralabaj native integration is planned for RC https://github.com/Particular/docs.particular.net/issues/886

yvesgoeleven:
@weralabaj I've added full config & queue config to the menu

menu/menu.yaml
```diff
@@ -140,7 +140,11 @@
         - Title: "Azure Service Bus Transport"
           Url: "nservicebus/azure-service-bus"
           Articles:
-            - "nservicebus/azure-service-bus/configuration"
+            - Title: "Azure Service Bus Transport"
+              Url: "nservicebus/azure-service-bus/configuration"
+              Articles:
+                - "nservicebus/azure-service-bus/configuration/full"
```
SimonCropp: this is too many indents for our menu structure

nservicebus/azure-service-bus/batching.md
```diff
@@ -12,16 +12,16 @@ reviewed: 2016-04-26
 
 ## Batching store access
 
-In order to increase messages throughput, the Azure Service Bus batches operations prior to writing to its internal store. This involves delaying write operations. When batching is enabled, the write operations are delayed by 20 ms. Read operations are not affected. This increases throughput.
+In order to increase messages throughput, the Azure Service Bus batches operations prior to writing to its internal store. This involves delaying write operations by 20 ms. Read operations are not affected.
```
SimonCropp: "messages throughput" => "message throughput"

menu/menu.yaml
yvesgoeleven: Hmm, that's too bad... how do you suggest we work around this limit? both articles logicall sit beneath the configuration article, just no entry in the menu and reference from the file? Or flatten it?

SimonCropp: either would work. its up to u. depends on how discoverable u want them to be

nservicebus/azure-service-bus/index.md
SimonCropp: @yvesgoeleven did u not think this was worth addressing?

yvesgoeleven: @SimonCropp dedupe? Dedupe in azure is pretty much an illusion, a false sense of ease of programming, ASB's feature is best effort and time limited so you still have to program idempotently even if it's on.

SimonCropp: so isnt seans comment valid then?

yvesgoeleven: (it is valid)

You meant, why I haven't changed the text yet, right?

Because I still need to go through these comments, tackled the ones on the issue conversation so far.

SimonCropp: ahh sorry. thought u were done

nservicebus/azure-service-bus/index.md
yvesgoeleven: Why would we not be FIFO? We don't reverse the queue do we :p

nservicebus/azure-service-bus/index.md
yvesgoeleven: indeed, we should favor the code only API

nservicebus/azure-service-bus/performance-tuning.md
yvesgoeleven: I'll change it to the new API: LimitMessageProcessingConcurrencyTo

yvesgoeleven: In hindsight, this is just V5 content, the update is scheduled for RC timeframe, we should just leave it as is for now

nservicebus/azure-service-bus/publisher-names-configuration.md
yvesgoeleven: marker interfaces are the default convention, how can we make that clear so uses don't start overridding unless it's really needed

nservicebus/upgrades/asb-6to7.md
yvesgoeleven: In order to ensure that subscribers are subscribed to the correct publisher

nservicebus/azure-service-bus/index.md
yvesgoeleven: changed it now :)

nservicebus/azure-service-bus/index.md
SeanFeldman: @yvesgoeleven we don't reverse the order, but we affect it. Example: messages that are deferred. If you think that's not an issue, np :)

nservicebus/azure-service-bus/publisher-names-configuration.md
SeanFeldman: Looks like the sentence is broken.

> In the latter case, the transport analyzes all types in the assembly to identify which are messages, according to the configured message conventions. Then for each message type, the transport registers a mapping between the type and the publisher name

The messages that are being looked at are solely events. 
How about this:

> In the latter case, the transport analyzes all types in the assembly to identify event messages. Then for each event message type, the transport registers a mapping between the type and the publisher name. Events are identified by NServiceBus marker interface `IEvent` or configured [`event convention`](/nservicebus/messaging/conventions.md).

weralabaj: Hm @melkio told me something else, I remember we talked about it specifically. Could you confirm?

SeanFeldman: What do you mean something else? Publishers API is only for events...
Please elaborate.

weralabaj: It's possible I misunderstood, but I thought it's checking all message types. I'll wait for Alessandro to clarify.

melkio: Current implementation uses `Conventions.IsMessageType()`
Can a user publish a "generic" message? Users can also disable core best practices and publish a command. How to cover this scenarios?

melkio:
> I'm not sure about "Secure credentials". The only thing we're securing is connection strings.

When we talked about the article title we decided for "Securing connection string", isn't it? Did @SimonCropp rename it?

weralabaj:
In ASQ it's called "Securing connection strings" http://docs.particular.net/nservicebus/azure-storage-queues/configuration#connection-strings-securing-connection-strings so it'd make sense to make it consistent? /cc: @SimonCropp

weralabaj:
I once more went through all comments, outstanding questions:

- Should we mention FIFO or is the order not guaranteed? https://github.com/Particular/docs.particular.net/pull/1494#discussion_r62353082
- Do we have code API to set up connection strings? if so let's provide a snippet here https://github.com/Particular/docs.particular.net/pull/1494#discussion_r62390589.
- Publisher namespaces - messages or events? https://github.com/Particular/docs.particular.net/pull/1494#discussion_r62584516
- Should we change "securing credentials" to "securing connection strings"?
- Should we mention new features in the upgrade guide? I'd say yes, just a list with links to feature description (e.g. batching, namespace partitioning).

nservicebus/azure-service-bus/publisher-names-configuration.md
yvesgoeleven: It is possible to publish non events, but should we really cover all this?

weralabaj: I think the issue is that the doco covers implementation correctly, so the question is this a problem with implementation or shoudl I just adjust doco as Sean suggested?

nservicebus/azure-service-bus/index.md
yvesgoeleven: FIFO does not mean order can never change, it just means that the first message in is probably the first message out, and that is true...

nservicebus/azure-service-bus/publisher-names-configuration.md
yvesgoeleven: Afaik the capability to publish non events is a backward compat feature from before core v5 timeframe, so I suggest to use the word events as we really want people to be literal about it

yvesgoeleven:
I went for:

1. FIFO is ok
2. Added snippet
3. Events
4. Securing connection strings
5. I'm on the fence, leaning towards no, I could probably live with a 'oh did you know about...' kind of link at the end of the article, but the article itself should be focussed on upgrade and not derail into other things

Changes 2, 3 & 4 have been committed

weralabaj:
OK, so I'm removing the WIP label.

weralabaj:
@SimonCropp the outstanding question is could we change "Securing credentials" to "Securing connection strings"?

SimonCropp:
I dont like the term connection strings since many apis we provide an explicit way to provide strong typed connections. Connection strings are an abstraction. So i prefer "securing connections" or "securing credentials".

yvesgoeleven:
@SimonCropp both are technically not what we do, we ensure that asb connectionstrings pointing to namespace which happen to contain credentials, do not get sent around. Change it to 'securing namespace connectionstrings'?

SimonCropp:
Ok so do we change asq to make the consistent?

yvesgoeleven:
I'll submit a PR to ASQ

SimonCropp:
@weralabaj @SeanFeldman @yvesgoeleven i did some very minor tweaks. but i think this is ready to merge. anything outstanding or objections?

SeanFeldman:
:shipit:

PR #1496, opened 2016-05-06
Raven DTC Configuration

DavidBoike:
This is a start at changing the ResourceManagerId article into a more comprehensive article about configuring RavenDB to have safe DTC settings.

So far I've written code snippets for NSB.Raven V4 only, where we have the new DocStore by Func<ReadOnlySettings, IDocumentStore> that gives us access to settings.LocalAddress() which makes it easy to define DTC settings in an INeedInitialization.

If we like this (and hence why I would like some review at WIP stage) then we have 2 options:

* Finish the other versions using an extension method instead of an INeedInitialization, where the user would have to pass in the unique value like:

```
busConfiguration.UsePersistence<RavenDBPersistence>()
    .WithDtcIdentifier("fed-into-DeterministicGuidBuilder-string");
```

* Backport the Func<ReadOnlySettings, IDocumentStore> to previous versions, which would be a feature, so we'd be talking 1.1, 2.3, and 3.2. That would also reset the clock on support for those minors for a year.

Comments please? @SimonCropp @Particular/ravendb-persistence-maintainers @andreasohlund @johnsimons


Snippets/Raven/Raven_4/RavenDtcConventions.cs
```diff
@@ -0,0 +1,62 @@
...
+        store.ResourceManagerId = resourceManagerId;
+        store.TransactionRecoveryStorage = new LocalDirectoryTransactionRecoveryStorage(txRecoveryPath);
+
+        return store;
+    }
+
+    static Guid DeterministicGuidBuilder(string input)
+    {
+        // use MD5 hash to get a 16-byte hash of the string
+        using (var provider = new MD5CryptoServiceProvider())
```
MarcinHoppe: Can we not spread MD5 any further? :smile:

nservicebus/ravendb/manual-dtc-settings.md
```diff
@@ -0,0 +1,62 @@
+---
+title: 'Setting RavenDB DTC settings manually'
+summary: 'Guidance on how to change the RavenDB ResourceManagerID and TransactionRecoveryStorage'
```
SimonCropp: add a full stop

nservicebus/ravendb/manual-dtc-settings.md
```diff
@@ -0,0 +1,62 @@
...
+tags: [RavenDB, Persistence]
+redirects:
+ - nservicebus/ravendb/how-to-change-resourcemanagerid
+ - nservicebus/ravendb/resourcemanagerid
+---
+
+In order to provide the most reliable support for distributed transactions in RavenDB, a custom DocumentStore must be provided to NServiceBus and configured to uniquely identify the NServiceBus endpoint to the Distributed Transaction Coordinator (DTC) and provide a storage location for uncommitted transaction recovery.
```
SimonCropp: isnt `the most` redundant?

Snippets/Raven/Raven_4/RavenDtcConventions.cs
```diff
@@ -0,0 +1,62 @@
...
+
+class RavenDtcConventions : INeedInitialization
+{
+    public void Customize(EndpointConfiguration configuration)
+    {
+        configuration.UsePersistence<RavenDBPersistence>()
+            .SetDefaultDocumentStore(CreateDocumentStore);
+    }
+
+    private IDocumentStore CreateDocumentStore(ReadOnlySettings settings)
```
SimonCropp: remove private

SimonCropp:
so SetDefaultDocumentStore will be marked as obsolete? and added to an upgrade guide?

SimonCropp:
So if that config is backported why r there no earlier version snippets for RavenDBManualDtcSettingExample  ?

SimonCropp:
missing menu change

nservicebus/ravendb/manual-dtc-settings.md
```diff
@@ -0,0 +1,62 @@
...
+
+To guard against the loss of a committed transaction, RavenDB requires a storage location to persist data in the event of a process crash immediately following a transaction commit.
+
+RavenDB provides the following methods for persisting transaction recovery storage information:
+
+* `VolatileTransactionRecoveryStorage` persists the information in memory, and since the information will be lost during a process restart, is not a safe storage method.
+* `IsolatedStorageTransactionRecoveryStorage` persists the information in [Isolated Storage](https://msdn.microsoft.com/en-us/library/system.io.isolatedstorage.aspx). This is the default for NServiceBus.RavenDB when connecting with a connection string, as it does not require any additional configuration. However, in high-contention scenarios, IsolatedStorage has been found to become unstable and may begin to throw exceptions. Therefore using LocalDirectoryTransactionRecoveryStorage is preferable.
+* `LocalDirectoryTransactionRecoveryStorage` stores transaction recovery information to a local path on disk, and is the safest option for production scenarios, but must be configured with a directory that is writeable by the application.
```
SimonCropp: writable

nservicebus/ravendb/manual-dtc-settings.md
```diff
@@ -0,0 +1,62 @@
...
+## Configuring safe settings
+
+In order to configure safe settings for production use of RavenDB, construct a `DocumentStore` instance and configure the `ResourceManagerId` and `TransactionRecoveryStorage` properties as shown in the following code:
+
+snippet:RavenDBManualDtcSettingExample
+
+In order to provide transaction safety, the following must be observed:
+
+* `ResourceManagerId` must be constant across process restarts, and uniquely identify the process running on the machine. (Do not use `Guid.NewGuid()`.) Otherwise, the transaction recovery process will fail when the process restarts.
+* `TransactionRecoveryStorage` must be set to an instance of `LocalDirectoryTransactionRecoveryStorage`, configured to a directory that is constant across process restarts, and writeable by the process.
```
SimonCropp: writable

SimonCropp: above u say application. here you say process. also why is that text duplicated?

nservicebus/ravendb/manual-dtc-settings.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Configuration by convention
+
+It can be cumbersome to manage these settings for multiple endpoints, so it is preferable to create a convention that will calculate a unique ResourceManagerId,
+and then use this value to create a storage location for TransactionRecoveryStorage as well.
```
SimonCropp: remove the newline

nservicebus/ravendb/manual-dtc-settings.md
```diff
@@ -0,0 +1,62 @@
...
+
+The ResourceMangaerId is a Guid that uniquely identifies a transactional resource on a machine, and must be both unique on that machine, and stable (deterministic) between restarts of the process utilizing the DTC. If more than one RavenDB document store attempts to use the same ResourceManagerId, it can result in the following error during a commit operation:
+
+> "A resource manager with the same identifier is already registered with the specified transaction coordinator"
+
+### TransactionRecoveryStorage
+
+To guard against the loss of a committed transaction, RavenDB requires a storage location to persist data in the event of a process crash immediately following a transaction commit.
+
+RavenDB provides the following methods for persisting transaction recovery storage information:
```
SimonCropp: why are we documenting raven storage options here? just link to the raven doco

nservicebus/ravendb/manual-dtc-settings.md
```diff
@@ -0,0 +1,62 @@
...
+and then use this value to create a storage location for TransactionRecoveryStorage as well. 
+
+snippet:RavenDBDtcSettingsByConvention
+
+It is important to keep a few things in mind when determining a convention.
+
+The string provided to DeterministicGuidBuilder will define the ResourceManagerId, and thus the identity of the endpoint. This string value must then be unique within the scope of the server. The EndpointName or LocalAddress provide attractive options as they are normally unique per server.
+
+An exception is side-by-side deployment, where an old version and new version of the same endpoint run concurrently, processing messages from the same queue, in order to vet the new version and enable easy rollback to the previous version. In this case using EndpointName or LocalAddress would result in duplicate ResourceManagerId values on the same server, which would lead to DTC exceptions. In this case, the a versioning scheme must be created, and the version included in the string provided to DeterministicGuidBuilder.
```
SimonCropp: vet => verify

SimonCropp: remove "easy "

SimonCropp: > In this case, the a versioning scheme must be created, and the version included in the string provided to DeterministicGuidBuilder

what version?

nservicebus/ravendb/manual-dtc-settings.md
```diff
@@ -0,0 +1,62 @@
...
+
+The exact convention used must be appropriately defined to match the deployment strategy in use for the endpoint.
```
SimonCropp: how does someone achieve this?

nservicebus/ravendb/manual-dtc-settings.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Safely decommissioning endpoints
+
+If an endpoint terminates unexpectedly for any reason, data can be left behind in the transaction recovery storage which represents transactions not yet committed to the RavenDB database, but which may be recovered when the endpoint restarts.
+
+In order to avoid losing data, it is important to ensure that endpoints that are decommissioned are taken offline through graceful endpoint shutdowns, and then the transaction recovery storage directory should be inspected to ensure no recovery files remain behind.
```
SimonCropp: "through graceful endpoint shutdowns"  how does someone ensure this?

SimonCropp: > then the transaction recovery storage directory should be inspected to ensure no recovery files remain behind.

Will there be other files there which are not recovery files? if so how does someone differentiate?

johnsimons:
Are we missing the configuration via connection string ?

johnsimons:
So am I correct in assuming that by default (with DTC) we are not safe ? Unless the users do changes ?

Snippets/Raven/Raven_4/RavenDtcConventions.cs
DavidBoike: This is a suggestion we're making to a customer to achieve a goal, not something we're baking into the product which must hold for all customers. If they're in a FIPS environment it's up to them to select a different hash algorithm. MD5 is the simplest one that results in exactly 16 bytes needed for the Guid.

DavidBoike:
@SimonCropp:

Great comments - thanks! Changes integrated.

> so SetDefaultDocumentStore will be marked as obsolete? and added to an upgrade guide?

There are now 3 overloads:

1. SetDefaultDocumentStore(IDocumentStore)
2. SetDefaultDocumentStore(ConnectionParameters)
3. SetDefaultDocumentStore(Func<ReadOnlySettings, IDocumentStore>)

We haven't considered obsoleting any of them. Are you referring to Overload 1 only? Or 1 & 2?

> So if that config is backported why r there no earlier version snippets for RavenDBManualDtcSettingExample ?

It's not, yet. Need to determine if it should be. If it is, the snippet I have here can basically match. If not, then you can't really do it by convention in an INeedInitialization at all. There just isn't enough information available. Then you have do it as an extension method SetDtcTransactionIdentity(string uniqueIdentifier) on PersistenceExtensions or some such thing.

**@Particular/ravendb-persistence-maintainers  That's a decision we need to make, and quickly.**

@johnsimons:

> Are we missing the configuration via connection string ?

Missing in what way? You can't set up LocalDirectoryTransactionRecoveryStorage that way, so not sure exactly what you're saying.

> So am I correct in assuming that by default (with DTC) we are not safe ? Unless the users do changes ?

That's correct. Since we can't "magic" in appropriate LocalDirectory based settings, we can't provide safe by default.

Snippets/Raven/Raven_4/RavenDtcConventions.cs
MarcinHoppe: Devs (some, not all) tend to copy & paste code from SO, MSDN, and other places. Our doco is no different. My point is that if we can avoid "Particular used MD5 in their sample so it must be legit" type of argument. If we can make developers better by showing them SHA1, let's do it. Just my 2 cents.

DavidBoike: Well, MD5 is legit, for this purpose, unless you're running FIPS, in which case the sample won't work for you, and in your environment you probably understand that about MD5, right?

SHA1 makes for a more complex example because it generates 4 more bytes than are needed, so you have to trim them.

I'm hesitant of getting right back into the same trap of having to create code that's perfectly appropriate for all customers, like we need to do in our libraries, in a documentation snippet

SimonCropp: Agreed. Md5 is fine in this context

johnsimons:
@DavidBoike I am referring to setting the `ResourceManagerId` via connection string, see https://ravendb.net/docs/article-page/3.0/csharp/client-api/setting-up-connection-string

DavidBoike:
Is there any point to that method when it doesn't also deal with the transaction recovery storage?

johnsimons:
> Is there any point to that method when it doesn't also deal with the transaction recovery storage?

Well, if you think there is no point of using the connection string without the other, then the doco should say that, right ?

nservicebus/ravendb/manual-dtc-settings.md
```diff
@@ -0,0 +1,57 @@
...
+
+In order to provide reliable support for distributed transactions in RavenDB, a custom DocumentStore must be provided to NServiceBus and configured to uniquely identify the NServiceBus endpoint to the Distributed Transaction Coordinator (DTC) and provide a storage location for uncommitted transaction recovery.
+
+## Definitions
+
+Support for the Distributed Transaction Coordinator (DTC) in RavenDB is dependent upon the **ResourceManagerId** and **TransactionRecoveryStorage** settings.
+
+### ResourceManagerId
+
+The ResourceMangaerId is a Guid that uniquely identifies a transactional resource on a machine, and must be both unique on that machine, and stable (deterministic) between restarts of the process utilizing the DTC. If more than one RavenDB document store attempts to use the same ResourceManagerId, it can result in the following error during a commit operation:
```
johnsimons: spelling mistake

mauroservienti:
> Well, if you think there is no point of using the connection string without the other, then the doco should say that, right ?

and it'll also result in a WARN logged and endpoint startup.

nservicebus/ravendb/manual-dtc-settings.md
mauroservienti: > If more than one RavenDB document store attempts to use the same ResourceManagerId

Shouldn't it be `DocumentStore`?

DavidBoike:
@johnsimons Not sure I'm following you.

> > Is there any point to that method when it doesn't also deal with the transaction recovery storage?

> Well, if you think there is no point of using the connection string without the other, then the doco should say that, right ?

Are you talking about use of a connection string at all? Or the specification of a ResourceManagerId within the connection string?

Currently this PR removes all mention of ResourceManagerId via connection string, so in that case, I'm not sure why the doco would want to say anything about it.

OR, are you saying that because you can't properly do ResourceManagerId and TransactionRecoveryStorage if we connect via connection string (best you could do is custom ResourceManagerId via connectionstring and IsolatedStorage by default, and still get a warning) then we should be more upfront about saying we REALLY want you to provide a DocumentStore?

And then if THAT's the case, the next question, I suppose, if if connection string approaches (including default localhost:8080) should be deprecated in 4, removed in 5? (And if so, is there any better way to obsolete a connection string mechanism other than to change the WARN inParticular/NServiceBus.RavenDB#232 to be an obsolete warning?)

DavidBoike:
Also, it would be very valuable to me for everyone following this issue (and @Particular/ravendb-persistence-maintainers) to provide input on whether the Func<ReadOnlySettings, IDocumentStore> configuration option should be backported, and if so, how far? Or in order to give that feedback, do you need to see from me how different/awful the "configuration by convention" approach would look in the V3 snippet?

I could see backporting it to V3 as NSB.Raven 3.2.0 (new minor as it's new public API) but I struggle to see the value in backporting it past that. Only difference down to 2.x is RavenDB 2.5, and 1.x supports NSB V4 which is about to go out of support in July anyway.

nservicebus/ravendb/manual-dtc-settings.md
DavidBoike: @johnsimons Spelling mistake? Which word?

@mauroservienti Do we always need to refer to a document store by its class name? I feel like if it's DocumentStore then it also has to be `DocumentStore` and excessive `CodeInText` style makes it harder to read.

SimonCropp: >  Do we always need to refer to a document store by its class name? I feel like if it's DocumentStore then it also has to be DocumentStore and excessive CodeInText style makes it harder to read.

@DavidBoike i have found the same re "CodeInText style makes it harder to read" but i have been struggling to on how to apply a rule to that. seems if we do it sometime we should do it all the time. appreciate any thoughts on when and when not to use code in style that can be applied generically

johnsimons:
@DavidBoike I have fixed the spelling

johnsimons:
> Are you talking about use of a connection string at all? Or the specification of a ResourceManagerId within the connection string?

The later, "specification of a ResourceManagerId within the connection string"

johnsimons:
> Currently this PR removes all mention of ResourceManagerId via connection string, so in that case, I'm not sure why the doco would want to say anything about it.

Because if a user is used to code against raven and reads raven doco (eg https://ravendb.net/docs/article-page/3.0/csharp/client-api/setting-up-connection-string), they could still use it. So it doesn't matter we don't talk about it.

nservicebus/ravendb/manual-dtc-settings.md
```diff
@@ -0,0 +1,57 @@
...
+
+> "A resource manager with the same identifier is already registered with the specified transaction coordinator"
+
+### TransactionRecoveryStorage
+
+To guard against the loss of a committed transaction, RavenDB requires a storage location to persist data in the event of a process crash immediately following a transaction commit.
+
+RavenDB provides [transaction recovery storage options](http://ravendb.net/docs/article-page/3.0/Csharp/client-api//session/transaction-support/dtc-transactions#transaction-storage-recovery) for volatile (in-memory) storage, IsolatedStorage, and local directory storage. `LocalDirectoryTransactionRecoveryStorage` is recommended as the only stable and reliable option.
```
SimonCropp: avoid deep linking into raven https://github.com/Particular/docs.particular.net#ravendb

DavidBoike:
@johnsimons I added a NOTE box to say that while you can set ResourceManagerId with the connection string, we don't recommend it.

@SimonCropp Integrated your notes, plus added corresponding snippets for all versions. Backporting the Func-based setup was moved out to an external issue so that this can be closed. That issue can update these snippets to use the Func-based config for the 3.x snippet of the "config by convention" section.

Also updated the Raven snippets to the Beta3 released today.

Removed the WIP and all checks are green.

PR #1505, opened 2016-05-10
make it more clear what the connectionstring is for

yvesgoeleven:
make it more clear what the connectionstring is for, to be consistent (and differentiating from) the same text in asb doco


Scooletz:
:+1:

PR #1502, opened 2016-05-09
Update for ASP beta 1 and 2

dbelcham:
Connects to https://github.com/Particular/PlatformDevelopment/issues/729

This PR is a result of configuration changes in https://github.com/Particular/NServiceBus.Persistence.AzureStorage/pull/73

PR #1506, opened 2016-05-10
Upgrade guide for the azure databus split

andreasohlund:
Went with a short section in the main upgrade guide and then a new page for the data bus specific details.

Depends on https://github.com/Particular/docs.particular.net/pull/1502 since I'm linking to the upgrade guide for ASP


andreasohlund:
Update after review with @SimonCropp

PR #1509, opened 2016-05-11
Add warning about duplicated SC notification events

mikeminutillo:
Replacement for https://github.com/Particular/docs.particular.net/pull/1475

### DIFF
--- a/servicecontrol/contracts.md
+++ b/servicecontrol/contracts.md
@@ -16,7 +16,7 @@ redirects:
 
 ServiceControl events allow to build notifications/integrations that will alert of something going wrong in the system.
 
-WARNING: ServiceControl sends external notification events in batches. If a problem is encountered part way through a batch, the entire batch will be resent. This can result in receiving mutliple events for a single notification. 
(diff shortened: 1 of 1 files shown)


servicecontrol/contracts.md
```diff
@@ -16,6 +16,8 @@ redirects:
 
 ServiceControl events allow to build notifications/integrations that will alert of something going wrong in the system.
 
+WARNING: ServiceControl sends external notification events in batches. If a problem is encountered part way through a batch, the entire batch will be resent. This can result in receiving mutliple events for a single notification.
```
SimonCropp: perhaps `re-sent`

mikeminutillo: Well technically they have a new message Id so they are different messages. Just dropping Outbox in won't fix it. For that reason I wanted to be very careful about this was phrased.

mikeminutillo: Ah. `re-sent` instead of `resent`. Yes. We don't resent batches :)

PR #1530, opened 2016-05-13
Step-by-step guide

DavidBoike:
Connects to Particular/DeveloperEducation#551


samples/step-by-step/sample.md
```diff
@@ -7,100 +7,190 @@ redirects:
 - nservicebus/nservicebus-step-by-step-guide
 ---
 
-In this sample shows a very simple ordering system that
+This sample illustrates most of the basic NServiceBus concepts in a step-by-step fashion. You can either download the code from the link above and run it, or you can build it yourself following the instructions here.
```
SimonCropp: i still dont see the value of personal voice here. eg the above could be 

> This sample illustrates most of the basic NServiceBus concepts in a step-by-step fashion. Either download the code from the link above and run it, or build it with the instructions.

It is even more concise

DavidBoike:
I'm torn on the personal voice. Since we're flagging it during build, I'm trying to remove it.

The soft benefit of personal voice is that it builds rapport / empathy with the reader, and a sense of "We're in this together."

The hard benefit I see is that it allows for a subject in a sentence, thus avoiding passive voice. Here's an example where that gets tough:

> First, we need to define the messages that will be used by the system.

"We" becomes the subject.

> First, it is required to define...

I don't like how that sounds at all. So the best I can come up with is:

> First, the messages that will be used by the system must be defined.

Passive voice.

DavidBoike:
@Particular/docs-maintainers ready for review

samples/step-by-step/Core_5/Client/Program.cs
```diff
@@ -3,24 +3,39 @@
 
 class Program
 {
-
+    #region StepByStep-ClientInit
     static void Main()
     {
+        // This makes it easier to tell console windows apart
```
indualagarsamy: I think it will keep the code snippet much cleaner, if the comments are part of the documentation that describes the Client and how to setup the client than individual line comments in the code. @Particular/docs-maintainers - thoughts?

samples/step-by-step/sample.md
```diff
@@ -7,100 +7,191 @@ redirects:
 - nservicebus/nservicebus-step-by-step-guide
 ---
 
-In this sample shows a very simple ordering system that
+This sample illustrates most of the basic NServiceBus concepts in a step-by-step fashion. Download the code from the link above and run it, or build it  following the instructions here.
```
indualagarsamy: > This sample illustrates most of the basic NServiceBus concepts in a step-by-step fashion. 

Why most of? Why not just:
This sample illustrates the basic NServiceBus concepts in a step-by-step fashion.

>Download the code from the link above and run it, or build it  following the instructions here.

Do we need to say run it, build it etc?
How about:
To download the code for this sample click the `Download` link. ( I honestly think that this is overkill also, after all there is a Download button)

samples/step-by-step/sample.md
```diff
@@ -7,100 +7,191 @@ redirects:
...
+
+
+## Concepts
+
+When complete, this sample will show a very simple ordering system that:
```
indualagarsamy: > When complete, this sample will show a very simple ordering system that:

Rather than start with When complete when the user has hardly begun, how about something like:

The goal of this step-by-step sample is to show how to build a simple system that incorporates using messaging to:
- Send a command from a  client to the server
- Receive a message from a client on the server that publishes a new event 
- Add a new subscriber that can receive published events.

samples/step-by-step/Core_5/Client/Program.cs
DavidBoike: Why does the snippet need to be clean exactly? Aren't comments supposed to explain "why"? What would not be clean would be 6 lines of "I don't have any idea what this does and I have to take this on faith" so of course we have to explain it in some way, at least a little bit.

If that explanation is outside the snippet then you have to repeat everything to refer back. Or you have to divide it into 6 different 1-line snippets, in which case you lose a feel for what it's supposed to look like all together.

mikeminutillo:
@DavidBoike 

> First, define the messages that will be used by the system.

Avoids personal voice and switches from passive to imperative.

DavidBoike:
@mikeminutillo Good idea. Used it.

samples/step-by-step/sample.md
```diff
@@ -7,100 +7,191 @@ redirects:
...
-snippet:OrderCreatedHandler
+When the Subscriber endpoint initializes, it will read this configuration, and knowing that it contains a message handler for `OrderPlaced`, it will send a special subscription message to the `Samples.StepByStep.Server` endpoint. When that endpoint receives the subscription request, it will store it locally (in this sample, using in-memory storage, but in a production system a database would be used) so that when it publishes a message, it can send a copy to every subscriber that expressed interest.
 
 
 ## Running the solution
 
-Run the solution.
+If following along with this tutorial, Visual Studio must be configured so that multiple projects will start when debugging.
+
+1. Right-click the solution file, and select **Properties**.
```
SimonCropp: remove these steps. if this is required it should be handled in a consistent way by the docs engine.

samples/step-by-step/sample.md
```diff
@@ -7,100 +7,191 @@ redirects:
...
- * a subscriber listens to, and handles, the event
 
+## Concepts
 
-## The Shared Project
+The goal of this step-by-step sample is to show how to build a simple system that uses messaging to:
 
-The `Shared` project is the container for shared classes including message definitions. This project will be shared between the client and server so both sides agree on the typed message descriptions. It is referenced by all projects in the solution.
+ * Send a command from a client to a server
+ * Receive a message from a client on the server that publishes a new event
```
SimonCropp: this statement seems confusing is it "the server that published the event" or "the server that then publishes a new event"

samples/step-by-step/sample.md
```diff
@@ -7,100 +7,191 @@ redirects:
...
+ * Add a new subscriber that can receive published events
 
+Completing these steps will serve as an introduction to many important NServiceBus concepts:
```
SimonCropp: in the below items link to articles where possible

samples/step-by-step/sample.md
```diff
@@ -7,100 +7,191 @@ redirects:
...
-snippet:OrderPlaced
+The shared project will be referenced by all of the applications and serve as a central location to store message definition classes and common pieces of configuration. This is not a best practice to combine all these things in one place, but serves to illustrate how things work for this simple example.
+
+
+### Setting dependencies
+
+Before beginning, set up the necessary dependencies.
+
+* Add a reference to the `Shared` project to the `Client`, `Server`, and `Subscriber` projects.
+* In each project, add a reference to the `NServiceBus` NuGet package, either using the Package Manager Console or the NuGet Package Manager visual tool.
```
SimonCropp: ", either using the Package Manager Console or the NuGet Package Manager visual tool." i think this is redundant give our target audience

samples/step-by-step/sample.md
```diff
@@ -7,100 +7,191 @@ redirects:
...
+
+
+## Defining messages
+
+First, define the messages that will be used by the system. The client will send a `PlaceOrder` command to the server, and in response, the server will publish an `OrderPlaced` event.
+
+There is a difference between commands and events. A **command** is a request to perform a task, which is sent to one specific location. In contrast, an **event** is an announcement that something has happened, which is published from one location and can be consumed by multiple subscribers.
```
SimonCropp: link to more info

samples/step-by-step/sample.md
```diff
@@ -7,100 +7,191 @@ redirects:
...
+
+The easiest way to differentiate commands and events is to use the `ICommand` and `IEvent` marker interfaces.
+
+Let's define the `PlaceOrder` command by creating a class that implements `ICommand` in the Shared project called `PlaceOrder`:
```
SimonCropp: remove "lets"

samples/step-by-step/sample.md
```diff
@@ -7,100 +7,191 @@ redirects:
...
+
+With a `PlaceOrder` command defined, and NServiceBus initialized, a loop can be created to send a new command message every time the Enter key is pressed.
+
+In the Client application, add this code to the Program class:
 
-snippet:SendOrder
+snippet:StepByStep-SendOrder
 
-This is a console blocking loop that is called the the console starts.
+In order to actually run, each endpoint needs to specify an error queue. This defines where messages are sent when they cannot be processed due to repetitive exceptions during message processing.
```
SimonCropp: "In order to actually run, " is redundant

samples/step-by-step/sample.md
```diff
@@ -7,100 +7,191 @@ redirects:
...
 
 ## The Server
 
-The `Server` project processes an Order. It receives `PlaceOrder` sent from `Client` and then publishes `OrderPlaced` on success.
+Like the client, the Server application is going to need to be configured as an NServiceBus endpoint as well.
+
+In the Server application, add the following code to the program's Main method:
 
-snippet:PlaceOrderHandler
+snippet:StepByStep-ServerInit
```
SimonCropp: u dont need to prefix snippets with StepByStep

samples/step-by-step/sample.md
```diff
@@ -7,100 +7,191 @@ redirects:
...
+
+In the previous snippet, notice the change in endpoint name, which differentiates this endpoint from the Client.
```
SimonCropp: "In the previous code"

samples/step-by-step/sample.md
```diff
@@ -7,100 +7,191 @@ redirects:
...
+
+Next, create a handler for the `PlaceOrder` command being sent by the Client.
+
+Create a new class in the Server project named `PlaceOrderHandler` using the following code:
+
+snippet:StepByStep-PlaceOrderHandler
+
+The handler class is automatically discovered by NServiceBus because it implements the `IHandleMessages<T>` interface. The dependency injection system (which supports constructor or property injection) injects the `IBus` instance into the handler to access messaging operations.
```
SimonCropp: link to containers

samples/step-by-step/sample.md
```diff
@@ -7,100 +7,191 @@ redirects:
...
+The important part is in the `MessageEndpointMappings` section. The `add` directive says that for messages of type `OrderPlaced` within the `Shared` assembly, subscription requests should be sent to the `Samples.StepByStep.Server` endpoint, which is the endpoint that is responsible for publishing `OrderPlaced` events.
 
-snippet:OrderCreatedHandler
+When the Subscriber endpoint initializes, it will read this configuration, and knowing that it contains a message handler for `OrderPlaced`, it will send a special subscription message to the `Samples.StepByStep.Server` endpoint. When that endpoint receives the subscription request, it will store it locally (in this sample, using in-memory storage, but in a production system a database would be used) so that when it publishes a message, it can send a copy to every subscriber that expressed interest.
 
 
 ## Running the solution
 
-Run the solution.
+If following along with this tutorial, Visual Studio must be configured so that multiple projects will start when debugging.
```
SimonCropp: i think it is enough to say "Set Client, Server, and Subscriber to be startup projects"

DavidBoike:
Ready for another review

samples/step-by-step/sample.md
```diff
@@ -7,100 +7,193 @@ redirects:
...
+
+The important part is in the `MessageEndpointMappings` section. The `add` directive says that for messages of type `OrderPlaced` within the `Shared` assembly, subscription requests should be sent to the `Samples.StepByStep.Server` endpoint, which is the endpoint that is responsible for publishing `OrderPlaced` events.
+
+When the Subscriber endpoint initializes, it will read this configuration, and knowing that it contains a message handler for `OrderPlaced`, it will send a special subscription message to the `Samples.StepByStep.Server` endpoint. When that endpoint receives the subscription request, it will store it locally (in this sample, using in-memory storage, but in a production system a database would be used) so that when it publishes a message, it can send a copy to every subscriber that expressed interest.
+
 
 ## Running the solution
 
-Run the solution.
+First, set Client, Server, and Subscriber to be startup projects:
```
SimonCropp: remove the points below. we r not teaching them how to use VS

DavidBoike: You're making the assumption that this is common knowledge. I didn't even know this was an option before NServiceBus.

SimonCropp: Then link to external doco. We should not own this nor polute every sample with it

DavidBoike:
@SimonCropp: Replaced multiple startup projects instructions with an MSDN link.

weralabaj:
@DavidBoike 
This part is awesome "Completing these steps will serve as an introduction to many important NServiceBus concepts:"!

> A Console Application named Subscriber: The subscriber will subscribe to

Something missing?

> A Console Application named Client: The client's job will be to send PlaceOrder commands to the Server.

I think we could skip the explanation and just leave "create x, y, and z". Especially that we explain it right below. It'll be slightly easier to follow.

> This is not a best practice to combine all these things in one place, but serves to illustrate how things work for this simple example.

Note: Storing all message definitions in a single location... 

> Add a reference to the Shared project to the Client, Server, and Subscriber projects.

In Client, Server and Subscriber projects  add a reference to the Shared project.

> In each project, add a reference to the NServiceBus NuGet package.

What version? I think that's important, since there're significant differences in APIs across versions and soon v6 will be the default in NuGet.

> Don't worry about the SendOrder method, that will be covered soon.

Don't worry about the missing `SendOrder` method, it'll be added it very soon.

Since we cover it right away, wouldn't it be easier to say "add clientinit code and also sendorder code" and then explain what it does? Or even just do it in a single snippet?

> Each endpoint needs to specify an error queue. This defines where messages are sent when they cannot be processed due to repetitive exceptions during message processing.

> While it is possible to configure the error queue in an App.config file, the IProvideConfiguration interface can be used to override app.config settings, which has the benefit able to be shared between all endpoints.

> In the Shared project, create a class named ConfigErrorQueue:

Keep it in a single paragraph. I think each paragraph should be a complete independent step, when the first one finished I felt a bit lost and only after a moment continued talking. Maybe reversing the order would also help - first do, then explain what you just did and why. 

Do we even need the "While it's possible..." section? If so, could it be moved to the Note section, because it interrupts the flow, and/or we could link to the doco on various ways it could be configured?

> The Client application is now complete, and could now be executed.

repetition with now

> In the Shared project, create a class named ConfigErrorQueue:

That could be moved to the section when we create messages. It's not really related to the client directly.

> Normally a handler is where information from the message would be saved **it to** a database,

into a database?

> Normally a handler is where information from the message would be saved it to a database, or used to call a web service, or some other business function, but in this example, it is simply logged to show that the message was received.

This is a very complex sentence, how about: Handlers are where the message is processed, very often that involved storing information from the message into a database, calling a web service or calling some business function. In this example the message is logged, so the information that the message was received will be visible in the Console window.

>  This won't actually send any messages, because there aren't any subscribers yet.

There are no subscribers... So what does it have to do with not sending messages exactly? (Let's assume I don't know how pub sub works) I'd skip that sentence and just leave "Next, the handler publishes a new OrderPlaced message. The next step is to create a subscriber for this message." and replace "message" with "event".

> Like the client, the Server application is going to need to be configured as an NServiceBus endpoint as well.
> Like the client and the server, the Subscriber application also needs to be configured as an NServiceBus endpoint.

We never explain what endpoint is, and why do we need it. Would it make sense to add note with a brief hosting explanation? Or we could stick to "initalizing NSB" as we say in the "Client" section, but consistency would help here.

> Next, create a message handler for the OrderPlaced event. Note that whether handling a command or an event, the syntax for a message handler remains the same.

> Create a new class in the Subscriber project named OrderCreatedHandler using the following code:

Again, having two separate paragraphs with lots of repetition is confusing and hard to read.

> The fact that these disparate tasks can be accomplished in completely separate message handlers instead of one monolithic process is one of the many strengths of the Publish/Subscribe messaging pattern.

I don't get that. What do I gain exactly by making it separate?

>  In a real system, a subscriber to OrderPlaced could charge a credit card for the order, start to prepare the order for shipment, or even update information in a customer loyalty database to track when the customer is eligible for different purchase-driven incentives.

That's largely a repetition from the previous section on what handler could do. Could we put it in a single box (e.g. using NOTE:) with explanation? I think having that visual difference between "follow those steps" and "more background/explanation" section would help a lot here.

> Next, the message publisher must be notified that the subscriber is interested in being notified when OrderPlaced events are published

Next, the message publisher needs to know that the subscriber is interested in (receiving) OrderPlaced events. or Next, the subscriber needs to inform the publisher that it wants to receive OrderPlaced events.

> The add directive says that for messages of type OrderPlaced within the Shared assembly, subscription requests should be sent to the Samples.StepByStep.Server endpoint, which is the endpoint that is responsible for publishing OrderPlaced events.

Break into shorter sentences.

> When the Subscriber endpoint initializes, it will read this configuration, and knowing that it contains a message handler for OrderPlaced, it will send a special subscription message to the Samples.StepByStep.Server endpoint

This one too.

> When that endpoint receives the subscription request, it will store it locally (in this sample, using in-memory storage, but in a production system a database would be used) so that when it publishes a message, it can send a copy to every subscriber that expressed interest.

And this.

> (in this sample, using in-memory storage, but in a production system a database would be used)

This could be moved outside the sentence (e.g. as a follow up sentence).

> Running the sample

Nowhere we mention pre-requisites. I know we have that mentioned in other places, but given that is a step-by-step-lets-start-here-guide I think it's worth a repetition. The exception I got wasn't very descriptive.

weralabaj:
Added a few comments, I'd like to run through this once again after the changes.

DavidBoike:
@weralabaj:

> > In each project, add a reference to the NServiceBus NuGet package.

> What version? I think that's important, since there're significant differences in APIs across versions and soon v6 will be the default in NuGet.

I don't want to add a specific version (like 5.2.14) because that text would be impossible to maintain. What I mean to say is "the current stable version" which at this moment would be 5.2.14. For the V6 launch I would want to have a PR prepped that would upgrade the entire thing to V6, and we'd merge it shortly after launch. Also, the sample infrastructure will pump out the NServiceBus version at the very top.

> Since we cover it right away, wouldn't it be easier to say "add clientinit code and also sendorder code" and then explain what it does? Or even just do it in a single snippet?

I'm a little worried that would be a lot to digest in one snippet.

> That (ConfigErrorQueue) could be moved to the section when we create messages. It's not really related to the client directly.

I view it as something related to setting up an endpoint, and so I think it has to be covered on the first endpoint, even though you're right you could refactor it to "This affects all so let's just do it first." I think that would leave the reader wondering, OK, why? What is this?

> There are no subscribers... So what does it have to do with not sending messages exactly? (Let's assume I don't know how pub sub works) I'd skip that sentence and just leave "Next, the handler publishes a new OrderPlaced message. The next step is to create a subscriber for this message." and replace "message" with "event".

I think it's an important concept that publishers can have zero subscribers. I reworked into a NOTE though.

> That's largely a repetition from the previous section on what handler could do. Could we put it in a single box (e.g. using NOTE:) with explanation? I think having that visual difference between "follow those steps" and "more background/explanation" section would help a lot here.

I didn't see the repetition, but I had already started to change those paragraphs so maybe I already edited it out. I don't want to overuse NOTE because it will already be broken up quite a bit by snippets, and I don't want the 1-line directions lost as the only things not in some sort of box. Trying to stick to using NOTE for "Here is something extra to think about."

> Nowhere we mention pre-requisites. I know we have that mentioned in other places, but given that is a step-by-step-lets-start-here-guide I think it's worth a repetition. The exception I got wasn't very descriptive

What exception did you get? And for prerequisites, were you talking about http://docs.particular.net/nservicebus/msmq/#nservicebus-configuration for MSMQ setup? Or .NET / Visual Studio versions?

DavidBoike:
What's the best canonical URL for a prerequisites section, for having MSMQ properly installed and configured?

We've got these:

* http://docs.particular.net/nservicebus/msmq/#nservicebus-configuration
* http://docs.particular.net/nservicebus/operations/running-on-windows

weralabaj:
> I don't want to add a specific version (like 5.2.14) because that text would be impossible to maintain. 

Fair enough, since I didn't have that rendered I forgot that'll be shown on the snippets. I hope that will be good enough.

> Exceptions

That was about NSB pre-requisites. Wouldn't be the easiest way to just use platform installer?
I'd be happy if we just said that - download PI, tick NSB box and click, click, click. Done.

DavidBoike:
So NSB 5.x requires .NET 4.5, and the sample downloadable code is .NET 4.5, but the Platform Installer (which we would strongly suggest they use to get MSMQ set up) requires 4.5.2. So in the requirements section, which should I list?

DavidBoike:
@weralabaj: Addressed your comments here: https://github.com/Particular/DeveloperEducation/issues/551#issuecomment-220612752

I think usually @SimonCropp leaves off namespaces to avoid unnecessary indentation and using statements, but I agree that in this context, where we're asking them to do things step by step, that it's going to be really confusing when they use default project conventions and it fails. So, I added namespaces to the classes **only** in the Shared project.

After I got done implementing that and pushed, I turned around and added a NOTE in the last commit to make sure the reader doesn't miss that, or if they try to do it with different conventions (CompanyName.Shared or something).

PR #1531, opened 2016-05-16
[WIP] Adding IIS reverse proxy rule 

WojcikMike:
This change in doco involves writing down a IIS Reverse Proxy rule that was created by @rlaveycal in the following [Service Pulse bug](https://github.com/Particular/ServicePulse/issues/343#issuecomment-219088021). This section will disappear when the SignalR will be fixed to properly work with reverse proxy hosting configuration.

I have questions/doubts regarding:
 * if we should state that this configuration is due to a bug in another library. 
 * if we should describe a bit more what was the problem and what this solution actually does (however for customers that didn't encounter this error it is just not important)
 * I struggle here how to name this situation, currently I use 2 terms:
  * non-root directory
  * sub directory
To refer to the same thing. And I still have doubts if customer would understand when this section refers to them, so any suggestions would be apreciated.

Connects to: Particular/ServicePulse#343

### DIFF
--- a/servicepulse/install-servicepulse-in-iis.md
+++ b/servicepulse/install-servicepulse-in-iis.md
@@ -102,7 +102,7 @@ It is also recommended that the IIS web site be configured to use SSL if an auth
 
 ### Configuring Reverse Proxy in a non-root directory
 
-Due to a bug in SignalR (as at Microsoft.AspNet.SignalR.JS version 2.2.0), usage of IIS as a reverse proxy in a virtual directory requires an additional URL Rewrite Rule on the `/api/` sub folder. This rule makes sure that SignalR uses the correct path when hosted within a virtual directory. This rule should look as follows: 
+Due to a [bug in SignalR](https://github.com/SignalR/SignalR/issues/3649) in Microsoft.AspNet.SignalR.JS version 2.2.0, usage of IIS as a reverse proxy in a virtual directory requires an additional URL Rewrite Rule on the `/api/` sub folder. This rule makes sure that SignalR uses the correct path when hosted within a virtual directory. This rule should look as follows: 
 
 ```xml
 <?xml version="1.0" encoding="UTF-8"?>
--- a/servicepulse/install-servicepulse-in-iis.md
+++ b/servicepulse/install-servicepulse-in-iis.md
@@ -100,6 +100,28 @@ WARNING: The default configuration for ServiceControl only allows access to REST
 
 It is also recommended that the IIS web site be configured to use SSL if an authorization provider is used.
 
+### Configuring Reverse Proxy in a non-root directory
+
+Currently due to the fact of a bug in JQuery SignalR JS library usage of reverse proxy in a non-root directory requires additional URL Rewrite Rule on api sub folder. This rule make sure that SignalR uses correct path when being hosted in a sub directory. This newly rule should look like the following: 
+
+```xml
+</rules>
+<outboundRules>
+    <rule name="Update Url property" preCondition="JSON" enabled="true" stopProcessing="true">
+        <match filterByTags="None" pattern="\&quot;Url\&quot;:\&quot;(.+?)\&quot;" />
+        <conditions>
+            <add input="{URL}" pattern="(.*)/api/" />
+        </conditions>
+        <action type="Rewrite" value="&quot;Url&quot;:&quot;{C:1}{R:1}&quot;" />
+    </rule>
+    <preConditions>
+        <preCondition name="JSON">
+            <add input="{URL}" pattern="/api/messagestream/negotiate" />
+            <add input="{RESPONSE_CONTENT_TYPE}" pattern="application/json" />
+        </preCondition>
+    </preConditions>
+</outboundRules>
+``` 
 
 ### Limitations
 
--- a/servicepulse/install-servicepulse-in-iis.md
+++ b/servicepulse/install-servicepulse-in-iis.md
@@ -105,22 +105,28 @@ It is also recommended that the IIS web site be configured to use SSL if an auth
 Due to a bug in SignalR (as at Microsoft.AspNet.SignalR.JS version 2.2.0), usage of IIS as a reverse proxy in a virtual directory requires an additional URL Rewrite Rule on the `/api/` sub folder. This rule makes sure that SignalR uses the correct path when hosted within a virtual directory. This rule should look as follows: 
 
 ```xml
-</rules>
-<outboundRules>
-    <rule name="Update Url property" preCondition="JSON" enabled="true" stopProcessing="true">
-        <match filterByTags="None" pattern="\&quot;Url\&quot;:\&quot;(.+?)\&quot;" />
-        <conditions>
-            <add input="{URL}" pattern="(.*)/api/" />
-        </conditions>
-        <action type="Rewrite" value="&quot;Url&quot;:&quot;{C:1}{R:1}&quot;" />
-    </rule>
-    <preConditions>
-        <preCondition name="JSON">
-            <add input="{URL}" pattern="/api/messagestream/negotiate" />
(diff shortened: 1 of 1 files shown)


servicepulse/install-servicepulse-in-iis.md
```diff
@@ -100,6 +100,28 @@ WARNING: The default configuration for ServiceControl only allows access to REST
 
 It is also recommended that the IIS web site be configured to use SSL if an authorization provider is used.
 
+### Configuring Reverse Proxy in a non-root directory
+
+Currently due to the fact of a bug in JQuery SignalR JS library usage of reverse proxy in a non-root directory requires additional URL Rewrite Rule on api sub folder. This rule make sure that SignalR uses correct path when being hosted in a sub directory. This newly rule should look like the following:
```
WilliamBZA: It's now `JQuery SignalR JS library`, it's just `SignalR`

WilliamBZA: > Currently due to the fact of a bug in JQuery SignalR JS library usage of reverse proxy in a non-root directory requires additional URL Rewrite Rule on api sub folder.

Should be:

> Due to a bug in SignalR (as at SignalR version 2.2.0), usage of reverse proxy in any IIS folder other than the root folder requires an additional URL Rewrite Rule on api sub folder.  This rule makes sure that SignalR uses the correct path when hosted within a sub directory. This rule should look as follows:

WilliamBZA: More tweaks (changing from root and folders and dirs)

> Due to a bug in SignalR (as at Microsoft.AspNet.SignalR.JS version 2.2.0), usage of IIS as a reverse proxy in a virtual directory requires an additional URL Rewrite Rule on the `/api/` sub folder. This rule makes sure that SignalR uses the correct path when hosted within a virtual directory. This rule should look as follows:

servicepulse/install-servicepulse-in-iis.md
```diff
@@ -100,6 +100,28 @@ WARNING: The default configuration for ServiceControl only allows access to REST
 
 It is also recommended that the IIS web site be configured to use SSL if an authorization provider is used.
 
+### Configuring Reverse Proxy in a non-root directory
+
+Due to a bug in SignalR (as at Microsoft.AspNet.SignalR.JS version 2.2.0), usage of IIS as a reverse proxy in a virtual directory requires an additional URL Rewrite Rule on the `/api/` sub folder. This rule makes sure that SignalR uses the correct path when hosted within a virtual directory. This rule should look as follows:
```
indualagarsamy: @WojcikMike - Is Configuring Reverse Proxy for SP a part of our documentation? In other words, are we documenting what needs to be part of MS doco?

indualagarsamy:
@WojcikMike @WilliamBZA - I feel like configuring reverse proxy / signal R bug is more MS doco than anything related to Particular doco? While it makes sense for the details to be captured in the issue, I am not sure that this belongs in doco. @Particular/docs-maintainers - thoughts?

SimonCropp:
@johnsimons @WojcikMike is https://github.com/Particular/ServicePulse/issues/343 going to be fixed? and if it is will this doco still be required?

johnsimons:
According to https://github.com/Particular/ServicePulse/issues/343#issuecomment-219423715 yes it is, but no eta hence this for now

SimonCropp:
>  yes it is, but no eta hence this for now

in that case this should not be documented. the issue should be discoverable via google. and have adequate information on how to work around. if it expected to impact many users than this should be justification for it to be fixed, not for a bug to be documented.

WojcikMike:
@SimonCropp regarding fixing I sure hope we will be able to fix it. However as it require change in SignalR JS lib I don't know if and when it will happen. I poked with the library and created a fix but I need ot first talk with them as they discourage pushing PR's out of the blue ;) So I am trying to follow their process.

In that case should we kill this PR?

servicepulse/install-servicepulse-in-iis.md
SimonCropp: link to the signalr bug

link to more info on URL Rewrite Rule

link to our bug

servicepulse/install-servicepulse-in-iis.md
```diff
@@ -100,6 +100,28 @@ WARNING: The default configuration for ServiceControl only allows access to REST
 
 It is also recommended that the IIS web site be configured to use SSL if an authorization provider is used.
 
+### Configuring Reverse Proxy in a non-root directory
+
+Due to a bug in SignalR (as at Microsoft.AspNet.SignalR.JS version 2.2.0), usage of IIS as a reverse proxy in a virtual directory requires an additional URL Rewrite Rule on the `/api/` sub folder. This rule makes sure that SignalR uses the correct path when hosted within a virtual directory. This rule should look as follows: 
+
+```xml
+</rules>
```
SimonCropp: why the hanging rule?

servicepulse/install-servicepulse-in-iis.md
johnsimons: @SimonCropp the issue is https://github.com/SignalR/SignalR/issues/3287, do you want to have a look and see how to best include the link in the text ?

servicepulse/install-servicepulse-in-iis.md
johnsimons: Fixed

servicepulse/install-servicepulse-in-iis.md
johnsimons: Actually probably this one is more relevant to this - https://github.com/SignalR/SignalR/issues/3649

WilliamBZA:
I vote kill. Including this type of doco makes us responsible for maintaining the doco and blurs the lines of where the actual issue is.

mikeminutillo:
How Pulse/Control specific is this solution? Can we add it to the SignalR bug and turn this into a Warning note with a link over there?

WilliamBZA:
@mikeminutillo It's not specific to us at all, we just happened to have been the catalyst that found it. It's 100% specific to SignalR.

:+1: to adding to the SignalR issue and closing this.

As @SimonCropp says, the original issue in ServicePulse that spawned this has the workaround and is easily discoverable by google.

SimonCropp:
@WojcikMike @johnsimons can you please respond to the comments by  @mikeminutillo and  @WilliamBZA

johnsimons:
@mikeminutillo and @WilliamBZA, unfortunately from a user stance, the bug is in our software, it just happens that we know that the issue lives in a third party lib we use but that is beside the point.
A customer that purchased our software is being affected by this issue and we need to help them as much as possible, I don't see the big issue in documenting this workaround (this is already a burden on our users) in our own documentation till we either get rid of SignalR or patch it.

WojcikMike:
I am torn between what @mikeminutillo and @WilliamBZA are suggesting and @johnsimons comment above. Currently in our doco we describe how to set up ServicePulse on IIS, and we include description of reverse proxy configuration with a xml snippet of rule that is created by IIS. 

At this point I personaly think we crosed the line of describing in to much detail how to configure IIS. Because we have that description customers who follow it and encounter software bug can reasonably perceive it as our fault. And to address it i wanted to augument our doco to show how to fix this problem. 

I am not sure what is good solution in this situation. However I believe that if we have detailed reverse proxy configuration in our doco we should add my change. Or we should remove IIS reverse proxy from our doco completely linking to MS documentation.

rlaveycal:
As a client I think the info around ARR should remain and be updated. The primary driver for wanting to host service pulse in IIS is to provide access over port 80. As Pulse itself is entirely in the browser that shifts the problem to connecting to Service Control over port 80, hence the need for ARR. I found it very useful to have the full story presented in your docs rather than "and now go a use ARR to set up a reverse proxy...good luck"

indualagarsamy:
After discussing this @SimonCropp, I will merge this PR, after I add the link to the signalR bug.

PR #1548, opened 2016-05-23
DataBus does not work with callbacks

ramonsmits:
Based on https://groups.google.com/forum/#!msg/particularsoftware/j5fKCILkD3o/qchIhaQWJAAJ

### DIFF
--- a/nservicebus/messaging/databus.md
+++ b/nservicebus/messaging/databus.md
@@ -22,7 +22,8 @@ The `DataBus` approach is to store a large payload in a location that both the s
 
 The FileShare DataBus also [leverage both serialization and headers](/nservicebus/messaging/headers.md#fileshare-databus-headers) to provide its functionality.
 
-NOTE: The DataBus does not work with callbacks. Please use regular handlers instead.
+NOTE: The DataBus does not support [callbacks](/nservicebus/messaging/handling-responses-on-the-client-side.md), and will not rehydrate `DataBusProperty<T>` properties in a callback message. Callback registrations are stored in-memory, and as a result are not durable and may never arrive if, for example, the server restarts before the callback arrives. Therefore, callbacks are not reliable for transporting large amounts of data using the DataBus. If an application requires a response message containing DataBus data, it should implement a normal [message handler](/nservicebus/handlers/) which will be handled reliably.
+
 
 ## Enabling the DataBus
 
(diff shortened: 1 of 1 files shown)


SimonCropp:
@Particular/nservicebus-maintainers should we throw an exception in this scenario

timbussmann:
> should we throw an exception in this scenario

not against that.

andreasohlund:
:+1: to throw

danielmarbach:
Can we lay out the why it doesn't work and why we don't want it to work?

And then when we agree on the reasoning implement the exception.

DavidBoike:
How's this (added commit) for rationale on no databus in callbacks?

Although I think, perhaps, it should be moved to further down the page. "Hey this doesn't work with callbacks" is probably not the first thing we should be saying on the topic.

indualagarsamy:
If we are going to throw an exception, do we need to document this as "Callbacks are not supported"? This seems more like a bug in the core that we should be fixing to me than something we should be documenting.

mikeminutillo:
> If we are going to throw an exception, do we need to document this as "Callbacks are not supported"? This seems more like a bug in the core that we should be fixing to me than something we should be documenting.

:+1: We used to document other feature combos that didn't work (i.e. No Transactions with SLR) but we don't seem to do that anymore. Is that because we throw at runtime when we detect invalid combinations?

> Can we lay out the why it doesn't work and why we don't want it to work?

Is "why it doesn't work" just because of behavior ordering? 

If the [DataBus Receive Behavior](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/DataBus/DataBusReceiveBehavior.cs) happens before the [Callbacks Receive Behavior](https://github.com/Particular/NServiceBus.Callbacks/blob/develop/src/NServiceBus.Callbacks/Reply/RequestResponseInvocationForMessagesBehavior.cs) wouldn't this work? They are both `Behavior<IIncomingLogicalMessageContext>` so the eventual order in the pipeline isn't deterministic is it?

danielmarbach:
I think it would be pretty weak if the reasoning behind it is: technical reasons. We either have strong arguments why we don't want to support it or we don't. if we don't have strong arguments then I would expect it to work

DavidBoike:
So do we want to try to support it then?

timbussmann:
I agree with @danielmarbach , this should be handled as a bug. Let's close this PR then?

danielmarbach:
@ramonsmits @DavidBoike I talked to @andreasohlund and we agreed it is a bug. Can you investigate the bug and see how much it takes to fix it? It might be an ordering issue of behaviors only.

(judging by the comments from @timbussmann he is on the same page as well)

danielmarbach:
@SimonCropp let's close this PR

danielmarbach:
Opened up a new issue https://github.com/Particular/NServiceBus.Callbacks/issues/29

PR #1549, opened 2016-05-24
[WIP] Detect Core Dependencies

mikeminutillo:
LINQPad Script to detect the core dependency of each NuGet package


weralabaj:
@mikeminutillo is that done or waiting for anything?

mikeminutillo:
@weralabaj Yeah. I need to sync up with @SimonCropp about persistence options so that other tooling can use the data to make decisions.

SimonCropp:
ok merging this. we can evolve it on the main branch since it is a stand alone script

PR #1551, opened 2016-05-24
Saga correlation

timbussmann:
If I understood @DavidBoike correctly, this line is invalid and users shouldn't set the correlation property manually?

### DIFF
--- a/Snippets/Core/Core_5/Core_5.csproj
+++ b/Snippets/Core/Core_5/Core_5.csproj
@@ -321,6 +321,7 @@
     <Compile Include="UpgradeGuides\4to5\Scheduler.cs" />
     <Compile Include="UpgradeGuides\4to5\Upgrade.cs" />
     <Compile Include="UpgradeGuides\5to6\HandlerOrdering.cs" />
+    <Compile Include="UpgradeGuides\5to6\Saga.cs" />
     <Compile Include="UpgradeGuides\5to6\StartupInterface.cs" />
     <Compile Include="UpgradeGuides\5to6\Upgrade.cs" />
     <Compile Include="UpgradeGuides\5to6\UpgradeMessageHandler.cs" />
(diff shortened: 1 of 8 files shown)


DavidBoike:
This raises the bigger question of how to handle this change to sagas in V6? Do we just silently take these mappings out of the V6 samples? Or do we leave comments in the V6 version that the thing that was necessary in all previous versions of NServiceBus no longer is?

timbussmann:
seems like we should add some more information to the upgrade guide instead of just "correlation property is read only" and a note/warning on the saga docs page?

indualagarsamy:
@DavidBoike - to your question. Does the current upgrade guide point the difference? If not it should.

DavidBoike:
Added a fix for the upgrade guide to this PR, but that doesn't really answer my question.

Put yourself in the seat of a user looking at the Sagas page, looking at these two snippets:

    // V3 - V5
    public void Handle(StartOrder message)
    {
        Data.OrderId = message.OrderId;
    }

    // V6
    public async Task Handle(StartOrder message, IMessageHandlerContext context)
    {
    }

Are you going to think, "Oh they removed that requirement in V6" or are you going to think, "Oh, the V6 snippet looks broken"? My guess is the latter.

So should we do this?

    // V6
    public async Task Handle(StartOrder message, IMessageHandlerContext context)
    {
        // Data.OrderId automatically set to value of message.OrderId
    }

I don't think that's something we should do _forever_ (Or maybe we should? You certainly can't tell that just by looking at text.) but it may be a useful educational device for the short term.

DavidBoike:
Also: I collapsed all the parts of the upgrade guide about "correlated properties" into one section, since they're all related, and added information about required mappings to correspond to https://github.com/Particular/NServiceBus/pull/3756.

indualagarsamy:
@DavidBoike - I think having snippets in the upgrade guide is a good idea.

To your question, I think it's weird to have just `Data.OrderId = message.OrderId`in the handler. Maybe a comment will clarify it better. So, in the v6 snippet, we can tweak that comment to clarify it better. 

```c#
// V3 - V5
public void Handle(StartOrder message)
{
  Data.OrderId = message.OrderId;
  // Other processing logic for the StartOrder message
}

// V6
public async Task Handle(StartOrder message, IMessageHandlerContext context)
{
  // Just the processing logic for the StartOrder message. 
}

```

If this change is captured well in the upgrade guide, I don't see why we can't go and remove the correlation lines from all the saga samples for v6. 

Does this answer your question?

PR #1554, opened 2016-05-25
Clarify when to use custom saga finder

weralabaj:
Inspired by a customer that used both mechanisms at the same time, and used custom finder even though they needed only simple property mapping. The doco could be more explicit and mention the main article (so if somebody lands on the custom finder page only, they can go to the main page for full context).

@Particular/docs-maintainers please, see if that's clear enough?

### DIFF
--- a/nservicebus/sagas/saga-finding.md
+++ b/nservicebus/sagas/saga-finding.md
@@ -11,7 +11,7 @@ related:
 - nservicebus/sagas/concurrency
 ---
 
-As described in [Correlating messages to a saga](/nservicebus/sagas/#correlating-messages-to-a-saga), every saga definition needs to specify how the existing saga instance should be found for an incoming message. The mechanisms described in that section should be the default choice and are suitable for many real-life scenarios. However, sometimes the mapping logic is too complex and it's necessary to define a custom saga finder.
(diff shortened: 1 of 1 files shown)


nservicebus/sagas/saga-finding.md
```diff
@@ -11,7 +11,7 @@ related:
 - nservicebus/sagas/concurrency
 ---
 
-Sometimes a saga handles certain message types without a single simple property that can be mapped to a specific saga instance. In those cases, finer-grained control of how to find a saga instance will be required.
+As described in [Correlating messages to a saga](/nservicebus/sagas/#correlating-messages-to-a-saga), every saga definition needs to specify how the existing saga instance should be found for an incoming message. The mechanisms described in that section should be the default choice and are suitable for many real-life scenarios. However, sometimes the mapping logic is too complex and it's necessary to define a custom saga finder.
```
andreasohlund: Not every incoming message requires a mapping, only the ones starting a saga require one. For others its optional since it may be a response to a request made the saga instance and those gets auto correlated. Can we reword to reflect this?

indualagarsamy:
@andreasohlund - I modified the intro more to why a CustomSagaFinder is required. I ran the new intro with @SimonCropp to polish it up. But any details on how default correlation works should be on the main saga article. Can you review that and open a PR it it doesn't capture it appropriately?

weralabaj:
@indualagarsamy @SimonCropp that PR shouldn't get merged. The point of the change was to say explicitly "go with the simple mapping by default, use this one ONLY if that's not enough". We kind of indicate that by saying "in some complex scenarios, one property in the message may not be sufficient to resolve a saga instance", but it's easy to to not catch those nuances when just skimming through the doco.
In my view after the changes it's no different from what we had before, only more text (link text doesn't indicate "this is the default you should use, check if that might be enough").

PR #1555, opened 2016-05-26
[WIP] Update rabbit client

bording:
This updates samples and snippets to the 3.6.2 release of the RabbitMQ client and the new 3.3.1 release of the RabbitMQ transport.

I've marked this as WIP for now because there is still a change in the Rabbit_All snippets project that needs to be made. The `ErrorQueue` class is still using the deprecated `QueueingBasicConsumer` class, and I'm not immediately sure how to remove it.

https://github.com/Particular/docs.particular.net/blob/master/Snippets/Rabbit/Rabbit_All/ErrorQueue.cs#L34

This could be switched to the recommended `EventingBasicConsumer`, but then some way to wait for the event handler firing would be needed. I could use a `TaskCompletionSource` if we don't mind introducing async here.

The other thing I'm curious about is the `query` used in the `BasicConsume` call. https://github.com/Particular/docs.particular.net/blob/master/Snippets/Rabbit/Rabbit_All/ErrorQueue.cs#L35

From the usage, it seems to imply that you can use that as a filter to only return the specified message id, but I can find absolutely zero documentation around this assumption, so I'm not even sure that it works.

### DIFF
--- a/Snippets/Rabbit/Rabbit_1/Rabbit_1.csproj
+++ b/Snippets/Rabbit/Rabbit_1/Rabbit_1.csproj
@@ -36,8 +36,8 @@
       <HintPath>..\..\..\packages\NServiceBus.RabbitMQ.1.1.5\lib\net40\NServiceBus.Transports.RabbitMQ.dll</HintPath>
       <Private>True</Private>
     </Reference>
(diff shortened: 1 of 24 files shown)


SimonCropp:
> From the usage, it seems to imply that you can use that as a filter to only return the specified message id, but I can find absolutely zero documentation around this assumption, so I'm not even sure that it works.

turns out it doesnt. hows this https://github.com/Particular/docs.particular.net/blob/master/Snippets/Rabbit/Rabbit_All/ErrorQueue.cs

and this PR has been manually merged. so closing

PR #1562, opened 2016-05-27
Rabbit client update tweaks

bording:
This includes a few tweaks, including a couple items that might have been missed from #1555.

@SimonCropp I took a look at the ErrorQueue stuff, and I came up with an approach that I think is a bit simpler. Let me know what you think! One other thing of note there, I changed the `BasicAck` call to use multiple = false, so that only the retried message is removed from the queue instead of every message up to and including the retried message.

One additional thing that hasn't been addressed here: The queue creation and deletion pieces are making some assumptions that aren't true for core v6 anymore. There is no more retries queue, and the callbacks queue won't always be `endpoint-machinename`, since it's now based on the user-supplied endpoint instance discriminator. Does this mean we need to split this out per version?


SimonCropp:
@bording can u rebase this?

bording:
> @bording can u rebase this?

Done!

SimonCropp:
> One additional thing that hasn't been addressed here: The queue creation and deletion pieces are making some assumptions that aren't true for core v6 anymore. There is no more retries queue, and the callbacks queue won't always be endpoint-machinename, since it's now based on the user-supplied endpoint instance discriminator. Does this mean we need to split this out per version?

Yes. do u want to work on that ?

bording:
Sure. Do you have any thoughts on what the best approach for that would be since those are all currently in the "all" project targeting core v6, and have tests?

PR #1573, opened 2016-06-01
Small sender-side distribution sample fixes

DavidBoike:
* Corrected spelling: UnawareCilent to UnawareClient
* InstanceIds of `1` and `2` are awful when presented as text, you have no idea what that means. Replaced with `instance1` and `instance2`


DavidBoike:
@Particular/docs-maintainers please review (should be quick)

PR #1574, opened 2016-06-01
Sender side distribution sample bug

DavidBoike:
Fix bug in sender-side distribution sample, add SequenceId to the message for clarity, so you can see the even and odd ones segregated when using the round-robin routing in the AwareClient.


DavidBoike:

The sample was not working correctly before because the routing XML file did not contain the correct endpoint name. Everything else is just for clarity.

In another PR I will probably tackle the sample MD text itself, as it's not really clear what's going on at the moment.

PR #1577, opened 2016-06-02
Update dependsOn documentation

timbussmann:
relates to https://github.com/Particular/NServiceBus/pull/3774

there was no statement about not removing the feature suffix in v6

@SimonCropp ping

### DIFF
--- a/Snippets/Core/Core_5/Features/Features.cs
+++ b/Snippets/Core/Core_5/Features/Features.cs
@@ -12,7 +12,7 @@ namespace Core5.Extending
     }
     #endregion
 
-    class DependencyAFeature : Feature
+    class ComponentBFeature : Feature
     {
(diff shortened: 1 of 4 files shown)


SimonCropp:
does this need to be touched https://github.com/Particular/docs.particular.net/blob/master/Snippets/Core/Core_6/Features/Features.cs#L28  ?

timbussmann:
not sure, since there is no real "DependencyB" feature declared in code. It would probably be more obvious if the feature is named "DependencyBFeature" so you can see how the suffix got removed from v5 to v6, but that just reads terrible?

SimonCropp:
why does it read terrible?

timbussmann:
IdK, I think `DependencyBFeature` reads weird. `DependencyB` or `FeatureB` read better. But `DependencyBFeature` would definitely help to point out the changes though

Issue #1347, opened 2016-03-23
Guidance content on security in SQL Server transport


weralabaj:
@ramonsmits could you list what you had in mind specifically?

SimonCropp:
@weralabaj @ramonsmits this does not have enough context to move forward. either give some more details or close the issue

ramonsmits:
Instead of relying on message encryption a customer can also use connection, table, database encryption.

Basically, differentiate between:

- Message
- Transport
- Storage

Together with some links to the relevant MSDN guidance.

ramonsmits:
Confused with a different part. This was probably about authorization on tables.

Mention that there is a difference between reads (receiver/sender) / writes (sender).

Endpoint A should R/W to its queues, but not R from others. How to set this up.

Mention that all endpoints only need write access to AUDIT and ERROR queues.

SimonCropp:
@ramonsmits can u submit a draft PR that include your above notes

PR #1579, opened 2016-06-06
SQL transport deployment guidance - deprecating multi-instance 

SzymonPobiega:
Connects to: https://github.com/Particular/V6Launch/issues/67


nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,85 @@
...
+## How it works
+
+The SQL Server transport is a hybrid queueing system which is neither store-and-forward (like MSMQ) nor a broker (like RabbitMQ). It treats the SQL Server as a storage infrastructure for the queues while the queue implementation logic is executed inside the client process.
+
+
+### Advantages of SQL Server Transport
+
+ * No additional licensing and training costs, as majority of Microsoft stack organizations already have SQL Server installed and have knowledge required to run it
+ * Great general-purpose tooling (SSMS)
+ * Throughput on par with MSMQ
```
johnsimons: :-1: to mention performance stats, it all depends as we know it

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,85 @@
...
+
+The queue tables are hosted in the same catalog as business data and the NServiceBus runtime can be hosted in the web worker. In some cases there is a need for a separate process for NServiceBus. In most cases these systems consists of 1 to 3 non-scaled out endpoints. Because they consist of a single logical service or bounded context, there is usually no need to create separate schemas for the queues.
+
+### Small
+
+The SQL Server transport is a good choice for a pilot project to prove feasibility of NServiceBus in a given organization as well as for a small, well-defined green field application. It usually requires nothing more than a database. Such a system usually consists of 2-10 fairly independent non-scaled out endpoints.
+
+The best option is to store the queues in the same catalog as the business data. Schemas can be used to make maintenance easier and allow for finer-grained security.
```
johnsimons: :-1: to recommend this, I think separate database for queues vs business data (including nsb tables) should be the way to go regardless of size, this way a user can scale up and there is no need to change it.

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,85 @@
...
+
+### Small
+
+The SQL Server transport is a good choice for a pilot project to prove feasibility of NServiceBus in a given organization as well as for a small, well-defined green field application. It usually requires nothing more than a database. Such a system usually consists of 2-10 fairly independent non-scaled out endpoints.
+
+The best option is to store the queues in the same catalog as the business data. Schemas can be used to make maintenance easier and allow for finer-grained security. 
+
+#### Security
+
+For best results, both in terms of security and maintainability, each endpoint should run in context of a different account. Each such account should own three schemas on the database level
```
johnsimons: This is interesting advice, but as a user how do I setup all of this ? The current creation of tables does not handle this

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,85 @@
...
+
+The best option is to have a dedicated catalog for the queues. This approach allows to use different scaling up strategies for the queue catalog. It also allows to use a dedicated back up policy for the queues (because data in queues has much more ephemeral). Unfortunately it leaves a problem of how to ensure that data modifications and queue operations are atomic and isolated from each other. 
+
+One option is to use Microsoft Distributed Transaction Coordinator and distributed transactions (NServiceBus with SQL Server transport is DTC-enabled by default). In this mode the message receive transaction and all data modifications that result from processing the message are part of one distributed transaction that spans two SQL catalogs.
+
+Another options is to use Outbox, which is a unique feature of NServiceBus, that provides exactly-once processing semantics over an at-least-once infrastructure. In this mode each individual database has to include the outbox table. Outgoing messages are stored in that table in the same transaction as the business data is modified and, only after that transaction commits successfully, they are dispatched to the destination. 
+
+#### Scaling up
```
johnsimons: i would only add this section if we actually have tested it and have seen improvements

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,85 @@
...
+
+**TODO**
+
+ * Separate file groups for hot queues?
+ * In mem OLTP? 
+
+#### Security
+
+For best results both in terms of security and maintainability each endpoint should run in context of a different account. Each such account should own two schemas in the transport catalog
+ * the incoming queue schema (e.g. `sales-incoming`) for which it has exclusive `DELETE` access. Other accounts should have `INSERT` permissions to this schema
```
johnsimons: "incoming queue schema" ?
'private queue schema" ?

Sorry I don't get it

nservicebus/transports/guidance.md
mikeminutillo: But doesn't this prevent the customer from using the outbox?

johnsimons: No, this is exactly how Rabbit works.

nservicebus/transports/guidance.md
SzymonPobiega: This is a placeholder. I think we should evolve the guidance in this direction (to include it) but I think we can release it initially without it.

nservicebus/transports/guidance.md
SzymonPobiega: Not sure about that. The difference is, nobody else needs to access private (satellite) queues. So by putting them in a separate schema we can make it more secure. But not sure if it is worth it.

nservicebus/transports/guidance.md
SzymonPobiega: @johnsimons so you say that we should discourage people from sharing connection between NH persistence and the transport and always use the outbox? It makes sense to me, I just want to make sure we are all on same page.

SzymonPobiega:
@tmasternak @MarcinHoppe please take a look.

nservicebus/transports/guidance.md
SzymonPobiega: For now we can advice users to at least specify the queue schema when configuring the transport. The private queue schema can be added later when the guidance and implementation matures.

nservicebus/transports/guidance.md
johnsimons: Not really, IMHO I think Outbox is too restrictive the way it works, it would mean all business data needs to live in a single database which in practice is never true.
So I am saying that unfortunately DTC is the way to go.

nservicebus/transports/guidance.md
johnsimons: I don't think is worth it

nservicebus/transports/guidance.md
SzymonPobiega: So you're saying even for small **single database** scenarios we should tell the users to use DTC and not the connection sharing (nor outbox)? Why would we say so? I get that connection sharing is problematic from our infrastructure point of view but the outbox seems to be better (faster, nothing to set up).

johnsimons: I guess if a user is using a single database for business data, then they can consider the usage of Outbox. The only issue is that once this system starts to grow and they use more databases they need to move out of it.

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,85 @@
...
+ * Throughput on par with MSMQ
+ * Scaling out of processing is simple because of support for competing consumers at the queue level
+
+## Usage scenarios - deployment size
+
+### Tiny
+
+The SQL Server transport is an ideal choice for extending an existing web application with asynchronous processing capabilities as an alternative for tedious batch jobs that tend to quickly get out of synch with the main codebase. Assuming the application already uses SQL Server as a database, this scenario does not require any additional infrastructure.
+
+The queue tables are hosted in the same catalog as business data and the NServiceBus runtime can be hosted in the web worker. In some cases there is a need for a separate process for NServiceBus. In most cases these systems consists of 1 to 3 non-scaled out endpoints. Because they consist of a single logical service or bounded context, there is usually no need to create separate schemas for the queues.
```
tmasternak: This is not clear for me. How did we know it's 1 to 3 endpoints?

tmasternak:
I like the guidance overall. :+1: 

I think however that we need to add introduction that will include information that's independent of deployment type/size. In my opinion those would be:
 * Making sure users think about endpoint state as a combination of business data, persistence data and transport data and not of MS Sql as a generic container for any data,
 * Express general invariant that transport data for all queues should be on a single instance of Sql Server no matter what deployment type one chooses,
  * Point out that there is possibility for transport data separation on the level of schema and catalog 
 * Point out that due to the fact that one can have business, persistence and transport data all in SQL (case of NHibernate and SqlServer) there are configuration options that are unique to such setup:
  * Local transaction between NHibernate and SqlServer transport
  * Local transaction between business data and NHibernate persister
  * DTC or Outbox for `exactly-once` message processing semantics

Having that in introduction I could see deployment types reference possible configurations mentioned in the introduction.

tmasternak:
And BTW I think we need diagrams :P.

nservicebus/transports/guidance.md
MarcinHoppe: I agree with @johnsimons. Throughput can be similar to MSMQ when SQL Server instance is local to the endpoint and they communicate over shared memory or named pipes transports. When the database is moved to a different node and we start to observe network latency, the performance drops.

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,85 @@
...
+
+
+## How it works
+
+The SQL Server transport is a hybrid queueing system which is neither store-and-forward (like MSMQ) nor a broker (like RabbitMQ). It treats the SQL Server as a storage infrastructure for the queues while the queue implementation logic is executed inside the client process.
+
+
+### Advantages of SQL Server Transport
+
+ * No additional licensing and training costs, as majority of Microsoft stack organizations already have SQL Server installed and have knowledge required to run it
```
MarcinHoppe: Maybe mention SQL Server Express as a free option to start with?

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,85 @@
...
+ * Great general-purpose tooling (SSMS)
+ * Throughput on par with MSMQ
+ * Scaling out of processing is simple because of support for competing consumers at the queue level
```
MarcinHoppe: Maybe explain that the simplicity comes from not having to use any load balancer mechanism like Distributor for MSMQ.

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,85 @@
...
+
+## Usage scenarios - deployment size
```
MarcinHoppe: Is our guidance really dependent on system size? I think customers might find it difficult to tell if their system is tiny, small, or rather big.

Also, they might start small and architect their system in a way that is not fit for growth they may not foresee.

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,85 @@
...
+
+The queue tables are hosted in the same catalog as business data and the NServiceBus runtime can be hosted in the web worker. In some cases there is a need for a separate process for NServiceBus. In most cases these systems consists of 1 to 3 non-scaled out endpoints. Because they consist of a single logical service or bounded context, there is usually no need to create separate schemas for the queues.
+
+### Small
+
+The SQL Server transport is a good choice for a pilot project to prove feasibility of NServiceBus in a given organization as well as for a small, well-defined green field application. It usually requires nothing more than a database. Such a system usually consists of 2-10 fairly independent non-scaled out endpoints.
+
+The best option is to store the queues in the same catalog as the business data. Schemas can be used to make maintenance easier and allow for finer-grained security. 
+
+#### Security
```
MarcinHoppe: There's a separate issue for security guidance (#1347). I suggest we do not mix security with deployment topology right now.

MarcinHoppe:
@SzymonPobiega @tmasternak @johnsimons @indualagarsamy Do you think it would be a good idea to start the document with a TL;DR section that the customer can just glance to get an idea of what the recommended approach is? Later on they could drill down in the rest of the document to understand the rationale behind the guidance.

nservicebus/transports/guidance.md
johnsimons: Exactly my point as explained in https://github.com/Particular/docs.particular.net/pull/1579/files#r65993071

tmasternak:
@SzymonPobiega @johnsimons @mikeminutillo I've added `Deployment considerations` sections that could highlight different aspects that should be taken into account and what are the possible options. In such approach `Tiny`, `Medium` and `Large` could be seen as sample deployments that choose different options based on their needs.

@MarcinHoppe I've added introduction sections that has `Note` which I think does the job of TL;DR.

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,114 @@
...
+ * Business data - data used for implementing business capabilities
+ * Persistence data - infrastructure level data including: saga data, timeout manager state and message driven pub/sub information
+ * Transport data - messaging infrastructure state
+
+SQL Server Transport manages transport data and puts no constraints on the type and configuration of storage technology used for persistence and business data. It can work with any of available persisters i.e. NHibernate and RavenDB, as well as any storage mechanisms used inside message handlers. However, sharing storage technology (SQL Server) between transport, persistence and/or business data offers unique advantages (see next section) and, as a result, this guidance will focus on such scenario.
+
+From data perspective, deployment model chosen by system architect answers the question of how business, persistence and transport data is stored in terms of SQL Server instances, catalogs and schemas.
+
+NOTE: No matter which deployment option is chosen, there is one general rule that should always apply: **All transport data (input, audit and error queues) should reside on a single SQL Server catalog**.
```
SzymonPobiega: Should we say "database" instead of catalog? Do SQL Server community use the term "catalog" these days?

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,114 @@
...
+
+SQL Server Transport manages transport data and puts no constraints on the type and configuration of storage technology used for persistence and business data. It can work with any of available persisters i.e. NHibernate and RavenDB, as well as any storage mechanisms used inside message handlers. However, sharing storage technology (SQL Server) between transport, persistence and/or business data offers unique advantages (see next section) and, as a result, this guidance will focus on such scenario.
+
+From data perspective, deployment model chosen by system architect answers the question of how business, persistence and transport data is stored in terms of SQL Server instances, catalogs and schemas.
+
+NOTE: No matter which deployment option is chosen, there is one general rule that should always apply: **All transport data (input, audit and error queues) should reside on a single SQL Server catalog**.  
+
+### Transactionality (Exactly once message delivery)
+When sharing single technology for different types of data it is possible to provide `exactly-once` message processing without need for DTC. In all other cases it is required to turn on DTC.
```
SzymonPobiega: Well, I don't think I agree. What about Outbox? 
Also, we should be talking about exactly once **processing**, not delivery

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,114 @@
...
+
+#### Business, Persistence and Transport data in single catalog
+Enables `exactly-once` message delivery with native transactions. Multi-catalog with local transactions are possible in principle but currently not supported by SQL Server transport. There is possibility to use different schemas.
```
SzymonPobiega: Multi-catalog is not defined about so readers would not know what does it mean in this context.

nservicebus/transports/guidance.md
SzymonPobiega: I wanted to provide some number to give users some kind of hint when they should be thinking about switching to higher tier/complexity models: from tiny to small (introduce schemas) and then to medium/large (move all transport to a dedicated catalog).

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,118 @@
+---
+title: SQL Server Transport usage guidance
```
MarcinHoppe: Is it usage guidance or deployment guidance?

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,118 @@
...
+summary: Usage guidance for SQL Server transport
+tags:
+- SQL Server
+---
+
+The SQL Server transport implements a message queueing mechanism on top of Microsoft SQL Server. It provides support for sending messages over [SQL Server](http://www.microsoft.com/en-au/server-cloud/products/sql-server/) tables. It does **not** make any use of ServiceBroker, a messaging technology built into the SQL Server.
```
MarcinHoppe: Maybe better as:

> The SQL Server transport implements a message queueing mechanism on top of Microsoft SQL Server. It provides support for sending messages over [SQL Server](http://www.microsoft.com/en-au/server-cloud/products/sql-server/) tables.

The SQL Server transport allows endpoints to exchange messages using the Microsoft SQL Server Database Engine. It implements queues on top of database tables and send / receive operations as `INSERT` and `DELETE` operations.

MarcinHoppe: Maybe:

> It does **not** make any use of ServiceBroker, a messaging technology built into the SQL Server.

would be better as:

Please note that the  SQL Server transport does **not** use or depend on SQL Server Service Broker.

SimonCropp:
note that this is currently red on TC since it does not follow our conventions

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,118 @@
...
+
+
+## How it works
+
+The SQL Server transport is a hybrid queueing system which is neither store-and-forward (like MSMQ) nor a broker (like RabbitMQ). It treats the SQL Server as a storage infrastructure for the queues while the queue implementation logic is executed inside the client process.
```
MarcinHoppe: I like this explanation, but it is a little bit contradictory to what we recommend, i.e. to treat SQL Server as a logical broker. Maybe we could state something like this:

SQL Server provides a central place to store queues and messages but the queue implementation is executed entirely within the endpoint process. As such, SQL Server transport is best thought of as a brokered transport like RabbitMQ rather than store-and-forward transport such as MSMQ.

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,118 @@
...
+ * Throughput on par with MSMQ
+ * Scaling out of processing is simple because of support for competing consumers at the queue level
+
+## Deployment considerations
+
+### Introduction
+Any NServiceBus endpoint operates and manages different types of data, those are:
+ * Business data - data used for implementing business capabilities
+ * Persistence data - infrastructure level data including: saga data, timeout manager state and message driven pub/sub information
+ * Transport data - messaging infrastructure state
```
MarcinHoppe: Could this

> Message infrastructure state

Be better put as:

Queues and messages

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,118 @@
...
+
+SQL Server Transport manages transport data and puts no constraints on the type and configuration of storage technology used for persistence and business data. It can work with any of available persisters i.e. NHibernate and RavenDB, as well as any storage mechanisms used inside message handlers.
+
+Building any system based on NServiceBus requires answering questions on how and where business, persistence and transport data will be stored. The purpose of this section is to define what needs to be considered from SQL Server Transport's perspective.
+ 
+NOTE: No matter what deployment options are chosen, there is one general rule that should always apply: **All transport data (input, audit and error queues) should reside on a single SQL Server instance**. Multi-instance deployments are technically possible but have been deprecated as of SQL Server Transport version 3. Please see: ###Section### for migration guidance.
```
MarcinHoppe: I believe we should move this to the beginning of the section.

nservicebus/transports/guidance.md
SimonCropp: isnt the file name for this wrong?

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Transactionality
+SQL Server Transport supports all [transaction modes](/nservicebus/transports/transactions.md) in particular `TransactionScope` which enables `exactly once` message processing with usage of distributed transactions. However when transport, persistence and business data are all stored in a single SQL Server catalog it is possible to achieve `exactly-once` message delivery without need for distributed transactions (###Link to Sample###).
+
+NOTE: `Exactly once` message processing without distributed transactions can be achieved with any transport (and SQL Server Transport in particular) using [Outbox](/nservicebus/outbox/). It requires business and persistence data to share the storage mechanism but does not put any requirements on transport data storage.
+     
+### Security 
+Security considerations for SQL Server Transport should follow minimal required access rights rule. For that each endpoint should use dedicated SQL Server account with SELECT and DELETE permissions on it's input queues as well as INSERT permissions on other endpoint's input queues, audit queue and error queue.
```
MarcinHoppe: minimal required access rights rule -> the [principle of least privilege](Principle of least privilege).

nservicebus/transports/guidance.md
MarcinHoppe: Maybe we should just state that standard techniques of tuning and scaling SQL Server for OLTP workloads apply?

nservicebus/transports/guidance.md
SzymonPobiega: @SimonCropp we'll figure it out later. It is [WIP] so not to be merged. It is still debatable if we are going to just include it in the doco content or generate something more like a whitepaper.

nservicebus/transports/guidance.md
SzymonPobiega: Not sure about that. Why?

nservicebus/transports/guidance.md
SzymonPobiega: You mean add a reference of that principle?

nservicebus/transports/guidance.md
SzymonPobiega: I removed the mention of number of endpoints.

nservicebus/transports/guidance.md
SzymonPobiega: I removed it because there is already a security section in the beginning of the document.

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,105 @@
...
+ * Great general-purpose tooling (SSMS)
+ * Free to start (Express edition)
+ * Scaling out of processing is simple because of support for competing consumers at the queue level
+
+## Deployment considerations
+
+### Introduction
+Any NServiceBus endpoint operates and manages different types of data, those are:
+ * Business data - data used for implementing business capabilities
+ * Persistence data - infrastructure level data including: saga data, timeout manager state and message driven pub/sub information
```
MarcinHoppe: I would argue that saga data is typicaly business data and not infrastructure level data. Maybe we should distinguish between business data (managed by the endpoint logic, i.e. handlers) and data managed by NServiceBus?

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,105 @@
...
+ * Transport data - queues and messages
+
+SQL Server Transport manages transport data and puts no constraints on the type and configuration of storage technology used for persistence and business data. It can work with any of available persisters i.e. NHibernate and RavenDB, as well as any storage mechanisms used inside message handlers.
+
+Building any system based on NServiceBus requires answering questions on how and where business, persistence and transport data will be stored. The purpose of this section is to define what needs to be considered from SQL Server Transport's perspective.
```
MarcinHoppe: Maybe this:

> Building any system based on NServiceBus requires answering questions on how and where business, persistence and transport data will be stored. The purpose of this section is to define what needs to be considered from SQL Server Transport's perspective.

Could be put better as:

Building an NServiceBus system requires that each kind of data is persisted in some way. This section explains what are the important factors to consider when choosing persistence technology to use in combination with the SQL Server transport.

nservicebus/transports/guidance.md
MarcinHoppe: After giving it some thought, highlighting this information should be enough. We're good.

MarcinHoppe: This:

> Multi-instance deployments are technically possible but have been deprecated as of SQL Server Transport version 3

Could be better as:

Multi-instance deployment topology is deprecated in version 3 of the SQL Server transport and will be removed in version 4.

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Transactionality
+SQL Server Transport supports all [transaction modes](/nservicebus/transports/transactions.md) in particular `TransactionScope` which enables `exactly once` message processing with usage of distributed transactions. However when transport, persistence and business data are all stored in a single SQL Server catalog it is possible to achieve `exactly-once` message delivery without need for distributed transactions (###Link to Sample###).
+
+NOTE: `Exactly once` message processing without distributed transactions can be achieved with any transport (and SQL Server Transport in particular) using [Outbox](/nservicebus/outbox/). It requires business and persistence data to share the storage mechanism but does not put any requirements on transport data storage.
+
+### Security 
+Security considerations for SQL Server Transport should follow the principle of least privilege. For that each endpoint should use dedicated SQL Server account with SELECT and DELETE permissions on it's input queues as well as INSERT permissions on other endpoint's input queues, audit queue and error queue.
```
MarcinHoppe: Account is probably not the right term when it comes to SQL Server security. We should probably use the term principal, which can be a user, a login, or a role.

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,105 @@
...
+
+NOTE: `Exactly once` message processing without distributed transactions can be achieved with any transport (and SQL Server Transport in particular) using [Outbox](/nservicebus/outbox/). It requires business and persistence data to share the storage mechanism but does not put any requirements on transport data storage.
+
+### Security 
+Security considerations for SQL Server Transport should follow the principle of least privilege. For that each endpoint should use dedicated SQL Server account with SELECT and DELETE permissions on it's input queues as well as INSERT permissions on other endpoint's input queues, audit queue and error queue. 
+
+Mulit-schema configuration can be used to ease the maintenance of security configuration because each endpoint may own a number of queues.
```
MarcinHoppe: Typo: mulit -> multi. I think section needs to be expanded. Example would be pure gold.

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,105 @@
...
+
+The best option is to store the queues in the same catalog as the business data. Schemas can be used to make maintenance easier. 
+
+### Medium to large
+
+For larger systems it usually makes more sense to have separate databases for each service or bounded context. It is also not unusual to see NoSQL databases in at least some of the services. The SQL databases might be hosted in a single instance of SQL Server or in separate instances depending on the IT policy.
+
+Such systems consist of tens to hundreds of endpoints. Some of the endpoints might be scaled out for higher throughput. 
+
+The best option is to have a dedicated catalog for the queues. This approach allows to use different scaling up strategies for the queue catalog. It also allows to use a dedicated back up policy for the queues (because data in queues has much more ephemeral). Unfortunately it leaves a problem of how to ensure that data modifications and queue operations are atomic and isolated from each other.
```
MarcinHoppe: has much more ephemeral -> are much more ephemeral

MarcinHoppe:
I think we're getting closer :smile:. Can we remove the sections we're not going to be covering in the first pass and create issues to cover them in the future?

tmasternak:
@ramonsmits @sfarmar Can you have a look at the guidance and add your review notes?

indualagarsamy:
@tmasternak @ramonsmits @sfarmar - any updates on this?

indualagarsamy:
@tmasternak @ramonsmits @sfarmar  @SzymonPobiega @MarcinHoppe - ping. 
Where are we on this? 
Is there anything I can help with?

tmasternak:
@indualagarsamy I've pinged @sfarmar to have a look and if he is good this is ready.

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,71 @@
...
+- SQL Server
+---
+
+The SQL Server transport allows endpoints to exchange messages using the Microsoft SQL Server Database Engine. It implements queues on top of database tables and send / receive operations as `INSERT` and `DELETE` operations.
+
+Please note that the SQL Server transport does not use or depend on SQL Server Service Broker.
+
+## How it works
+
+SQL Server provides a central place to store queues and messages but the message queuing implementation resides entirely within the endpoint process. SQL Server transport is best thought of as a brokered transport like RabbitMQ rather than store-and-forward transport such as MSMQ.
```
sfarmar: Maybe explain the difference between a broker style architecture to a distributed one, mentioning the single point of failure? Also the Store and Forward mechanism (maybe a link to a document describing store and forward in MSMQ?)

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Advantages of SQL Server Transport
+
+ * No additional licensing and training costs, as majority of Microsoft stack organizations already have SQL Server installed and have knowledge required to run it
+ * Great general-purpose tooling (SSMS)
+ * Free to start (Express edition)
+ * Scaling out of processing is simple because of support for competing consumers at the queue level
```
sfarmar: Scaling out using the Competing Consumer pattern?

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,71 @@
...
+
+The SQL Server transport allows endpoints to exchange messages using the Microsoft SQL Server Database Engine. It implements queues on top of database tables and send / receive operations as `INSERT` and `DELETE` operations.
+
+Please note that the SQL Server transport does not use or depend on SQL Server Service Broker.
+
+## How it works
+
+SQL Server provides a central place to store queues and messages but the message queuing implementation resides entirely within the endpoint process. SQL Server transport is best thought of as a brokered transport like RabbitMQ rather than store-and-forward transport such as MSMQ.
+
+### Advantages of SQL Server Transport
```
sfarmar: Add the cons? mainly single point of failure and the mitigation using clustering/redundancy

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,71 @@
...
+
+ * No additional licensing and training costs, as majority of Microsoft stack organizations already have SQL Server installed and have knowledge required to run it
+ * Great general-purpose tooling (SSMS)
+ * Free to start (Express edition)
+ * Scaling out of processing is simple because of support for competing consumers at the queue level
+
+## Deployment considerations
+
+### Introduction
+Any NServiceBus endpoint operates and manages different types of data, those are:
```
sfarmar: Any process hosting NServiceBus?

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,71 @@
...
+ * Business data - data managed by NServiceBus user, usually via code executed from inside message handlers
+ * Persistence data - data managed by infrastructure including: saga data, timeout manager state and message driven pub/sub information
```
sfarmar: Subscription state?

nservicebus/transports/guidance.md
SimonCropp: i dont understand why you would put this content elsewhere

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,71 @@
...
+ * Transport data - queues and messages managed by the transport
+
+SQL Server Transport manages transport data and puts no constraints on the type and configuration of storage technology used for persistence and business data. It can work with any of available persisters i.e. NHibernate and RavenDB, as well as any storage mechanisms used from inside message handlers.
```
sfarmar: > It can work with any of available persisters i.e. NHibernate and RavenDB, as well as any storage mechanisms used from inside message handlers.

Separate the persistence mechanism from the transport?

sfarmar: > available persisters i.e. NHibernate and RavenDB

available persisters i.e. NHibernate or RavenDB

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,71 @@
...
+
+Building an NServiceBus system requires that each kind of data is persisted in some way. This section explains what are the important factors to consider when choosing persistence technology to use in combination with the SQL Server transport.
```
sfarmar: > Building an NServiceBus system
Building a system using NServiceBus?

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,71 @@
...
+## Deployment considerations
+
+### Introduction
+Any NServiceBus endpoint operates and manages different types of data, those are:
+ * Business data - data managed by NServiceBus user, usually via code executed from inside message handlers
+ * Persistence data - data managed by infrastructure including: saga data, timeout manager state and message driven pub/sub information
+ * Transport data - queues and messages managed by the transport
+
+SQL Server Transport manages transport data and puts no constraints on the type and configuration of storage technology used for persistence and business data. It can work with any of available persisters i.e. NHibernate and RavenDB, as well as any storage mechanisms used from inside message handlers.
+
```
sfarmar: Add a title like " Persistence Considerations when choosing SQL transport"?

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,71 @@
...
+Building an NServiceBus system requires that each kind of data is persisted in some way. This section explains what are the important factors to consider when choosing persistence technology to use in combination with the SQL Server transport.
+ 
+NOTE: No matter what deployment options are chosen, there is one general rule that should always apply: **All transport data (input, audit and error queues) should reside in a single SQL Server catalog**. Multi-instance/multi-catalog deployment topology is deprecated in version 3 of the SQL Server transport and will be removed in version 4.
+
+### Transactionality
+SQL Server Transport supports all [transaction modes](/nservicebus/transports/transactions.md) in particular `TransactionScope` which enables `exactly once` message processing with usage of distributed transactions. However when transport, persistence and business data are all stored in a single SQL Server catalog it is possible to achieve `exactly-once` message delivery without need for distributed transactions. [SQL Server native integration sample](/samples/sqltransport/native-integration/) contains more information on such setup.
```
sfarmar: > achieve `exactly-once` message delivery without need for distributed transactions

Add this to the Advantages section?

sfarmar: > However when transport

However, when transport

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,71 @@
...
+
+NOTE: `Exactly once` message processing without distributed transactions can be achieved with any transport (and SQL Server Transport in particular) using [Outbox](/nservicebus/outbox/). It requires business and persistence data to share the storage mechanism but does not put any requirements on transport data storage.
+
+### Security 
+Security considerations for SQL Server Transport should follow the principle of least privilege. For that each endpoint should use dedicated SQL Server principal with `SELECT` and `DELETE` permissions on it's input queues as well as `INSERT` permissions on other endpoint's input queues, audit queue and error queue. 
+
+Multi-schema configuration can be used to ease the maintenance of security configuration because each endpoint may own a number of queues.
+
+### Monitoring
```
sfarmar: Why do we need to mention this?

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,71 @@
...
+All deployment options for SQL Server Transport described in this section are supported by Service Control.
+
+## Sample usage scenarios
+### Tiny
+
+The SQL Server transport is an ideal choice for extending an existing web application with asynchronous processing capabilities as an alternative for tedious batch jobs that tend to quickly get out of synch with the main codebase. Assuming the application already uses SQL Server as a data store, this scenario does not require any additional infrastructure.
+
+The queue tables are hosted in the same catalog as business data and the NServiceBus runtime can be hosted in the web worker. In some cases there might be a need for a separate process for hosting NServiceBus. Because system consists of a single logical service or bounded context, there is usually no need to create separate schemas for the queues.
```
sfarmar: This doesn't read very well...

nservicebus/transports/guidance.md
SimonCropp: dont we suffer from some of these cons http://mikehadlow.blogspot.com.au/2012/04/database-as-queue-anti-pattern.html

nservicebus/transports/guidance.md
SimonCropp: so isnt almost every article "guidance"? can we have a better file name?

SimonCropp: shouldnt this file exist under `nservicebus\sqlserver`?

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,71 @@
+---
+title: SQL Server Transport usage guidance
+summary: Usage guidance for SQL Server transport
+tags:
+- SQL Server
+---
+
+The SQL Server transport allows endpoints to exchange messages using the Microsoft SQL Server Database Engine. It implements queues on top of database tables and send / receive operations as `INSERT` and `DELETE` operations.
+
+Please note that the SQL Server transport does not use or depend on SQL Server Service Broker.
```
SimonCropp: remove please

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,71 @@
...
+ * No additional licensing and training costs, as majority of Microsoft stack organizations already have SQL Server installed and have knowledge required to run it
+ * Great general-purpose tooling (SSMS)
+ * Free to start (Express edition)
+ * Scaling out of processing is simple because of support for competing consumers at the queue level
+
+## Deployment considerations
+
+### Introduction
+Any NServiceBus endpoint operates and manages different types of data, those are:
+ * Business data - data managed by NServiceBus user, usually via code executed from inside message handlers
```
SimonCropp: in this context what is a "NServiceBus user"?

nservicebus/transports/guidance.md
SimonCropp: perhaps link to persisters?

nservicebus/transports/guidance.md
SimonCropp: "what are" is redundant

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,71 @@
...
+ * Persistence data - data managed by infrastructure including: saga data, timeout manager state and message driven pub/sub information
+ * Transport data - queues and messages managed by the transport
+
+SQL Server Transport manages transport data and puts no constraints on the type and configuration of storage technology used for persistence and business data. It can work with any of available persisters i.e. NHibernate and RavenDB, as well as any storage mechanisms used from inside message handlers.
+
+Building an NServiceBus system requires that each kind of data is persisted in some way. This section explains what are the important factors to consider when choosing persistence technology to use in combination with the SQL Server transport.
+ 
+NOTE: No matter what deployment options are chosen, there is one general rule that should always apply: **All transport data (input, audit and error queues) should reside in a single SQL Server catalog**. Multi-instance/multi-catalog deployment topology is deprecated in version 3 of the SQL Server transport and will be removed in version 4.
```
SimonCropp: > Multi-instance/multi-catalog deployment topology is deprecated in version 3 of the SQL Server transport and will be removed in version 4.

link to more info

SimonCropp: > No matter what deployment options are chosen, there is one general rule that should always apply

how does this add value? if i remove it then the statement still makes sense

nservicebus/transports/guidance.md
SimonCropp: "such a setup"

SimonCropp: > SQL Server Transport supports all [transaction modes](/nservicebus/transports/transactions.md). `TransactionScope` mode is particularly useful as it enables `exactly once` message processing with usage of distributed transactions.

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Transactionality
+SQL Server Transport supports all [transaction modes](/nservicebus/transports/transactions.md) in particular `TransactionScope` which enables `exactly once` message processing with usage of distributed transactions. However when transport, persistence and business data are all stored in a single SQL Server catalog it is possible to achieve `exactly-once` message delivery without need for distributed transactions. [SQL Server native integration sample](/samples/sqltransport/native-integration/) contains more information on such setup.
+
+NOTE: `Exactly once` message processing without distributed transactions can be achieved with any transport (and SQL Server Transport in particular) using [Outbox](/nservicebus/outbox/). It requires business and persistence data to share the storage mechanism but does not put any requirements on transport data storage.
```
SimonCropp: > and SQL Server Transport in particular

isnt this redundant?

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Security 
+Security considerations for SQL Server Transport should follow the principle of least privilege. For that each endpoint should use dedicated SQL Server principal with `SELECT` and `DELETE` permissions on it's input queues as well as `INSERT` permissions on other endpoint's input queues, audit queue and error queue.
```
SimonCropp: isnt "For that" redundant?

SimonCropp: perhaps link to audit and error more info?

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,71 @@
...
+
+Multi-schema configuration can be used to ease the maintenance of security configuration because each endpoint may own a number of queues.
+
+### Monitoring
+All deployment options for SQL Server Transport described in this section are supported by Service Control.
+
+## Sample usage scenarios
+### Tiny
+
+The SQL Server transport is an ideal choice for extending an existing web application with asynchronous processing capabilities as an alternative for tedious batch jobs that tend to quickly get out of synch with the main codebase. Assuming the application already uses SQL Server as a data store, this scenario does not require any additional infrastructure.
```
SimonCropp: remove tedious

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,71 @@
...
+
+The queue tables are hosted in the same catalog as business data and the NServiceBus runtime can be hosted in the web worker. In some cases there might be a need for a separate process for hosting NServiceBus. Because system consists of a single logical service or bounded context, there is usually no need to create separate schemas for the queues.
+
+### Small
+
+The SQL Server transport is a good choice for a pilot project to prove feasibility of NServiceBus in a given organization as well as for a small, well-defined green field application. It usually requires nothing more than a single shared SQL Server instance.
```
SimonCropp: why is "tiny" a real system. but small is a pilot? i would prefer to see pilot mentioned in its own section

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,71 @@
...
+
+The best option is to store the queues in the same catalog as the business data. Schemas can be used to make maintenance easier. 
+
+### Medium to large
+
+For larger systems it usually makes more sense to have dedicated catalogs for each service or bounded context. It is also not unusual to see NoSQL data stores in at least some of the services. The SQL catalogs might be hosted in a single instance of SQL Server or in separate instances depending on the IT policy.
```
SimonCropp: "not unusual" can the double negative be removed?

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,71 @@
...
+
+Is such case local transactions are not enough to ensure `exactly-once` message processing. When required, one option is to use Microsoft Distributed Transaction Coordinator and distributed transactions (NServiceBus with SQL Server transport is DTC-enabled by default). In this mode the message receive transaction and all data modifications that result from processing the message are part of one distributed transaction that spans two SQL catalogs (possibly stored on two different instances).
+
+Another options is to use [Outbox](/nservicebus/outbox/), which is a unique feature of NServiceBus, that provides `exactly-once` processing semantics over an at-least-once message delivery infrastructure. In this mode each individual endpoint has to store bussiness data and persistence data in the same catalog. Outgoing messages are stored by persistece infrastructure in a dedicated table. Single local transaction handles outgoing message persistence together with business data modifications. Only after such transaction commits successfully, messages are dispatched to the final destination.
```
SimonCropp: perhaps

> After the local transaction commits successfully the messages are dispatched to the final destination. 

and do we need an "if not"?

nservicebus/transports/guidance.md
tmasternak: I've added reference to store-and-forward description from doco. I don't think we should explain the difference between broker and distributed styles but can't find any relevant page about that.

nservicebus/transports/guidance.md
tmasternak: This is contrast to MSMQ when one needs Distributor

nservicebus/transports/guidance.md
tmasternak: I think that we should probably remove this section. It's duplicates what is written [here](http://docs.particular.net/nservicebus/sqlserver/#advantages-and-disadvantages-of-choosing-sql-server-transport). We can link to that page at the beginning of this document.

nservicebus/transports/guidance.md
tmasternak: Fixed

nservicebus/transports/guidance.md
tmasternak: Changed.

nservicebus/transports/guidance.md
MarcinHoppe: @SimonCropp Why? I think this is valuable information. Maybe we could move this to SQL Server Transport Design doco?

SimonCropp: sorry. that is me using too few words.

please rephrase to 

> Note that the SQL Server transport does not use or depend on SQL Server Service Broker.

MarcinHoppe: Ah, OK :smile:. I assumed you are just being very kind in your removal requests :wink:.

nservicebus/transports/guidance.md
sfarmar: What i meant was to change to "Scaling out using the Competing Consumer pattern"

MarcinHoppe: @sfarmar :+1: I'm working on it and will change it to use your wording.

nservicebus/transports/guidance.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Monitoring
+All deployment options for SQL Server Transport described in this section are supported by Service Control.
+
+## Sample usage scenarios
+### Tiny
+
+The SQL Server transport is an ideal choice for extending an existing web application with asynchronous processing capabilities as an alternative for batch jobs that tend to quickly get out of sync with the main codebase. Assuming the application already uses SQL Server as a data store, this scenario does not require any additional infrastructure.
+
+The queue tables can be hosted in the same SQL Server catalog as business and persistence data. NServiceBus endpoint can be hosted in the ASP.NET worker process. In some cases there might be a need for a separate process for hosting the NServiceBus endpoint. Because system consists of a single logical service or bounded context, there is usually no need to create separate schema for the queues.
```
MarcinHoppe: I think this:

> In some cases there might be a need for a separate process for hosting the NServiceBus endpoint.

could use an explanation about when to consider moving NServiceBus out of the worker.

tmasternak: I think @SzymonPobiega meant the scenario in which [IIS Idle Time-out Setting](https://technet.microsoft.com/en-us/library/cc771956(v=ws.10).aspx)

SzymonPobiega: I meant, hosting NSB in iis carries some constraints and most people like hosting it on other machines behind some firewalls.

MarcinHoppe: Can we then extend this sentence to hint at some things to watch out for when hosting in IIS?

tmasternak:
@SzymonPobiega @MarcinHoppe I've been looking at other documents for sql server that we have in docs. And I think that this document has a lot of overlap and a bit of inconsistency with http://docs.particular.net/nservicebus/sqlserver/. I propose to merge this one into the already existing in the following way:
 * Keep the `How it works` and `Adv + Disadv` section from the existing one
 * Move `Deployment considerations` from this PR as the second section to the existing one
 * Replace `Usage Scenarios` from the existing one with `Sample usage scenarios` from this one, keeping `Messaging framework for an application using SQL Azure` section.

Thoughts?

nservicebus/sqlserver/index.md
```diff
@@ -1,7 +1,6 @@
 ---
 title: SQL Server Transport
 summary: High-level description of NServiceBus SQL Server Transport.
-reviewed: 2016-03-22
```
SimonCropp: why?

tmasternak: @SimonCropp that might be my misunderstanding but I did the full review of this page after are merged in additional content.

tmasternak: Oh wait. I've removed that! Not my intention - fixing it now.

nservicebus/sqlserver/index.md
```diff
@@ -28,38 +28,63 @@ The SQL Server transport is a hybrid queueing system which is neither store-and-
...
+ * Free to start (Express edition)
  * Queues support competing consumers (multiple instances of same endpoint feeding off of same queue) so scale-out doesn't require using [distributor](/nservicebus/scalability-and-ha/distributor/)
 
 
 ### Disadvantages
 
  * No local store-and-forward mechanism, meaning that when SQL Server instance is down the endpoint cannot send nor receive messages
- * In centralized deployment maximum throughput applies for the whole system, not individual endpoints. If SQL Server on a given hardware can handle 2000 msg/s, each one of the 10 endpoints deployed on this machine can only receive 200 msg/s (on average).
+ * In centralized deployment maximum throughput applies for the whole system, not individual endpoints. If SQL Server on a given hardware can handle 2000 msg/s, each one of the 10 endpoints deployed on this machine can only receive 200 msg/s (on average)
+ * When empty queue gets polled periodically (every 1 second) by the receiver to detect new messages. This can lead to processing delays for rarely used endpoints
```
MarcinHoppe: This:

>  When empty queue gets polled periodically (every 1 second) by the receiver to detect new messages. This can lead to processing delays for rarely used endpoints 

could be better as:

Queue tables are polled periodically to check if receive operation should be started. This may lead to delays in message delivery for endpoints that experience low rate of incoming messages.

nservicebus/sqlserver/index.md
```diff
@@ -28,38 +28,63 @@ The SQL Server transport is a hybrid queueing system which is neither store-and-
...
  * No local store-and-forward mechanism, meaning that when SQL Server instance is down the endpoint cannot send nor receive messages
- * In centralized deployment maximum throughput applies for the whole system, not individual endpoints. If SQL Server on a given hardware can handle 2000 msg/s, each one of the 10 endpoints deployed on this machine can only receive 200 msg/s (on average).
+ * In centralized deployment maximum throughput applies for the whole system, not individual endpoints. If SQL Server on a given hardware can handle 2000 msg/s, each one of the 10 endpoints deployed on this machine can only receive 200 msg/s (on average)
+ * When empty queue gets polled periodically (every 1 second) by the receiver to detect new messages. This can lead to processing delays for rarely used endpoints  
 
+## Deployment considerations 
 
-## Usage Scenarios
+### Introduction
+Any process hosting NServiceBus operates and manages different types of data, those are:
```
MarcinHoppe: This can be safely removed:

> , those are

nservicebus/sqlserver/index.md
```diff
@@ -28,38 +28,63 @@ The SQL Server transport is a hybrid queueing system which is neither store-and-
...
+ * Business data - data managed by NServiceBus user, usually via code executed from inside message handlers
```
MarcinHoppe: This:

> data managed by NServiceBus user, usually via code executed from inside message handlers

could be better as:

data managed by user code, typically message handlers and sagas

nservicebus/sqlserver/index.md
```diff
@@ -28,38 +28,63 @@ The SQL Server transport is a hybrid queueing system which is neither store-and-
...
+ * Persistence data - data managed by infrastructure including: saga data, timeout manager state and subscriptions state
+ * Transport data - queues and messages managed by the transport
 
+SQL Server Transport manages transport data and puts no constraints on the type and configuration of storage technology used for persistence and business data. It can work with any of available persisters i.e. [NHibernate](/nservicebus/nhibernate) or [RavenDB](/nservicebus/ravendb/), as well as any storage mechanisms used from inside message handlers.
```
MarcinHoppe: message handlers. --> message handlers and sagas.

nservicebus/sqlserver/index.md
```diff
@@ -28,38 +28,63 @@ The SQL Server transport is a hybrid queueing system which is neither store-and-
...
 
-### Messaging framework for the enterprise
+Building a system using NServiceBus requires that each kind of data is persisted in some way. This section explains the important factors to consider when choosing persistence technology to use in combination with the SQL Server transport.
+ 
+NOTE: No matter what deployment options are chosen, there is one general rule that should always apply: **All transport data (input, audit and error queues) should reside in a single SQL Server catalog**. Multi-instance/multi-catalog deployment topology although supported has been deprecated in version 3 of the SQL Server transport and will be removed in version 4.
```
MarcinHoppe: This:

> Multi-instance/multi-catalog deployment topology although supported has been deprecated in version 3 of the SQL Server transport and will be removed in version 4.

could be revised better as:

Multi-instance/multi-catalog deployment topology is still available but is deprecated in version 3 of the SQL Server transport and will be removed in version 4.

nservicebus/sqlserver/index.md
tmasternak: I would not include sagas here. The intention was to point to data that is managed in terms of persistence (CRUD) by user written code. Does not apply to saga which are handled by NServiceBus

nservicebus/sqlserver/index.md
```diff
@@ -28,38 +28,63 @@ The SQL Server transport is a hybrid queueing system which is neither store-and-
...
+NOTE: `Exactly once` message processing without distributed transactions can be achieved with any transport using [Outbox](/nservicebus/outbox/). It requires business and persistence data to share the storage mechanism but does not put any requirements on transport data storage.
 
+### Security 
+Security considerations for SQL Server Transport should follow the principle of least privilege. Each endpoint should use a dedicated SQL Server principal with `SELECT` and `DELETE` permissions on its input queue tables and `INSERT` permission on input queue tables of endpoints it sends messages to. Each endpoint should also have permissions to insert rows to audit and error queue tables.
 
-### Background worker replacement
+Multi-schema configuration can be used to manage fine-grained access control to various database objects used by the endpoint, including its queue tables.
```
MarcinHoppe: Link to multi-schema section in SQL Server Transport configuration doco page?

nservicebus/sqlserver/index.md
```diff
@@ -28,38 +28,63 @@ The SQL Server transport is a hybrid queueing system which is neither store-and-
...
 
-SQL Server transport works well for small web applications that require reliable background processing capabilities. Instead of firing a background thread or doing `async` processing the web application would send a message on the bus. The message would be picked up by the backend service, processed and the response returned. SQL Server transport will work out-of-the-box in that scenario without requiring DTC. 
+### Monitoring
```
MarcinHoppe: Rename subsection header to "Service Control"?

nservicebus/sqlserver/index.md
tmasternak: :+1:. The new wording is way better. I've only changed 
> receive operation should be started

to
> receive operation should be performed

nservicebus/sqlserver/index.md
tmasternak: I would avoid mentioning sagas here. AFAIK the best practice that we share with clients is to only send messages from sagas and rely on NSericeBus persistence for storing saga's state.

nservicebus/sqlserver/index.md
MarcinHoppe: What I meant was message handler code within the saga process, not saga data itself that is managed by NSB. The fact that I confused you probably means that it would confuse readers as well. Let's stay with handlers only.

nservicebus/sqlserver/index.md
MarcinHoppe: :+1:

nservicebus/sqlserver/index.md
```diff
@@ -17,7 +17,7 @@ The SQL Server transport implements a message queueing mechanism on top of Micro
 
 ## How it works
 
-The SQL Server transport is a hybrid queueing system which is neither store-and-forward (like MSMQ) nor a broker (like RabbitMQ). It treats the SQL Server as a storage infrastructure for the queues. The queueing logic itself is implemented and executed as part of the transport code running in an NServiceBus endpoint.
+SQL Server provides a central place to store queues and messages but the message queuing implementation resides entirely within the endpoint process. The SQL Server transport is best thought of as a brokered transport like RabbitMQ rather than [store-and-forward](/nservicebus/architecture/principles.md#drilling-down-into-details-store-and-forward-messaging) transport such as MSMQ.
```
weralabaj: Wouldn't it be simpler to say that in SQL Server transport we use SQL Server only to store queues and messages? If we really want to be explicit then we can say that we don't use queueing services provided by it or that we implement queueing logic ourselves.

nservicebus/sqlserver/index.md
```diff
@@ -28,38 +28,63 @@ The SQL Server transport is a hybrid queueing system which is neither store-and-
...
+ * Free to start (Express edition)
  * Queues support competing consumers (multiple instances of same endpoint feeding off of same queue) so scale-out doesn't require using [distributor](/nservicebus/scalability-and-ha/distributor/)
 
 
 ### Disadvantages
 
  * No local store-and-forward mechanism, meaning that when SQL Server instance is down the endpoint cannot send nor receive messages
- * In centralized deployment maximum throughput applies for the whole system, not individual endpoints. If SQL Server on a given hardware can handle 2000 msg/s, each one of the 10 endpoints deployed on this machine can only receive 200 msg/s (on average).
+ * In centralized deployment maximum throughput applies for the whole system, not individual endpoints. If SQL Server on a given hardware can handle 2000 msg/s, each one of the 10 endpoints deployed on this machine can only receive 200 msg/s (on average)
+ * Queue tables are polled periodically to check if receive operation should be performed. This may lead to delays in message delivery for endpoints that experience low rate of incoming messages.
```
weralabaj: > This may lead to delays in message delivery for endpoints that experience low rate of incoming messages.

Isn't the poll period constant? Or we try to be smart about this?  If the latter then I'd add something like "because the polling period is determined dynamically/optimized for the average rate". Can the user control it, at least the default? If so then also worth mentioning.

tmasternak: The user cannot control it (we have a plan of exposing the setting if any user asks for that). The value is not adjusted in any way at runtime. Currently it is pretty arbitrary.

nservicebus/sqlserver/index.md
tmasternak: Would 
> SQL Server transport uses SQL Server to store queues and messages. I doesn't use any queuing services provided by SQL Server - queuing logic is implemented within SQL Server transport and executed separately by each endpoint.  

sound better?

nservicebus/sqlserver/index.md
```diff
@@ -28,38 +28,63 @@ The SQL Server transport is a hybrid queueing system which is neither store-and-
...
+### Introduction
+Any process hosting NServiceBus operates and manages different types of data:
+ * Business data - data managed by NServiceBus user, usually via code executed from inside message handlers
+ * Persistence data - data managed by infrastructure including: saga data, timeout manager state and subscriptions state
+ * Transport data - queues and messages managed by the transport
 
+SQL Server Transport manages transport data and puts no constraints on the type and configuration of storage technology used for persistence and business data. It can work with any of available persisters i.e. [NHibernate](/nservicebus/nhibernate) or [RavenDB](/nservicebus/ravendb/), as well as any storage mechanisms used from inside message handlers.
 
-### Messaging framework for the enterprise
+Building a system using NServiceBus requires that each kind of data is persisted in some way. This section explains the important factors to consider when choosing persistence technology to use in combination with the SQL Server transport.
```
weralabaj: > Building a system using NServiceBus requires that each kind of data is persisted in some way

That's not completely accurate (it depends on what exactly you do), and I don't think it adds useful information, can we skip that sentece and just say "When choosing a persistence technology to use in combination with the SQL Server transport, consider the following factors:"?

nservicebus/sqlserver/index.md
weralabaj: > pretty arbitrary

hardcoded? then why do we say it impacts endpoints that have low rate of incoming messages?

nservicebus/sqlserver/index.md
weralabaj: To me yes, more precise :D

nservicebus/sqlserver/index.md
```diff
@@ -28,38 +28,63 @@ The SQL Server transport is a hybrid queueing system which is neither store-and-
...
- * [**Outbox**](/nservicebus/outbox/) The *exactly-once* delivery guarantee is provided by the outbox. No DTC service is required. Messages addressed to a database that is temporarily down will by retried by means of [Second-Level Retries](/nservicebus/errors/automatic-retries.md) so SLR timeout larger than maintenance windows should be configured. The downside is that all the systems need to have Outbox enabled which requires them to use NServiceBus 5.x or later.
+NOTE: `Exactly once` message processing without distributed transactions can be achieved with any transport using [Outbox](/nservicebus/outbox/). It requires business and persistence data to share the storage mechanism but does not put any requirements on transport data storage.
 
+### Security 
+Security considerations for SQL Server Transport should follow the principle of least privilege. Each endpoint should use a dedicated SQL Server principal with `SELECT` and `DELETE` permissions on its input queue tables and `INSERT` permission on input queue tables of endpoints it sends messages to. Each endpoint should also have permissions to insert rows to audit and error queue tables.
```
weralabaj: Since we're mentioning a specific thing, link https://en.wikipedia.org/wiki/Principle_of_least_privilege ?

nservicebus/sqlserver/index.md
```diff
@@ -28,38 +28,63 @@ The SQL Server transport is a hybrid queueing system which is neither store-and-
...
+[Multi-schema](/nservicebus/sqlserver/configuration.md#multiple-custom-schemas) configuration can be used to manage fine-grained access control to various database objects used by the endpoint, including its queue tables.
 
-SQL Server transport works well for small web applications that require reliable background processing capabilities. Instead of firing a background thread or doing `async` processing the web application would send a message on the bus. The message would be picked up by the backend service, processed and the response returned. SQL Server transport will work out-of-the-box in that scenario without requiring DTC. 
+### Service Control
+All deployment options for SQL Server Transport described in this section are supported by Service Control.
 
-Using messaging in that scenario is more reliable than using additional threads. Even if the processing of the message fails, the message itself is not lost and it can be retired at a later time. 
+## Sample usage scenarios
 
+### Tiny
```
weralabaj: > Tiny, Small, Medium to large

From the outsider perspective what does it mean? I feel something more descriptive like "Replacing batch jobs", "Pilot project with the existing system" etc. would guide use better.

nservicebus/sqlserver/index.md
```diff
@@ -28,38 +28,63 @@ The SQL Server transport is a hybrid queueing system which is neither store-and-
...
+The SQL Server transport is a good choice for a pilot project to prove feasibility of NServiceBus in a given organization as well as for a small, well-defined green field application. It usually requires nothing more than a single shared SQL Server instance.
 
-The SQL Azure's limitations include no support for distributed transactions and relatively small database size. This makes it an ideal candidate for a [multi-database mode](deployment-options.md) SQL Server transport with [Outbox](/nservicebus/outbox/). Each endpoint could have its own database that meets the size limit. Outbox would guarantee *exactly-once* processing of messages.
\ No newline at end of file
+The best option is to store the queues in the same catalog as the business data. Schemas can be used to make maintenance easier. 
+
+### Medium to large
+
+For larger systems it usually makes more sense to have dedicated catalogs for each service or bounded context. NoSQL data stores can be used in some services, depending on data access requirements. The SQL catalogs might be hosted in a single instance of SQL Server or in separate instances depending on the IT policy.
```
weralabaj: > For larger systems it usually makes more sense to have dedicated catalogs for each service or bounded context.

Didn't you just say multi-instance/multi-catalog is being removed? Confused.

nservicebus/sqlserver/index.md
```diff
@@ -28,38 +28,63 @@ The SQL Server transport is a hybrid queueing system which is neither store-and-
...
+
```
weralabaj: Also that doesn't seem to be "Usage scenario" in the sense "when should I use that transport" like the previous are, but rather explaining how to approach a specific problem. IMO that should be a separate section, and better defined.

nservicebus/sqlserver/index.md
```diff
@@ -28,38 +28,63 @@ The SQL Server transport is a hybrid queueing system which is neither store-and-
...
+The best option is to have a dedicated catalog for the queues. This approach allows to use different scaling up strategies for the queue catalog. It also allows to use a dedicated back up policy for the queues (because data in queues are much more ephemeral). 
+
+Is such case local transactions are not enough to ensure `exactly-once` message processing. When required, one option is to use Microsoft Distributed Transaction Coordinator and distributed transactions (NServiceBus with SQL Server transport is DTC-enabled by default). In this mode the message receive transaction and all data modifications that result from processing the message are part of one distributed transaction that spans two SQL catalogs (possibly stored on two different instances).
```
weralabaj: > Is such case local transactions are not enough to ensure `exactly-once` message processing.

Didn't you just tell me to keep all queues in one catalog, how did we end up here?!

weralabaj: I think that should be covered in the section dedicated to transactions, it feels really out of place here.

nservicebus/sqlserver/index.md
tmasternak: What this sentence was supposed to say is that one might want to have dedicated catalog/sql-instance per bounded context for persistence+business data.

nservicebus/sqlserver/index.md
tmasternak: I think that problem comes from the polling itself. It requires to make a trade-off between:
 * Worst-case processing delay - the smaller polling interval the lower worst case (a single message is put on a queue after longer period of zero load) processing delay  
 * Load put on input table - frequent polling means executing potentially unnecesary queries

nservicebus/sqlserver/index.md
tmasternak: Rephrased. Does it look better?

nservicebus/sqlserver/index.md
tmasternak: :+1:

nservicebus/sqlserver/index.md
tmasternak: Changed the names of scenarios to reflect the usage and not the size.

nservicebus/sqlserver/index.md
tmasternak: Changed from `usage` to `deployment`. The assumption here is that someone already decided that want to go with SQL Server transport. Comparison of different transports is out of scope.

nservicebus/sqlserver/index.md
tmasternak: Changed.

nservicebus/sqlserver/index.md
tmasternak: Changed to:
> For larger systems it usually makes more sense to have dedicated catalogs for business and persistence data per service or bounded context.

Hope that exposes intent more clearly.

tmasternak:
@weralabaj I've added changes after you review

nservicebus/sqlserver/index.md
weralabaj: hm... i don't see that

tmasternak: I think this is a commit based dif. See this for newest version https://github.com/Particular/docs.particular.net/pull/1579/files

Issue #1247, opened 2016-03-03
Document our backwards compatibility policy


particularbot:
http://docs.particular.net/search?q=compatibility => comes up pretty empty

Document how we treat:

* Wire compatibility
* Data compatibility
* API compatibility

Here's a video of me talking about it http://fast.wistia.net/embed/iframe/49x5srdsta?popover=true

particularbot:
**Comment by [andreasohlund](https://github.com/andreasohlund)** _https://github.com/Particular/PlatformDevelopment/issues/611#issuecomment-181756309_

----

cc @Particular/platform-dev-squad @udidahan

mikeminutillo:
I think it would make sense to reference this in the upgrade guides as well. Something simple like:

> Endpoints running different versions of NServiceBus are all compatible (LINK). To minimize risk, upgrade one endpoint at a time.

andreasohlund:
@mikeminutillo and me just have the following answer to a [GG question](https://groups.google.com/d/msg/particularsoftware/GnmI5B3rS5k/woQccoX9AgAJ) on this topic

>Regarding upgrades we strongly recommend users to go "major by major". This will give you the best upgrade experience as you will be guided by compiler errors and our upgrade guides.

>Keep in mind that NServiceBus is backwards compatible on the wire all the way back to Version 3. This means that you can upgrade one endpoint at a time to lower the risk of the upgrade.

>While we guarantee wire compatibility back to Version 3 you need to be aware that we only guarantee backwards compatibility for persisted data (sagas, timeouts, subscriptions etc) between sequential major versions. For this reason we recommend that you go live with a v5 version before moving to v6 to ensure that your data is migrated forwards. 

cc @MarcinHoppe

weralabaj:
@andreasohlund @mikeminutillo I've started a doco PR for persistence tests, but seems like it'd make more sense to have all compatibility aspects covered in one doco, could that be part of this issue? https://github.com/Particular/docs.particular.net/pull/1580/files/9f0c568597dd9ec7294b7e06626c7740af713b0b

Can we prioritize it and get started? It'll require getting a few people involved to review and provide their input.

As for the task itself, would that Readme section https://github.com/Particular/EndToEnd#persistence-compatibility-tests be sufficient?

mikeminutillo:
[...]

PR #1580, opened 2016-06-06
[WIP] Document compatibility tests

weralabaj:
Connects to https://github.com/Particular/docs.particular.net/issues/1247

### DIFF
--- a/nservicebus/testing/cross-version-testing.md
+++ b/nservicebus/testing/cross-version-testing.md
@@ -1,0 +1,24 @@
+---
+title: Cross-version compatibility testing with NServiceBus
+summary: Develop service layers and long-running processes using test-driven development.
+reviewed: 2016-06-06
+related:
+ - samples/unit-testing
+---
+
+In order to ensure that NServiceBus upgrades are safe and that endpoints in various versions can communicate with each other, it is necessary to extensively test multiple aspects of cross-version compatibility. As much as possible, those tests are automated. Some of them are available publicly in the [Particular/EndToEnd](https://github.com/Particular/EndToEnd/) repository.
+
+Those tests might be useful for NServiceBus users in order to verify that the framework works correctly in an unusual scenario, for example when using an unusually complex data structure, in order to understand in detail how specific feature works, or to reproduce an issue they experienced. The provided code can be run on a local machine, the pre-requisites are listed in the [Readme](https://github.com/Particular/EndToEnd/blob/master/README.md) file in the repository.
+
+
+## How it works?
+
+
+### Pulling packages from NuGet
+
+The tests cover all supported versions of NServiceBus and related packages (persistences, transports, etc.), including all minor versions. The available version numbers within a specified range are automatically determined based on the NuGet feeds. They are also automatically pulled to be stored locally before running the tests.
+
+
+### Versions combinations
+
+In order to ensure cross-version compatibility versions combinations are generated automatically within a given range. The test cases are executed across all combinations of minor versions within a specified range of versions.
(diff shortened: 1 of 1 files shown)


nservicebus/testing/cross-version-testing.md
```diff
@@ -0,0 +1,24 @@
...
+title: Cross-version compatibility testing with NServiceBus
+summary: Develop service layers and long-running processes using test-driven development.
+reviewed: 2016-06-06
+related:
+ - samples/unit-testing
+---
+
+In order to ensure that NServiceBus upgrades are safe and that endpoints in various versions can communicate with each other, it is necessary to extensively test multiple aspects of cross-version compatibility. As much as possible, those tests are automated. Some of them are available publicly in the [Particular/EndToEnd](https://github.com/Particular/EndToEnd/) repository.
+
+Those tests might be useful for NServiceBus users in order to verify that the framework works correctly in an unusual scenario, for example when using an unusually complex data structure, in order to understand in detail how specific feature works, or to reproduce an issue they experienced. The provided code can be run on a local machine, the pre-requisites are listed in the [Readme](https://github.com/Particular/EndToEnd/blob/master/README.md) file in the repository.
```
HEskandari: I don't like this paragraph as a whole. It reads as if there's not enough work is done to test the edge-cases and the user will be on their own, which obviously is not the case. Do we need to justify why releasing this source code to public is a good thing or how it can be used? isn't that already implied?

SimonCropp: > Those tests might be useful for NServiceBus users in order to verify that the framework works correctly in an unusual scenario, for example when using an unusually complex data structure, in order to understand in detail how specific feature works, or to reproduce an issue they experienced.

I would prefer some runnable code that shouw them how they can verify this stuff between versions. it would take a fair bit of effort to groc our EndToEnd enough to make them useful inside a endpoint

weralabaj: So @HEskandari @SimonCropp would it be better just to skip the whole justification and start with running on local?

> I would prefer some runnable code that shouw them how they can verify this stuff between versions.

Not sure what you have in mind? Snippets?

SimonCropp: well i think u need to phrase this in terms of what they want to do and what we want to achieve

one scenario is people wanting to evolve saga data/messages moving forward. how would u write tests for this? and repeat those test over multiple iterations. while there is much code in out endtoend we could use to do that , leaving that to the user to work out how to do that is a big ask

another scenario is someone building a persister/transport/serializer. in that case there is a much higher amount of code that could be reused from endtoend

the 3rd point we want to get across is that "we do backwards compat test and we take it seriously". this is more pitched at the enterprise architect level and is almost marketing speak.

weralabaj: the 3rd will be addressed by a a story anyway, so I think focusing on option 2 makes the most sense at the moment. the 1st - not sure how often anybody would use it? also probably would be better to describe it, when we isolate the re-usable parts (generating versions pairs, pulling nugets, etc.) into separate libraries?

SimonCropp: >  when we isolate the re-usable parts (generating versions pairs, pulling nugets, etc.) into separate libraries?

yes. in fact this has already been happening. samples and snippets regularly drive the design moving forwrd. for example most of the scripting snippets (http://docs.particular.net/nservicebus/msmq/operations-scripting) i would eventually like to see them as reusable libs.

weralabaj: Might be worth opening an issue for it and getting a TF to move this forward faster?

weralabaj:
@SimonCropp @HEskandari As per our discussion, I connected this issue to the documenting backward compat policy. I don't think that should be a blocker for closing https://github.com/Particular/PlatformDevelopment/issues/737.

tmasternak:
We closing this. The content should be a good starting point to TF that will create backwards compatibility policy document.

PR #1599, opened 2016-06-15
[WIP] routing pubsub refactoring

SzymonPobiega:
Updates the routing doco for https://github.com/Particular/NServiceBus/pull/3816

### DIFF
--- a/Snippets/Core/Core_6/EndpointName/Usage.cs
+++ b/Snippets/Core/Core_6/EndpointName/Usage.cs
@@ -31,8 +31,8 @@
         {
             #region InputQueueOverrideRouting
 
-            var unicastRouting = endpointConfiguration.UnicastRouting();
-            unicastRouting.RouteToEndpoint(typeof(MyMessage), "MyEndpoint");
+            var routing = endpointConfiguration.Routing();
+            routing.RouteToEndpoint(typeof(MyMessage), "MyEndpoint");
             var transport = endpointConfiguration.UseTransport<MyTransport>();
             transport.AddAddressTranslationException(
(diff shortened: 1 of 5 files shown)


johnsimons:
Some inconsistencies in the API:
```C#
var routing = endpointConfiguration.Routing();
routing.RouteToEndpoint(typeof(MyMessage), "MyEndpoint");
```
```c#
var transport = endpointConfiguration.UseTransport<UnicastTransport>();           
transport.RegisterPublisherForType("Sales", typeof(MyEvent));           
transport.RegisterPublisherForAssembly("Sales", typeof(OtherEvent).Assembly);
```
So in the first code block we have `routing.RouteToEndpoint(typeof(MyMessage), "MyEndpoint");` and in the second the parameter order is inverse `transport.RegisterPublisherForType("Sales", typeof(MyEvent));`

I think it would make sense to make them the same order, thoughts ?

johnsimons:
Also I wonder if we could have a simpler syntax like:
```
routing.RouteToEndpoint<MyMessage>("MyEndpoint");
```

johnsimons:
This is confusing, and the doco api does not help (it is the same for both):
image
and
image

johnsimons:
image

These seem to be "advanced apis" do we really need to expose them at the root level ?

SimonCropp:
@johnsimons your feedback should be targeted at the NSB repo not the doco repo

SimonCropp:
merged manually here https://github.com/Particular/docs.particular.net/commit/a4c4507417d7e48d530552f08ece7e16ae2f06f9

PR #1606, opened 2016-06-17
Add a note that behaviors are now effectively singletons

SzymonPobiega:


nservicebus/upgrades/5to6.md
```diff
@@ -234,6 +234,7 @@ The `IMessageMutator` and `IMutateTransportMessages` interfaces are no longer av
 
  * `IncomingContext` and `OutgoingContext` have been replaced with multiple contexts ("stages") per pipeline offering more useful information depending on the requirements. [Pipeline Steps Stages and Connectors](/nservicebus/pipeline/steps-stages-connectors.md) provides more details about the available stages.
  * `Behavior<TContext>` now provides the base class for implementing custom behaviors and replaces `IBehavior<TContext>`.
+ * All behaviors are effectively singletons, no matter how they are registered in the container.
```
SimonCropp: in this context isnt "singletons" ambiguous? most people assume singletons means single instance per appdomain, which is not true in this case.

SimonCropp: "no matter how they are registered in the container"  what is the value of this? do we want to say "behaviors are not resolved from the container and hence should not be registered in the container"?

nservicebus/upgrades/5to6.md
```diff
@@ -234,6 +234,7 @@ The `IMessageMutator` and `IMutateTransportMessages` interfaces are no longer av
 
  * `IncomingContext` and `OutgoingContext` have been replaced with multiple contexts ("stages") per pipeline offering more useful information depending on the requirements. [Pipeline Steps Stages and Connectors](/nservicebus/pipeline/steps-stages-connectors.md) provides more details about the available stages.
  * `Behavior<TContext>` now provides the base class for implementing custom behaviors and replaces `IBehavior<TContext>`.
```
capone3691: Behavior

nservicebus/upgrades/5to6.md
mikeminutillo: Behavior instances only created once and then cached for the life of the endpoint?

SimonCropp: i like it. although "only" is redundant

SzymonPobiega: @SimonCropp they are resolved, but only once, when building the pipeline.

SimonCropp: @SzymonPobiega sorry yes u r correct

SzymonPobiega: @mikeminutillo @SimonCropp I've updated the proposal.

nservicebus/upgrades/5to6.md
SimonCropp: @capone3691 not sure what you mean by this comment. can u raise a new issue with some more details. Thanks

nservicebus/upgrades/5to6.md
```diff
@@ -234,6 +234,7 @@ The `IMessageMutator` and `IMutateTransportMessages` interfaces are no longer av
...
+ * Behavior instances are created once, when the endpoint starts up, and then cached for the life of the endpoint. This is true even if the behavior type is registered as `InstancePerCall` or `InstancePerUnitOfWork`.
```
weralabaj: > even if the behavior type is registered as `InstancePerCall` or `InstancePerUnitOfWork`

That seems confusing, you mean even if the user registers them as such? When they create new behaviors? Or that it's in the core code?

SzymonPobiega: @weralabaj I tried to explain it but I failed. There is actually no reason to have these APIs if we need to explain why they don't work as they are expected to. I raised an issue to remove them: https://github.com/Particular/NServiceBus/issues/3844.

weralabaj: OK, I like your proposition of change, so I see two options:

1. for now we'll go with 

"This is true even if the behavior type is registered using `RegisterComponents` method as `InstancePerCall` or `InstancePerUnitOfWork`."

and add note to the issue in NSB repo

2. we'll mark this as WIP and clarify after the NSB issue is resolved

I'd say it depends on when do we expect the code change will get done.

SzymonPobiega: I strongly believe in removing the APIs now that I see how flawed they are. Let's wait with this PR.

SimonCropp:
marked as WIP until core issue is resolved

indualagarsamy:
@SzymonPobiega - any updates on this?

indualagarsamy:
@SzymonPobiega - ping

SzymonPobiega:
@indualagarsamy I pinged core maintainers on the blocking issue.

SimonCropp:
@andreasohlund what was the outcome on this one?

andreasohlund:
The decision was to not remove the api in this version since the `IBuilder` overload would have the same problem so we're not "plugging the hole"

https://github.com/Particular/NServiceBus/issues/3844

@SzymonPobiega can we go ahead and put that note in without removing the api?

SzymonPobiega:
@andreasohlund the problem is, I have no idea how to phrase it correctly. Tried a couple of times but it proved to be very difficult to explain.

andreasohlund:
How about:

> In version 6 behavior instances are created once, when the endpoint starts up, and then cached for the life of the endpoint. This is true even if the behavior type is registered as `InstancePerCall` or `InstancePerUnitOfWork` making all pipeline behaviors effectively singletons. Make sure to always use the passed behavior context to store state information to avoid multiple instances of the same behavior from sharing state.

or something along those lines?

SzymonPobiega:
@andreasohlund @SimonCropp pointed out that by singletons people usually understand something that has a single instance per app domain. So does not work here. I was thinking about saying that the lifetime scope is ignored for behavior registrations but could not make it work.

SzymonPobiega:
@SimonCropp @andreasohlund  Does this sound acceptable?

SimonCropp:
LGTM. merged manually here https://github.com/Particular/docs.particular.net/commit/ca24930cf960ca9ecf662de9c222a8314335bbe3

PR #1607, opened 2016-06-20
Fix for multi instance so that satellite queues are created using the correct instance

ramonsmits:
* Both V2 and V3 sample now use a *StartsWith*
* Fixed behavior of V3 sample that previously returned sender connection for receiver endpoint and vice versa.


SimonCropp:
@Particular/sqlserver-transport-maintainers can one of you review this

tmasternak:
I've already synced with @ramonsmits on that. LGTM

SimonCropp:
@tmasternak @ramonsmits thanks

Issue #1609, opened 2016-06-20
SQL Transport - Configuration push mode not documented for V3


ramonsmits:
Looking at http://docs.particular.net/nservicebus/upgrades/sqlserver-2to3 there is no mention of *push mode configuration* and in the docs at http://docs.particular.net/nservicebus/sqlserver/configuration#multiple-connection-strings-via-the-configuration-api-push-mode there is no V3 sample. Either the sample needs to be added or the upgrade guide needs to mention that this feature is dropped.

weralabaj:
@ramonsmits it says that now you need to use `EnableLegacyMultiInstanceMode` method to achieve that effect, is that not enough?

ramonsmits:
@weralabaj I don't see any reference to push in the upgrade.

Pull is not mentioned either but I see `EnableLegacyMultiInstanceMode`as the replacement for Pull

weralabaj:
Hm that's one way to look at it, I guess not what we expected. It says "The configuration API for multi-instance support has now changed. Multiple connection strings have to be provided by connection factory method passed to EnableLagacyMultiInstanceMode method." so that kind of indicates this is the only supported method. I see your point though, we could be more explicit here.

PR #1610, opened 2016-06-20
Mention push mode for sql transport upgrade guide

weralabaj:
Connects to https://github.com/Particular/docs.particular.net/issues/1609


weralabaj:
@ramonsmits does it address your concern?

Issue #928, opened 2015-11-16
NHibernate doco doesn't cover setting connection string in code well


DavidBoike:
Related to [this StackOverflow question](http://stackoverflow.com/questions/33726745/exception-when-trying-to-configure-nservicebus-to-use-nhibernate-persistence):

The main [NHibernate persistence page](http://docs.particular.net/nservicebus/nhibernate/) has a section for configuration via code but that doesn't cover the simple example of setting the connection string.

PR #1615, opened 2016-06-21
NHibernate persistence - add info about passing connection string via code

weralabaj:
Connects to https://github.com/Particular/docs.particular.net/issues/928.


weralabaj:
@DavidBoike does that cover what you wanted to do here?

DavidBoike:
Yes! LGTM. Thanks for picking this up!

Issue #366, opened 2014-07-28
expand on the differences between audit and forwarding


mauroservienti:
Atm there is no difference. And that's the issue, we want to transition forwarding to mean:
"Forward the processed message untouched to X other queues" (note the untouched and X)
The current audit/forward actually modifies/adds headers to the message.

A possible scenario is:

"As a devop I'm moving a service from endpoint A to endpoint B and I do not want to be required to update immediately/at the same time all the senders to the new configuration so I want to configure endpoint A to forward all messages of type XYZ to endpoint B so to be able to gradually update senders configuration or simply to temporarily move something transparently."

--

related to #327

mauroservienti:
backlogged until there is a difference.

weralabaj:
@mauroservienti is that still relevant? can we close?

mauroservienti:
If my last statement, there is no difference yet, is still true, then I see no reason to keep this open.

weralabaj:
Confirmed with Andreas that there's a difference now. Need to investigate the details.

PR #1625, opened 2016-06-22
Clarify the difference between auditing and forwarding

weralabaj:
Resolves https://github.com/Particular/docs.particular.net/issues/366.

### DIFF
--- a/nservicebus/messaging/forwarding.md
+++ b/nservicebus/messaging/forwarding.md
@@ -9,16 +9,14 @@ redirects:
 
 Use this feature to forward successfully processed messages from an endpoint to a specified destination endpoint. Forwarding messages is particularly useful in complex upgrade scenarios, when the old version and new version of a particular endpoint are running side-by-side.
 
-Note that forwarding is a solution to a specific problem and generally should be used for a limited transition period. In case of long-term need to duplicate messages the [Publish-Subscribe](/nservicebus/messaging/publish-subscribe/) pattern will be usually more appropriate.
-
 
 ## Auditing vs Fowarding
 
 Technically [Auditing](/nservicebus/operations/auditing.md) and Forwarding are very similar, both send a copy of the processed message to another queue. The main difference are intended usage scenarios.
 
-Auditing is used for collecting information on what is happening in the system, therefore the audited message is enriched with additional [information regarding processing it](/nservicebus/operations/auditing.md#message-headers). The forwarded message is a copy of the message, without the additional auditing information.
+Auditing is used for collecting information on what is happening in the system, therefore the audited message is enriched with additional [information regarding processing it](/nservicebus/operations/auditing.md#message-headers). The forwarded message is a copy of the processed message, without the additional auditing information.
 
-Note: In Version 5 and below the forwarded messages contain some of the headers added during auditing. Those headers are no longer added in Version 6 and above.
+Note: In Versions 5 and below some of the audit headers were available for the forwarded messages. Starting from Version 6 the forwarded messages will no longer contain the audit message headers.
 
 
 ## Configuring Message Forwarding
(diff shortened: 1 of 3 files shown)


weralabaj:
@mauroservienti @andreasohlund can you have a look?

andreasohlund:
:+1: good stuff! :shipit:

nservicebus/messaging/forwarding.md
```diff
@@ -7,7 +7,18 @@ redirects:
 ---
 
 
-Use this feature to forward successfully processed messages from an endpoint to a specified destination endpoint. If there is a need to collect all successfully processed messages from all endpoints, then use the [Auditing Feature](/nservicebus/operations/auditing.md) instead.
+Use this feature to forward successfully processed messages from an endpoint to a specified destination endpoint. Forwarding messages is particularly useful in complex upgrade scenarios, when the old version and new version of a particular endpoint are running side-by-side.
+
+Note that forwarding is a solution to a specific problem and generally should be used for a limited transition period. In case of long-term need to duplicate messages the [Publish-Subscribe](/nservicebus/messaging/publish-subscribe/) pattern will be usually more appropriate.
```
indualagarsamy: >Note that forwarding is a solution to a specific problem and generally should be used for a limited transition period.

I am not sure if this is true. You can use forwarding when you have a workflow where after a message is processed by endpoint 1, you want endpoint 2 to do more processing. Sort of like an itinerary pattern.  I think the original use case was meant for this.

nservicebus/messaging/forwarding.md
```diff
@@ -7,7 +7,18 @@ redirects:
...
+
+
+## Auditing vs Fowarding
+
+Technically [Auditing](/nservicebus/operations/auditing.md) and Forwarding are very similar, both send a copy of the processed message to another queue. The main difference are intended usage scenarios.
+
+Auditing is used for collecting information on what is happening in the system, therefore the audited message is enriched with additional [information regarding processing it](/nservicebus/operations/auditing.md#message-headers). The forwarded message is a copy of the message, without the additional auditing information.
+
+Note: In Version 5 and below the forwarded messages contain some of the headers added during auditing. Those headers are no longer added in Version 6 and above.
```
indualagarsamy: >In Version 5 and below

Should be `In Versions 5 and below`,

indualagarsamy: Note: In Version 5 and below the forwarded messages contain some of the headers added during auditing. Those headers are no longer added in Version 6 and above.

How about:

NOTE: While some of the audit headers were available for the forwarded messages in Versions 5 and below, starting from Version 6, forwarded messages will no longer contain the audit message headers.

nservicebus/messaging/forwarding.md
indualagarsamy: @SimonCropp - I think you have more context on the original message forwarding. Any insights?

weralabaj: That was based on the conversation with Andreas, I'll reword.

weralabaj: > Sort of like an itinerary pattern.

Can't you use sagas in that situation?

mauroservienti: It's much more similar to a routing slip approach

weralabaj: @indualagarsamy I removed that sentece, since I'm not 100% what are the possible usage scenarios I don't want the description to be too limiting. Do you think it'd make sense to suggest considering Pub-Sub or sagas as alternatives or that also assumes too much about how people might want to use that feature?

SimonCropp: history is here https://github.com/Particular/NServiceBus/issues/1591

weralabaj:
@SimonCropp @mauroservienti I've removed that sentence. I think it'd be better, if we want to go into details and list possible usage scenarios then please, help with specific suggestions :D

Issue #1244, opened 2016-03-02
Advantages / Disadvantages of RabbitMQ


boblangley:
https://twitter.com/codesennin/statuses/705052643824148480

Create an Advantages / Disadvantages section for RabbitMQ similar to [documentation in the SQL Transport doco](http://docs.particular.net/nservicebus/sqlserver/).

PR #1626, opened 2016-06-22
RabbitMQ transport - advantages and disadvantages

weralabaj:
Resolves https://github.com/Particular/docs.particular.net/issues/1244.

### DIFF
--- a/nservicebus/rabbitmq/index.md
+++ b/nservicebus/rabbitmq/index.md
@@ -13,3 +13,16 @@ Provides support for sending messages over [RabbitMQ](http://www.rabbitmq.com/) 
 The NuGet package is available here https://www.nuget.org/packages/nservicebus.rabbitmq.
 
     PM> Install-Package NServiceBus.RabbitMQ
+
+
+## Advantages and Disadvantages of choosing RabbitMQ Transport
+
+
+### Advantages
+- RabbitMQ provides dedicated reliability and high-availability features.
+- Offers native Publish-Subcribe mechanism, therefore it doesn't require NServiceBus persistence for storing event subscriptions.
+- Wide range of supported clients allows for integrating the system with applications written in other languages using native RabbitMQ features.
+
+### Disadvantages
+- The additional cost of commercial RabbitMQ license and support.
+- Many organizations don't have the same level of expertise with RabbitMQ as for example with SQL Server, so it might require additional training.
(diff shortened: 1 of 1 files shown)


weralabaj:
@Particular/rabbitmq-transport-maintainers please see what else could I include here

bording:
The list you've got looks good to me. Here are a couple other items I can think of.

Possible other advantages:

- Automatically supports multiple instances of an endpoint in a round-robin fashion without any additional configuration
- I need to double check the perf numbers, but I think it's one of the fastest transports now. Not sure we want to get into perf stuff here, though
 

Possible disadvantages:
- You have to deal with duplicate messages (so you have to use the Outbox, or make everything idempotent). This can be tricky around dealing with saga timeouts due to how the timeout manager works.

weralabaj:
Thanks @bording

WojcikMike:
@bording could you please elaborate on this:
> Automatically supports multiple instances of an endpoint in a round-robin fashion without any additional configuration

As I don't know what exactly do you mean.

When it comes to the list of + and - I know that @andreasohlund on his talk was ranting a lot about how badly Rabbit handle network partitioning which might cause messages loses:
https://www.rabbitmq.com/partitions.html

I don't know how to describe it in clear way.

MarcinHoppe:
@WojcikMike Can we use something mentioning scale-out using the Competing Consumers pattern? I think this is a phrase that we've used in the past.

WojcikMike:
@MarcinHoppe I think this is a question to Brandon, as I am almost sure that as a broker we support competing consumers OOB.

weralabaj:
Regarding performance, this is what we say for SQL Server "Maximum throughput for any given endpoint is on par with MSMQ.".

SimonCropp:
> "Maximum throughput for any given endpoint is on par with MSMQ.".

i dont know how anyone can make that claim. broker versus distributed queue, they have very difference performance under various types of load

weralabaj:
well... we do :P possibly sql then needs editing.

MarcinHoppe:
@SimonCropp @weralabaj I agree we should probably remove this claim. PR is on the way...

weralabaj:
ping @Particular/docs-maintainers

bording:
@WojcikMike The point I was trying to make is that since RabbitMQ pushes messages to the endpoints instead of the endpoints polling, the broker is responsible for distributing messages to the available consumers, and it does so in a round-robin fashion. To get this behavior with RabbitMQ, you just start another instance of your same endpoint.

To get similar behavior with MSMQ, for example, you have to use the distributor (V5), or sender-side distribution (V6), and that requires additional infrastructure or configuration changes to the endpoint.

SimonCropp:
@bording re this https://github.com/Particular/docs.particular.net/pull/1626#issuecomment-228191647  so is this extra content u still think this PR requires?

bording:
@SimonCropp Seems like this line:
>  Supports [Competing consumers](http://www.enterpriseintegrationpatterns.com/patterns/messaging/CompetingConsumers.html) pattern. Messages are received by instances in a round-robin fashion.

Covers its well enough, but maybe something that speaks to the part about not needing any extra config/modification to the endpoint to gain the benefit?

weralabaj:
@SimonCropp I added what @bording suggested

PR #1630, opened 2016-06-23
Remove vague sentence

MarcinHoppe:
We don't really have firm data to back this sentence. Performance of the SQL Server transport depends on so many factors and can be subject to various database tuning techniques.

/cc @SimonCropp @weralabaj @Particular/sqlserver-transport-maintainers


tmasternak:
@MarcinHoppe I was wandering about this one as well. However after talking with @ramonsmits it looks that `EndToEnd` test suite result actually confirm that.

MarcinHoppe:
Then we must rethink how we communicate performance characteristics of the SQL Server transport. Comparing it to MSMQ is... you know, apples to oranges.

I just wanted to remove this sentence for now because it sets wrong expectations. Someone might migrate off MSMQ to the SQL Server transport and expect the performance to be the same and most likely it will not be the same.

tmasternak:
@MarcinHoppe :+1: for removing it

PR #1632, opened 2016-06-23
Document ServiceControl with SQL Server transport multi-instance topology

MarcinHoppe:
Connects to Particular/PlatformDevelopment#555


MarcinHoppe:
@ramonsmits @mauroservienti @mikeminutillo @johnsimons @tmasternak @SzymonPobiega @indualagarsamy Can you take a look at this initial doco for SC / SQL transport / multi-instance?

It's still WIP but I'd like your early feedback. Thanks!

Snippets/SqlServer/SqlServer_2/MultiDb.cs
```diff
@@ -45,4 +45,34 @@ void OtherEndpointConnectionParamsPull(BusConfiguration busConfiguration)
...
+            {
+                return ConnectionInfo.Create()
+                    .UseConnectionString("Server=DbServerB;Database=BillingDB;");
+            }
+            if (transportAddress == "Sales")
+            {
+                return ConnectionInfo.Create()
+                    .UseConnectionString("Server=DbServerC;Database=SalesDB;");
+            }
+            return null;
```
SimonCropp: is null correct here? or should we throw?

Snippets/SqlServer/SqlServer_3/MultiDb.cs
```diff
@@ -22,7 +21,39 @@ await connection.OpenAsync()
...
+            if (transportAddress == "Billing")
+            {
+                connectionString = "Server=DbServerB;Database=BillingDB;";
+            }
+            if (transportAddress == "Sales")
+            {
+                connectionString = "Server=DbServerC;Database=SalesDB;";
+            }
+
+            var connection = new SqlConnection(connectionString);
```
SimonCropp: should we throw if connectionString  is null?

nservicebus/sqlserver/service-control.md
```diff
@@ -0,0 +1,42 @@
...
+
+![](servicecontrol-multiinstance.png)  
+
+
+## ServiceControl configuration
+
+
+### Distributed transactions
+
+Multi-instance deployment of the SQL Server transport requires that the Distributed Transaction Coordinator (DTC) is used by all endpoints. This is also required by ServiceControl to support retry of failed messages. Default configuration of ServiceControl disables support for distributed transactions and it has to be turned of explicitly using `EnableDtc` configuration setting:
```
SimonCropp: > be turned of explicitly using `EnableDtc` configuration setting:

should this be

> be turned on explicitly using `EnableDtc` configuration setting:

johnsimons:
@mikeminutillo I thought we were not going to document this hack ?

mikeminutillo:
@johnsimons that's a good point. 

How far do we want to go in supporting multi-instance mode officially?
Do we doco it fully and trust that customers won't go down this path unless they are already in this situation?
Should we doco it but not include it in the menu so we can send direct links and you can find it via search? _And we can send a link when answering support cases._
Should we not document this at all and provide answers to customers who contact us via support?
Should we doco it on the ServiceControl repo and point customers who ask about it there?

mikeminutillo:
@MarcinHoppe the content looks good though :)

johnsimons:
:+1: to move this doco out to either the sqltransport or servicecontrol

The reasons for not hosting this doco in docs website is because this is a hack, we actually want to know how many users are affected by this issue, so we really need them to contact us via support so we can guide them out of this deployment. Also by contacting support we will start having an idea on numbers of customers affected and whether we need to provide the some kind of automation or other ways to migrate out of the incorrect deployment.

tmasternak:
:100: :+1:  for the diagram

tmasternak:
@MarcinHoppe what is the purpose of those snippets? I though that it's possible to configure SC with mulit-instance only with app.config settings?

MarcinHoppe:
@tmasternak SC can be configured via app.config. The endpoints need to be configured via the pull API.

MarcinHoppe:
@SimonCropp @johnsimons @mikeminutillo I see the concern. @ramonsmits has a sample in his private repo that we used for testing. We were thinking about moving the sample under the labs org.

Idea: what if we put the sample and this doco under labs. I think it is clear that whatever is in the labs is unsupported but we still have some documentation to use internally and to point customers to, if need be.

Thoughts?

tmasternak:
@mikeminutillo Here are my thoughs
> Do we doco it fully and trust that customers won't go down this path unless they are already in this situation?
Should we doco it but not include it in the menu so we can send direct links and you can find it via search? And we can send a link when answering support cases.
Should we not document this at all and provide answers to customers who contact us via support?
Should we doco it on the ServiceControl repo and point customers who ask about it there?

I think we should provide full doco and add exactly the comment you mentioned: "If you don't already have this kind of deployment, don't do it".

tmasternak:
@MarcinHoppe I don't like the idea of putting docs in labs TBH. Yes we are deprecating the multi-instance mode but it's still a fully supported part of the platform including SC.

Snippets/SqlServer/SqlServer_2/MultiDb.cs
MarcinHoppe: Probably better to throw with a good message. I will fix this.

Snippets/SqlServer/SqlServer_3/MultiDb.cs
MarcinHoppe: Again, :+1: for throw.

nservicebus/sqlserver/service-control.md
```diff
@@ -0,0 +1,37 @@
...
+reviewed: 2016-06-24
+tags:
+ - SQL Server
+ - Transports
+ - ServiceControl
+---
+
+
+## Overview
+Even though it is [recommended that all SQL Server transport queue tables are stored in a single SQL Server catalog](/#deployment-considerations), it is possible to use ServiceControl to monitor a multi-instance deployment of the SQL Server transport. A requirement for such a configuration is that all endpoints share `error` and `audit` queues and that these queues are stored in the same catalog as ServiceControl queues. Other queues used by individual endpoints may be stored in different SQL Server instances. The following diagram shows an example system configuration:
```
SzymonPobiega: Maybe "multi-catalog"? I think we should stay consistent in a single statement and talk about instances or catalogs. Actually I think that an ambiguous term 'database' would be a good fit here.

MarcinHoppe: @SzymonPobiega I rephrased it a little bit.

MarcinHoppe:
@tmasternak If it is fully supported then we should put it in the doco. The guidance we recently released says:

> All deployment options for SQL Server Transport described in this section are supported by Service Control.

If we say so, then we should also say how to configure such a setting.

Also, addressed some review remarks.

SimonCropp:
fyi it can still be in docs and hidden  https://github.com/Particular/docs.particular.net#hidden

MarcinHoppe:
@SimonCropp Didn't know that is possible. It looks like a reasonable middle ground. I will mark this doco page as hidden.

SimonCropp:
can the assets that were used to generate the image also be added to this PR

SimonCropp:
since 555 is still open should this one be a WIP?

MarcinHoppe:
Good point, fixed. I will take it out of WIP when SC is released.

MarcinHoppe:
@Particular/docs-maintainers Given that SC 1.19.0 has been [released](https://github.com/Particular/ServiceControl/releases/1.19.0), this is ready to be merged.

PR #1639, opened 2016-06-27
Update disadvantages section with low latency processing note

tmasternak:
@Particular/sqlserver-transport-maintainers /cc: @SimonCropp

### DIFF
--- a/nservicebus/sqlserver/index.md
+++ b/nservicebus/sqlserver/index.md
@@ -35,7 +35,7 @@ SQL Server transport uses SQL Server to store queues and messages. It doesn't us
 
  * No local store-and-forward mechanism, meaning that when SQL Server instance is down the endpoint cannot send nor receive messages.
  * In centralized deployment maximum throughput applies for the whole system, not individual endpoints. If SQL Server on a given hardware can handle 2000 msg/s, each one of the 10 endpoints deployed on this machine can only receive 200 msg/s (on average).
- * Queue tables are polled periodically to check if receive operation should be performed. This may lead to delays in message delivery for endpoints that experience low rate of incoming messages. As a result SQL Server transport show not be used when there are requirements for worst case message processing latency.
+ * Queue tables are polled periodically to check if receive operation should be performed. This may lead to delays in message delivery for endpoints that experience low rate of incoming messages. As a result SQL Server transport should not be used when there are requirements for worst case message processing latency.
 
 
 ## Deployment considerations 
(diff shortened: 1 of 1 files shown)


nservicebus/sqlserver/index.md
```diff
@@ -35,7 +35,7 @@ SQL Server transport uses SQL Server to store queues and messages. It doesn't us
 
  * No local store-and-forward mechanism, meaning that when SQL Server instance is down the endpoint cannot send nor receive messages.
  * In centralized deployment maximum throughput applies for the whole system, not individual endpoints. If SQL Server on a given hardware can handle 2000 msg/s, each one of the 10 endpoints deployed on this machine can only receive 200 msg/s (on average).
- * Queue tables are polled periodically to check if receive operation should be performed. This may lead to delays in message delivery for endpoints that experience low rate of incoming messages.
+ * Queue tables are polled periodically to check if receive operation should be performed. This may lead to delays in message delivery for endpoints that experience low rate of incoming messages. As a result SQL Server transport should not be used when there are requirements for worst case message processing latency.
```
SimonCropp: > worst case message processing latency

seems a bit ambiguous

tmasternak: Maybe something like:
> As a result SQL Server transport should not be used when processing latency lower than 10s of miliseconds has to be guaranteed for every arriving message.

SimonCropp: TBH i dont like that either. but i cant think of anything better.  paging @Particular/docs-maintainers @Particular/sqlserver-transport-maintainers

MarcinHoppe: Maybe we should make a recommendation to look at other transports (Rabbit) if latency in the milisecond range is a requirement?

indualagarsamy: @tmasternak @MarcinHoppe @SimonCropp - How about something like this:

When using SqlTransport, a database table serves as a queue for the messages for the endpoints. These tables are polled periodically to see if messages need to be processed by the endpoints.  Although the polling interval is in the order of milliseconds, this may still lead to delays in when a message is picked up for processing. For environments where low latency is a requirement, rather than using SQL as a queuing transport, consider using other transports that use queuing technologies like RabbitMQ, MSMQ, etc.

MarcinHoppe: I like it. Just to make our records straight, the polling interval is in the order of seconds (exactly one second right now), not milliseconds. Other than that I think this explains the issue well.

indualagarsamy: @MarcinHoppe - thanks. I will push that change in to the PR.

tmasternak:
@indualagarsamy @MarcinHoppe Thanks for helping out. :+1:

PR #1664, opened 2016-07-05
Clarify what it means to "send a message based on the ReplyToAddress header"

udidahan:


SimonCropp:
@weralabaj @udidahan ibus doesnt exist anymore

MarcinHoppe:
@SimonCropp It exists in the majority of patched versions. We can make it more precise by adding "or `IMessageHandlerContext`".

weralabaj:
@SimonCropp that's why it says "for example" ;)

PR #1665, opened 2016-07-05
Add IMessageHandlerContext info for v6

MarcinHoppe:
A little bit more v6-proof.

/cc @SimonCropp @weralabaj @udidahan

### DIFF
--- a/security-advisories/sqlserver-sqlinjection.md
+++ b/security-advisories/sqlserver-sqlinjection.md
@@ -33,7 +33,7 @@ Attackers can use this vulnerability to force an endpoint to execute arbitrary S
 The exploitation of this vulnerability requires that all of the conditions below are met at the same time:
 
  1. An attacker must have the ability to send a malicious message to an endpoint OR an attacker must be able to modify a message that is in transit to an endpoint such that the `ReplyToAddress` header can be manipulated,
- 1. The receiving endpoint sends a message based on `ReplyToAddress` header - for example, by using the `Reply` method of `IBus`.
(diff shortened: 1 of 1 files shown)


security-advisories/sqlserver-sqlinjection.md
```diff
@@ -33,7 +33,7 @@ Attackers can use this vulnerability to force an endpoint to execute arbitrary S
 The exploitation of this vulnerability requires that all of the conditions below are met at the same time:
 
  1. An attacker must have the ability to send a malicious message to an endpoint OR an attacker must be able to modify a message that is in transit to an endpoint such that the `ReplyToAddress` header can be manipulated,
- 1. The receiving endpoint sends a message based on `ReplyToAddress` header - for example, by using the `Reply` method of `IBus`.
+ 1. The receiving endpoint sends a message based on `ReplyToAddress` header - for example, by using the `Reply` method of `IBus` or `IMessageHandlerContext`.
```
indualagarsamy: @MarcinHoppe - instead of 
>`IBus` or `IMessageHandlerContext`.

how about:
The receiving endpoint sends a message based on `ReplyToAddress` header - for example, by using the `Reply` method of `IBus` or using the `IMessageHandlerContext` parameter available in the message handlers when using NServiceBus Version 6 and above.

Thoughts?

MarcinHoppe: @indualagarsamy I like it a lot, it's more precise. I improved wording per your suggestion.

PR #1674, opened 2016-07-07
apply async practices

SimonCropp:


SimonCropp:
@danielmarbach review please

danielmarbach:
@SimonCropp these changes don't make sense. The APIs used return a string as a return type. The sample shows how to wrap APIs with return types. With your changes you'd just ignore the return type. I propose to close this

SimonCropp:
@danielmarbach ahhh i get it. so i agree with u for HandlerWhichIntegratesWithRemotingWithAPM and will fix that. how about the other two?

danielmarbach:
Look closer. They also are calling APIs with return types

PR #1675, opened 2016-07-07
Update best-practice-enforcement.md

timbussmann:
> While this worked it caused other features like [auto subscribe](publish-subscribe/controlling-what-is-subscribed.md) to stop working since only `Events` are auto subscribed.

This sentence is confusing. The autosubscribe doc explicitly states that it only works for messages marked as IEvent or with an appropriate convention.

> These enforcements can be bypassed by defining all messages as plain messages.

Using plain messages without any conventions throws exceptions at the message handler registry, so that sentence doesn't seem to be right?

### DIFF
--- a/nservicebus/messaging/best-practice-enforcement.md
+++ b/nservicebus/messaging/best-practice-enforcement.md
@@ -7,9 +7,7 @@ tags:
 - Event
 ---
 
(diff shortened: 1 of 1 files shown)


SimonCropp:
@andreasohlund given you added that text (https://github.com/Particular/docs.particular.net/commit/b88f381b24cd834c83b3f62a23678b98fe4b87e4) can u comment on this PR

SimonCropp:
ok merging this since it seems correct to me. @andreasohlund can u revisit later u u want to adjust

PR #1678, opened 2016-07-08
Update index.md

dvdstelt:
Just a couple of links that I noticed were incorrect, because the info was moved to separate and more detailed pages.


SimonCropp:
seems to have some broken links

PR #1690, opened 2016-07-11
Some clarification and caveats for Task.Run / Factory.StartNew in handlers

danielmarbach:
Should NOTE become WARN?

Or should we just drop the `Task.Run` samples because there are potentially too many variables at play? 

In my personal opinion we should show it for completeness reasons.

### DIFF
--- a/nservicebus/handlers/async-handlers.md
+++ b/nservicebus/handlers/async-handlers.md
@@ -59,16 +59,19 @@ From a server perspective if asynchronous is compared to synchronous by just loo
 
 ## Calling short running compute-bound code
 
-Short running compute-bound code (a few milliseconds) that is executed in the handler should be executed directly on the IO-thread that is executing the handler code.
+Short running compute-bound code that is executed in the handler should be executed directly on the IO-thread that is executing the handler code.
 
 snippet: ShortComputeBoundMessageHandler
 
 Call the code directly and **do not** wrap it with a [`Task.Run`](https://msdn.microsoft.com/en-us/library/system.threading.tasks.task.run.aspx) or [`Task.Factory.StartNew`](https://msdn.microsoft.com/en-au/library/dd321439.aspx).
 
+For the majority of business scenarios this approach is just fine since many of the asynchronous base class library methods in the .NET Framework will schedule continuations on the worker thread pool the likelyhood that no IO-thread is blocked is high.
 
 ## Calling long-running compute-bound code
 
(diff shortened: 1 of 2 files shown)


distantcam:
I'd just call the method `BlocksForALongTime` or even `BlockingCall`.

weralabaj:
IMO it could be a warning, to make it even more visible. 

> The last thing one should be doing is wrapping code inside the handler with `Task.Run` or `Task.Factory.StartNew`. It could harm the throughput more than it helps if wrongly applied.

That sounds a bit informal, how about

Wrapping code inside the handler with `Task.Run` or `Task.Factory.StartNew` can seriously harm the throughput if applied incorrectly. It should be used only in special cases.

danielmarbach:
@weralabaj @distantcam thanks for the suggestions! Applied and pushed

nservicebus/handlers/async-handlers.md
```diff
@@ -59,16 +59,19 @@ From a server perspective if asynchronous is compared to synchronous by just loo
...
 snippet: ShortComputeBoundMessageHandler
 
 Call the code directly and **do not** wrap it with a [`Task.Run`](https://msdn.microsoft.com/en-us/library/system.threading.tasks.task.run.aspx) or [`Task.Factory.StartNew`](https://msdn.microsoft.com/en-au/library/dd321439.aspx).
 
+For the majority of business scenarios this approach is just fine since many of the asynchronous base class library methods in the .NET Framework will schedule continuations on the worker thread pool the likelyhood that no IO-thread is blocked is high.
 
 ## Calling long-running compute-bound code
 
-Long running compute-bound code (more than a hundred milliseconds) that is executed in a handler could be offloaded to the worker thread pool.
+WARN: This approach should only be used after a thorough analysis of the runtime behavior and the code involved in the call hierarchy of a handler. Wrapping code inside the handler with Task.Run or Task.Factory.StartNew can seriously harm the throughput if applied incorrectly. It should be used only in special cases.
```
ramonsmits: > It should be used only in special cases.

Maybe as at least one special case?

It should be used only in special cases like allowing parallelism for multiple CPU-bound tasks. Within a handler that only has sequential logic there would be no need to use it.

danielmarbach: @ramonsmits good like it. I try to incorporate it

danielmarbach: @ramonsmits made updates as well as changed one of the snippets

Issue #1327, opened 2016-03-17
Explain our reasoning for the transition from "Bus" to "Endpoint" or "Message"


boblangley:
Would it be helpful to provide a short explaination, either in the 5-6 upgrade guide or elsewhere why we changed the names of some of our core objects?

We have adequately covered that they have changed, but should we provide some context / reasoning?

Relevant discussions:

* https://github.com/Particular/PlatformDevelopment/issues/600
* https://github.com/Particular/NServiceBus/issues/3411
* https://github.com/Particular/NServiceBus/issues/3401
* https://github.com/Particular/NServiceBus/issues/3399
* https://github.com/Particular/NServiceBus/issues/3425

/cc: @Particular/tf-v6-launch @indualagarsamy

DavidBoike:
I vote absolutely. We will need to have an answer for this. Better to provide it proactively.

indualagarsamy:
@Particular/nservicebus-maintainers - can you please add a list of relevant documents that needs to be updated based on the above?

timbussmann:
@indualagarsamy That seems to be a task for the taskforce working on this issue?

SzymonPobiega:
I think the best place is the intro to the upgrade guide where we outline key changes.

indualagarsamy:
What I'd like to do is:
Have a proper introduction for the upgrade guide. 
Outline the major changes that are in v6. Namely the async and moving away from IBus to more into Contexts. And provide why we went that route. 
Rearrange some of the content, such as Endpoint Naming, IBus being obsolete etc to make more cohesive sense.

SimonCropp:
@indualagarsamy still working on this? or should i remove your name?

indualagarsamy:
What's major in v6 is more of a release notes / announcement thing.
What i've done in the PR is covered the IBus obsoletes and the various scenarios where IBus was previously used and what to do in each of the cases instead.

PR #1693, opened 2016-07-13
Added reasons as to why the bus was deprecated.

indualagarsamy:
Fixes: https://github.com/Particular/docs.particular.net/issues/1327

### DIFF
--- a/nservicebus/upgrades/5to6-moving-away-from-ibus.md
+++ b/nservicebus/upgrades/5to6-moving-away-from-ibus.md
@@ -8,11 +8,11 @@ Starting with Version 6, the `IBus` interface has been deprecated.
 The `IBus` interface was one of the key interfaces when using previous versions of NServiceBus. It provided access to many of the operations like sending messages, subscribing to events, and manipulating headers. The `IBus` interface was available through dependency injection in message handlers, any custom components that were registered in the container and through other mechanisms like the saga base class.
 
 
-## Why was IBus deprecated?
+## Reasons for deprecating IBus
 
-The problem with `IBus` is that the interface contained numerous methods and properties. However, not all of them are valid in certain scenarios. For example, methods like `Reply` and `ForwardCurrentMessageTo` are always available on the `IBus` interface, but it only makes sense to use those in the context of handling an incoming message. Using them in other scenarios will throw an exception.
+The `IBus` interface contained numerous methods and properties. However, not all of them are valid in all scenarios where an instance of `IBus` was exposed and this can lead to confusion. For example, the methods `Reply` and `ForwardCurrentMessageTo` are always available on the `IBus` interface, but it is only relevant to use it in the context of handling an incoming message. Using them in other scenarios will throw an exception.
 
-For users who are new to the NServiceBus API, the fact that `IBus` was available in message handlers using Dependency Injection wasn't obvious, especially when trying to send or publish messages from inside message handlers. Rather than using dependency injection, a `IMessageHandlerContext` parameter is now available in all the message handlers. This parameter will expose all of the appropriate actions to the message handler. Methods that aren't applicable will not be available making the API simpler. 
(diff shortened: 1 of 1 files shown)


indualagarsamy:
@boblangley @Particular/docs-maintainers - please review

nservicebus/upgrades/5to6-moving-away-from-ibus.md
```diff
@@ -7,28 +7,38 @@ Starting with Version 6, the `IBus` interface has been deprecated.
 
 The `IBus` interface was one of the key interfaces when using previous versions of NServiceBus. It provided access to many of the operations like sending messages, subscribing to events, and manipulating headers. The `IBus` interface was available through dependency injection in message handlers, any custom components that were registered in the container and through other mechanisms like the saga base class.
 
+
+## Why was IBus deprecated?
```
SimonCropp: i know this approach is in other places but i dislike rhetorical questions as titles. how about "Reasons for IBus deprecated"

nservicebus/upgrades/5to6-moving-away-from-ibus.md
```diff
@@ -7,28 +7,38 @@ Starting with Version 6, the `IBus` interface has been deprecated.
...
+
+The problem with `IBus` is that the interface contained numerous methods and properties. However, not all of them are valid in certain scenarios. For example, methods like `Reply` and `ForwardCurrentMessageTo` are always available on the `IBus` interface, but it only makes sense to use those in the context of handling an incoming message. Using them in other scenarios will throw an exception.
```
SimonCropp: > The `IBus` interface contained numerous methods and properties. However, not all of them are valid in all scenarios where an instance of `IBus` was exposed. This lead to confusion over which API were valid or recommended in these scenarios. For example, the methods `Reply` and `ForwardCurrentMessageTo` are always available on the `IBus` interface, but it only makes sense to use those in the context of handling an incoming message. Using them in other scenarios will throw an exception.

nservicebus/upgrades/5to6-moving-away-from-ibus.md
```diff
@@ -8,11 +8,11 @@ Starting with Version 6, the `IBus` interface has been deprecated.
...
 
 
-## Why was IBus deprecated?
+## Reasons for deprecating IBus
 
-The problem with `IBus` is that the interface contained numerous methods and properties. However, not all of them are valid in certain scenarios. For example, methods like `Reply` and `ForwardCurrentMessageTo` are always available on the `IBus` interface, but it only makes sense to use those in the context of handling an incoming message. Using them in other scenarios will throw an exception.
+The `IBus` interface contained numerous methods and properties. However, not all of them are valid in all scenarios where an instance of `IBus` was exposed and this can lead to confusion. For example, the methods `Reply` and `ForwardCurrentMessageTo` are always available on the `IBus` interface, but it is only relevant to use it in the context of handling an incoming message. Using them in other scenarios will throw an exception.
```
boblangley: >but it is only relevant to use it in the context of handling an incoming message.

The subject are the methods not the bus, I would change to:

>but they are only relevant to use in the context of handling an incoming message.

Switching from **it** to **they** also ties this sentence better into the following one.

indualagarsamy: yep. thanks. done.

nservicebus/upgrades/5to6-moving-away-from-ibus.md
```diff
@@ -7,28 +7,38 @@ Starting with Version 6, the `IBus` interface has been deprecated.
...
 
+
+## Reasons for deprecating IBus
+
+The `IBus` interface contained numerous methods and properties. However, not all of them are valid in all scenarios where an instance of `IBus` was exposed and this can lead to confusion. For example, the methods `Reply` and `ForwardCurrentMessageTo` are always available on the `IBus` interface, but they are only relevant to use it in the context of handling an incoming message. Using them in other scenarios will throw an exception.
+
+For users who are new to the NServiceBus API, the fact that `IBus` was available in message handlers using Dependency Injection wasn't obvious, especially when trying to send or publish messages from inside message handlers. Rather than using dependency injection, a `IMessageHandlerContext` parameter is now available in all the message handlers. This parameter will expose all of the appropriate actions to the message handler. Methods that aren't applicable will not be available making the API simpler.
```
danielmarbach: There is also a technical reason. Previously IBud impl was bound to a per request scope which sometimes was implemented as PerThreadScope. This doesn't work with async see my floating state post. Should we mention it?

PR #1703, opened 2016-07-19
Don't hide upgrading existing handlers under a separate heading

DavidBoike:
I went looking for this link and couldn't find it, in part because the section "Message handlers" that I thought should have linked to it did not contain the link. There's no reason for a separate heading containing one line of text.

### DIFF
--- a/nservicebus/upgrades/5to6.md
+++ b/nservicebus/upgrades/5to6.md
@@ -45,15 +45,15 @@ WARNING: Do not `return null` from the message handlers. A `null` will result in
 
 snippet:5to6-messagehandler
 
+For a step by step upgrade of existing handlers or sagas, see: [Migrate existing handlers/sagas](/nservicebus/upgrades/5to6-migrate-existing-handlers.md).
+
+
 ### Bus Send and Receive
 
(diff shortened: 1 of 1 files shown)


nservicebus/upgrades/5to6.md
```diff
@@ -45,15 +45,15 @@ WARNING: Do not `return null` from the message handlers. A `null` will result in
 
 snippet:5to6-messagehandler
 
+For a step by step upgrade of existing handlers or sagas, see: [Migrate existing handlers/sagas](/nservicebus/upgrades/5to6-migrate-existing-handlers.md).
```
indualagarsamy: Rather than remove the heading, why not put the "Migrate existing handlers" right below the handlers, i.e. right above, "Bus Send and Receive" ?

DavidBoike: So in a document about **migrating** from V5 to V6, in a section named **Message handlers**, we need a subheading called **Migrating existing handlers**?

If that adds understanding, that's cool. I question the need to expand the (already mammoth) TOC any further though.

indualagarsamy: @DavidBoike - fair enough.

PR #1704, opened 2016-07-20
Update viewing-message-content-in-msmq.md

martys:
I can't upload an image. Could someone that will be approving this, also add an image to the Mqueue viewer section from http://www.mqueue.net/


SimonCropp:
@martys thanks. how does this look http://docs.particular.net/nservicebus/msmq/viewing-message-content-in-msmq#mqueue-viewer  ?

martys:
Great, thank You

Martynas

Issue #1694, opened 2016-07-14
Document serializer and deserializer extension model


danielmarbach:
See https://github.com/Particular/NServiceBus/pull/3923

PR #1706, opened 2016-07-21
[WIP] Additional settings for serializers.

Scooletz:
Connects to #1694 

Requires at least `NServiceBus  6.0.0-Beta6`.

### DIFF
--- a/Snippets/Core/Core_6/Serialization/RegisterCustomSerializer.cs
+++ b/Snippets/Core/Core_6/Serialization/RegisterCustomSerializer.cs
@@ -1,4 +1,4 @@
-﻿namespace Core6
+﻿namespace Core6.Serialization
 {
     using System;
     using System.Collections.Generic;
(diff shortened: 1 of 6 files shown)


Scooletz:
@danielmarbach Is that describes this capability properly or should I move it to the https://github.com/Particular/docs.particular.net/blob/master/nservicebus/serialization/custom-serializer.md and describe it from the custom serializer point of view?

danielmarbach:
I think it would not hurt to show for the custom serializer how to provide such a settings extension method

danielmarbach:
Apart from that I think this addition here is already a good start

Scooletz:
Ping @danielmarbach  I've added settings to the custom serializer. Could you take a look?

Snippets/Core/Core_6/Serialization/RegisterCustomSerializer.cs
```diff
@@ -26,13 +27,19 @@ class MyCustomSerializerDefinition :
...
             }
         }
 
         class MyCustomSerializer :
             IMessageSerializer
         {
+            public MyCustomSerializer(string settingsValue)
+            {
+                // Add code initializing serializer on the basis of settingsValue
+                throw new NotImplementedException();
```
danielmarbach: I would remove the throw Exception

PR #1710, opened 2016-07-22
Add reference to Microsoft.WindowsAzure.SDK package

mstolk:


SimonCropp:
I think the added text needs to be moved to in front of the period

ramonsmits:
@mstolk As we cannot edit your branch you will need to make the edit before we can merge this.

PR #1724, opened 2016-07-28
Update showcase sample

### LINKED ISSUE #1717: Update the Showcase / on-premise sample to the latest (client) dependencies
The [on-premise sample](http://docs.particular.net/samples/show-case/on-premise/) `AngularJS` version is 1 year old and should be updated to the latest 1.5.x.
`Bootstrap` has just released a new version that is compatible with `JQuery` 3, I'll double check that `SignalR` is compatible with that as well.

- [ ] V5
- [ ] V6

mauroservienti:
Connects to #1717 
Fix #1717 

Do you prefer to have a single squashed commit?


weralabaj:
> Do you prefer to have a single squashed commit?

We can squash when merging, no worries :)

PR #1726, opened 2016-07-28
Safeguard message retries

WojcikMike:
Connects to: https://github.com/Particular/V6Launch/issues/77

Waits for https://github.com/Particular/ServicePulse/pull/364

This PR is to describe new functionalities of pending retry screen and redirects functionalities.


WojcikMike:
I will update the exact error messages and screenshots when we finish work on the UI

boblangley:
Added images and updated some of the wording. Rebased with master and forced push. Please re-checkout this branch.

boblangley:
@johnsimons Can you review?

SimonCropp:
@boblangley @WojcikMike is this ready for a final review?

WojcikMike:
@SimonCropp I would say yes. But it still need to wait for the SC+SP being released as we would need to provide proper version numbers.

indualagarsamy:
@WojcikMike - There is a build failure on this. Can you check the missing snippet error?

```
System.Exception : Could not render page: C:\BuildAgent\work\542e5eb63d1ae8b\nservicebus\hosting\cloud-services-host\critical.md
  ----> System.Exception : Failed to render article. PackageId:NServiceBus.Hosting.Azure. Version:[7.0.0-pre, 8.0.0).
  ----> CaptureSnippets.MissingSnippetsException : MissingSnippet. 
  Line: 14
  Key: DefineCriticalErrorActionForAzureHost
```

servicepulse/intro-pending-retries.md
```diff
@@ -0,0 +1,59 @@
+---
+title: Pending Retries Message Management in ServicePulse
```
indualagarsamy: Would `Managing Archived Messages in ServicePulse` be better for the title?

servicepulse/intro-pending-retries.md
```diff
@@ -0,0 +1,59 @@
...
+summary: Describes how ServicePulse monitors archived failed messages, and allows unarchiving archived failed messages.
```
indualagarsamy: Is this summary the correct summary?

servicepulse/intro-pending-retries.md
indualagarsamy: Isn't pending just one stage of retrying a message. If so, can we have the title as "Retrying failed messages in Service Pulse" and have the pending status and explanations for it be part of this? The title seems to be very implementation specific and not sure how friendly it is to the user.

indualagarsamy: I think if we have add a brief introduction in the [intro to failed messages page](http://docs.particular.net/servicepulse/intro-failed-messages#failed-message-retry) about retries and point the users to this page, then this page can focus on retrying a message including what happens to the pending retries. Thoughts?

servicepulse/intro-pending-retries.md
```diff
@@ -0,0 +1,59 @@
...
+tags:
+ - ServicePulse
+reviewed: 2016-07-28
+---
+
+
+In Versions **1.6.6** and above ServicePulse includes a screen to view and manage failed messages that have been sent for retry but are pending final status.
```
indualagarsamy: >In Versions **1.6.6** and above ServicePulse includes a screen to view and manage failed messages that have been sent for retry but are pending final status.

How about:

In Versions **1.6.6** and above, ServicePulse includes an additional screen to view and manage failed messages that have been requested to be retried but have not completed yet.

servicepulse/intro-pending-retries.md
```diff
@@ -0,0 +1,59 @@
...
+
+Pending retries messages can be found by going to the pending retries screen.
+
+![Pending Retries Tab](images/pending-retries.png 'width=500')
```
indualagarsamy: Do we need a screen shot of the whole Failed Messages screen to show the user the Pending Retries link?

servicepulse/intro-pending-retries.md
```diff
@@ -0,0 +1,59 @@
...
+
+
+### Pending Retries Messages
+
+The Pending Retries Messages screen shows failed messages which have been retried but the status of that retry is pending. The status of retried failures is updated when either the message is processed again as either an Audited message (success) or as a failed message.
```
indualagarsamy: >The Pending Retries Messages screen shows failed messages which have been retried

needs a comma after the retried.

servicepulse/intro-pending-retries.md
```diff
@@ -0,0 +1,59 @@
...
+
+Failed messages that are retried may stay in the pending state for the following reasons:
+ - The retrying endpoint is not working (e.g. crashed or is scaled-out) and the retried messages is waiting in the queue and has not yet been processed.
+ - The retrying endpoint does not have auditing enabled but has successfully processed the retried message.
```
indualagarsamy: What is the scaled-out scenario? I understand the crashed scenario, i.e. process not running. What about the scaled out scenario. In this case, there are 2 instances running, won't one of them process it? Can you clarify?

I am not sure if calling the endpoint as the retrying endpoint is right. How about:

Failed messages that are retried may stay in the pending state for the following reasons:

- If the endpoint where the retried message is sent to is not running, e.g. crashed, then the retried messages would be waiting in the queue. 
- If auditing is not setup for the endpoint where the retried message is sent to and the message has been successfully processed by that endpoint.

servicepulse/intro-pending-retries.md
```diff
@@ -0,0 +1,59 @@
...
+ * **Failure Timestamp:** Timestamp when the message have failed.
+ * **Endpoint** Processing endpoint.
+ * **Machine** Machine on which the processing endpoint was operating.
+ * **Redirect** Information if redirect is created for this queue.
+ * **StackTrace:** Displays the full .NET exception stacktrace.
+ * **Headers:** Displays a complete set of message headers.
+ * **Body:** Displays the serialized message body.
+ * **Copy Message Id:** Copies the message unique ID to the clipboard.
+ * **Open in ServiceInsight:** Launches [ServiceInsight](/serviceinsight/), focusing on the message for in-depth analysis of the failure causes. This only works if ServiceInsight is installed on the local machine.
+
```
indualagarsamy: Do we need to have the info such as "Failure timestamp", "endpoint", "machine", etc., here? Doesn't this belong in the failed messages content? I think a link is sufficient.
If the failed messages doco doesn't have it, IMO, this should move to that page.

servicepulse/intro-pending-retries.md
```diff
@@ -0,0 +1,59 @@
...
+To limit the time period of displayed messages select an option from the available predefined time periods, by default pending messages for all time are displayed.
+
+![Period Filter](images/pending-retries-period-selection.png 'width=500')
+
+Results can be filtered by queue name using the search functionality:
+
+![Queue Filter](images/pending-retries-filter-queues.png 'width=500')
+
+Displayed messages display information in the same way as on [Failed Message](intro-failed-messages.md).
```
indualagarsamy: >Displayed messages display information in the same way as on [Failed Message](intro-failed-messages.md).

How about:
The information about the message such as Failure timestamp, endpoint, stack trace of the error, etc.,  is displayed in the same manner as it on the [Failed Messages]((intro-failed-messages.md) page.

servicepulse/intro-pending-retries.md
```diff
@@ -0,0 +1,59 @@
...
+ * **Redirect** Information if redirect is created for this queue.
+ * **StackTrace:** Displays the full .NET exception stacktrace.
+ * **Headers:** Displays a complete set of message headers.
+ * **Body:** Displays the serialized message body.
+ * **Copy Message Id:** Copies the message unique ID to the clipboard.
+ * **Open in ServiceInsight:** Launches [ServiceInsight](/serviceinsight/), focusing on the message for in-depth analysis of the failure causes. This only works if ServiceInsight is installed on the local machine.
+
+### Message Retry
+
+WARNING: Pending retried messages can be retried however this feature should be used with care. It will create a duplicate of a message to be processed within the system. Person should ensure that this message was not processed or was processed incorrectly before using retry feature.
```
indualagarsamy: >Pending retried messages can be retried however this feature should be used with care. It will create a duplicate of a message to be processed within the system.

How about:

Failed messages that are currently in the pending status can be retried again, however this feature should be used with care. Retrying pending messages can cause the endpoint to reprocess the same message multiple times.

indualagarsamy: > Person should ensure that this message was not processed or was processed incorrectly before using retry feature.

What does a person need to do to ensure that this message wasn't processed again?

servicepulse/intro-pending-retries.md
```diff
@@ -0,0 +1,59 @@
...
+
+### Pending Retries Messages
+
+The Pending Retries Messages screen shows failed messages which have been retried but the status of that retry is pending. The status of retried failures is updated when either the message is processed again as either an Audited message (success) or as a failed message.
+
+Failed messages that are retried may stay in the pending state for the following reasons:
+ - The retrying endpoint is not working (e.g. crashed or is scaled-out) and the retried messages is waiting in the queue and has not yet been processed.
+ - The retrying endpoint does not have auditing enabled but has successfully processed the retried message.
+
+To limit the time period of displayed messages select an option from the available predefined time periods, by default pending messages for all time are displayed.
```
indualagarsamy: > To limit the time period of displayed messages select an option from the available predefined time periods, by default pending messages for all time are displayed.

How about:

The messages displayed in this screen can be filtered based on the time period by selecting one of the options, such as messages in the last 2 hours, messages in the previous day or week. The default option is set to display all of the pending messages.

servicepulse/intro-pending-retries.md
```diff
@@ -0,0 +1,59 @@
...
+
+![Period Filter](images/pending-retries-period-selection.png 'width=500')
```
indualagarsamy: Do we need this screen shot? Screen shots are nice, but a pain when it comes to maintenance.

servicepulse/intro-pending-retries.md
```diff
@@ -0,0 +1,59 @@
...
+
+Results can be filtered by queue name using the search functionality:
```
indualagarsamy: I think this should be 
Results can also be filtered by queue name...

servicepulse/intro-pending-retries.md
```diff
@@ -0,0 +1,59 @@
...
+ * **Copy Message Id:** Copies the message unique ID to the clipboard.
+ * **Open in ServiceInsight:** Launches [ServiceInsight](/serviceinsight/), focusing on the message for in-depth analysis of the failure causes. This only works if ServiceInsight is installed on the local machine.
+
+### Message Retry
+
+WARNING: Pending retried messages can be retried however this feature should be used with care. It will create a duplicate of a message to be processed within the system. Person should ensure that this message was not processed or was processed incorrectly before using retry feature.
+
+To retry a pending retry message, select the failed message(s) in the list and click the "Retry Selected" button.
+
+Alternately a queue can be selected and "Retry All" option chosed which will retry all messages for that queue.
```
indualagarsamy: >Alternately a queue can be selected and "Retry All" option chosed which will retry all messages for that queue.

How about:
Alternately a queue can be selected and the "Retry All" option can be used to retry all the messages targeted for the queue.

servicepulse/intro-pending-retries.md
```diff
@@ -0,0 +1,59 @@
...
+
+Message retry will use [message redirects](redirect.md) if the original endpoint has been redirected in ServicePulse.
```
indualagarsamy: What does this mean, can you expand this?

servicepulse/intro-pending-retries.md
```diff
@@ -0,0 +1,59 @@
...
+
+WARNING: A pending retry message that is sent for retry will remain in the pending retry list until it is resolved or fails again.
```
indualagarsamy: >WARNING: A pending retry message that is sent for retry will remain in the pending retry list until it is resolved or fails again.

Isn;t that true also for messages that have the "Pending" status that the user hasn't explicitly retried?

servicepulse/intro-pending-retries.md
```diff
@@ -0,0 +1,59 @@
...
+
+### Mark as complete
+
+When audits are disabled pending retries will remain in that state indefinitely, even when successfully processed. Use the `Mark as complete` feature to manually mark the failed message as resolved. Once the message is marked as completed it will no longer appear in the pending retries message list.
```
indualagarsamy: How about:

When the audit feature is disabled in the endpoint that processes the failed message, the entry will remain in the pending state indefinitely even after the message has been successfully reprocessed by that endpoint. In this scenario, use the `Mark as complete` feature to manually mark the failed message as resolved. Once the message is marked as resolved, it will no longer appear in the pending retries message list.

servicepulse/redirect.md
```diff
@@ -0,0 +1,45 @@
...
+summary: Describes what message redirects are and how to use them.
+tags:
+- ServicePulse
+related:
+- servicepulse/troubleshooting
+---
+
+## Message Redirects
+
+In Versions **1.6.6** and above ServicePulse includes a screen to view and manage failed message redirects that have been created to send failed messages to an alternate queue when the original processing endpoint address is no longer available.
```
indualagarsamy: How about:

When a failed message needs to be retried, but the destination endpoint no longer exists and the message needs to be routed to a different endpoint than what is specified in the failed message headers, ServicePulse offers the redirect feature. This feature is only available from Versions 1.6.6 and above.

servicepulse/redirect.md
```diff
@@ -0,0 +1,45 @@
...
+
+INFO: Message redirects are only a feature of ServicePulse/ServiceControl and will not alter the routing for NServiceBus endpoints.
```
indualagarsamy: should this be INFO or a WARNING. I am not sure.

servicepulse/redirect.md
```diff
@@ -0,0 +1,45 @@
...
+
+To manage redirects go to Configuration page and select tab called `Retry Redirects`.
```
indualagarsamy: How about:
Redirects can be managed from the Configuration page by selecting the "Retry Redirects" tab.

servicepulse/redirect.md
```diff
@@ -0,0 +1,45 @@
...
+
+![Redirects Tab](images/redirects.png 'width=500')
+
+### Managing redirects
+On the redirect page a list of redirects is being displayed containing following information.
```
indualagarsamy: How about:

This page displays the configured redirects that are currently in effect along with the following information:

servicepulse/redirect.md
```diff
@@ -0,0 +1,45 @@
...
+ - **Destination Queue** This is the queue that will be new destination when doing retry.
+ - **Last modified** This is a timestamp of last modification.
+
+To create a redirect click the `Create Redirect` button. A dialog will appear.
+
+![Create Redirects Dialog](images/redirects-create.png 'width=500')
+
+Choose a source queue from the dropdown. Enter the queue name for the target of the redirect in `To Physical Address` input. To immediately retry all unresolved failed messages from the sources address check the box with the label `Immediately retry any matching failed messages`.
+
+Once created every redirect can be deleted using `End Redirect` link. This action requires an confirmation.
```
indualagarsamy: > Once created every redirect can be deleted using `End Redirect` link. This action requires an confirmation. 

How about:
Any existing redirects can be deleted using ...

servicepulse/redirect.md
```diff
@@ -0,0 +1,45 @@
...
+
+To create a redirect click the `Create Redirect` button. A dialog will appear.
+
+![Create Redirects Dialog](images/redirects-create.png 'width=500')
+
+Choose a source queue from the dropdown. Enter the queue name for the target of the redirect in `To Physical Address` input. To immediately retry all unresolved failed messages from the sources address check the box with the label `Immediately retry any matching failed messages`.
+
+Once created every redirect can be deleted using `End Redirect` link. This action requires an confirmation. 
+
+To change the target of a redirect click the `Modify Redirect` link. Only the target queue can be modified, the source queue remains read-only.
```
indualagarsamy: missing a comma after: `To change the target of a redirect`

servicepulse/redirect.md
```diff
@@ -0,0 +1,45 @@
...
+
+#### Validation errors
+
+When adding and editing redirects following validation rules are checked:
```
indualagarsamy: >When adding and editing redirects following validation rules are checked:

When adding and editing redirects the following validation rules are applied:

servicepulse/redirect.md
```diff
@@ -0,0 +1,45 @@
...
+ - **Duplicate** There can not be more than one rule for given Source Queue. In that case the following message will be displayed: "Can not create redirect to a queue *QueueName* as it already has a redirect. Provide a different queue or end the redirect."
+ - **Dependent** There can never be two redirects that one destination queue in one is a destination queue in another. To use an example: when there is a rule QueueA -> QueueB. There can not be a second rule QueueB -> QueueC or QueueC -> QueueA. When this error occur the following message will be displayed: "Failed to create a redirect, can not create a redirect to a queue that already has a redirect or is a target of a redirect."
```
indualagarsamy: >- **Dependent** There can never be two redirects that one destination queue in one is a destination queue in another. To use an example: when there is a rule QueueA -> QueueB. There can not be a second rule QueueB -> QueueC or QueueC -> QueueA. When this error occur the following message will be displayed: "Failed to create a redirect, can not create a redirect to a queue that already has a redirect or is a target of a redirect."

Does chaining explain the concept better? Can you check the following?

- **Dependent** - Redirects cannot be chained i.e., a destination queue in one redirect cannot be the source queue for another redirect. For example, if there is a configured redirect that redirects messages from QueueA -> QueueB, there cannot be a second rule that's configured to redirect from QueueB -> QueueC. An error message will be displayed when attempting to chain multiple redirects.

servicepulse/redirect.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Immediately retry any matching failed messages
```
indualagarsamy: Can this be explained before the Validation errors?, since its common to adding and modifying a redirect?

servicepulse/redirect.md
```diff
@@ -0,0 +1,45 @@
...
+When editing or adding a redirect a checkbox is present allowing to immediately retry all failed messages to which this rule will apply. This will apply only to messages that are failed and not the ones that are pending for retry.
```
indualagarsamy: > When editing or adding a redirect a checkbox is present allowing to immediately retry all failed messages to which this rule will apply. This will apply only to messages that are failed and not the ones that are pending for retry.

Selecting this option will immediately retry any failed messages that match the redirect rule. This does not apply to failed messages whose retry status is in the `Pending` state.

indualagarsamy:
@WojcikMike - I've reviewed the PR and provided some suggestions for some text. Please review. 
@Particular/docs-maintainers - do we have any guidelines on the screen shots and how liberally we are to use it etc? My thoughts on screen shots is to use it minimally absolutely where needed, as updating them is painful. Consider the scenario where the article is correct, but the UI changes and now we have to update this. Can you please weigh in?

weralabaj:
@indualagarsamy I agree that too many is a lot of paing, but generally we don't have that many, mainly for SP and SI. I'm ok with using more, if they make the text easier to follow.

servicepulse/intro-pending-retries.md
WojcikMike: Change introduced

servicepulse/intro-pending-retries.md
WojcikMike: Would you prefer just a screen shot of top part?

servicepulse/intro-pending-retries.md
WojcikMike: introduced

servicepulse/intro-pending-retries.md
WojcikMike: With regards to scale-out what you say is true if one is using competing consumers, than at lease 1 consumer makes the trick. However if one is using distributor or Sender Side Distribution like here:
http://docs.particular.net/nservicebus/msmq/scalability-and-ha/sender-side-distribution#mapping-physical-endpoint-instances
So basically for MSMQ scale-out, than you need to move the message from 1 queue to another. Having said that what if we change your text to something like:

If the endpoint where the retried message is sent to is not running, e.g. crashed or scaled-out using [Sender-side distribution](/nservicebus/msmq/scalability-and-ha/sender-side-distribution), then the retried messages would be waiting in the queue.

servicepulse/intro-pending-retries.md
WojcikMike: There is a difference, In here there is Redirect information present which doesn't exist on failed screen. Should we link failed message screen and say on top of that there is a redirect information?

servicepulse/intro-pending-retries.md
WojcikMike: I added this:

The information about the message such as Failure timestamp, endpoint, stack trace of the error, etc.,  is displayed in the same manner as it on the [Failed Messages](intro-failed-messages.md) page providing additional information as follows:

followed by Redirect point

servicepulse/intro-pending-retries.md
WojcikMike: Agreed, introduced

servicepulse/intro-pending-retries.md
WojcikMike: Done

servicepulse/intro-pending-retries.md
WojcikMike: Applied

servicepulse/intro-pending-retries.md
WojcikMike: If you defined a redirect for a given queue A such that A -> B

Then when using retry the message will be send to B.

servicepulse/intro-pending-retries.md
WojcikMike: Message will be in pending retry state after being retried by the used (both from failed and from pending retry screens).

servicepulse/intro-pending-retries.md
WojcikMike: introduced

servicepulse/redirect.md
WojcikMike: I am not sure as well. I don't want anyone to look at it as a routing override. Let me change it for warning

servicepulse/redirect.md
WojcikMike: introduced

servicepulse/redirect.md
WojcikMike: introduced

servicepulse/redirect.md
```diff
@@ -0,0 +1,45 @@
...
+Choose a source queue from the dropdown. Enter the queue name for the target of the redirect in `To Physical Address` input. To immediately retry all unresolved failed messages from the sources address check the box with the label `Immediately retry any matching failed messages`.
+
+Once created every redirect can be deleted using `End Redirect` link. This action requires an confirmation. 
+
+To change the target of a redirect click the `Modify Redirect` link. Only the target queue can be modified, the source queue remains read-only.
+
+#### Validation errors
+
+When adding and editing redirects following validation rules are checked:
+ - **Duplicate** There can not be more than one rule for given Source Queue. In that case the following message will be displayed: "Can not create redirect to a queue *QueueName* as it already has a redirect. Provide a different queue or end the redirect."
```
WojcikMike: As the message can be confusing by keeping it mostly the same I allow the user to google our doco page, which sheds more light. I vote for leaving the error message

servicepulse/redirect.md
WojcikMike: Yes chain is the good explanation

servicepulse/redirect.md
WojcikMike: Moved

servicepulse/redirect.md
WojcikMike: Done

WojcikMike:
@indualagarsamy I applied most of the messages raised. I am not sold on division between failed messages and retry as you suggest, but let me pick @boblangley brain on that. I looked for the snipper error but it seems to work on my end (code + doco), let's see if next build failes as well

WojcikMike:
I rebased the changes to fix the build, as this is probable solution

WojcikMike:
@Particular/docs-maintainers the Task Force believe that this is good to go, please have a look

SimonCropp:
Should these pages appear in the menu?

SimonCropp:
Seem there are two outstanding comments from indu. Is that just a github rendering issue?

WojcikMike:
@SimonCropp thanks for the question, I completely forgot about that. Yes I added them

SimonCropp:
now needs a rebase :)

WojcikMike:
Yeah I saw that :). Done

WojcikMike:
@SimonCropp 
> Seem there are two outstanding comments from indu. Is that just a github rendering issue?

You are right, let me talk to @boblangley about it today

servicepulse/intro-pending-retries.md
boblangley: @indualagarsamy I think the second suggestion works better, I have added a commit with that link.

boblangley:
@WojcikMike and @SimonCropp I added a commit which adopted the second suggestion for change by @indualagarsamy. 

@indualagarsamy I agree in general regarding using the minimum of images to communicate effectively with our customers, but with both Service Puse and Service Insight having UI's they will require more than the average doco page. Of course we think we struck the appropriate balance. It would help more if you can point to a specific image we included that you do not think is necessary so we can discuss.

boblangley:
From @indualagarsamy:
> Hi. Regarding the images, there was a full screenshot of a page, just to show a button and there was no red pointer to the button. I'll see if I can find that screenshot

I can see now what you mean. The intent was to just show them how to get to the page, I will adjust.

boblangley:
I focused the screenshot on the navigation elements and spoke with @indualagarsamy about it.

SimonCropp:
@boblangley @WojcikMike should this still be marked WIP?

boblangley:
That was an oversight. I think it is ready to merge.

PR #1734, opened 2016-08-02
Update NServiceBus.NHibernate to 7.0.0-Beta5

MarcinHoppe:
Connects to Particular/V6Launch#65


MarcinHoppe:
@Particular/docs-maintainers I think this can be merged now, it's only bump of NHibernate packages to 7.0.0-Beta5.

SzymonPobiega:
The `reference` nodes in the `csproj`s now contain the assembly name but I guess that's fine, right?

MarcinHoppe:
You mean assembly info such as version, etc:

`NServiceBus.NHibernate, Version=7.0.0.0, Culture=neutral, PublicKeyToken=9fc386479f8a226c, processorArchitecture=MSIL`

I think this is not necessary. I will remove this and leave just assemby name (`NServiceBus.NHibernate`).

PR #1736, opened 2016-08-02
FeatureStartupTask note clarification

SeanFeldman:
The differences of internal details of implementation between version 5 and 6 should not matter. Final result to the users is the same. In v6, we're awaiting for all the `FeatureStartupTask`s to complete, so in essence delaying the endpoint startup. Just like in v5.

@Particular/nservicebus-maintainers please confirm and let know if the change makes sense.

### DIFF
--- a/nservicebus/pipeline/features.md
+++ b/nservicebus/pipeline/features.md
@@ -124,9 +124,7 @@ snippet:FeatureStartupTaskRegistration
 
 The task will only be created and called if the feature is enabled. The `FeatureStartupTask`s are activated and started in random order.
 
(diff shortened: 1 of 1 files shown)


danielmarbach:
Why not leave as is?

bording:
> Why not leave as is?

The problem with leaving the docs unchanged is that it very heavily implies that long-running tasks are ok now, but that isn't true, because all startup tasks have to complete before we complete the endpoint startup process.

timbussmann:
I'm good with that. The previous text was probably a bit too detailed on that level.

PR #1751, opened 2016-08-09
WIP: Implement a custom serializer

johnsimons:
Since the interface is so simple and this is an advance feature, not sure it is worth saying more about it ?

### DIFF
--- a/nservicebus/serialization/custom-serializer.md
+++ b/nservicebus/serialization/custom-serializer.md
@@ -1,21 +1,28 @@
 ---
-title: How to register a custom serializer
-reviewed: 2016-03-17
+title: Custom serializers
+summary: How to create and register custom serializer for NServiceBus.
+component: Core
+reviewed: 2016-08-17
 tags:
 - serialization
 redirects:
 - nservicebus/how-to-register-a-custom-serializer
 ---
 
-An example how to register a custom serializer. 
-
 
 ### Create a custom serializer
-Create a custom serializer which additionally requires a configuration value.
+
+A custom serializer needs to implement `IMessageSerializer` interface:
 
 snippet: CustomSerializer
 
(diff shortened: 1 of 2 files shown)


johnsimons:
Feedback from @SimonCropp:

also in the scope of someone building a serializer how can they possibly ensure "does not clash with any existing registered serializer". they have no idea what other serializers could possible be registered
u have not mentioned the versions
u have not addressed transitioning between versions

johnsimons:
@SimonCropp do you have any suggestions how to better address the points you raised ?

SimonCropp:
@johnsimons no i dont know how to address those. you are the expert on that API. so if u dont have the answers then the API is broken

johnsimons:
Versioning is a tough one, but swapping serializers (xml to json) works reasonably well.
I think we can just say that versioning is not supported ?

weralabaj:
Some ideas on how to move it forward:

> Community run serializers

Apart from this I think it's worth either linking or at least adding as a related header the samples we have, e.g. http://docs.particular.net/samples/serializers/.

> ## How to implement your own serializer
 
 > To implement a serializer all it needs to be done is implement `IMessageSerializer`.
 > The only tricky bit is to ensure that the string returned by `ContentType` does not clash with any existing registered serializer, the `ContentType` string is used internally as a dictionary key lookup.

How about:

- Move it to http://docs.particular.net/nservicebus/serialization/#specifying-additional-deserializers
- Phrase as: "Note: When using multiple deserializers make sure that there's only one type registered per `ContentType`."

Not sure if that also matters in other scenarios? Like when there's clash with default serializer?

weralabaj:
@johnsimons do you plan on updating it or should I take over?

johnsimons:
@weralabaj modified according to your review.

johnsimons:
all your now

weralabaj:
rebased

PR #1754, opened 2016-08-09
Update azure host to core v7

andreasohlund:


andreasohlund:
Should be good to go once the build is green

PR #1762, opened 2016-08-10
Update routing samples to beta7

SzymonPobiega:
@timbussmann @DavidBoike please review.


samples/routing/fair-distribution/Core_6/Client/Program.cs
```diff
@@ -83,9 +83,9 @@ static void AddRouting(EndpointConfiguration endpointConfiguration)
         #region Routing
 
         const string server = "Samples.FairDistribution.Server";
-        var routing = endpointConfiguration.Routing();
+        var routing = endpointConfiguration.UseTransport<MsmqTransport>().Routing();
         routing.RouteToEndpoint(typeof(PlaceOrder), server);
-        routing.Mapping.Physical.Add(
+        routing.GetSettings().GetOrCreate<EndpointInstances>().Add(
```
SimonCropp: this is not a valid API

SimonCropp: ie we cant, with a straight face, ask people to use this

SzymonPobiega: Would an instance mapping file be better here? I didn't want to include it here to not broaden the scope of the sample.

timbussmann: that sample would now require a feature using `RoutingPluginSettingsExtensions`?

SzymonPobiega: This is just showing another way of distributing so I think we should bite the bullet and add the file-based mapping. In real life that's what users would do.

weralabaj: > we should bite the bullet and add the file-based mapping. In real life that's what users would do.

If that's more realistic then I'm :+1:, even if that will make the sample a bit more complicated.

SimonCropp: what makes this routing API special that it should only be used from inside a feature?

timbussmann: @SimonCropp we wanted to avoid users hardcoding routing information. That API can be used by plugins to provide routing information from a custom information source and in that case you will most probably already have a feature or it's ok to expect the users plugin this functionality with a feature.

SimonCropp: how does restricting its usage to feature stop people from hard coding routing?

timbussmann: the goal was separate these more advanced APIs from the basic API so it does not confuse users which just try to setup basic routes. It does of course not prevent users from hard coding it, but we wanted to prevent them from going "ah, I'll just use this "RouteToAddress" overload here. We're very open to suggestions.

SimonCropp: i prefer the previous API

timbussmann: @SimonCropp are we talking about a scenario where you want to configure a machine name for a route or do you have another scenario in mind which would use this API?

SimonCropp: it doesnt matter what scenario we r talking about. obscuring an api into only being accessible through a feature, when that api is not a "feature api", makes no sense.

timbussmann: I understand the following two concerns:
* the api itself ("i prefer the previous API")
* the location of the "advanced routing API"

again, we're open for good ideas to make an API available in a feature and the endpoint config without directly "competing" with the basic routing API.

samples/routing/custom/Core_6/Sales/Program.cs
```diff
@@ -14,8 +14,7 @@ static void Main()
     {
         Console.Title = "Samples.CustomRouting.Sales.1";
         var endpointConfiguration = new EndpointConfiguration("Samples.CustomRouting.Sales");
-        var scaleOut = endpointConfiguration.ScaleOut();
-        scaleOut.InstanceDiscriminator("1");
+        endpointConfiguration.MakeInstanceUniquelyAddressable("1");
```
DavidBoike: Can we make the string "instance1" or something a little more meaningful?

samples/routing/fair-distribution/Core_6/Server/Program.cs
```diff
@@ -13,8 +13,7 @@ static void Main()
     {
         Console.Title = "Samples.FairDistribution.Server.1";
         var endpointConfiguration = new EndpointConfiguration("Samples.FairDistribution.Server");
-        var scaleOut = endpointConfiguration.ScaleOut();
-        scaleOut.InstanceDiscriminator("1");
+        endpointConfiguration.MakeInstanceUniquelyAddressable("1");
```
DavidBoike: Another instance of "1" as a discriminator that should be more descriptive.

samples/routing/fair-distribution/Core_6/Server2/Program.cs
```diff
@@ -13,8 +13,7 @@ static void Main()
     {
         Console.Title = "Samples.FairDistribution.Server.2";
         var endpointConfiguration = new EndpointConfiguration("Samples.FairDistribution.Server");
-        var scaleOut = endpointConfiguration.ScaleOut();
-        scaleOut.InstanceDiscriminator("2");
+        endpointConfiguration.MakeInstanceUniquelyAddressable("2");
```
DavidBoike: More descriptive descriminator

samples/routing/fair-distribution/Core_6/Client/Program.cs
DavidBoike: Agree with bringing in instance mapping file.

The key here is the distinction between logical mapping to endpoints, which should be accomplished in code by a developer, vs. the physical mapping of logical endpoints to physical machines, which is a deployment time concern and should always be malleable by an operations person. We're very opinionated that you shouldn't be hardcoding that, which means it should not be available/discoverable starting from EndpointConfiguration.

Just putting it behind an `.Advanced.` in the API is just putting the gun on a slightly taller shelf.

So availability through a feature is a way to provide it to a (different use of this word) customized routing feature that would need it. We don't intend for customers to find this themselves. They would see it in a sample or ask us directly. This seems preferable to yet another abstraction along the lines of `IControlLogicalToPhysicalRouting` or something like that.

SimonCropp: -1 to any `.Advanced.` api. the fact of an api being advanced or not is only relative to the consumer. their experience level will be the determining factor on if an api seems advanced.

also there are many other apis in NSB that would also classify as "advanced" based on the above logic. we will just end up polluting people code with verbose APIs in an effort to optimize for people who dont know when to not hard code a machine name. 

This should be handled with documentation. not with changes to the API that will make it more complex for people who know what they are doing

SimonCropp:
@DavidBoike re `MakeInstanceUniquelyAddressable("2")` comments. can u raise a new PR for that since that feedback spans more than just this sample

samples/routing/fair-distribution/Core_6/Client/Program.cs
SzymonPobiega: I think we should just use the file-based mapping as the most obvious option. As for the extensibility APIs, they are in the place they are most intuitive-- on a feature context because extending the routing should be done via a feature.

SzymonPobiega:
I replaced discriminators here with multi-machine simulation hack. I think it is more similar to production usage.

timbussmann:
> I replaced discriminators here with multi-machine simulation hack

tbh I don't think this makes the sample easier to understand

timbussmann:
I updated the sample to beta7 and it seems to be working so far, but it seems to have some issues about the "single route per command" rule. In some of the endpoints you can see how the routes for `OrderAccepted` are constantly switching forth and back. Since this seems to be directly related to your persistence mechanism, do you want to have a closer look at this @SzymonPobiega ?

SzymonPobiega:
@timbussmann I'll take a look. Which sample are you referring to? The automatic routing?

timbussmann:
@SzymonPobiega the custom routing sample, yes

SzymonPobiega:
@timbussmann i've updated the remaining samples to beta7. I fixed the instable route (the automatic routing was registering (unused) command routes for event types. I also changed the queue translation hack to operate on `queue` attribute rather then overriding machine name. When the change in the MSMQ translation in core is done I'll be able to remove the hacks.

weralabaj:
@SzymonPobiega is that ready for review and merge, or are we still waiting for anything? more tests?

SzymonPobiega:
@weralabaj ready.

PR #1764, opened 2016-08-10
Update custom transport sample to core beta6

timbussmann:
basic sample works, but error handling is not working properly. Need to figure out how to work with these transactions.

### DIFF
--- a/samples/custom-transport/Core_6/Endpoint1/Endpoint1.csproj
+++ b/samples/custom-transport/Core_6/Endpoint1/Endpoint1.csproj
@@ -30,7 +30,7 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="NServiceBus.Core">
-      <HintPath>..\..\..\..\packages\NServiceBus.6.0.0-beta0005\lib\net452\NServiceBus.Core.dll</HintPath>
+      <HintPath>..\..\..\..\packages\NServiceBus.6.0.0-beta0006\lib\net452\NServiceBus.Core.dll</HintPath>
     </Reference>
     <Reference Include="System" />
     <Reference Include="System.Core" />
--- a/samples/custom-transport/Core_6/Endpoint1/Endpoint1.csproj
+++ b/samples/custom-transport/Core_6/Endpoint1/Endpoint1.csproj
@@ -29,8 +29,9 @@
     <StartupObject />
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="NServiceBus.Core">
-      <HintPath>..\..\..\..\packages\NServiceBus.6.0.0-beta0006\lib\net452\NServiceBus.Core.dll</HintPath>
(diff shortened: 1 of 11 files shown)


timbussmann:
@SimonCropp @andreasohlund do you think having a transaction mechanism adds a real value to this sample in the first place? (looking at the current state, the pasted code is already quite a lot to read through and not easy to grasp, by implementing the `onError` contract properly. this will potentially add quite a bit of code). It makes sense to have a sample how to use the transaction API, but could we also come up with a simpler sample for that?

andreasohlund:
Since the sample is focued on the how to implement a transport perhaps we can just support "ReceiveOnly" and just delete all the (kid of irrelevant) transaction stuff?

SimonCropp:
@andreasohlund @timbussmann isnt the transaction hoops relevant since in any real world transport people would have to do similar

andreasohlund:
It depends, eg Rabbit and ASQ doesn't have any "transactions" so they just
supports `ReceiveOnly`.

Since transactions would be provided by the underlying queuing system I'd
say that this sample should focus on the seam and not how to implement TX
on top of a file system?

Is there a middle ground, perhaps just put a `//todo: implement
transactions` comment or something?

timbussmann:
> Is there a middle ground, perhaps just put a `//todo: implement
transactions` comment or something?

we could point out at what place you would have to pass down your transactions with the `TransportTransaction` class.

Personally I'm :+1: on removing transactions. As @SimonCropp pointed out correctly, if we want to show how to implement transactional behavior, this would probably be the sample for, but I still think it makes the sample so much more complicated. Maybe we can show transaction implementation with a simpler in-memory transport instead as it's own sample?

SimonCropp:
either way we have to solve the transactions. @andreasohlund plans on adding the file transport to the core in a future version. so if we already need to solve the error handling/transactions for that change i think we should just also include them in here

timbussmann:
but wouldn't it make more sense to point to @andreasohlund production ready file transport implementation if interested how transactions work instead of having different implementations around?

How about removing the SendsAtomicWithReceive capabilities and use ReceiveOnly instead, that would allow us to do in-memory FLR which would make the transaction implementation simpler will still showing how to use transactions?

SimonCropp:
> but wouldn't it make more sense to point to @andreasohlund production ready file transport implementation if interested how transactions work instead of having different implementations around?

Ok i am happy with that

andreasohlund:
I'll make sure to update it to the new seam next week

timbussmann:
@andreasohlund @Particular/docs-maintainers changed transaction handling to ReceiveOnly mode and implemented "proper" immediate retries. Please review.

SzymonPobiega:
@timbussmann is it ready? If so, can you remove the `[WIP]` tag

SzymonPobiega:
Also, it needs a rebase because the build is failing due to title restrictions on other pages that already have been fixed on master.

timbussmann:
ah, will rebase

PR #1766, opened 2016-08-10
Change AddressTranslationException to OverrideLocalAddress

timbussmann:
Update to the changes introduced by https://github.com/Particular/NServiceBus/pull/3970

/cc @SzymonPobiega


timbussmann:
does it make sense to have a section in the upgrade document about the renaming from `OverrideLocalAddress` to `OverrideInputQueueName` ?

weralabaj:
@timbussmann If that's a change between 5 and 6, and it's public API, then absolutely!

IMO it'd make most sense close to http://docs.particular.net/nservicebus/upgrades/5to6#endpoint-name-is-mandatory, maybe right after?

timbussmann:
yeah sounds good to me. will add this to this PR

nservicebus/upgrades/5to6.md
```diff
@@ -37,6 +37,11 @@ snippet: 5to6-endpointNameRequired
 The endpoint name is used as a logical identifier when sending or receiving messages. It is also used for determining the name of the input queue the endpoint will be bound to. See [Derived endpoint name](/nservicebus/upgrades/5to6-endpoint-name-helper.md) for the algorithm used in Versions 5 and below to select endpoint name if backwards compatibility is a concern.
 
 
+## OverrideLocalAddress is removed
+
+`OverrideLocalAddress` was replaced with `OverrideInputQueueName`. The specified parameter will be used as the name of the local input queue.
```
SimonCropp: can u show some snippets here to give some more context. and link to the doco for OverrideLocalAddress

nservicebus/endpoints/specify-endpoint-name.md
```diff
@@ -57,19 +57,5 @@ By default the endpoint's input queue name is the same as endpoint's name. The i
...
-
-snippet:InputQueueOverrideRouting
-
-
-### Override in app.config
-
-snippet:InputQueueOverrideRoutingXml
-
-
-In Versions 5 and below the overridden input queue is passed directly to the routing configuration. In Versions 6 and above the logical routing works on an endpoint name level. The endpoint name is passed to the routing configuration, and a separate API call is required to specify the name of its input queue.
```
SimonCropp: why is this doco deleted?

timbussmann:
potential suggestion to just rename the method with v5 to avoid migration effort: https://github.com/Particular/NServiceBus/pull/4049

nservicebus/endpoints/specify-endpoint-name.md
timbussmann: the description of the v6 behavior holds no longer true. The snippets just show that you have to adjust routing to the new input queue name, I'm not sure this adds any real value? Maybe have text stating that you also have to use the override instead of the endpoint name when addressing this endpoint (big surprise?).

timbussmann:
removed the upgrade guide section again, since there is now the same API and behavior as in v5

nservicebus/endpoints/specify-endpoint-name.md
SzymonPobiega: So if we override an input queue for an endpoint, in V6 there is no way to send messages to that endpoint other than specifying the transport address in `SendOptions.SetDestination`. I think it is worth mentioning here.

timbussmann: @SzymonPobiega you can also change the value in the normal routing API, as long as you treat the overriden queue name as the new logical name, so there shouldn't be a difference to v5? Also if you look at line 60, do you think that needs a different text? (but makes me think whether we should include some docs about `SendOptions.SetDestination` on the routing page instead?

SzymonPobiega: I'd say there is no good option here, which leads me to questioning if we should keep the `OverrideLocalAddress` API at al:
 * `SetDestination` is awkward
 * If we treat the overridden name as the logical endpoint name in other endpoints, we are not consistent and we may confuse people what is the logical and physical address.

timbussmann: > If we treat the overridden name as the logical endpoint name in other endpoints, we are not consistent and we may confuse people what is the logical and physical address.

I think in most scenarios users don't care and don't see a difference. As far as I can see the same rules and actions are rquired as in v5. I agree it's a bit incorrect from a logical routing perspective, but I'd rather solve that by specifying the address of a specific endpoint instance (which currently not supported by EndpointInstance) because if you configure an endpoint "sales" to use the queue "bla", that is essentially an instance specific information. Seems to be out of scope here, do you want to raise an issue?

SzymonPobiega:
@timbussmann , since https://github.com/Particular/NServiceBus/pull/4049 got merged, do wee still need this PR?

timbussmann:
@SzymonPobiega I removed some docs which I found more confusing than helpful, so in my opinion, yes, this PR is still valid.

PR #1770, opened 2016-08-12
Update Custom Checks sample and snippets to beta 6 (Core beta 7)

mikeminutillo:
Connects to https://github.com/Particular/V6Launch/issues/65

To be merged when the Custom Checks plugin is pushed to NuGet.

### DIFF
--- a/Snippets/CustomChecks/CustomChecks6_1/CustomChecks6_1.csproj
+++ b/Snippets/CustomChecks/CustomChecks6_1/CustomChecks6_1.csproj
@@ -28,8 +28,9 @@
     <Reference Include="NServiceBus.Core">
       <HintPath>..\..\..\packages\NServiceBus.6.0.0-beta0007\lib\net452\NServiceBus.Core.dll</HintPath>
     </Reference>
-    <Reference Include="ServiceControl.Plugin.Nsb6.CustomChecks">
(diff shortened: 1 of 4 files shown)


weralabaj:
@mikeminutillo generally if the package is on myget, we merge related docs. can we do that with this one too?

mikeminutillo:
@weralabaj if someone downloads the sample and attempts a NuGet restore without having the MyGet feed configured it will fail.

I may need to revisit this one tomorrow anyway as the plugin was updated for the Beta6 release but not yet for Beta7 (it will probably work just fine but I need to verify)

weralabaj:
OK

>  if someone downloads the sample and attempts a NuGet restore without having the MyGet feed configured it will fail

That would happen for other packages anyway, so I wouldn't worry too much. Might be also good to raise it as a concern for Launch. I think the reason is that nobody expected those updates will take so much time overall :/

mikeminutillo:
No worries. Like I said, I need to quickly revisit this one tomorrow and I suspect that by the time I'm done the core change will be pushed and I can push the updated plugins to NuGet as well.

SimonCropp:
Confused. All samples have the myget feed configured

mikeminutillo:
Ah yes. @SimonCropp is correct :) I withdraw my concern. The fact that this still needs a potential update remains though.

mikeminutillo:
Good to go

weralabaj:
rebased, waiting for TC checks to pass before merging

PR #1771, opened 2016-08-12
Repair representation error

fangqifan:
i think the
‘To enable this feature at the endpoint level using:’ 
should be 
‘To disable this feature at the endpoint level using:’


weralabaj:
Ah you're right! Thanks a lot for catching it! <3 I'm merging now.

PR #1782, opened 2016-08-16
Update Raven doco to the latest

weralabaj:


weralabaj:
the script messes up the upgrade, I'll do it manually

PR #1791, opened 2016-08-18
Add note about reply behavior when using bus.defer

timbussmann:
relates to https://github.com/Particular/NServiceBus/issues/2710

Do transports support native deferral behave the same? @andreasohlund

### DIFF
--- a/nservicebus/messaging/delayed-delivery.md
+++ b/nservicebus/messaging/delayed-delivery.md
@@ -18,6 +18,7 @@ NOTE: Only send operations can be deferred. Publish and reply operations cannot 
 
 partial:intro
 
+Note: Similar to `bus.SendLocal`, `bus.Defer` will also change the message's reply-to address to the endpoint deferring the message. Calling `bus.Reply` on a deferred message will send the reply to itself.
 
 ## Using a TimeSpan
 
--- a/nservicebus/messaging/delayed-delivery.md
+++ b/nservicebus/messaging/delayed-delivery.md
@@ -18,7 +18,7 @@ NOTE: Only send operations can be deferred. Publish and reply operations cannot 
 
 partial:intro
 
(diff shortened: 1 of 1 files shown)


andreasohlund:
Pretty sure they do since that header is controlled by the core so I'm pretty sure that it would have the same issue (needs to be verified though).

nservicebus/messaging/delayed-delivery.md
```diff
@@ -26,6 +26,7 @@ Calling `bus.Defer(...)` the handling of the message is deferred by certain amou
 
 Using this mechanism delayed delivery can be achieved by calling `bus.Defer(...)` which can be seen in [Delayed Delivery Sample](/samples/delayed-delivery).
 
+Note: Similar to `bus.SendLocal`, `bus.Defer` will also change the message's reply-to address to the endpoint deferring the message. Calling `bus.Reply` on a deferred message will send the reply to itself.
```
SimonCropp: `bus` or `endpoint`?

weralabaj: @SimonCropp this page doesn't use version switching yet, so this fragment is only about v5. I can take care of it during review.

@timbussmann do we need to add anything about v6?

andreasohlund: Bus.Defer is no longer present in v6 but users can still do

```
 var options = new SendOptions();

options.RouteToLocalEndpoint();
options.DelayDeliveryWith(TimeSpan.FromSeconds(30));

context.Send(options, message)

```

and then run into the same issue?

andreasohlund:
We probably should go ahead and doco this but I also raised a bigger discussion about "reusing" the incoming message that might be relevant https://github.com/Particular/NServiceBus/issues/4074

SzymonPobiega:
Rebased. Good to merge?

PR #1793, opened 2016-08-19
Gateway version compatibility

DavidBoike:
Attempting to document Gateway version compatibility, related to https://github.com/Particular/NServiceBus.Gateway/issues/55#issuecomment-240630778.

I admit, it gets awkward when describing about the versions, especially since Gateway got split out from Core V5.

The last paragraph is also a bit of hand-wavyness on my part. Please bash freely.

//cc @Particular/docs-maintainers @Particular/nservicebus-maintainers @indualagarsamy @andreasohlund @johnsimons @boblangley

### DIFF
--- a/nservicebus/gateway/index.md
+++ b/nservicebus/gateway/index.md
@@ -69,3 +69,12 @@ snippet:GatewayConfiguration
 ### Recoverability
 
 partial: recoverability
+
+
+### Version compatibility
+
+The Gateway component ensures only forward compatibility for one major version. So a message sent by the NServiceBus 3.x Gateway can be understood by the NServiceBus 4.x Gateway, and a message sent by the NServiceBus 4.x Gateway can be understood by NServiceBus.Gateway 1.x (the independent package targeting NServiceBus 5.x).
+
+However, a message sent by the NServiceBus 3.x Gateway will not be understood by NServiceBus.Gateway 1.x (NServiceBus 5.x) as this skips a major version. Likewise, a message sent by NServiceBus.Gateway 2.x will not be understood by NServiceBus.Gateway 1.x, as backwards communication is not supported.
+
+Since a Gateway on the receiving end exists to receive HTTP requests and translate them to messages on the queue, it's possible to deploy multiple Gateway instances using different major versions, either on different servers, or using different ports on the same server. As long as these Gateway instances contain the same message routing settings, they don't require any handler logic. This can be useful during a gradual, endpoint-by-endpoint upgrade to ensure that each deployed endpoint can communicate with a remote Gateway instance that is compatible with it.
--- a/nservicebus/gateway/index.md
+++ b/nservicebus/gateway/index.md
@@ -73,8 +73,8 @@ partial: recoverability
 
 ### Version compatibility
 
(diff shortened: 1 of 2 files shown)


nservicebus/gateway/index.md
```diff
@@ -97,3 +97,12 @@ To disable retries in the gateway use the `DisableRetries` setting:
...
+
+However, a message sent by the NServiceBus 3.x Gateway will not be understood by NServiceBus.Gateway 1.x (NServiceBus 5.x) as this skips a major version. Likewise, a message sent by NServiceBus.Gateway 2.x will not be understood by NServiceBus.Gateway 1.x, as backwards communication is not supported.
+
+Since a Gateway on the receiving end exists to receive HTTP requests and translate them to messages on the queue, it's possible to deploy multiple Gateway instances using different major versions, either on different servers, or using different ports on the same server. As long as these Gateway instances contain the same message routing settings, they don't require any handler logic. This can be useful during a gradual, endpoint-by-endpoint upgrade to ensure that each deployed endpoint can communicate with a remote Gateway instance that is compatible with it.
```
andreasohlund: >As long as these Gateway instances contain the same message routing settings, they don't require any handler logic. 

Is this true? The GW always delivers to its "local" endpoint if I'm not mistaken?

https://github.com/Particular/NServiceBus.Gateway/blob/develop/src/NServiceBus.Gateway/Routing/Endpoints/EndpointRouter.cs#L11 

so I think the only option is to upgrade the GW enpoints "major to major" which shouldn't be a huge deal since gateways communicate within the serivice boundary so it shouldn't cause ripple effects.

nservicebus/gateway/index.md
```diff
@@ -97,3 +97,12 @@ To disable retries in the gateway use the `DisableRetries` setting:
 snippet: GatewayDisableRetriesConfiguration
 
 When retries are disabled, any messages that fail to be sent to another site will be immediately routed to the configured error queue.
+
+
+### Version compatibility
+
+The Gateway component ensures only forward compatibility for one major version. So a message sent by the NServiceBus 3.x Gateway can be understood by the NServiceBus 4.x Gateway, and a message sent by the NServiceBus 4.x Gateway can be understood by NServiceBus.Gateway 1.x (the independent package targeting NServiceBus 5.x).
```
SimonCropp: what about backwards compat?

nservicebus/gateway/index.md
DavidBoike: You caught me. I've never used the Gateway IRL before. I thought it received via HTTP and then did a Bus.Send on them. If it really only handles them locally then this text makes absolutely no sense.

nservicebus/gateway/index.md
DavidBoike: Next paragraph talks about (lack of) backwards compat. Perhaps too indirectly?

For instance when the V4 Gateway sends a message, it's going to be using the new SendSingleRequest type that the V3 Gateway knows nothing about. V3 would also be expecting an Ack that would never come.

andreasohlund: We do have some kind of legacy mode 

https://github.com/Particular/NServiceBus.Gateway/blob/da5198f5bc0393965e366b917bbdc22976cb6411/src/NServiceBus.Gateway/HeaderManagement/GatewayIncomingBehavior.cs#L31

not sure how that's supposed to work. @janovesk do you know?

weralabaj: @andreasohlund do we need to wait for @janovesk before we can close this? or merge as it is and then review once he gets back?

andreasohlund: Let's go ahead , we can always update when he's back

weralabaj:
rebased

PR #1796, opened 2016-08-23
Default concurrency limit in V6

ramonsmits:


weralabaj:
@ramonsmits I think that could be a single note, like this:

NOTE: The default concurrency limit for NServiceBus Versions 5 and below is `1`, for Versions 6 and above `100`.

PR #1807, opened 2016-08-25
Clarify upgrade policy

weralabaj:
Resolves https://github.com/Particular/PlatformDevelopment/issues/917.

Connects to https://github.com/Particular/V6Launch/issues/99

### DIFF
--- a/nservicebus/upgrades/support-policy.md
+++ b/nservicebus/upgrades/support-policy.md
@@ -1,7 +1,7 @@
 ---
 title: Support Policy
 summary: Versions that are currently supported
-reviewed: 2016-06-29
+reviewed: 2016-09-24
 related:
  - nservicebus/licensing
  - nservicebus/upgrades/release-policy
@@ -11,7 +11,7 @@ related:
 
 Supported versions include any major version released within the last 3 years. Within a supported major version, the latest minor version is supported, or any minor version released within the last year.
 
-WARNING: If a version is not listed or has expired, but support is required contact [Particular support](http://particular.net/support). 
+WARNING: If a version is not listed or has expired, but support is required contact [Particular support](http://particular.net/support).
 
 Particular will apply critical bug fixes to all supported versions. If the bug also impacts a non supported version, the recommended approach is to upgrade to a supported version which contains the fix.
 
@@ -42,8 +42,8 @@ All new features are backwards compatible by default. In the rare cases when it'
 
 ## Upgrading
 
-The NServiceBus versions are wire-compatible, meaning the endpoints using different versions can exchange messages with each other. 
+NServiceBus versions are wire-compatible, meaning the endpoints using different versions can exchange messages with each other.
 
-However, some features might require data migration (e.g. converting from old to new format), and the conversion might be performed in the background after upgrade. 
+However, some features might require data migration (e.g. converting from old to new format), and the conversion might be performed in the background after upgrade.
 
 Therefore the recommended approach is to upgrade 1 major version at a time, including a full regression testing of the system and deployment to production.
--- a/nservicebus/upgrades/support-policy.md
+++ b/nservicebus/upgrades/support-policy.md
@@ -31,3 +31,12 @@ As of 11 July 2016, Version 4 is no longer supported.
 ### NServiceBus 3.x
 
 As of 10 October 2015, Version 3 is no longer supported.
+
+
+## Compatibility guarantees
+
+The compatibility guarantees cover two aspects: upgrades and communication between endpoints in various versions (so called `wire compatibility`).
+
+The recommended approach is to upgrade 1 major version at a time, with deployment to production. That guarantees that the necessary transformations of data are performed and no information will be lost.
+
+The wire compatibility is guaranteed across 1 major version. Additionally, there's the best effort and extensive testing to cover compatibility across all versions, starting with NServiceBus Version 3.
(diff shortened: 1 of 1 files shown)


nservicebus/upgrades/support-policy.md
```diff
@@ -30,4 +30,13 @@ If a version is not listed or has expired but support is required [contact Parti
...
+As of 10 October 2015, Version 3 is no longer supported. If support is required [contact Particular](http://particular.net/contactus).
+
+
+## Compatibility guarantees
+
+The compatibility guarantees cover two aspects: upgrades and communication between endpoints in various versions (so called `wire compatibility`).
+
+The recommended approach is to upgrade 1 major version at a time, with deployment to production. That guarantees that the necessary transformations of data are performed and no information will be lost.
+
+The wire compatibility is guaranteed across 1 major version. Additionally, there's the best effort and extensive testing to cover compatibility across all versions, starting with NServiceBus Version 3.
```
andreasohlund: Just to make sure I get it, this means we're wire compatible btw

3-5
4-6
5-7

and so on?

andreasohlund: >there's the best effort and extensive testing to cover compatibility across all versions

Where would a user find out what's the status? Do we need another page?

andreasohlund: It feels a bit odd to lock the wire compat down to versions (since we want to be more aggressive with releasing major in the future if possible)

Should we consider using dates instead? or "we guarantee wire compatibility between all supported versions" ?

weralabaj: > Just to make sure I get it, this means we're wire compatible btw
> 3-5
> 4-6
> 5-7

I thought that means 3-4, 4-5, 5-6, etc. Either way I agree, we need to specify with example.

weralabaj: >> there's the best effort and extensive testing to cover compatibility across all versions
> Where would a user find out what's the status? Do we need another page?

Not sure if we're ready to mention end to end tests here?

nservicebus/upgrades/support-policy.md
```diff
@@ -30,4 +30,13 @@ If a version is not listed or has expired but support is required [contact Parti
...
 ### NServiceBus 3.x
 
-As of 10 October 2015, Version 3 is no longer supported. If support is required [contact Particular](http://particular.net/contactus).
\ No newline at end of file
+As of 10 October 2015, Version 3 is no longer supported. If support is required [contact Particular](http://particular.net/contactus).
+
+
+## Compatibility guarantees
+
+The compatibility guarantees cover two aspects: upgrades and communication between endpoints in various versions (so called `wire compatibility`).
```
mikeminutillo: Should we give these aspects both names? i.e. Upgrade compatibility and Wire compatibility.

Upgrade Compatibility says that you can always upgrade an endpoint to the next major. Skipping majors is never supported. Downgrades are not supported. It's one-way. One step at a time.

Wire Compatibility says that endpoints of 2 adjacent majors can communicate. It might be worth noting that majors can introduce new features that break Wire Compatibility but these are always opt in (we could probably do a better job of pointing these out).

weralabaj: > Should we give these aspects both names? i.e. Upgrade compatibility and Wire compatibility.

That was the idea, but after we agree on what they exactly mean :)

indualagarsamy:
@weralabaj - This needs a rebase. Need any help to get this done?

andreasohlund:
We have a call on this topic tomorrow so this is blocked for now

weralabaj:
@andreasohlund would that be all based on our today's discussion? we'll probably evolve in the future, but just for now

nservicebus/upgrades/support-policy.md
```diff
@@ -11,9 +11,11 @@ related:
 
 Supported versions include any major version released within the last 3 years. Within a supported major version, the latest minor version is supported, or any minor version released within the last year.
 
-If a version is not listed or has expired but support is required [contact Particular](https://particular.net/contactus). For more information on the relationship between releases and support see [the release policy](/nservicebus/upgrades/release-policy.md).
+Particular guarantees patching all supported versions with critical bug fixes, even if the specific issue issue was raised only for one version. In case the bug also impacts non supported version, the recommended approach is to upgrade to the supported version which contains the fix.
```
andreasohlund: >impacts non supported version

impacts a non supported version?

andreasohlund: >issue issue

  remove one of them

andreasohlund: >even if the specific issue issue was raised only for one version

Do we need to say this? The previous statement already says that we "even if the specific issue issue was raised only for one version" so if the issue fits that it goes without saying that we'll do it?

andreasohlund: >recommended approach is to upgrade to the supported version which contains the fix.

Can we specify what "the version" is? (latest/first/lowest) or just to with "a supported version that contains the fix"

nservicebus/upgrades/support-policy.md
```diff
@@ -31,3 +33,17 @@ As of 11 July 2016, Version 4 is no longer supported.
 ### NServiceBus 3.x
 
 As of 10 October 2015, Version 3 is no longer supported.
+
+
+## Compatibility guarantees
+
+All new features are backwards compatible by default. In the rare cases when it's not true, enabling the new features requires an explicit opt-in.
```
andreasohlund: >enabling the new features requires an explicit opt-in. 

the feature would be disabled by default.?

PR #1808, opened 2016-08-25
[WIP]Guide users towards scripting when they need full control over queue creation

andreasohlund:
Based on a discussion with @SimonCropp. The idea is that we should provide api's for the users to create their own queues outside what they need for endpoints and satellites and instead guide them to use native transport scripting instead.

@Particular/docs-maintainers Thougts on the todos?

* Should we have a scripting landing page?
* Should we add a scripting page for ASB? (ping @Particular/azure-service-bus-maintainers)
* Should we add a scripting page for ASQ? (ping @Particular/azure-storage-queues-maintainers )


nservicebus/azure-service-bus/token-provider.md
```diff
@@ -14,7 +14,7 @@ This is useful when, for example, delegating authentication and authorization to
 
 Changing the authentication mechanism is done using the [Azure Service Bus SDK's TokenProvider class](https://msdn.microsoft.com/en-us/library/microsoft.servicebus.tokenprovider.aspx). The Azure Service Bus SDK requires an instance of this class at 2 different levels.
 
- * `NamespaceManager`: requires a `TokenProvider` that issues tokens with manage rights on the namespace. Note that this is only needed if queue creation is enabled, so that it can list, create and update entities in the namespace.
+ * `NamespaceManager`: requires a `TokenProvider` that issues tokens with manage rights on the namespace. Note that this is only needed if [queue creation](/nservicebus/transports/queuecreation) is enabled, so that it can list, create and update entities in the namespace.
```
SeanFeldman: I'm not sure what value that link is providing. The context here is `NamespaceManage` and having managed rights to create queues. Not manually create queues.

SeanFeldman:
Good question @andreasohlund 

In my past life I didn't have to create queue in Azure, always relied on NSB to do it for me. Including scaled-out queues (queue per instance, scenario where it didn't even make sense to do it manually). If we think that this adds value, raise an issue?

nservicebus/azure-service-bus/token-provider.md
andreasohlund: Got it, will remove!

SimonCropp:
@andreasohlund bump. need any help?

andreasohlund:
@yvesgoeleven I could add that link you mentioned in this PR? (for ASB/ASQ scripting)

yvesgoeleven:
@andreasohlund yes, that doco is online now, you can refer to it http://docs.particular.net/nservicebus/azure-service-bus/operational-scripting

andreasohlund:
@yvesgoeleven @Particular/azure-storage-queues-maintainers should we create a similar page for ASQ? (or would the content be the same?)

yvesgoeleven:
@andreasohlund we would have to look up the existing doco in msdn to see if it covers our needs, but as queues are mostly not the focus of the storage team I would not be surprised if we need to provide the guidance ourselves

andreasohlund:

SimonCropp:
@andreasohlund do the result of merging this is we will have a public API (IWantQueueCreated) that is no longer documented. is this really a situation you want to be in?

SimonCropp:
also. broke the build

andreasohlund:
I'll fix the build

>do the result of merging this is we will have a public API (IWantQueueCreated) that is no longer documented. is this really a situation you want to be in?

My current stance is that we should remove the api since the satellites cover all use cases we want to support.

There are currently 2 usages outside the core (https://github.com/search?l=&q=user%3Aparticular+IWantQueueCreated&ref=advsearch&type=Code&utf8=%E2%9C%93)

* Distributor - OK since we no longer provide a distributor
* ServceControl - https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/Infrastructure/Installers/ErrorLogQueueInstaller.cs

ServiceControl is more  tricky, do we keep the api just for SC? do we create custom scripts for SC? custom abstractions for SC?

I'd say we merge this (once the build is green) and raise a new issue to kill the API? (since end users should not use it)

SimonCropp:
> I'd say we merge this (once the build is green) and raise a new issue to kill the API? (since end users should not use it)

agreed on "kill the api". but there may be people using the existing API. removing the doco for it makes it a pain for anyone providing support for that API

andreasohlund:
Is this some kind of middle ground? API is still there for at least this major and our doco guides towards scripting.

New users: Will not find and use it
Existing users: Already knows about it and uses it, if they get stuck upgrading they will talk to us and we can get more info and perhaps sway them to switch to scripting?

Not optimal but perhaps good enough?

Scooletz:
@yvesgoeleven in regards to ASQ, is it about using? https://www.nuget.org/packages/Microsoft.Azure.Management.Storage/5.0.2-preview

andreasohlund:
Replaced by https://github.com/Particular/docs.particular.net/pull/1874 based on discussions with @SimonCropp

PR #1809, opened 2016-08-25
[WIP] Legacy Retries Satellite

tmasternak:
Connects to: https://github.com/Particular/V6Launch/issues/99


nservicebus/upgrades/5to6-recoverability.md
```diff
@@ -72,4 +72,14 @@ snippet: 5to6-configureImmediateRetriesViaXml
...
 
-snippet: 5to6-configureDelayedRetriesViaXml
\ No newline at end of file
+snippet: 5to6-configureDelayedRetriesViaXml
+
+## Legacy `.Retries` queue
+
+In NServiceBus 5 all messages scheduled for delayed retry where forwarded to `.Retries` queue. In version 6 this queue is no longer used, however it is possible that after upgrading from version 5 the queue will contain unprocessed messages.
+
+To prevent message loss, version 6 runs a dedicated receiver that processes any messages left in `.Retries` queue. The receiver is enabled by default. It is safe to disable it and delete `.Retries` queue if the endpoint did not use Second Level Retries or after upgrade when the queue is empty.
```
SimonCropp: "Second Level Retries " should be "Delayed Retries"

SimonCropp:
how about these 

 * http://docs.particular.net/nservicebus/msmq/operations-scripting#create-queues-the-create-queue-helper-methods-in-c-35
 * http://docs.particular.net/nservicebus/rabbitmq/operations-scripting#create-queues-the-create-queue-helper-methods
 * http://docs.particular.net/nservicebus/sqlserver/operations-scripting#create-queues-the-create-queue-helper-methods

nservicebus/upgrades/5to6-recoverability.md
```diff
@@ -72,4 +72,14 @@ snippet: 5to6-configureImmediateRetriesViaXml
 
 snippet: 5to6-configureDelayedRetriesViaCode
 
-snippet: 5to6-configureDelayedRetriesViaXml
\ No newline at end of file
+snippet: 5to6-configureDelayedRetriesViaXml
+
+## Legacy `.Retries` queue
+
+In NServiceBus 5 all messages scheduled for delayed retry where forwarded to `.Retries` queue. In version 6 this queue is no longer used, however it is possible that after upgrading from version 5 the queue will contain unprocessed messages.
```
weralabaj: > In NServiceBus 5 

and below

weralabaj:
@SimonCropp do you mean they need to be updated?

weralabaj:
@tmasternak that needs to wait for the next release (RC?) before the snippet can be merged

SimonCropp:
@weralabaj i think so

tmasternak:
@SimonCropp I've added changes to integration scripts. 
@weralabaj and yest this needs to wait for new version of Core. I've done local patching to test it.

nservicebus/upgrades/5to6-recoverability.md
```diff
@@ -72,4 +72,14 @@ snippet: 5to6-configureImmediateRetriesViaXml
 
 snippet: 5to6-configureDelayedRetriesViaCode
 
-snippet: 5to6-configureDelayedRetriesViaXml
\ No newline at end of file
+snippet: 5to6-configureDelayedRetriesViaXml
+
+## Legacy .Retries queue
```
indualagarsamy: ## Legacy .Retries queue

The `.Retries` Satellite queue is no longer necessary when running Version 6 of NServiceBus.  However, when starting a Version 6 endpoint, unless explicitly configured not to use the retry satellite, a dedicated receiver that watches the .retries queue will still be created. This default configuration is mainly for a one-time scenario necessary to prevent message loss when upgrading from Version 5 to Version 6. 

### When Upgrading from Versions 5 and below

In NServiceBus Versions 5 and below, the [Second Level Retries feature](link to SLR article) of NServiceBus used the `[endpoint_name].Retries` queue as a placeholder to durably store the messages before persisting them for retries.  When upgrading a Version 5 endpoint to Version 6, during the process of stopping the endpoint, there is a possibility that the .retries queue may contain some of these messages that were meant to be delayed and retried. Therefore to prevent message loss, when starting up a Version 6 endpoint, the .retries satellite receiver runs to serve a one-time purpose of forwarding those messages from the `.retries` queue to the endpoint's main queue to be retried appropriately.  Once this is done, while the satellite receiver continues to run, it is not necessary as Version 6 does not utilize this queue any longer. 

Letting this satellite run after the upgrade process depends on the choice of transport and the upgrade policies.   

For endpoints that use SqlServer as transport, running the satellite receiver might have a performance implication, as it involves a constant internal SQL Query to check for messages in the .retries queue.  If this is not desired, Version 6 introduces a configurable option to turn this satellite off.  A subsequent deployment that disables the satellite and deletes the .retries queue can be done once the endpoint is running on Version 6.

### When Creating New Endpoints using Version 6

In Version 6, the only reason that the .retries queue exists is so that Version 5 endpoints can be migrated to Version 6 without any message loss.  Any endpoints that are written purely using Version 6, can safely use the following configuration API to disable the satellite from even starting. 

### Disabling the Retry Satellite via Configuration.

To disable the  `.Retries` satellite via code:

snippet: 5to6-DisableLegacyRetriesSatellite 

As part of the deployment of this change, the Ops team can safely delete the `.Retries` queue. 

INFO: This configuration API will be obsoleted and removed in Version 7.

tmasternak: @indualagarsamy I like it :+1: . I've incorporated it with minor changes to upgrade guide.

indualagarsamy:
@tmasternak - Looks like the build is failing. Can you have a look?

tmasternak:
@indualagarsamy this will not build until the RC package is out. Current one is missing `DisableLegacyRetriesSatellite` method.

tmasternak:
@Particular/docs-maintainers I've split scripting for msmq to addressing @SimonCropp [comment](https://github.com/Particular/docs.particular.net/pull/1809#issuecomment-242371166). I haven't done it yet for sql and rabbit because I want to know your opinion about such approach.

Here is what I did:
 * I've split scripting snippets into two parts: utils + endpoint queue creation method
 * moved endpoint queue creation method to version specific projects (V6 and V4 - because my understanding is that V3 will soon be gone)
 * moved endpoint creation tests to version specific project
 * fixed header validation tests to use specific queue creation methods (previously they were using endpoint creation methods)

Can you review and confirm we are good with such approach so that I can make the similar changes to Sql and Rabbit scripting?

SimonCropp:
merged manually

@tmasternak i went with a simpler solution than the instance Id param http://docs.particular.net/nservicebus/msmq/operations-scripting?version=Core_6#create-queues

tmasternak:
@SimonCropp thanks a lot!

PR #1814, opened 2016-08-26
Update to new core default for concurrency

andreasohlund:
I removed the part that talked about "allow the transport to optimize" since that isn't true.

The current statement isn't 100% correct but perhaps close enough?

Related to https://github.com/Particular/NServiceBus/pull/4092

### DIFF
--- a/nservicebus/operations/tuning.md
+++ b/nservicebus/operations/tuning.md
@@ -9,7 +9,7 @@ related:
 - nservicebus/licensing/licensing-limitations
 ---
 
-NServiceBus will by default try to optimize for maximum performance when it comes to message processing. While this is usually the preferred mode of operation there are situations where tuning needs to be applied.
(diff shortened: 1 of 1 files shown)


nservicebus/operations/tuning.md
```diff
@@ -9,12 +9,12 @@ related:
 - nservicebus/licensing/licensing-limitations
 ---
 
-NServiceBus will by default allow the transport to optimize for maximum performance when it comes to message processing. While this is usually the preferred mode of operation there are situations where tuning needs to be applied.
+NServiceBus will by default try to optimize for maximum performance when it comes to message processing. While this is usually the preferred mode of operation there are situations where tuning needs to be applied.
```
weralabaj: That sounds like we have some automatic adjustments built-in in NSB, that's not correct, is it?

weralabaj: How about something like "NServiceBus uses defaults that ensure good performance in most common cases"?

andreasohlund: I like it, pushed an update

PR #1821, opened 2016-08-26
TransportType addition to configuration API

SeanFeldman:
Connects to https://github.com/Particular/NServiceBus.AzureServiceBus/issues/306

### DIFF
--- a/nservicebus/azure-service-bus/configuration/full.md
+++ b/nservicebus/azure-service-bus/configuration/full.md
@@ -110,6 +110,7 @@ The following settings determine how NServiceBus will connect to Azure Service B
 
  * `NumberOfClientsPerEntity(int)`: NServiceBus maintains a pool of receive and send clients for each entity. This setting determines how big that pool is. Defaults to 5.
  * `ConnectivityMode(ConnectivityMode)`: Determines how NServiceBus connects to Azure Service Bus, using TCP or HTTP. Defaults to TCP.
+ * `TransportType(TransportType)`: Determines what transport protocol NServiceBus is using for Azure Service Bus, `TransportType.NetMessaging` or `TransportType.Amqp`. Defaults to `TransportType.NetMessaging`.
  * `BrokeredMessageBodyType(SupportedBrokeredMessageBodyTypes)`: Controls how the body of a brokered message will be serialized, either as a byte array or as a stream. Defaults to byte array.
  * `MessagingFactories()`: Provides access to settings of the used native instances of the class `MessagingFactory`. These settings will automatically apply to all `MessageReceiver` and `MessageSender` instances created by the `MessagingFactory`.
  * `MessageReceivers()`: Provides access to the settings of the `MessageReceiver` instances.
--- a/nservicebus/azure-service-bus/configuration/full.md
+++ b/nservicebus/azure-service-bus/configuration/full.md
@@ -6,8 +6,11 @@ tags:
 - Cloud
 redirects:
 - nservicebus/azure-service-bus/configuration/configuration
+reviewed: 2016-08-29
 ---
 
+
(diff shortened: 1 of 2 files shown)


nservicebus/azure-service-bus/configuration/full.md
```diff
@@ -110,6 +110,7 @@ The following settings determine how NServiceBus will connect to Azure Service B
 
  * `NumberOfClientsPerEntity(int)`: NServiceBus maintains a pool of receive and send clients for each entity. This setting determines how big that pool is. Defaults to 5.
  * `ConnectivityMode(ConnectivityMode)`: Determines how NServiceBus connects to Azure Service Bus, using TCP or HTTP. Defaults to TCP.
+ * `TransportType(TransportType)`: Determines what transport protocol NServiceBus is using for Azure Service Bus, `TransportType.NetMessaging` or `TransportType.Amqp`. Defaults to `TransportType.NetMessaging`.
```
SimonCropp: is this only for a specific version range?

yvesgoeleven: @SimonCropp not sure, what you're refering to. But here are both answers: 
* the API as a whole is new for V7 and so is this setting
* NetMessaging has been the protocol used in the past, in theory it should have been possible for users to swithc to amqp using the connectionstring notation, but I don't think anyone ever did that. Going forward AMQP will become the defacto standard for ASB clients, so we'll have to transition eventually as well, but as long as there is feature imparity we'll stick to netmessaging as default (but support amq)

SeanFeldman:
@Particular/docs-maintainers this is ready to go if you're OK with the review.

nservicebus/azure-service-bus/send-atomic-with-receive-note.include.md
```diff
@@ -1 +1 @@
-Note: `TransportTransactionMode.SendsAtomicWithReceive` level is only supported in Versions 7 and above of the Azure Service Bus transport. It requires destination and receive queues to be in the same namespace.
\ No newline at end of file
+Note: `TransportTransactionMode.SendsAtomicWithReceive` level is only supported in Versions 7 and above of the Azure Service Bus transport. It requires destination and receive queues to be in the same namespace and the use of the NetMessaging transport type.
```
indualagarsamy: >Note: `TransportTransactionMode.SendsAtomicWithReceive` level is only supported in Versions 7 and above of the Azure Service Bus transport. It requires destination and receive queues to be in the same namespace and the use of the NetMessaging transport type.

How about:

INFO: `TransportTransactionMode.SendsAtomicWithReceive` level is only supported in Versions 7 and above of the Azure Service Bus transport. It requires both destination and receive queues to be in the same namespace. Also, the transport type must be set to `TransportType.NetMessaging` instead of `TransportType.Amqp`.

nservicebus/azure-service-bus/configuration/full.md
indualagarsamy: @yvesgoeleven - How about this?

>* `TransportType(TransportType)`: Determines what transport protocol NServiceBus is using for Azure Service Bus, `TransportType.NetMessaging` or `TransportType.Amqp`. Defaults to `TransportType.NetMessaging`.

* `TransportType(TransportType)`:  This setting determines what transport protocol NServiceBus is using for Azure Service Bus. The available options are: `TransportType.NetMessaging` and `TransportType.Amqp`. The defaults option is set to `TransportType.NetMessaging`.

indualagarsamy:
As discussed with @SeanFeldman, for the interim, add INFO will be added to the top of this article stating the versions that its applicable. But this document and the other version doc needs to include partials, so as to just highlight what's new in the versions as opposed to repeating content. 

Also, please add the `reviewed` tag on this page.

PR #1826, opened 2016-08-29
Added missing routing content

SzymonPobiega:
Added some missing content (from the previous `message-oweneship` page)

### DIFF
--- a/nservicebus/messaging/routing.md
+++ b/nservicebus/messaging/routing.md
@@ -3,7 +3,6 @@ title: Message routing
 summary: How NServiceBus routes messages between the endpoints.
 reviewed: 2016-08-17
 component: Core
-versions: '[6,]'
 tags:
 - routing
 - message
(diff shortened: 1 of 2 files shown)


SzymonPobiega:
@SimonCropp @SeanFeldman please check if that covers the gaps.

SzymonPobiega:
It also fixes the bug that http://docs.particular.net/nservicebus/messaging/routing was declared as V6-only.

nservicebus/messaging/routing.md
```diff
@@ -3,7 +3,6 @@ title: Message routing
 summary: How NServiceBus routes messages between the endpoints.
 reviewed: 2016-08-17
 component: Core
-versions: '[6,]'
```
SeanFeldman: :+1:

SeanFeldman:
Thank you @SzymonPobiega 
@SimonCropp with this we can redirect people to the routing and not worry that MEM specific doco is gone.

SimonCropp:
where is this content https://github.com/Particular/docs.particular.net/blob/433ee500d3ebdf836d8674673e194a6140ffda78/nservicebus/messaging/message-owner.md#some-example-mappings  ?

SzymonPobiega:
Isn't this snippet equivalent https://github.com/Particular/docs.particular.net/edit/master/Snippets/Core/Core_6/Routing/EndpointMapping.config.xml#L3. I'd say it is better because it show the actual syntax, not only the mapping. Do you think it would be better if I put some comments inside the XML file?

PR #1830, opened 2016-08-30
Make InMemoryPersistence summary clear

SeanFeldman:
Clearly indicate it's NOT for production

@Particular/docs-maintainers


SimonCropp:
summary only appears on search results and index pages. not in the actual page content

SeanFeldman:
Exactly. The page itself has this info, but the summary doesn't. When this
is the top search result to show up, it looks like something to use...

SimonCropp:
ahh ok

SeanFeldman:
Merge? 😃

PR #1860, opened 2016-09-02
Adding the info that NServiceBus does work with other transports.

indualagarsamy:
The Getting Started is linked from a lot of places and is geared for the new user. When a user gets here, the user can have the impression that NServiceBus only works with MSMQ as the pre-requisite says install MSMQ. This can put off users who dislike MSMQ and/or already using other transports.


indualagarsamy:
cc / @Particular/docs-maintainers - I added this because I had a recent desk case and the user assumed that NSB requires MSMQ.

PR #1862, opened 2016-09-04
Removed redundant implementations of `void Handle`

qujck:
Removed redundant implementations of `void Handle` from the v6 step-by-step sample code


SimonCropp:
@qujck good catch. thanks.

PR #1873, opened 2016-09-07
Add a not about mulitschema configuration constraints

tmasternak:


nservicebus/sqlserver/connections-settings_multi-schema-config-notes_sqlserver[3,).partial.md
```diff
@@ -1,2 +1,4 @@
 NOTE: Starting from Versions 3 the table and schema names can be passed either using common convention with square brackets, or without them. 
- 
\ No newline at end of file
+NOTE: Starting from Version 3 custom schema needs to configured either via app.config or code api. The two approaches cannot be used at the same time.
```
SimonCropp: since this and the above not will only show up when version 3  is selected isn the "Starting from Version 3" redundant?

PR #1874, opened 2016-09-07
Guiding v6 users towards using scripts for custom queue creation

andreasohlund:
After talking to @SimonCropp we decided to 

1. Warn users to not use the `IWantQueuesCreated` interface since we're going to deprecate it in the future
2. Keep the existing snippets but remove the v6 ones

Replaces #1808


andreasohlund:
@Particular/docs-maintainers getting a 

```
  ----> System.Exception : Failed to render article. PackageId:NServiceBus. Version:[6.0.0-pre, 7.0.0).
  ----> CaptureSnippets.MissingSnippetsException : MissingSnippet. 

```

since I removed the v6 snippet. What do I need to do to work around this?

PR #1875, opened 2016-09-07
More realistic connection lookup implementation

tmasternak:
Connects to https://github.com/Particular/NServiceBus.SqlServer/issues/264


samples/sqltransport/multi-instance/sample.md
```diff
@@ -63,11 +63,11 @@ It receives `ClientOrder` message sent by Sender and replies to them with `Clien
 snippet:Reply
 
 
-### EndpointConnectionLookup project
+### Multi-instance connection lookup
 
-The EndpointConnectionLookup provides an open `SqlConnection` for address requested either by Sender or Receiver. Note that in Version 2 it provides connections for remote endpoints only, in Version 3 it provides connections for all endpoints.
+Both sender and receiver provide a custom lookup mechanism for providing connection information for given destination. The following snippet shows lookup logic used by Sender.
```
indualagarsamy: Are there no version dependent behavior any longer?

samples/sqltransport/multi-instance/SqlServer_3/Receiver/Program.cs
```diff
@@ -38,4 +39,22 @@ await endpointInstance.Stop()
...
     }
 
+    static class ConnectionProvider
+    {
+        const string DefaultConnectionString = @"Data Source=.\SqlExpress;Database=ReceiverCatalog;Integrated Security=True";
+        const string SenderConnectionString = @"Data Source=.\SqlExpress;Database=SenderCatalog;Integrated Security=True";
+
+        public static async Task<SqlConnection> GetConnection(string transportAddress)
+        {
+            var connectionString = transportAddress.StartsWith("Samples.SqlServer.MultiInstanceSender")
```
indualagarsamy: Shouldn't the check be  `Samples.SqlServer.MultiInstanceReceiver` instead of `Samples.SqlServer.MultiInstanceSender` as this is in the `Receiver/Program.cs` ?

Wouldn't the transport address be MultiInstanceReceiver in this project?

samples/sqltransport/multi-instance/SqlServer_3/Sender/Program.cs
```diff
@@ -68,4 +69,25 @@ await endpoint.Send("Samples.SqlServer.MultiInstanceReceiver", order)
...
 
+    #region SenderConnectionProvider
+    static class ConnectionProvider
+    {
+        const string DefaultConnectionString = @"Data Source=.\SqlExpress;Database=SenderCatalog;Integrated Security=True";
+        const string ReceiverConnectionString = @"Data Source=.\SqlExpress;Database=ReceiverCatalog;Integrated Security=True";
+
+        public static async Task<SqlConnection> GetConnection(string transportAddress)
+        {
+            var connectionString = transportAddress.StartsWith("Samples.SqlServer.MultiInstanceReceiver")
```
indualagarsamy: I don't understand why the check is inverse? i.e. the project is Sender, but it checks the transport address to that of the Receiver. 
Shouldn't the check be  `Samples.SqlServer.MultiInstanceSender` instead of `Samples.SqlServer.MultiInstanceReceiver` as this is in the `Sender/Program.cs` ?

Wouldn't the transport address be MultiInstanceReceiver in this project?

samples/sqltransport/multi-instance/sample.md
tmasternak: @indualagarsamy That is a good question :). Here is what I know:
 * There is not difference in SqlServer transport behavior. The callback in both v2 and v3 will be called with addresses of local queues (main queue + any satellite), error/audit or remote queues
 * I think the purpose of this comment was that there was a difference in callback implementation in sample for v2 and v3 - which TBH was misleading

In other words I think we should remove to comment. It brings more confusion than value.

samples/sqltransport/multi-instance/SqlServer_3/Receiver/Program.cs
tmasternak: I think this is ok. The receiver endpoint will return default connection string in all cases, except when message is targeted to any destination that starts with `Samples.SqlServer.MultiInstanceSender` (which is the remote endpoint transport address)

samples/sqltransport/multi-instance/SqlServer_3/Sender/Program.cs
tmasternak: This is the reverse to Sender. The receiver returns default connection string, except when message is destined to Receiver.

In general this the approach that we want to support and that's why the sample was changed. The approach is:
  * By default return a (no surprise  here ;)) default connection string
  * For each **remote** destination have an explicit match and return non default connection string.

SimonCropp:
needs a rebase

tmasternak:
@SimonCropp done

PR #1877, opened 2016-09-07
Distributor upgrade guide

timbussmann:
connects to Particular/V6Launch#68

ping @SzymonPobiega @DavidBoike @bording please have a look


SimonCropp:
@timbussmann awesome to see this happening :)

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,87 @@
+---
+title: Deprecated Distributor in Version 6
+tags:
+ - upgrade
+ - migration
+ related:
+ - samples/scaleout/distributor-upgrade
+---
+
+The [distributor](/nservicebus/scalability-and-ha/distributor.md) has been deprecated and replaced by [sender-side distribution]((/nservicebus/msmq/scalability-and-ha/sender-side-distribution.md)) with NServiceBus Version 6.
```
SzymonPobiega: Double parentheses in the link. Intentional?

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,87 @@
...
+
+Use this process when shutting down all endpoints at once is no option. This process aims to allow upgrade in production environments at no downtime and without missing events.
+
+* Upgrade all endpoints communicating to the distributor (by commands or event subscription messages) to NServiceBus Version 6 first.
+* Apply the following steps for each scaled out instance enlisted to the Distributor, one instance after another:
+  * Shut down the instance.
+  * [Upgrade the instance to NServiceBus Version 6](#upgrade-endpoint-to-version-6).
+  * [Enlist it with the distributor](#enlist-version-6-endpoints-with-a-distributor).
+  * Start the instance again
+* Once all instances are upgraded to Version 6, configure all endpoints communicating to the distributor (by commands or event subscription messages) to use [sender-side distribution](/nservicebus/msmq/scalability-and-ha/sender-side-distribution.md) to send messages directly to the scaled out instances instead of the distributor.
```
SzymonPobiega: "by commands or event subscription messages" -> "by sending commands or publishing events"

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,87 @@
...
+* Detach some of the instances (but not all) from the Distributor by removing `endpointConfiguration.EnlistWithLegacyMSMQDistributor` and restarting them.
+* [Unsubscribe the Distributor from all publishers](#remove-subscriptions-for-the-distributor)
```
SzymonPobiega: How about the trick with replacing the distributor with a V6 worker? No unsubscribe is required in that case.

nservicebus/upgrades/5to6/distributor.md
timbussmann: not intentional, good catch :+1:

nservicebus/upgrades/5to6/distributor.md
timbussmann: would be a reasonable approach when running the distributor in master config which means that it's also a worker node (if I understood that correctly). Depending on the exact scenario you could come up with a simpler process (e.g. when not using pub/sub), but it makes the upgrade much more complicated in terms of "if your setup is like x, then do these steps, if it is like y, then use these steps, otherwise go there or there". I think it's probably safest to just always unsubscribe as general guideline?

SzymonPobiega: How about we suggest to :fire: one of the workers and install a worker on distributor VM? Then we don't need to unsubscribe unless the distributor queue on clustered MSMQ which is no longer necessary after moving to sender-side distribution.

timbussmann: > Then we don't need to unsubscribe unless the distributor queue on clustered MSMQ which is no longer necessary after moving to sender-side distribution.

I wonder about that "workaround", since a v5 subscription is subscribing without an endpoint and a v6 subscription contains an endpoint. So from a persistence perspective, the v5 subscription may not be removed and this only works properly because we deduplicate receiver-addresses (you'd find 2 subscriptions, both going to the distributor queue). Technically this would work, but I'm not sure this is a clean solution?

SzymonPobiega: Hmmm no. The key in the subscription store is `subscriber address + type`. So when a V6 subscription message arrives, it replaces/updates the subscription store entry that came from V5 previously instead of creating a new one.

timbussmann: ah, alright. Maybe we could add a section about running the distributor with worker/without a worker and the differences while upgrading, since these step currently assume a distributor with no workers (or the worker to be removed)?

SimonCropp:
needs a rebase

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,87 @@
...
+Use this process when shutting down all endpoints at once is no option. This process aims to allow upgrade in production environments at no downtime and without missing events.
+
+* Upgrade all endpoints communicating to the distributor (by commands or event subscription messages) to NServiceBus Version 6 first.
+* Apply the following steps for each scaled out instance enlisted to the Distributor, one instance after another:
+  * Shut down the instance.
+  * [Upgrade the instance to NServiceBus Version 6](#upgrade-endpoint-to-version-6).
+  * [Enlist it with the distributor](#enlist-version-6-endpoints-with-a-distributor).
+  * Start the instance again
+* Once all instances are upgraded to Version 6, configure all endpoints communicating to the distributor (by commands or event subscription messages) to use [sender-side distribution](/nservicebus/msmq/scalability-and-ha/sender-side-distribution.md) to send messages directly to the scaled out instances instead of the distributor.
+* Detach some of the instances (but not all) from the Distributor by removing `endpointConfiguration.EnlistWithLegacyMSMQDistributor` and restarting them.
```
timbussmann: now that I'm going again through this, wouldn't this configuration cause duplicate handling of published events? @SzymonPobiega 

If I don't miss anything, the only proper way to avoid duplicate events, would be to migrate the distributor to a worker node after all "regular" workers were updated. So the process would probably be simpler by:
* Shutting down all workers
* Migrate the distributor to a regular node
* Migrate workers

But this approach would require a short downtime of an endpoint because the distributor endpoints needs to shut down (e.g. callbacks could timeout). Am I missing something? (The good part would be, that there is no need to remove subscriptions unless you want to decommission the migrated distributor later).

nservicebus/upgrades/5to6/distributor.md
janovesk: The distributor is usually running on a Windows Failover Cluster because it is a single point of failure. Simply replacing a distributor with a worker would avoid the need to unsubscribe the distributor queue, at first, but it is not really a long term solution. The result would be a worker running on a fail over cluster, which does not make much sense. You'd probably want to remove that from the fail over cluster eventually, forcing you to do the very unsubscribe operation you avoided in the first place anyway.

timbussmann: > You'd probably want to remove that from the fail over cluster eventually, forcing you to do the very unsubscribe operation you avoided in the first place anyway.

I agree in that specific scenario. But then this can be handled by a regular "decomissioning endpoints" documentation which is not specific to the distributor upgrade?

SzymonPobiega: @janovesk convinced me that it would be only a temporary and incomplete solution, esp for deployments with a HA distributor.

nservicebus/upgrades/5to6/distributor.md
SzymonPobiega: Yeah, that would work. The downtime would be minimal if done properly.

timbussmann:
I merged the two processes to a single process which requires the migration of the distributor endpoint to a worker node. All in all it makes the process and choice much simpler and eliminates risks of duplicate events at the cost of a short downtime. I think this is good enough as a general guideline, we may want to consider adding a note to contact support if further help is required. /cc @SzymonPobiega @janovesk

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,45 @@
...
+## Client-side distribution
+
+With sender-side distribution, senders are aware of scaled out receivers, allowing senders to distribute messages directly to the receivers instead of using an intermediary distributor.
+
+
+## Upgrade scenario
+
+This process aims to allow upgrade in production environments without missing events and minimal downtime.
+
+WARN: To avoid duplicate events, the Distributor node needs to be upgraded to a worker node using NServiceBus 6 (it can be decommissioned after all workers are upgraded to Version 6).
```
SimonCropp: `WARNING:` not `WARN:`

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,45 @@
+---
+title: Deprecated Distributor in Version 6
```
SimonCropp: Version 6 is ambiguous. is it v6 of the core or v6 of Distributor ?

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,45 @@
...
+* Once all instances are upgraded to Version 6, configure all endpoints sending commands or receiving events from the Distributor to use [sender-side distribution](/nservicebus/msmq/scalability-and-ha/sender-side-distribution.md) to send messages directly to the scaled out instances.
+
+
+## Upgrade endpoint to Version 6
+
+* Remove the `NServiceBus.Distributor.MSMQ` package.
+* Remove Distributor specific configuration options.
+  * Remove the `MasterNodeConfig` section from the application configuration file.
+  * Remove the `DistributorControlAddress` and `DistributorDataAddress` attributes from the `UnicastBusConfig` configuration section in the application configuration file.
+* Upgrade the endpoint to Version 6 (See the [Upgrade Guide](/nservicebus/upgrades/5to6)).
```
SimonCropp: `](/nservicebus/upgrades/5to6)).`  missing slash?

timbussmann: hm I'd say a duplicate parantheses or are trailing slashes required?

nservicebus/upgrades/5to6/distributor.md
janovesk: I think we are just postponing the real job by recommending converting the distributor to a worker. You don't want workers running on the Windows fail over cluster.

It's not necessarily a quick fix either. The code running the distributor does not necessarily have any message handlers deployed.

timbussmann: > I think we are just postponing the real job by recommending converting the distributor to a worker. 

It's the only approach I could come up with in order to not loose or receive duplicate events. Do you have a suggestion?

> The code running the distributor does not necessarily have any message handlers deployed.

that's a good point. Need to add this to the guide in case we stick with the current proposal.

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,45 @@
...
+## Client-side distribution
+
+With sender-side distribution, senders are aware of scaled out receivers, allowing senders to distribute messages directly to the receivers instead of using an intermediary distributor.
+
+
+## Upgrade scenario
+
+This process aims to allow upgrade in production environments without missing events and minimal downtime.
+
+WARNING: To avoid duplicate events, the Distributor node needs to be upgraded to a worker node using NServiceBus 6 (it can be decommissioned after all workers are upgraded to Version 6).
```
DavidBoike: Why would not doing this cause duplicate events? May need an overview of what these steps hope to achieve and why?

DavidBoike: And honestly I'm not sure I know what this text means, especially since it's not clear if you're talking about starting from a Master+Worker or Distributor Only configuration.

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,45 @@
...
+
+* Upgrade all endpoints communicating to the Distributor (by commands or event subscription messages) to NServiceBus Version 6 first.
+* Shut down the Distributor instance.
+* [Upgrade the Distributor to NServiceBus Version 6](#upgrade-endpoint-to-version-6). This will migrate the Distributor endpoint to a regular worker instance.
```
DavidBoike: So, this assumes that it's a Master+Worker node, correct? If the user was running a Distributor only on a failover cluster, for instance, they probably wouldn't want to do this.

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,45 @@
+---
+title: Deprecated Distributor in NServiceBus Version 6
+tags:
+ - upgrade
+ - migration
+related:
+ - samples/scaleout/distributor-upgrade
```
DavidBoike: Should there be a more explicit link in this article to this sample than just a related link?

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,45 @@
+---
+title: Deprecated Distributor in NServiceBus Version 6
```
indualagarsamy: Should the title be:

Upgrading the Distributor for NServiceBus Version 6

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,45 @@
...
+The [distributor](/nservicebus/scalability-and-ha/distributor) has been deprecated and replaced by [sender-side distribution](/nservicebus/msmq/scalability-and-ha/sender-side-distribution.md) with NServiceBus Version 6.
+
+## Client-side distribution
+
+With sender-side distribution, senders are aware of scaled out receivers, allowing senders to distribute messages directly to the receivers instead of using an intermediary distributor.
+
+
+## Upgrade scenario
+
+This process aims to allow upgrade in production environments without missing events and minimal downtime.
```
indualagarsamy: >This process aims to allow upgrade in production environments without missing events and minimal downtime.

This process aims to allow upgrade in production environments without losing messages and to facilitate minimal downtime.

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,45 @@
...
+
+WARNING: To avoid duplicate events, the Distributor node needs to be upgraded to a worker node using NServiceBus 6 (it can be decommissioned after all workers are upgraded to Version 6).
+
+* Upgrade all endpoints communicating to the Distributor (by commands or event subscription messages) to NServiceBus Version 6 first.
```
indualagarsamy: What about the workers? The workers communicate to the distributor sending ready messages. Will that cause any problems?

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,45 @@
...
+* Shut down the Distributor instance.
+* [Upgrade the Distributor to NServiceBus Version 6](#upgrade-endpoint-to-version-6). This will migrate the Distributor endpoint to a regular worker instance.
+* Start the instance again.
+* [Upgrade the remaining worker nodes to NServiceBus Version 6](#upgrade-endpoint-to-version-6).
+* Once all instances are upgraded to Version 6, configure all endpoints sending commands or receiving events from the Distributor to use [sender-side distribution](/nservicebus/msmq/scalability-and-ha/sender-side-distribution.md) to send messages directly to the scaled out instances.
```
indualagarsamy: Now what happens to the distributor instance that was migrated to a regular worker instance? Do you need to decommission it or does that serve as an extra node for sender side distribution?

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,45 @@
...
+
+
+## Upgrade endpoint to Version 6
```
indualagarsamy: Isn't this more like Upgrading a Distributor master node or worker node to Version 6?

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,45 @@
...
+
+* Remove the `NServiceBus.Distributor.MSMQ` package.
+* Remove Distributor specific configuration options.
+  * Remove the `MasterNodeConfig` section from the application configuration file.
+  * Remove the `DistributorControlAddress` and `DistributorDataAddress` attributes from the `UnicastBusConfig` configuration section in the application configuration file.
+* Upgrade the endpoint to Version 6 (See the [Upgrade Guide](/nservicebus/upgrades/5to6).
+
+
+## Mixing NServiceBus Versions
```
indualagarsamy: Should this be Use of the Distributor in Mixed Version environments?

nservicebus/upgrades/5to6/distributor.md
timbussmann: because the tricky part is about event subscriptions which are pointing to the distributor endpoint. You can't just reconfigure them on the publisher using the routing APIs to send to upgraded endpoints but you need new subscriptions while you usually want to avoid duplicate events (and we don't have a proper way of unsubscribing endpoints from events)

nservicebus/upgrades/5to6/distributor.md
timbussmann: the issue are remaining subscriptions.
* If you take down the distributor, you need to unsubscribe it first, which is quite tricky and you're facing a time where no subscription may be active which results in message loss
* If you upgrade a worker before shutting down the distributor you may get duplicate events (one for the v6 endpoint, one for the distributor)

Therefore the approach is to migrate the distributor to a worker node, so that the subscriptions in the publishers remain valid. The goal was to avoid "decommissioning steps" out of this guide (because it would require manually fiddling with the subscription storages which I really don't want to include in this guide since this alone is a dedicated doco page)

nservicebus/upgrades/5to6/distributor.md
timbussmann: It doesn't have to be one, but it needs to become a "worker" node (for the subscription reasons mentioned before). I'd say in that scenario the process would be upgrade -> decommission.

nservicebus/upgrades/5to6/distributor.md
timbussmann: depends on your needs. You can decommission it or keep it running as a regular node in which case it needs to be configured in sender-side distribution.

nservicebus/upgrades/5to6/distributor.md
timbussmann: as far as I understood, worker nodes also require the distributor package to enlist with a distributor, so this should be valid for all endpoints in a distributor setup.

nservicebus/upgrades/5to6/distributor.md
janovesk: Hmm, this is not necessarily possible. If A subscribes to B and B subscribes to A, you will have to start with one of them.

nservicebus/upgrades/5to6/distributor.md
timbussmann: I'd prefer `Migrating the Distributor to use Sender Side Distribution in NSB Version 6` but do we want to write "NSB"? (also it seems to be very long). Do we need to state versions,since it's part of the NServiceBus6 upgrade section?

e.g. how about "Migrating the Distributor to use Sender Side Distribution" ?

janovesk:
Consider the scenario where A subscribes to B, and B to A. Both run on a machine called "dist".

Following this guide, you would get:

1. v5 A subscription store contains : **(B@dist)**

   v5 B subscription store contains: **(A@dist)**
2. Shut down **A distributor** 

3. Upgrade A distributor to v6, run it as worker, let's call it **v6 A distworker**

4. **v6 A distworker** sends **subscribe (A@dist,A)** to **B**  

5. v5 B subscription store now contains: **(A@dist)** (ignores the endpoint name in the new subscription)

6. Upgrade other worker for A to v6, let's call it **v6 A worker1**

7. **v6 A worker1** sends **subscribe (A@worker1,A)** to B

8. v5 B now contains sub: **(A@dist)**, **(A@worker1)**. This will generate duplicate events.

Basically v5 endpoints can never receive subscribe messages from multiple instances of v6 send-side endpoints. v5 endpoints are unable to duplicate these subscriptions and will always generate duplicated events. 

I suggest we change the recommendation to shut down all the distributors first, upgrading them to v6 workers and then restarting them all. This will result in duplicates in the subscription storage, but it will be in v6 endpoints. It handles this gracefully and makes sure only ONE event is published per logical endpoint.  Once the distributors have been converted and started, the rest of the workers can be converted.

If you do all the distributors first, you get: 

1. v5 A subscription store contains : **(B@dist)**

   v5 B subscription store contains: **(A@dist)**
2. Shut down **A distributor** 

3. Shut down **B distributor** 

4. Upgrade A distributor to v6, run it as worker, let's call it **v6 A distworker**
5. Upgrade B distributor to v6, run it as worker, let's call it **v6 B distworker**

6. **v6 A distworker** sends **subscribe (A@dist,A)** to **B**  

7. **v6 B distworker** sends **subscribe (B@dist,B)** to **A**  

8. v6 A subscription store contains : **(B@dist, B)**

   v6 B subscription store contains: **(A@dist, A)**
9. Upgrade A worker1 to v6: let's call it **v6 A worker1**

10. **v6 A worker1** sends **subscribe (A@worker1,A)** to **B**

11. 
   v6 B subscription store contains: **(A@dist, A)**,**(A@worker1, A)**- Duplicates, but fine in a v6 endpoint

nservicebus/upgrades/5to6/distributor.md
indualagarsamy: @timbussmann - I am good with `Migrating the Distributor to use Sender Side Distribution`

nservicebus/upgrades/5to6/distributor.md
indualagarsamy: @timbussmann - i think this is good info and needs to be on the doco.

nservicebus/upgrades/5to6/distributor.md
indualagarsamy: @janovesk - Let me clarify. What I meant is, 

>Upgrade all endpoints communicating to the Distributor (by commands or event subscription messages) to NServiceBus Version 6 first.

In here, not sure if this includes the worker, because technically workers also send messages to the distributor. 

Perhaps clarify it as:

Upgrade all endpoints that interact with the distributor to Version 6 first. For example, upgrade all endpoints that send command messages to the Distributor endpoint for the load to be distributed, or endpoints that send subscription messages to the distributor,  subscribing to events when a worker node publishes an event. At this stage, do not upgrade the workers to Version 6.

timbussmann:
@janovesk 

> v6 A distworker sends subscribe (A@dist,A) to B

> v5 B subscription store now contains: (A@dist) (ignores the endpoint name in the new subscription)

I don't think that behavior is correct. the expected behavior is that the new subscription overwrites the old one. This works with inmemory and nhibernate persistence and I'm going to raise issues for Raven/Azure/MSMQ Persistence today.

janovesk:
It is correct that the new subscription overwrites the old one. It overwrites the old **"(A@dist)"** with a new "**(A@dist)"**, making no difference. It discards the endpoint name however, because the v5 subscription does not have that concept. It was introduced in v6. The problem then pops up when a A worker 1 sends **"subscribe (A@worker1,A)"**. The v5 subscription store will then store **(A@worker1)**, again dropping the endpoint name. End result: **(A@distworker), (A@worker1)** in the v5 B subscription store. That's going to produce duplicates.

We can't have multiple entries for the same logical endpoint in a v5 subscription store, that will always lead to duplicates because it does not know about endpoint names and thus can not do de-duplication.

SzymonPobiega:
@janovesk @timbussmann that's why my initial idea was to first upgrade all workers to V6 and restart them as enlisted in the distributor. This does not change the overall behavior of the system. 

I think in order to ensure that all subscriptions are V6 we need either to:
a) shut down, upgrade and start up all workers in all endpoints at once
b) shut down, upgrade and srat up all workers in each endpoint at once and repeat that procedure twice to ensure all subscriptions contain the endpoint name.

timbussmann:
I agree with @SzymonPobiega here. Once all workers are upgraded to v6, the subscriptions should be "correct" (containing a endpoint). I think the previous version of this PR contained the process described by szymon. I'll revert it this version which at the end leaves you with the following situation:
* All workers are upgraded to v6
* All workers except at least one are disconnected from the distributor. We need to keep the distributor and one worker running because of leftover subscriptions.

From there, you can now:
* Either decommission the distributor (which means to remove subscriptions manually from publishers)
* Or migrate the distributor to a regular worker

I think with this approach we can handle all situations include the windows cluster scenario mentioned by @janovesk and @DavidBoike ?

timbussmann:
@janovesk @SzymonPobiega please have a look again. I reverted back the process to the previous approach which now allows to either decommission the distributor or to upgrade it to a regular worker. Do you see any issues with the general approach?

I had to draw a lot the steps and configurations to make the steps really understandable to myself and I think it would add a lot of value to convert these drawing to diagrams alongside the steps to visualize the steps to readers, thoughts?

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,94 @@
...
+
+With sender-side distribution, senders are aware of scaled out receivers, allowing senders to distribute messages directly to the receivers instead of using an intermediary distributor.
+
+
+## Upgrade scenario
+
+This process aims to allow upgrade in production environments without missing events and minimal downtime.
+
+* Upgrade all endpoints that interact with the distributor to Version 6 first. For example, upgrade all endpoints that send command messages to the Distributor endpoint for the load to be distributed, or endpoints that send subscription messages to the distributor, subscribing to events when a worker node publishes an event. At this stage, do not upgrade the workers to Version 6.
+* Apply the following steps for each scaled out instance enlisted to the Distributor, one instance after another:
```
indualagarsamy: >Apply the following steps for each scaled out instance enlisted to the Distributor, one instance after another:

@timbussmann - I think rather than `scaled out instance`, people who use the distributor understand the term worker better. Did you mean worker node here?

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,94 @@
...
+  * Shut down the instance.
+  * [Upgrade the instance to NServiceBus Version 6](#upgrade-endpoint-to-version-6).
+  * [Enlist it with the distributor](#enlist-version-6-endpoints-with-a-distributor).
+  * Start the instance again.
+* Configure [sender-side distribution](/nservicebus/msmq/scalability-and-ha/sender-side-distribution.md) for all endpoints sending commands or receiving events from the distributor.
+* Detach the instances from the Distributor by  applying the following steps to the instances enlisted to the Distributor. But *skip this step for at least one instance* to ensure some instances are still attached to the distributor.
+  * Shut down the instance.
+  * Detach the instance from the Distributor by removing the `EnlistWithLegacyMSMQDistributor` configuration.
```
indualagarsamy: any relevant snippets we can add here?

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,94 @@
...
+  * Start the instance again.
+
+Most instances should now be detached from the Distributor. The Distributor can now either be upgraded to a regular worker node or be decommissioned:
+
+### Upgrade the Distributor to a worker node
```
indualagarsamy: Should this be, "Downgrade the distributor to a worker node" instead?

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,94 @@
...
+
+This approach allows you to continue using the resources used for the Distributor as a regular worker node.
+
+* Endpoint need to have the same endpoint name as other workers. Change queue somehow?
```
indualagarsamy: >Endpoint need to have the same endpoint name as other workers. Change queue somehow?
Is this work in progress? Question to @janovesk or @SzymonPobiega ?

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,94 @@
...
+
+* Shut down the Distributor.
+* Shut down the remaining instances attached to the Distributor.
+* Replace the Distributor with a worker endpoint.
```
indualagarsamy: what is involved in doing this?

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,94 @@
...
+* Detach the attached instances from the Distributor.
```
indualagarsamy: how does the user do this?

nservicebus/upgrades/5to6/distributor.md
timbussmann: yes. I can change that to worker if that makes it clearer.

nservicebus/upgrades/5to6/distributor.md
timbussmann: We could add a snippet for enlisting with the distributor using `EnlistWithLegacyMSMQDistributor`. I wonder whether we should add that to some distributor doc file instead and link to that?

nservicebus/upgrades/5to6/distributor.md
timbussmann: hm, how about "migrate"?

nservicebus/upgrades/5to6/distributor.md
timbussmann: I added that as a reminder to think the scenario through where the distributor has a different name than the workers, which means that the input queue is different too. But since we now support a "queue" attribute in the instance mapping file, I think this shouldn't be an issue and I can remove that point. Maybe @SzymonPobiega can confirm?

nservicebus/upgrades/5to6/distributor.md
timbussmann: I think the easiest would be to remove the endpoints binaries with those of a worker node, but we need to make sure that the worker uses the same input queue as the distributor, so we probably need to make this clearer.

nservicebus/upgrades/5to6/distributor.md
timbussmann: this is described above, since this is just the raw sequence of steps, I'll copy the full explanation here if we agree the process itself works.

indualagarsamy:
@SzymonPobiega @timbussmann  - is this still WIP?

timbussmann:
@indualagarsamy yes it misses the persistence specific sections to manually remove subscriptions. Otherwise it should be good.

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Removing subscriptions from NHibernate persistence
+
+Execute the following script against the database which is configured for NHibernate persistence:
+
+```
+DELETE
+FROM dbo.Subscription
+WHERE SubscriberEndpoint = '<distributorAddress>'
+```
```
timbussmann: @Particular/nhibernate-persistence-maintainers is that a valid/correct approach?

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,128 @@
...
+  * upgrade all endpoints that send subscription messages to the Distributor, 
+  * upgrade all endpoints subscribing to events published by worker nodes.
+* Apply the following steps for each worker, one after another:
+  * Shut down the worker.
+  * [Upgrade to NServiceBus Version 6](#upgrade-endpoint-to-version-6).
+  * [Configure it to enlist it with the distributor](#enlist-version-6-endpoints-with-a-distributor).
+  * Start it again.
+* Configure [sender-side distribution](/nservicebus/msmq/scalability-and-ha/sender-side-distribution.md) for all endpoints sending commands or publishing events to the scaled out endpoint.
+* Detach the workers from the Distributor by applying the following steps to the instances enlisted to the Distributor. But *skip this step for at least one instance* to ensure some workers remain attached to the distributor.
+  * Shut down the worker.
```
janovesk: As long as you have instances of the same endpoint BOTH running standalone AND as joined through their distributor AT THE SAME TIME, you will never drain the distributor queue.  Worker NOT joined with distributor will subscribe with (A@worker1,A) etc. Worker joined with distributor will subscribe (A@DIST,A). Both are stored in the subscription store  and both are included in the round robin for A. Meaning the distributor QUEUE will never really drain out.

If you run workers for the same endpoint both joined and not joined to a distributor, you get duplicated subscriptions and thus never empty distributor queue. On the other hand.. if you DON'T run at least one joined with the distributor, you risk having messages left in the distributor queue.

janovesk:
These are two alternate solutions I need you guys to poke holes in. In my head, they should both work.

Scenario:
A v5, B v5. Both behind distributors, both with two workers. They each subscribe to each other. Both want to upgrade to v6 and use SSD.

**a)** Offline and upgrade everything at the same time possible. Willing to manually change/delete subscriptions. Willing to manually move messages between queues.

  1. Upgrade endpoints to v6, NOT JOINING the legacy distributors.
  2. Set up message routing
  3. Create instance file
  4. Delete all subscriptions
  5. Deploy new v6 workers.
  6. Start up all workers.
  ASSUMPTION: Subscriptions now B: (A@worker1,A), (A@worker2,A), A: (B@worker1,B), (B@worker2,B)
  7. Move messages from A@dist to A@worker1/A@worker2. Discarding any subscribe messages
  8. Move messages from B@dist to B@worker1/B@worker2. Discarding any subscribe messages
  9. Remove A dist & B dist
  
  While the endpoints start up in 6, there is actually a time window where things might happen without the subscriptions having stableized. 

  Greatly helps if we create tools to support 4 and 7/8. (Queue Explorer makes 7/8 easy)

  Number of deploys: 1 + dist deletion


**b)** Offline and upgrade everything at the same time possible.

  1. Upgrade endpoints to v6, JOINING the legacy distributors.
  2. Set up message routing
  3. Create instance file 
  4. Deploy new v6 workers. 
  5. Start up all workers, joining distributors as v6 workers.
  ASSUMPTION: Subscriptions B: (A@dist,A), A: (B@dist,B)
  6. Let distributor queues dry out completely. New messages sent should be SSD, but publish will use subscribed A@dist and B@dist unless we pick instances from instance file on publish.
  7. Change workers to NOT JOIN legacy distributors.
  8. Deploy v6 workers. 
  9. Start up all workers.
  10. Remove A dist & B dist

  WILL NOT WORK WITHOUT PICKING INSTANCES FROM THE INSTANCE FILE!  

  Number of deploys: 2 + dist deletion

  Somewhat safer than a) because there is NO point in time without subscriptions correct.

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,128 @@
...
+
+## General upgrade scenario
+
+This process aims to allow upgrade in production environments without missing any messages and minimal downtime. If a scaled-out endpoint only receives commands and does not subscribe to messages consider following the [Simple upgrade scenario](#simple-upgrade-scenario) instead.
+
+* Upgrade all endpoints that interact with the distributor to Version 6 first. At this stage, do not upgrade the workers to Version 6.
+  * upgrade all endpoints that send command messages to the Distributor endpoint to be distributed,
+  * upgrade all endpoints that send subscription messages to the Distributor, 
+  * upgrade all endpoints subscribing to events published by worker nodes.
+* Apply the following steps for each worker, one after another:
```
janovesk: > upgrade all endpoints subscribing to events published by worker nodes.

I'm afraid this is not realistic. A can subscribe to B, just as B can subscribe to A. The final solution needs to work without assuming there are no such cycles. In my experience they happen all the time in the wild.

nservicebus/upgrades/5to6/distributor.md
timbussmann: > As long as you have instances of the same endpoint BOTH running standalone AND as joined through their distributor AT THE SAME TIME, you will never drain the distributor queue.

that's not required in this step. As there is still at least one instance which will handle incoming messages. depending on the next step (migrate to worker or remove), there are different actions to take as described after this section.

> If you run workers for the same endpoint both joined and not joined to a distributor, you get duplicated subscriptions and thus never empty distributor queue. 

I don't think you do. As all the workers were upgraded, they will subscribe with (A@worker1, A) for detached and (A@Dist, A) for attached workers, the publisher will round-robbin events between those, not duplicate as both provide a logical endpoint name.

nservicebus/upgrades/5to6/distributor.md
timbussmann: the scenario would work too I think. You would have to migrate all workers behind the distributor first while keeping the distributors in place.

nservicebus/upgrades/5to6/distributor.md
janovesk: >I don't think you do. As all the workers were upgraded, they will subscribe with (A@worker1, A) for detached and (A@Dist, A) for attached workers, **the publisher will round-robbin events between those**, not duplicate as both provide a logical endpoint name.

That's the problem. You have duplicate **subscription** for that event, both directly to worker and through the distributor. The publisher will round robin between them, because they both point to the same logical endpoint. If you have a round robin that includes the distributor, you will never drain the distributor queue. It will have continous incoming traffic. @timbussmann

nservicebus/upgrades/5to6/distributor.md
janovesk: If you migrate all the workers, how do you make sure no messages are left in the distributor queues? If you migrate all of them to not use the distributor, nobody will pick from the distributor queue. If you migrate some of them to use the distributor and some to not, the distributor queue will still be round robined to, meaning it will never empty out because there will be fresh incoming traffic continously.

nservicebus/upgrades/5to6/distributor.md
timbussmann: > That's the problem. You have duplicate subscription for that event, both directly to worker and through the distributor.

Sorry I'm not getting that statement. When a worker is attached to the distributor, it subscribers with the distributors address to the publisher. When a worker is detached it uses it's own endpoint address. So I can't see the issue with the event being distributed round-robbin between detached and attached workers? Also, this step does not require to drain the distributor queue, again, that's done in the next part where you unsubscribe the distributor manually and drain the queue? Am I missing something?

nservicebus/upgrades/5to6/distributor.md
timbussmann: this depends on the chosen option between
* migrating the distributor to a worker => in that case, draining the queue is not necessary
* decommissioning the distributor => the next section explains how to unsubscribe the distributor, drain the queue and remove it.

nservicebus/upgrades/5to6/distributor.md
janovesk: Your are right, @timbussmann. I don't know how I managed to miss the later drain step. Sorry! :)

SzymonPobiega:
@janovesk I don't like *a)* because of that stabilization period where we might miss an event. I prefer *b)* even in a variant where we remove the distributor subscription manually after the workers are started up as not enlisted.

SzymonPobiega:
I think this whole discussion proves that message-driven pub sub is non appropriate for the SSD scenarios at all. I think we might try to gather a task force to work on the spike @timbussmann started as an alternative for message-driven pub-sub migration.

nservicebus/upgrades/5to6/distributor.md
timbussmann: I'm not too familiar with the sample and whether it needs some more related to text to it. Maybe @SzymonPobiega can chime in on that question?

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,188 @@
...
+
+### Removing subscriptions from RavenDB persistence
+
+For every endpoint the Distributor (including its workers) subscribes to:
+* Find the publishers database in the RavenDB Management Studio.
+* Go to the Query view.
+* From the indexes select `Subscriptions`
+* Enter the following Query: `Subscribers:*"<endpointAddress>"*` where you replace `<endpointAddress>` with the address of your endpoint. Make sure to escape special characters like `.` or `-` with `\`. E.g. `my\.endpoint@distributor\-machine`.
+* Run the query.
+* On all listed documents, remove the entry with the Distributor's subscriber address and save it.
```
timbussmann: @Particular/ravendb-persistence-maintainers please confirm that this approach is correct

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,188 @@
...
+```
+DELETE
+FROM dbo.Subscription
+WHERE SubscriberEndpoint = '<distributorAddress>'
+```
+
+where `<distributorAddress>` is the address of the Distributor. E.g. `My.Endpoint@distributor-machine`.
+
+
+### Removing subscriptions from Azure Storage Persistence
```
timbussmann: @Scooletz 👇

timbussmann:
@Particular/docs-maintainers @SzymonPobiega @janovesk removed WIP label, please do another round of reviews.

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,188 @@
...
+        } else {
+            Write-Host 'Deletion aborted'
+        }
+    }
+}
+```
+
+Then execute the script:
+
+`Remove-Subscriptions -accountStorageName {storageAccountName} -accountStorageKey {storageAccountKey} -transportAddressToRemove {distributorAddress}`
```
Scooletz: I'd add the following for non-default subscription storage:

> If the subscriptions are configured to be stored in a non-default table, include it in the script:
`Remove-Subscriptions -accountStorageName {storageAccountName} -accountStorageKey {storageAccountKey} -subscriptionTableName {subscriptionTable} -transportAddressToRemove {distributorAddress}`

Scooletz: I'd consider dropping the brackets around parameter values `{ }`. The most common way of providing examples I've seen so far is to use `your` prefix like:

`-accountStorageName yourStorageAccountName`

SimonCropp:
this PR deletes nservicebus/scalability-and-ha/distributor/upgrading-the-distributor but provides no [redirect](https://github.com/Particular/docs.particular.net#redirects) for it.  is that intentional?

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,188 @@
...
+* Enter the following Query: `Subscribers:*"<endpointAddress>"*` where you replace `<endpointAddress>` with the address of your endpoint. Make sure to escape special characters like `.` or `-` with `\`. E.g. `my\.endpoint@distributor\-machine`.
+* Run the query.
+* On all listed documents, remove the entry with the Distributor's subscriber address and save it.
+
+
+### Removing subscriptions from NHibernate persistence
+
+Execute the following script against the database which is configured for NHibernate persistence:
+
+```
```
SimonCropp: can we add a `sql` key after the 3 ticks

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,188 @@
...
+FROM dbo.Subscription
+WHERE SubscriberEndpoint = '<distributorAddress>'
+```
+
+where `<distributorAddress>` is the address of the Distributor. E.g. `My.Endpoint@distributor-machine`.
+
+
+### Removing subscriptions from Azure Storage Persistence
+
+The following Powershell script removes all subscriptions of a specific subscriber from the configured Azure storage account. Copy & Paste the script to a Powershell and hit Enter:
```
SimonCropp: use `and` not `&` http://english.stackexchange.com/a/3145/1227

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,188 @@
...
+
+```
```
SimonCropp: add a `ps` key to the ricks

SimonCropp: why is this not in a `.ps` file?

nservicebus/upgrades/5to6/distributor.md
SimonCropp: single ticks should only be used for shorter inline code. can u make this a triple  tick with a `ps` key

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,188 @@
...
+* When the Distributor input queue is empty, shut down the Distributor.
+* When the attached instances input queues are empty, shut down the attached instances.
+* Remove the Distributor and the attached instances. Alternatively you can detach the attached instances from the Distributor and start them again (make sure to include them in the sender-side distribution configuration too).
+
+
+## Simple upgrade scenario
+
+This process describes a faster way to migrate a scaled-out endpoint using the Distributor to NServiceBus Version 6. This approach can only be applied if the endpoint does not subscribe to events and only handles incoming commands. Do not use this approach when endpoints use `Reply` to send messages back to workers attached to a Distributor.
+
+WARNING: Following this process when endpoints subscribe to events may cause duplicate events or message loss!
```
SimonCropp: no `!`. it is already a warning

SimonCropp: in fact lets make this a `DANGER:`

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,188 @@
...
+The [distributor](/nservicebus/scalability-and-ha/distributor) has been deprecated and replaced by [sender-side distribution](/nservicebus/msmq/scalability-and-ha/sender-side-distribution.md) with NServiceBus Version 6.
+
+## Client-side distribution
+
+With sender-side distribution, senders are aware of scaled out receivers, allowing senders to distribute messages directly to the receivers instead of using an intermediary distributor.
+
+
+## General upgrade scenario
+
+This process aims to allow upgrade in production environments without missing any messages and minimal downtime. If a scaled-out endpoint only receives commands and does not subscribe to messages consider following the [Simple upgrade scenario](#simple-upgrade-scenario) instead.
```
SimonCropp: "missing any messages" is strange phrasing. usually we use the term "message loss"

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,188 @@
...
+
+When decommissioning the Distributor, it is necessary to manually remove remaining subscriptions in case the Distributor receives events to avoid issues on publishers or events being lost.
+
+Follow the steps specified by the used subscription storage:
+
+
+### Removing subscriptions from RavenDB persistence
+
+For every endpoint the Distributor (including its workers) subscribes to:
+* Find the publishers database in the RavenDB Management Studio.
```
SimonCropp: link for `RavenDB Management Studio`

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,188 @@
...
+* Go to the Query view.
```
SimonCropp: link for `Query view.` if it exists

nservicebus/upgrades/5to6/distributor.md
SimonCropp: would a screen shot help here?

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,188 @@
...
+
+
+## Remove Distributor subscriptions
+
+When decommissioning the Distributor, it is necessary to manually remove remaining subscriptions in case the Distributor receives events to avoid issues on publishers or events being lost.
+
+Follow the steps specified by the used subscription storage:
+
+
+### Removing subscriptions from RavenDB persistence
```
SimonCropp: perhaps mention, and link to backup doco, here?

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,188 @@
...
+* Find the publishers database in the RavenDB Management Studio.
+* Go to the Query view.
+* From the indexes select `Subscriptions`
+* Enter the following Query: `Subscribers:*"<endpointAddress>"*` where you replace `<endpointAddress>` with the address of your endpoint. Make sure to escape special characters like `.` or `-` with `\`. E.g. `my\.endpoint@distributor\-machine`.
+* Run the query.
+* On all listed documents, remove the entry with the Distributor's subscriber address and save it.
+
+
+### Removing subscriptions from NHibernate persistence
+
```
SimonCropp: perhaps mention, and link to backup doco, here?

nservicebus/upgrades/5to6/distributor.md
SimonCropp: perhaps mention, and link to backup doco, here?

nservicebus/upgrades/5to6/distributor.md
SimonCropp: should add a related to https://docs.particular.net/nservicebus/scalability-and-ha/distributor/

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,188 @@
...
+* Upgrade all endpoints that interact with the distributor to Version 6 first. At this stage, do not upgrade the workers to Version 6.
+  * upgrade all endpoints that send command messages to the Distributor endpoint to be distributed,
+  * upgrade all endpoints that send subscription messages to the Distributor, 
+  * upgrade all endpoints subscribing to events published by worker nodes.
+* Apply the following steps for each worker, one after another:
+  * Shut down the worker.
+  * [Upgrade to NServiceBus Version 6](#upgrade-endpoint-to-version-6).
+  * [Configure it to enlist it with the distributor](#enlist-version-6-endpoints-with-a-distributor).
+  * Start it again.
```
SimonCropp: keep the language consistent. "Start the worker again"

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,188 @@
...
+* Configure [sender-side distribution](/nservicebus/msmq/scalability-and-ha/sender-side-distribution.md) for all endpoints sending commands or publishing events to the scaled out endpoint.
+* Detach the workers from the Distributor by applying the following steps to the instances enlisted to the Distributor. But *skip this step for at least one instance* to ensure some workers remain attached to the distributor.
```
SimonCropp: i think `skip this step for at least one instance` should be bold not italic

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,188 @@
...
+  * Shut down the worker.
+  * Remove the `EnlistWithLegacyMSMQDistributor` configuration.
+  * Start the worker again.
+
+Most instances should now be detached from the Distributor. The Distributor can now either be upgraded to a regular worker node or be decommissioned:
```
SimonCropp: "Most instances should now be detached from the Distributor"  why the vague language? did we instruct them to detach a random number of workers?

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,188 @@
...
+
+### Migrate the Distributor to a worker node
+
+This approach allows you to continue using the resources used for the Distributor as a regular worker node. This step only works when the Distributor endpoint has the same name as the worker node.
```
SimonCropp: "This approach enables the continued utilization of the resources used for the Distributor as a regular worker node"

SimonCropp:
Does the error queue and retries have any impact on this documentation. if it does not can it be explicitly called out that nothing need to be done in that area?

nservicebus/upgrades/5to6/distributor.md
SimonCropp: `production` is redundant. they will want to test this in lower environments anyway

SimonCropp:
can some info be added as to verifying their chosen approach in a lower environment?

nservicebus/upgrades/5to6/distributor.md
SimonCropp: PowerShell not Powershell

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,188 @@
...
+        $entries = $table.CloudTable.ExecuteQuery($query)
+
+        Write-Host "Following subscriptions will be removed for " $transportAddressToRemove
+
+        $entries | Format-Table @{ Label = "MessageType"; Expression={$_.PartitionKey}}, @{ Label = "EndpointName"; Expression={$_.Properties["EndpointName"].StringValue }} -AutoSize
+
+        $confirmation = Read-Host "Are you sure you want to remove this entries? [y/n]"
+
+        if ($confirmation -eq 'y') {
+            # proceed
```
SimonCropp: proceed?

nservicebus/upgrades/5to6/distributor.md
SimonCropp: link to Azure Storage Persistence doco

nservicebus/upgrades/5to6/distributor.md
```diff
@@ -0,0 +1,188 @@
...
+For every endpoint the Distributor (including its workers) subscribes to:
+* Find the publishers database in the RavenDB Management Studio.
+* Go to the Query view.
+* From the indexes select `Subscriptions`
+* Enter the following Query: `Subscribers:*"<endpointAddress>"*` where you replace `<endpointAddress>` with the address of your endpoint. Make sure to escape special characters like `.` or `-` with `\`. E.g. `my\.endpoint@distributor\-machine`.
+* Run the query.
+* On all listed documents, remove the entry with the Distributor's subscriber address and save it.
+
+
+### Removing subscriptions from NHibernate persistence
```
SimonCropp: link to nh doco

nservicebus/upgrades/5to6/distributor.md
SimonCropp: link to raven doco

nservicebus/upgrades/5to6/distributor.md
SimonCropp: i would actually prefer this to be added to a scripting document under here https://docs.particular.net/nservicebus/azure-storage-persistence/ and then linked to

nservicebus/upgrades/5to6/distributor.md
SimonCropp: can this be moved to a new scripting document under here https://docs.particular.net/nservicebus/nhibernate/

nservicebus/upgrades/5to6/distributor.md
SimonCropp: is there really no way to script this? if not i would prefer to see a custom tool written for it.

nservicebus/upgrades/5to6/distributor.md
SzymonPobiega: I think it would be easier to understand if we changed the guidance to:
 * Keep workers attached
 * Remove the distributor entry from publishers' subscription storages
 * Update all timeout to point to one of the worker nodes instead of the distributor
 * Wait till the queue dries out
 * Restart the workers as not enlisted

nservicebus/upgrades/5to6/distributor.md
timbussmann: @SimonCropp what link would you use? http://ravendb.net/docs/article-page/3.5/Csharp/studio/accessing-studio (seems to be too coupled to the version)?

nservicebus/upgrades/5to6/distributor.md
timbussmann: I think that could help. Alternatively we could give an example for the document JSON before and after instead of a screenshot?

nservicebus/upgrades/5to6/distributor.md
timbussmann: hm suggestions where to link? Since this mainly relates to the used SQL Server, I'm not sure where to point at.

nservicebus/upgrades/5to6/distributor.md
timbussmann: @SzymonPobiega so you're saying we should always guide on how to remove subscriptions, even when the distributor will be migrated to a worker node?

nservicebus/upgrades/5to6/distributor.md
timbussmann: the guidance states that you should remain at least one worker attached to the distributor (`skip this step for at least one instance`) but of course it could be more, depending on the message pressure. Maybe we can leave out the first sentence altogether?

timbussmann:
> Does the error queue and retries have any impact on this documentation. if it does not can it be explicitly called out that nothing need to be done in that area?

* error queue should have no impact
* retries from v5 workers are sent to the the distributors "Retries" queue. This means that we could loose timeouts when the distributor is decommissioned. If the distributor is migrated to a worker, I think there should be no issue. So I guess we need to add some statements to the decommissioning section which describes how to check for remaining timeouts and how to avoid message loss? /cc @SzymonPobiega

timbussmann:
> this PR deletes nservicebus/scalability-and-ha/distributor/upgrading-the-distributor but provides no redirect for it. is that intentional?

did add a redirect 👍

timbussmann:
I think i addressed most of the comments. Github won't remove review comments :(

indualagarsamy:
@timbussmann @janovesk @SzymonPobiega - is this content good to be reviewed?

timbussmann:
@indualagarsamy yes it is

SzymonPobiega:
> So I guess we need to add some statements to the decommissioning section which describes how to check for remaining timeouts and how to avoid message loss?

@WojcikMike prepared a timeouts re-routing guide which we can reference.

timbussmann:
> @WojcikMike prepared a timeouts re-routing guide which we can reference.

I can't state how happy I am to hear, that at least that I don't have to write ❤️ 😆

nservicebus/upgrades/5to6/distributor.md
SzymonPobiega: If it will be migrated to the worker node, we don't need to keep one worker enlisted either. So I'd say we should:
 * Start workers as attached to distributor
 * Decide if distributor is going to me turned into worker (masternode scenario) or decommissioned (pure distributor, probably in HA mode)
 * if former, detached the workers, upgrade distributor to worker and done
 * if latter, proceed with removing the subscriptions and re-routing timeouts.

Scooletz:
@timbussmann I need to review the Azure scripts again. The current version might not work properly. Will keep you posted.

SimonCropp:
OK i am calling this one "good enough"

@Scooletz if u need to update the azure scripts can u raise a new PR

SzymonPobiega:
🎉

PR #1880, opened 2016-09-08
Remove paragraph about running multiple versions of the same gateway

janovesk:
Running multiple gateways is unfortunately not possible ATM. Gateways currently route to their local input queues only. If you run multiple versions with the same endpoint name on the same machine, they will all put the incoming messages on the same queue. The different versions will then use competing consumer against this queue, meaning you have no control if the incoming vX message will have handlers run on the vX endpoint or the vX+1 endpoint. This is not safe, as there is no guarantee that vX and vX+1 will handle the same message type and if they do, will not necessarily have the same handlers.


janovesk:
Have a look at the explanation and pull, @DavidBoike.

SimonCropp:
LGTM

PR #1884, opened 2016-09-08
Clarify that we no longer allow correlated properties to have a null …

andreasohlund:
…or default value

And how to workaround it using a custom finder

### DIFF
--- a/nservicebus/upgrades/5to6/handlers-and-sagas.md
+++ b/nservicebus/upgrades/5to6/handlers-and-sagas.md
@@ -107,6 +107,8 @@ snippet:5to6-ConfigureHowToFindSaga
 
 Version 6 automatically correlates incoming message properties to its saga data counterparts. Any saga data correlation in the message handler code can be safely removed. Correlated properties (for existing saga instances) will not be changed once set.
 
+NOTE: NServiceBus will no longer allow a correlated property to have a null/default value. Use a custom finder for the relevant message to override this validation.
+
 snippet:5to6-NoSagaDataCorrelationNeeded
 
 Version 6 will only support correlating messages to a single saga property. Correlating on more than one property is still supported by creating a custom [saga finder](/nservicebus/sagas/saga-finding.md).
--- a/nservicebus/upgrades/5to6/handlers-and-sagas.md
+++ b/nservicebus/upgrades/5to6/handlers-and-sagas.md
@@ -107,11 +107,11 @@ snippet:5to6-ConfigureHowToFindSaga
 
 Version 6 automatically correlates incoming message properties to its saga data counterparts. Any saga data correlation in the message handler code can be safely removed. Correlated properties (for existing saga instances) will not be changed once set.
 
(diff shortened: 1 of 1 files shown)


nservicebus/upgrades/5to6/handlers-and-sagas.md
```diff
@@ -107,6 +107,8 @@ snippet:5to6-ConfigureHowToFindSaga
 
 Version 6 automatically correlates incoming message properties to its saga data counterparts. Any saga data correlation in the message handler code can be safely removed. Correlated properties (for existing saga instances) will not be changed once set.
 
+NOTE: NServiceBus will no longer allow a correlated property to have a null/default value. Use a custom finder for the relevant message to override this validation.
```
SimonCropp: "No longer" is ambiguous.

SimonCropp: Perhaps link here https://msdn.microsoft.com/en-us/library/83fhsxwc.aspx

SimonCropp: Add the exception message

ramonsmits:
Just thinking about this... Do we raise an exception if we receive a message that has a value in the correlation property that is a default value? Its a bit strange that we would not allow to persist it but still query using the default value and execute the handler.

SimonCropp:
@Particular/nservicebus-maintainers can one of you chime in on @ramonsmits comment above

andreasohlund:
We don't check the message so it would blow up "with default value not
allowed" when persisting the saga.

But again I don't see the use case for correlating on a null/default value?

The same workaround still applies: Create a custom finder

andreasohlund:
And we can always put out a minor release that would provide a "opt-out"
setting should users complain. This way we will be getting the feedback at
least? (perhaps there is a valid use case)

That said there might be one for "singleton" saga where they might have
used a Guid.Default or something to always hit the same saga but lets see?

SimonCropp:
@andreasohlund but if we get a message and the saga config says "use this prop on message" and that message prop is default shouldnt that be an exception state?

andreasohlund:
What if the user override? (they can still change the corr id if they want)

```
 
if(message.OrderId == null)
{
    this.Data.OrderId = someOtherId;
}

```

The current code will not throw in the above scenario. I do agree that this is a funky scenario?

SimonCropp:
@andreasohlund i would say thats is a bug since if they have duplicate messages with the same OrderId  then every one of them will create a new saga

andreasohlund:
I agree, lets raise an issue to track this? (since this is unlikely to
happen in the current version given that we never persist a saga with a
null/default corr id)

SimonCropp:
@ramonsmits since u raised the initial question can u raise an issue in core.

@andreasohlund still some outstanding comments on the content of this PR

andreasohlund:
@SimonCropp your comments have been addressed

PR #1899, opened 2016-09-12
Simplify behavior registration for v6

timbussmann:
directly register a behavior without using a registrationstep.


nservicebus/pipeline/manipulate-with-behaviors_registernew_core_[5.0,6.0).partial
```diff
@@ -0,0 +1,8 @@
+
```
timbussmann: is the version range correct this way?

PR #1908, opened 2016-09-14
Switch PrefetchCount to PrefetchCountPerMessageProcessor for RabbitMQ…

adamralph:
… transport 4.0


adamralph:
I'm not sure if I'm using the partials correctly here...

PR #1910, opened 2016-09-14
Clarify that legacy multi-instance topology is for backwards compat only

### LINKED ISSUE #1819: SQL Server transport legacy multi-instance mode is for backwards compatibility only
We should document that legacy support for multi-instance topology in SQL Server transport V3 is only to allow smooth interop with V2 endpoints using this topology and we do not recommend its usage for new deployments.

MarcinHoppe:
Fixes #1819


MarcinHoppe:
@Particular/docs-maintainers This is ready for review.

PR #1911, opened 2016-09-14
Update to ASB 7.0.0-RC2

andreasohlund:


ramonsmits:
@andreasohlund as asked, tested a few and worked.

PR #1927, opened 2016-09-20
Core RC2 and more Callback RC2 updates

danielmarbach:
@Particular/docs-maintainers as mentioned in slack missing updates


SimonCropp:
@danielmarbach good catch

PR #1931, opened 2016-09-21
Adjust MSMQ subscription storage doco to actual behavior

timbussmann:
relates to https://github.com/Particular/NServiceBus/issues/4147

/cc @andreasohlund

### DIFF
--- a/nservicebus/msmq/subscription-persistence_config_core_[5.0,].partial.md
+++ b/nservicebus/msmq/subscription-persistence_config_core_[5.0,].partial.md
@@ -6,4 +6,4 @@ By default NServiceBus uses a queue called `NServiceBus.Subscriptions`. In order
 
 snippet:MsmqSubscriptionAppConfig
 
-WARNING: When using MSMQ Subscription Persistence on multiple endpoints running on the same machine, every endpoint requires it's own dedicated subscription storage queue.
+WARNING: When using MSMQ Subscription Persistence on multiple endpoints running on the same machine, every endpoint requires a dedicated subscription storage queue.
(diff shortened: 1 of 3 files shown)


andreasohlund:
LGTM

nservicebus/msmq/subscription-persistence_config_core_[5.0,].partial.md
```diff
@@ -0,0 +1,9 @@
+To configure MSMQ as the subscription persistence:
+
+snippet:ConfiguringMsmqPersistence
+
+By default NServiceBus uses a queue called `NServiceBus.Subscriptions`. In order to specify a different queue to store the subscriptions, add the following `configSections` and subsequent config entry:
+
+snippet:MsmqSubscriptionAppConfig
+
+WARNING: When using MSMQ Subscription Persistence on multiple endpoints running on the same machine, every endpoint requires it's own dedicated subscription storage queue.
```
Comment (SimonCropp) [inline nservicebus/msmq/subscription-persistence_config_core_[5.0,].partial.md]: "it is own"?

Comment (andreasohlund) [inline nservicebus/msmq/subscription-persistence_config_core_[5.0,].partial.md]: perhaps "a dedicated"?

weralabaj:
@timbussmann @andreasohlund it's not clear from this doco taht "This may cause different logical endpoints to use the same subscription storage." (mentioned in the linked issue), should we state more explicitly that if you don't provide the name, the default will be used, which will have that effect?

timbussmann:
@weralabaj that was the reason why I added that warning. I'm not sure whether explaining it in more detail adds more weight or value to it?

weralabaj:
Let me rephrase slightly.

PR #1934, opened 2016-09-21
Describe difference between correlation and conversation IDs

MarcinHoppe:
@Particular/docs-maintainers @andreasohlund @dvdstelt Please have a look at my attempt to describe the difference between correlation and conversation IDs. I feel it is not perfect yet (hence [WIP]) but would love to get some realy feedback.


nservicebus/messaging/message-correlation.md
```diff
@@ -7,14 +7,18 @@ related:
  - nservicebus/messaging/headers
 ---
 
-Message correlation is the act of connecting instances of request messages with instances of response messages. Since all messages have a `Message Id` the `Correlation Id` is just a reference back to the specific message instance that caused this message to be sent back as a reply.
+Message correlation connects instances of request messages with instances of response messages. Each message has a `Message Id` and the `Correlation Id` is a reference back to the specific message instance that caused this message to be sent back as a reply.
```
weralabaj: How's that different from `NServiceBus.RelatedTo`?

Also in https://docs.particular.net/nservicebus/messaging/headers we say "A string used to correlate reply messages to their corresponding request message.", so we probably need to udpate that doco too?

weralabaj: BTW I've always thought this is some remain from old version used for the sake of SI, so clarification would be very useful :)

MarcinHoppe: What's incorrect about the statement in the headers doco?

The differences between `NServiceBus.RelatedTo` and `NServiceBus.CorrelationId` do exist. One of them is that the correlation ID is customizable by the user. Do you think it would be good to expand on this as well?

weralabaj: Yes @MarcinHoppe, more detail can't hurt ;)

andreasohlund: >How's that different from NServiceBus.RelatedTo?

RelatedTo is a pointer to the message that caused a given message to be sent. Ie if you process `PlaceOrderCommand` and as a result publish the `OrderPlacedEvent`, the event will have:

RelatedTo= PlaceOrderCommand.MessageId

This allows us , together with the conversationId to create the message flow diagram in SI.

Does this make sense?

andreasohlund: >the `Correlation Id` is a reference back to the specific message instance that caused this message to be sent back as a reply.

Should we link to http://www.enterpriseintegrationpatterns.com/patterns/messaging/CorrelationIdentifier.html ?

weralabaj: > RelatedTo is a pointer to the message that caused a given message to be sent. 

> the Correlation Id is a reference back to the specific message instance that caused this message to be sent back as a reply.

I still don't get what's the difference. Is correlationId used only for reply messages and RelatedTo to all "preceding" messages, even if it's not a reply message?
Pointer/reference, at the end of the day isn't that all just message id?

weralabaj:
Summary of the call with @andreasohlund and my current understanding:

- Conversation Id - used to tie messages together in a flow, utilized in SI, might be overriden by user if they need to combine two flows or have a "business-level flow" (e.g. by using OrderId)
- Correlation Id - used to tie messages for callbacks and request-response (?), might be overriden by user, not sure for what reason or whether it's a good idea in the first place (can they mess up callbacks? if not then why conversation id and related to are not sufficient?)
- RelatedTo - used to tie messages to message that was sent directly before them, different from conversation id which points to the first message in the flow, not the previous one, not sure how's that used internally

So the open questions:
- What's the use case for correlation id?
- Why do we expose that to the user? What's the use case?
- Is exposing correlation id safe?
- What we woulnd't be able to do without correlation id (what features depend on it)?
- What we woulnd't be able to do without related to (what features depend on it)?

nservicebus/messaging/message-correlation.md
dvdstelt: Good questions, if it's not clear to us, how can it be clear to our users?

@MarcinHoppe CorrelationId and MessageId can be the same, if there's no originating message. From what I understand, this is always the case with commands?

Also it's focused on the reply, but isn't the CorrelationId different when publishing events?

nservicebus/messaging/message-correlation.md
```diff
@@ -7,14 +7,18 @@ related:
...
 
 This is a pattern from the [Enterprise Integration Patterns book](http://www.enterpriseintegrationpatterns.com/patterns/messaging/CorrelationIdentifier.html).
 
-One example usage is that it allows NServiceBus to find out which callback to invoke should the user have requested a callback when the request message was sent out.
+One example usage within NServiceBus is to find out which callback to invoke should the user have requested a callback when the request message was sent out.
```
dvdstelt: When publishing a message and the CorrelationId is MessageId of the original incoming, should we not add "Another example" here, so users understand the usages of the ids?

nservicebus/messaging/message-correlation.md
```diff
@@ -7,14 +7,18 @@ related:
...
 
-From a message payload perspective the `Correlation Id` is just an ordinary [message header](/nservicebus/messaging/headers.md) that defaults to the same value as the message ID for new messages but automatically gets set to the `Message Id` of the incoming message when doing a `Reply`.
+From a message payload perspective the `Correlation Id` is transmitted as a [message header](/nservicebus/messaging/headers.md) that defaults to the same value as the message ID for new messages but automatically gets set to the `Message Id` of the incoming message when doing a `Reply`.
```
dvdstelt: or publish?

nservicebus/messaging/message-correlation.md
```diff
@@ -7,14 +7,18 @@ related:
...
 
 NServiceBus handles this automatically. Should full control be needed over the `Correlation Id` for integration purposes use the following code when sending the message
 
 snippet:custom-correlationid
+
+### Correlation vs conversations
+
+Correlation allows NServiceBus to match requests with responses. In addition to this NServiceBus tracks message flows than span more than one message exchange. First message that is sent in a new flow is assigned a unique `Conversation Id` that is then propagated to all the messages that are subsequently sent, thus forming a _conversation_. Similar to the `Correlation Id`, the `Conversation Id` is transmitted in `NServiceBus.ConversationId` message header.
```
dvdstelt: > ~~than span~~ that span

danielmarbach:
> - What's the use case for correlation id?
Like you pointed out the callback API uses the correlation id to be able to correlate back a request which was sent out to the task completion source which will complete the request when a message comes in as a reply. Any tampering from the users on the replying side (meaning the conversation ID might not longer be the message id) will break the callbacks API.

- Why do we expose that to the user? What's the use case? Why can't they use conversation id or custom header?

MessageId and ConversationId are a fundamental pattern in messaging. A message id is used as a correlation id when reply to. This means be can always correlate back. Some transport layers expose the correlation id as a property on their messaging entity. This means from a native integration perspective the message id and correlation id has to be transported and carried over all the time. In my old active mq days we used this principle to encode reply addresses into the message id which then when a third party integrator replied was always present. This meant we didn't have to teach third parties to concept of: Oh btw. these headers from NServiceBus you have to send back always, if not our clients will break.

- Is exposing correlation id safe?
Nope it is not. @timbussmann, @andreasohlund and I had longer conversations about that. As soon as you tamper with it there might be dragons. We even contemplated removing it in v6 (my Github google foo fails me here).

- What we woulnd't be able to do without correlation id (what features depend on it)?
Nope see explanation above in point two.
- What we woulnd't be able to do without related to (what features depend on it)?

andreasohlund:
>Is exposing correlation id safe?
>Nope it is not. @timbussmann, @andreasohlund and I had longer conversations about that. As soon as you tamper with it there might be dragons. We even contemplated removing it in v6 (my Github google foo fails me here).

I wonder if we mixed up conversation and correlation ids since it does makes sense to allow users to customize the conversation id (ant not the correlationid like you say). 

Eg in the scenario where you are sending lets say PlaceOrder + CancelOrder commands they would show up as different message flows in SI since both commands will get a fresh conversation id. Should the user instead override and use the same "orderid" for both commands the message flow view in SI would show the entire flow properly.

Action points seems to be:

1. Deprecate the SendOptions.SetCorrelationId() in the next major
2. Introduce a .SetConversationId(...), this can be done in a minor if needed

Does this make sense?

indualagarsamy:
@andreasohlund - can we raise new issues for the action points? From a documentation perspective, what can be changed now?

andreasohlund:
@Particular/nservicebus-maintainers are you good with the actions points in https://github.com/Particular/docs.particular.net/pull/1934#issuecomment-249604674 ? (I'll raise the relevant issues)

weralabaj:
@indualagarsamy @MarcinHoppe from the doco perspective, let's capture what's been clarified in the comments without saying anything about the future changes. we should be clear that for nice flows in SI one should use ConversationId.

MarcinHoppe:
@weralabaj I will now incorporate all the comments posted so far.

andreasohlund:
Core issues raised:

https://github.com/Particular/NServiceBus/issues/4170
https://github.com/Particular/NServiceBus/issues/4171

nservicebus/messaging/message-correlation.md
MarcinHoppe: @dvdstelt When we publish an event from within the handler, the correlation Id is the message Id of the published message, not the correlation Id of the incoming message. Correct?

nservicebus/messaging/message-correlation.md
MarcinHoppe: Ditto.

MarcinHoppe:
@Particular/docs-maintainers I refined the description a little bit. I omitted the part about customizing conversation IDs as we currently don't have an API for this.

If we feel this is important, I'd be happy to create a sample that hooks into the pipeline to do this and link the sample from this page.

weralabaj:
@MarcinHoppe I updated based on the conversation with Andreas, I agree we can create a sample to show how to override a conversation Id, but doesn't seem to be a high-prio. On the other hand my understanding is that we DON'T WANT users to manipulate correlation ids and that API will be removed. So I'd rather not mention it in docs.

See if it still makes sense after my review.

MarcinHoppe:
@weralabaj Makes sense. I improved the wording a little bit and made the build green.

Issue #1829, opened 2016-08-30
Make AAA style testing more prominent on testing page


danielmarbach:
Based on https://github.com/Particular/NServiceBus.Testing/issues/90#issuecomment-243380979 I think we should make AAA style testing more prominent on the main testing page. For example first show a quick example of AAA style testing with further information link to the sample page and then show the fluent style syntax. Thoughts @Particular/docs-maintainers ?

SimonCropp:
sounds good to me

weralabaj:
@danielmarbach do you think that this issue could be addressed by having version switching for this page, rather than a dedicated page for v6 (which is a sample, sigh....)?

IMO that would be simpler. see related https://github.com/Particular/docs.particular.net/issues/1746, didn't manage to get to it before vacation though :)

danielmarbach:
Not sure @weralabaj. As far as I understand version switches is that they should be used if the content of the snippet is similar. But here I'm talking about revamping the page to show AAA style testing first and make it more prominent.

SimonCropp:
@danielmarbach 

> As far as I understand version switches is that they should be used if the content of the snippet is similar.

nope. some pages switch all (or almost all) their content. eg https://docs.particular.net/nservicebus/messaging/routing-extensibility

danielmarbach:
Ok then I'm fine with that approach

PR #1935, opened 2016-09-21
Update testing page

weralabaj:
Resolves https://github.com/Particular/docs.particular.net/issues/1829


weralabaj:
@timbussmann looking at https://docs.particular.net/nservicebus/testing/ do you think it makes sense to provide samples of AAA style for anything else than the basic handler and saga tests?

BTW I'm not sure how to get working saga test when saga in the handler calls ReplyToOriginator and RequestTimeout. I might hack it, but that requires knowledge of NSB internals, what's the recommended approach?

timbussmann:
with that approach, it looks like we have a lot of duplicated code from this page: https://docs.particular.net/samples/unit-testing/ ?

weralabaj:
true, but I don't think we can have sample and article combined on a one page :/

timbussmann:
There is no real way of automatically trigger timeouts with the unit testing approach. You would just have to handle the timeout message. You can verify whether a timeout has been requested by something like:

```
var timeout= context.SentMessages.Containing<TimeoutMessage>().Single();
Assert.That(timeout.Options.GetDeliveryDelay(), Is.EqualTo(TimeSpan.FromMinutes(5)));
```

or similar. I think we could provide an extension method which checks the `NServiceBus.Headers.IsSagaTimeoutMessage` header. e.g. something like `Where(m => m.IsSagaTimeout())`. Thoughts?

For ReplyToOriginator you would write:

```
var reply = context.SentMessages.Containing<SubmitOrder>().Single();
Assert.That(reply.Options.GetDestination(), Is.EqualTo(saga.Data.Originator));
```

weralabaj:
@timbussmann I think it might be useful to extend the sample to cover at least the same scenarios as the current article describes for all versions, I'll raise a separate issue though, ok?

weralabaj:
Notes 
[ ] ExpectSagaData is not covered

weralabaj:
@timbussmann  https://github.com/Particular/docs.particular.net/issues/1938

danielmarbach:
Awesome sauce @weralabaj

PR #1936, opened 2016-09-21
Fix casing in msmq routing file snippet

andreasohlund:
@timbussmann @SzymonPobiega should we consider making the comparison case insensetive to avoid issues with casing? (see https://groups.google.com/d/msgid/particularsoftware/a3312514-c2bf-4f5c-aa6e-d9f988441f41%40googlegroups.com?utm_medium=email&utm_source=footer)

### DIFF
--- a/Snippets/Core/Core_6/Routing/MSMQ/InstanceMappingFile.xml
+++ b/Snippets/Core/Core_6/Routing/MSMQ/InstanceMappingFile.xml
@@ -3,13 +3,13 @@
   <!-- startcode InstanceMappingFile -->
   <endpoints>
     <endpoint name="Sales">
-      <instance Machine="VM-1"/>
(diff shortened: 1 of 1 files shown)


indualagarsamy:
@andreasohlund - for the case comparison, can you open a new issue? I figured, since this is incorrect and you fixed it, it might be better to get this merged first.

andreasohlund:
Will do after I sync with @timbussman

indualagarsamy:
Awesome, thanks.

timbussmann:
> @timbussmann @SzymonPobiega should we consider making the comparison case insensetive to avoid issues with casing

+ 💯

andreasohlund:
@indualagarsamy issue raised https://github.com/Particular/NServiceBus/issues/4154

PR #1937, opened 2016-09-22
Clarify how we backport

andreasohlund:
Thoughts?


andreasohlund:
I saw now that we're repeating this in the `Backport important bugfixes` section below. Perhaps just remove the sentence about backporting from the first paragraph?

PR #1948, opened 2016-09-23
Add a sample for version 5

WilliamBZA:
Adding a sample for version 5 of NServiceBus.NHibernate.

Not sure if there is value in this, but I needed to do it for a support case so perhaps it can just be included anyway.


SzymonPobiega:
@WilliamBZA looks good. I'll merge it. I know we don't support NSB 4 but the doco in general extends to 4 or even to 3 in some places.

PR #1949, opened 2016-09-23
Asb perf tuning

yvesgoeleven:


yvesgoeleven:
ping @SeanFeldman

SeanFeldman:
|Concurrency  |PrefetchCount| Result|
|---------------|----------------|-------|
|     Nx2          |           Cx1       | |
...

N = number of logical cores
C = concurrency level

Concurrency to try : N x 2, 4, 8, 16
prefetch count to try: C x 1, 2(best, but still poor), 3

SeanFeldman:
Multiple receivers, each with a certain concurrency, total to have a rate limiter just like [ASQ does](https://docs.particular.net/nservicebus/azure-storage-queues/configuration#configuration-parameters-degreeofreceiveparallelism) it or using a custom task scheduler to be passed into ASB client, _if possible_.

indualagarsamy:
@SeanFeldman @yvesgoeleven - build seems to be failing. Can you have a look? Is this one ready for review?

SeanFeldman:
Not yet. That's why we still keep the WIP on it and don't ping the docs
maintainers.

indualagarsamy:
@SeanFeldman - thanks. I'll look for the ping before I bother.

SeanFeldman:
- `MessagingFactoryLifeCycleManager` is creating a ring buffer of messaging factories per namespace (5 by default).  
- When a `MessageReceiverCreator` is created, we get a factory. `MessageReceiverLifeCycleManager` is asked to provide a `IMessageReceiver`, it will have by default 5 clients created that are using the same factory. 
- Down the road, when we need to have a `MessageReceiveNotifier`, we spawn up a single one, which uses a single `IMessageReceiver` created by `MessageReceiverLifeCycleManager`, with OnMessage API concurrency set to what we need.

So what we have is:

OnMessageCallbacks X concurrency -->MessageReceiver (single) --> MessageFactory (single)

Am I missing anything @yvesgoeleven

SeanFeldman:
Update: we should try having a single (MsgReceiver per MsgFactory) X concurrency level.

yvesgoeleven:
@SeanFeldman I'll try that tomorrow.

SeanFeldman:
Results I'm getting:

| Sample | Transaction mode | Factories | Clients | Batch Flash Interval | Prefetch Count | Total Concurrency| Rate|
|----------|----------|----------|----------|----------|----------|----------|----------|----------|
| Send     | `ReceiveOnly`    | 5            | 5           | 100 mSec| 200  (default) | default | 2600             |
| Receive    | `Atomic`    | 16            | 16           | 500 mSec (default) | 0/20 | 4 x 16 | 23/63             |
| Receive    | `ReceiveOnly`    | 32            | 32           | default | 128 | 128 x 32 | 258             |
| Receive & Send   | `Atomic`    | 16            | 16           | default | 0/20 | 4 x 16 | S-R23/S-R120       |
| Receive & Send   | `ReceiveOnly`    | 16            | 16           | default | default | 128  | S-R258          |

For `Atomic` prefetch of 0 didn't work for me well, but a small (20) prefetch did with ~x3-x4 performance increase.

SeanFeldman:
@yvesgoeleven updated the [table](https://github.com/Particular/docs.particular.net/pull/1949#issuecomment-251863310) with my results. We're much more aligned now. Need to agree on these values as defaults and roll that into RTM.

SeanFeldman:
For the record, prior to ASB RC6 results (rate) where: 2600, **23-63**, **258**, **23-120**, 250

yvesgoeleven:
My results

| Sample | Transaction mode | Factories | Clients | Batch Flash Interval | Prefetch Count | Total Concurrency| Rate|
|----------|----------|----------|----------|----------|----------|----------|----------|----------|
| Send     | `ReceiveOnly`    | 5            | 5           | 100 mSec| 200  (default) | default | 1900             |
| Receive    | `Atomic`    | 16            | 16           | 500 mSec (default) | **0/20** | 4 x 16 | **xxx/xxx**    |
| Receive    | `ReceiveOnly`    | 32            | 32           | default | 128 | 128 x 32 | xxxx             |
| Receive & Send   | `Atomic`    | 16            | 16           | default | 0 | 4 x 16 | S-R xxx       |
| Receive & Send   | `ReceiveOnly`    | 16            | 16           | default | default | 128  | S-R xxx          |

yvesgoeleven:
Results on azure VM

| Sample | Transaction mode | Factories | Clients | Batch Flash Interval | Prefetch Count | Total Concurrency| Rate|
|----------|----------|----------|----------|----------|----------|----------|----------|----------|
| Send     | `ReceiveOnly`    | 5            | 5           | 100 mSec| 200  (default) | default | xxxx             |
| Receive    | `Atomic`    | 16            | 16           | 500 mSec (default) | **0/20** | 4 x 16 | **xxx/xxx**    |
| Receive    | `ReceiveOnly`    | 32            | 32           | default | 128 | 128 x 32 | xxxx             |
| Receive & Send   | `Atomic`    | 16            | 16           | default | 0 | 4 x 16 | S-R xxx       |
| Receive & Send   | `ReceiveOnly`    | 16            | 16           | default | default | 128  | S-R xxx          |

nservicebus/azure-service-bus/performance-tuning_body_asb_[7,).partial.md
```diff
@@ -0,0 +1,61 @@
...
+### Receive only
+
+This mode has a lot less constraints and will work better with high numbers for concurrency and prefetch count, which should result in a pretty high throughput (likely over 1000/s even for remote connections)
+
+  * Concurrency: Values in the range of 128 per receive client seems to work well.
+  * PrefetchCount: Set this to 1x or 2x the per receiver concurrency, so 128 or 256.
+
+But, sending messages out at the same time does influence this mode more. If the average handler sends out only a single message then the above values still hold true. However if the handler sends out more messages, let's say 10 or 100 messages per receive, the concurrency and prefetchcount should be dialed down to allow more bandwidth to the senders.
+
+  * Concurrency: Values in the range of 16 per receive client seem to work well in this scenario.
```
ramonsmits: What is a receive client?

nservicebus/azure-service-bus/performance-tuning_body_asb_[7,).partial.md
```diff
@@ -0,0 +1,61 @@
...
+  * All send operations that occur in the scope of a single receive (inside a handler), must fit into the transaction. Therefore a maximum of 100 messages can be sent this way. The good news is that doing so has little additional implications on performance tuning as 1 send operation or 100 is actually the same thing except for the size of the operation.
+
+### Receive only
+
+This mode has a lot less constraints and will work better with high numbers for concurrency and prefetch count, which should result in a pretty high throughput (likely over 1000/s even for remote connections)
+
+  * Concurrency: Values in the range of 128 per receive client seems to work well.
+  * PrefetchCount: Set this to 1x or 2x the per receiver concurrency, so 128 or 256.
+
+But, sending messages out at the same time does influence this mode more. If the average handler sends out only a single message then the above values still hold true. However if the handler sends out more messages, let's say 10 or 100 messages per receive, the concurrency and prefetchcount should be dialed down to allow more bandwidth to the senders.
```
ramonsmits: > If the average handler sends out only a single message then the above values still hold true. However if the handler sends out more messages, let's say 10 or 100 messages per receive, the concurrency and prefetchcount should be dialed down to allow more bandwidth to the senders.

What does this mean? Why reduce the prefetching/concurrency because handlers are chatty? its data going out instead of in.

I can imagine that if these are very large messages that in the beginning ingress could be hindering exgress but once the prefect buffer is warmed up it should not really matter.

nservicebus/azure-service-bus/performance-tuning_body_asb_[7,).partial.md
```diff
@@ -0,0 +1,61 @@
+## Tuning send performance outside the scope of a handler
+
+Only settings at the ASB SDK level impact this scenario, more specifically
+
+  * `MessagingFactories().BatchFlushInterval()`: Controls the time that the ASB SDK batches client side requests before sending them out. A value of TimeSpan.Zero turns this feature off. Values between 100 and 500 milliseconds appear to work best.
+  * `MessagingFactories().NumberOfMessagingFactoriesPerNamespace()`: Each factory manages a tcp connection to a front end node in the broker, this connection has throughput limits. Opening more factories significantly improves send performance. Values between 2 and 8 seem to work best for sending.
+  * `NumberOfClientsPerEntity()`: Keep this value equal to the number of factories, this ensures there is 1 send client per factory.
+
+The way the send api is invoked matters a lot as well. Using `await` on each `Send()` operation will make the ASB SDK wait until the send operation completes and therefore makes client side batching useless. It's better to maintain a list of send tasks and await them all together instead.
```
SeanFeldman: > api

API

nservicebus/azure-service-bus/performance-tuning_body_asb_[7,).partial.md
```diff
@@ -0,0 +1,61 @@
+## Tuning send performance outside the scope of a handler
+
+Only settings at the ASB SDK level impact this scenario, more specifically
```
SeanFeldman: "The following ASB SDK settings impact out of handler context sending operations"?

nservicebus/azure-service-bus/performance-tuning_body_asb_[7,).partial.md
```diff
@@ -0,0 +1,61 @@
...
+
+  * `MessagingFactories().BatchFlushInterval()`: Controls the time that the ASB SDK batches client side requests before sending them out. A value of TimeSpan.Zero turns this feature off. Values between 100 and 500 milliseconds appear to work best.
```
SeanFeldman: > A value of TimeSpan.Zero turns this feature off. Values between 100 and 500 milliseconds appear to work best.

Optimal values are between 100 and 500 milliseconds. To turn this feature off, provide `TimeSpan.Zero`.

nservicebus/azure-service-bus/performance-tuning_body_asb_[7,).partial.md
```diff
@@ -0,0 +1,61 @@
...
+  * `MessagingFactories().NumberOfMessagingFactoriesPerNamespace()`: Each factory manages a tcp connection to a front end node in the broker, this connection has throughput limits. Opening more factories significantly improves send performance. Values between 2 and 8 seem to work best for sending.
```
SeanFeldman: > tcp 
TCP

SeanFeldman: > seem to work

Optimal values...

nservicebus/azure-service-bus/performance-tuning_body_asb_[7,).partial.md
```diff
@@ -0,0 +1,61 @@
...
+## Tuning receive performance
+
+Tuning receive performance depends on quite a few things that vary per scenario, but the following settings bring improvements for all scenarios.
+
+  * `MessagingFactories().NumberOfMessagingFactoriesPerNamespace()`: Each factory manages a tcp connection to a front end node in the broker, this connection has throughput limits. Opening more factories significantly improves receive performance. Values between 16 and 40 seem to work best for receiving.
+  * `NumberOfClientsPerEntity()`: Keep this value equal to the number of factories, this ensures there is 1 receive client per factory.
+
+Other settings that impact performance a lot, but their values differ quite a bit per scenario, are:
+
+  * `LimitMessageProcessingConcurrencyTo()`: This is the global concurrency limit across all receive entities. Ensure this value matches the per receiver desired concurrency for the scenario, multiplied by the number of receivers.
```
SeanFeldman: No mentioning of the concurrency we found and correlation with number of clients/factories? (4x16 and 128x32)

nservicebus/azure-service-bus/performance-tuning_body_asb_[7,).partial.md
```diff
@@ -0,0 +1,61 @@
...
+  * `MessageReceivers().PrefetchCount()`: Value is per receiver, in general the following rule applies: The lower the throughput rate of an individual receiver, the lower this value should be.
+  * `Transactions()`: The most significant performance impact comes from the transactional guarantees needed, `SendsAtomicWithReceive` will offer less performance then `ReceiveOnly`.
+
+### Sends Atomic With Receive
+
+In order to guarantee that sends occur as an atomic operation with the receive confirmation, the ASB SDK requires that both operations are enclosed in a `Serializable` transaction. This implies that:
+
+  * All atomic operations are executed in a 'Serial' fashion, aka 1 by 1, therefore it does not make sense to allow many concurrent operations per client, nor does it make much sense to prefetch a lot of messages.
```
SeanFeldman: > aka
i.e.

nservicebus/azure-service-bus/performance-tuning_body_asb_[7,).partial.md
SeanFeldman: `NumberOfClientsPerEntity()` value

Snippets/ASB/ASB_7/Performance.cs
```diff
@@ -0,0 +1,43 @@
...
+    async Task SendPerformance()
+    {
+        var endpointConfiguration = new EndpointConfiguration("whatever");
+        var endpointInstance = await Endpoint.Start(endpointConfiguration).ConfigureAwait(false);
+
+        #region asb-slow-send
+
+        for (var i = 0; i < 100; i++)
+        {
+            // by awaiting each individual send, no client side batching can take place
```
ramonsmits: This is only applicable when sending/publishing outside of an incoming message handler context. That is not clear in the sample.

samples/azure/performance-tuning-asb/ASB_7/FastSender/Program.cs
```diff
@@ -0,0 +1,115 @@
...
+                {
+                    var task = endpointInstance.Send(new SomeMessage());
+                        //.ContinueWith(t => Console.WriteLine("Batched a message..."));
+                    tasks.Add(task);
+                }
+
+                Console.WriteLine("Waiting for completion...");
+                // by awaiting the sends as one unit, this code allows the ASB SDK's client side batching to kick in and bundle sends
+                // this results in less latency overhead per individual sends and thus higher performance
+                await Task.WhenAll(tasks).ContinueWith(t =>
```
ramonsmits: Why use the ContinueWIth here as the code is using await?.

samples/azure/performance-tuning-asb/ASB_7/FastSender/Program.cs
```diff
@@ -0,0 +1,115 @@
...
+                {
+                    stopwatch.Stop();
+                    var elapsedSeconds = stopwatch.ElapsedTicks / (double)Stopwatch.Frequency;
```
ramonsmits: Why not:

var elapsedSeconds = stopwatch.Elapsed.TotalSeconds();

ramonsmits:
I would just not add the 'slow' sender. Probably not even add the fast sender except for all the initialization.

The `Task.WhenAll` approach is not only applicable to ASB but to ASB, ASQ and RabbitMQ. In the future potentially also SQL. I think it would be best to put that in a separate docs section.

SeanFeldman:
@ramonsmits agree that slow sender example doesn't add any value. Fast sender, let's keep it. A) it adds clarity B) it helps with seeding needed for receivers in the sample.

yvesgoeleven:
The slow samples are in there so that I can explain why they are slow, see the associated sample text

nservicebus/azure-service-bus/performance-tuning_body_asb_[7,).partial.md
yvesgoeleven: @ramonsmits @SeanFeldman should we repeat explanation of settings here? We already explain the setting in the configuration api, but should we repeat it here? Or maybe link from here to the explanation?

nservicebus/azure-service-bus/performance-tuning_body_asb_[7,).partial.md
yvesgoeleven: The machine/connection has a limit to it's bandwidth, the more data goes out, the less can come in, by reducing the amount of concurrent operations you reduce the amount of incoming data. 

I'll try to reword this

nservicebus/azure-service-bus/performance-tuning_body_asb_[7,).partial.md
yvesgoeleven: @SeanFeldman what do you mean exactly? The ratio is really different per transaction mode, can't generalize it

Snippets/ASB/ASB_7/Performance.cs
yvesgoeleven: @ramonsmits In the samples I have written all sends this way, even the ones inside handlers, because it is also applicable when  sendoptions ask for immediate dispatch. Only for default dispatch the await doesn't matter.

yvesgoeleven: I'll add a note about this

yvesgoeleven:
@SeanFeldman @ramonsmits feedback included

samples/azure/performance-tuning-asb/sample.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Sender
+
+`Sender` sends `SomeMessage` to `Receiver`.
+
+snippet: SomeMessage
+
+The sender sends a large amount of messages to the receiver and times how long it takes to do so. The resulting time will be divided by the number of messages sent in order to compute the average throughput. 
+
+There are two variations of `Sender` to show different (good and bad) ways of sending messages, outside the scope of a receive operation, and explains what the impact is on throughput.
```
weralabaj: > outside the scope of a receive operation

Not sure what that means? Is that about transaction mode? If so then say it explicitly what mode it uses.

weralabaj: > to show different (good and bad) ways of sending messages, outside the scope of a receive operation, and explains what the impact is on throughput.

to show different ways of sending messages and illustrate what impact they have on performance.

samples/azure/performance-tuning-asb/sample.md
```diff
@@ -0,0 +1,133 @@
...
+
+## Receiver
+
+`Receiver` receives `SomeMessage` and computes every second how many it was able to receive in that timeframe
+
+Again there are different permutations of settings that have impact on the combines receive throughput.
```
weralabaj: > that have impact on the combines receive throughput

combines? You mean average? Total?

samples/azure/performance-tuning-asb/sample.md
```diff
@@ -0,0 +1,133 @@
...
+
+ * Run the sender standalone to test send performance, but also to fill up the receive queue.
+ * Run the receiver standalone to test receive performance, if the queue empties, just run the sender again.
+ * Run the sender receiver standalone
+ 
+## Variations
+ 
+### Slow Sender
+
+In the slow sender sample, everything is done wrong:
```
weralabaj: > everything is done wrong

I'd skip that, the name speaks for itself.

samples/azure/performance-tuning-asb/sample.md
```diff
@@ -0,0 +1,133 @@
...
+ * Only 1 factory is defined, this means that only 1 tcp connection is maintained with the broker.
+ * Client side batching of the ASB SDK is turned off, so the SDK will send message by message.
+
+Snippet: slow-send-config
+
+And each send is awaited individually, another cause that would prevent batching (even if batching client side batching were on).
```
weralabaj: > And each send is awaited individually, another cause that would prevent batching (even if batching client side batching were on)

In this scenario each send is awaited individually, which prevents batching (even if batching is enabled on the client side).

samples/azure/performance-tuning-asb/sample.md
```diff
@@ -0,0 +1,133 @@
...
+
+Snippet: slow-send  
+
+### Fast Sender
+
+The fast sender is configured much better:
```
weralabaj: >  is configured much better

I'd skip that fragment and just said "In the fast sender scenario:"

samples/azure/performance-tuning-asb/sample.md
```diff
@@ -0,0 +1,133 @@
...
+
+Snippet: fast-send-config
+
+Individual sends are not awaited, instead the entire batch is awaited. Code execution will continue when all batches have been sent.
+
+Snippet: fast-send
+
+### Slow Atomic Receiver
+
+In the slow atomic receiver sample, everything is configured wrong:
```
weralabaj: as above: "In the slow atomic receiver scenario/sample:"

samples/azure/performance-tuning-asb/sample.md
```diff
@@ -0,0 +1,133 @@
...
+ * Only 1 factory is defined, this means that only 1 tcp connection is maintained with the broker.
+ * Prefetching is turned off, meaning that the receiver will fetch message by message
```
weralabaj: > will fetch message by message

one message at a time

samples/azure/performance-tuning-asb/sample.md
```diff
@@ -0,0 +1,133 @@
...
+ * No concurrent operations allowed, messages are processed one by one.
+
+Snippet: slow-atomic-receiver-config
+
+### Fast Atomic Receiver
+
+The fast atomic receiver is configured much better
```
weralabaj: as above

samples/azure/performance-tuning-asb/sample.md
```diff
@@ -0,0 +1,133 @@
...
+ * Multiple connections are established with the broker.
+ * Each connection is matched with exactly 1 receive client.
+ * Ensures that at least 16 connections are established when using partitioned queues, as each partitioned queue consists of 16 partitions. Depending on available bandwidth configure more until the network saturates.
+ * TransactionMode AtomicSendsWithReceive will execute complete operations in a serializable transaction, which means one by one. Therefore it is not beneficial to allow high per receiver concurrency. A value of 2 allows one receive to start while another is allowed to complete. The more time execution takes the higher this value can be, but never use really high values here.
+ * PrefetchCount is set to a relatively low number, as per receiver concurrency and throughput is low it makes no sense to load many messages at once.
```
weralabaj: > PrefetchCount is set to a relatively low number, as per receiver concurrency and throughput is low it makes no sense to load many messages at once.

PrefetchCount is set to a relatively low number, since receiver concurrency and throughput are also low.

samples/azure/performance-tuning-asb/sample.md
```diff
@@ -0,0 +1,133 @@
...
+### Fast Atomic Sender Receiver
+
+The fast atomic sender receiver scenario is similar to the fast atomic receiver scenario, except:
+ * It established twice the amount of connections, so that the receive clients and send clients can leverage different connections.
+
+Snippet: fast-atomic-send-receive-config
+
+### Slow Non Atomic Receiver
+
+In the slow none atomic receiver sample, everything is configured wrong:
```
weralabaj: as above

samples/azure/performance-tuning-asb/sample.md
```diff
@@ -0,0 +1,133 @@
...
+
+Snippet: slow-non-atomic-receiver-config
+
+### Fast Non Atomic Receiver
+
+The fast non atomic receiver is configured much better
+ * Multiple connections are established with the broker.
+ * Each connection is matched with exactly 1 receive client.
+ * Ensures that at least 16 connections are established when using partitioned queues, as each partitioned queue consists of 16 partitions. Depending on available bandwidth configure more until the network saturates.
+ * TransactionMode ReceiveOnly can batch message completions, allowing receive operations to work independently. Therefore it is very beneficial to allow high per receiver concurrency. For example 128 operations per receiver if the operation can handle this kind of concurrency.
```
weralabaj: > For example 128 operations per receiver if the operation can handle this kind of concurrency.

how can I find out? not sure what do you mean by "operation can handle"?

nservicebus/azure-service-bus/performance-tuning_body_asb_[7,).partial.md
```diff
@@ -0,0 +1,63 @@
...
+
+Fast:
+
+Snippet: asb-fast-send
+
+## Tuning receive performance
+
+Tuning receive performance depends on quite a few things that vary per scenario, but the following settings bring improvements for all scenarios.
+
+  * `MessagingFactories().NumberOfMessagingFactoriesPerNamespace()`: Each factory manages a TCP connection to a front end node in the broker, this connection has throughput limits. Opening more factories significantly improves receive performance. Optimal value ares between 16 and 40 for receiving.
```
ramonsmits: I've seen improvements for as many as the network allows. I saturated my uplink during receives when using 64 factories. 

Probably better written as:

The number of factories depends on how much bandwidth Azure Service Bus is allowed to consume and the size of your messages.  Monitor your network received and sent bytes per second. If you see that your network isn't congested you can choose to increase the number of factories and clients.

nservicebus/azure-service-bus/performance-tuning_body_asb_[7,).partial.md
```diff
@@ -0,0 +1,63 @@
...
+  * `MessageReceivers().PrefetchCount()`: Value is per receiver, in general the following rule applies: The lower the throughput rate of an individual receiver, the lower this value should be.
+  * `Transactions()`: The most significant performance impact comes from the transactional guarantees needed, `SendsAtomicWithReceive` will offer less performance then `ReceiveOnly`.
+
+### Sends Atomic With Receive
+
+In order to guarantee that sends occur as an atomic operation with the receive confirmation, the ASB SDK requires that both operations are enclosed in a `Serializable` transaction. This implies that:
+
+  * All atomic operations are executed in a 'Serial' fashion, i.e. 1 by 1, therefore it does not make sense to allow many concurrent operations per client, nor does it make much sense to prefetch a lot of messages.
+	  - Concurrency: Optimal values are 2 or 4 per client
+	  - PrefetchCount: Optimal value is 20
```
ramonsmits: My optimal prefetch size was 18.

nservicebus/azure-service-bus/performance-tuning_body_asb_[7,).partial.md
```diff
@@ -0,0 +1,63 @@
...
+  * Setting up the transactional connection takes up quite some time, this causes the first operational minutes to perform very 'bumpy', but after a while the throughput becomes consistent.
```
ramonsmits: What about:

> this causes the first operational minutes to perform very 'bumpy'

If you are in an environment that has a *large* number of messages per second (exceeding 5,000msg/s) the endpoint throughput will not be optimal just after starting the endpoint. Throughput slowly ramps up and after a few minutes the throughput should reach its maximum based on the configured settings and available network bandwidth and should remain stable.

Snippets/ASB/ASB_7/Performance.cs
ramonsmits: Yes, please be sure to mention this as customer might think that this must also be done withing a incomiming message context.

samples/azure/performance-tuning-asb/ASB_7/FastAtomicSenderReceiver/Program.cs
```diff
@@ -0,0 +1,90 @@
...
+        transportConfiguration.NumberOfClientsPerEntity(numberOfReceivers);
+
+        #endregion
+
+
+        var endpointInstance = await Endpoint.Start(endpointConfiguration)
+            .ConfigureAwait(false);
+        try
+        {
+            Console.WriteLine("Receiver is ready to receive messages");
```
ramonsmits: await Console.Out.WriteLineAsync

samples/azure/performance-tuning-asb/ASB_7/FastNonAtomicReceiver/Program.cs
```diff
@@ -0,0 +1,73 @@
...
+        endpointConfiguration.SendFailedMessagesTo("error");
+        endpointConfiguration.UseSerialization<JsonSerializer>();
+        endpointConfiguration.EnableInstallers();
+        endpointConfiguration.UsePersistence<InMemoryPersistence>();
+
+        #region fast-non-atomic-receiver-config
+
+        transportConfiguration.Transactions(TransportTransactionMode.ReceiveOnly);
+
+        transportConfiguration.Queues().EnablePartitioning(true);
```
ramonsmits: I'm missing:

```
transport.Topics().EnablePartitioning(true);
```

samples/azure/performance-tuning-asb/sample.md
```diff
@@ -0,0 +1,133 @@
...
+`Receiver` receives `SomeMessage` and computes every second how many it was able to receive in that timeframe
+
+Again there are different permutations of settings that have impact on the combines receive throughput.
+
+## SenderReceiver
+
+`SenderReceiver` is a variation on `Receiver` that does not only receive messages, but it also sends messages to a destination queue.
+
+## Running the sample
+
```
ramonsmits: Run without debugging, press CTRL+F5

nservicebus/azure-service-bus/performance-tuning_body_asb_[,7).partial.md
```diff
@@ -0,0 +1,14 @@
+### CPU vs IO bound processing
+
+The following settings are used in order to tune performance in Azure Service Bus version 6 and below:
+
+ - `MaximumConcurrencyLevel`
+ - `BatchSize`
+ - `LockDuration`
+ - `MaxDeliveryCount`
+
+In scenarios where handlers are CPU intense and have very little IO, it is advised to lower the number of threads to one and increase the `BatchSize`. `LockDuration` and `MaxDeliveryCount` might require an adjustment to match the batch size, taking into account the number of messages that end up in the dead letter queue.
```
ramonsmits: > it is advised to lower the number of threads

We don't have threads, only maximum concurrency. IO bound benefits from allowing more concurrency at the trad e of or increasing latency. Increasing concurrency means it's likely that you also require to increase the lock duration.

Snippets/ASB/ASB_7/Performance.cs
```diff
@@ -0,0 +1,43 @@
...
+
+        #region asb-fast-send
+        var tasks = new List<Task>();
+        for (var i = 0; i < 10000; i++)
+        {
+            var task = endpointInstance.Send(new SomeMessage());
+            tasks.Add(task);
+        }
+
+        // by awaiting the sends as one unit, this code allows the ASB SDK's client side batching to kick in and bundle sends
```
SimonCropp: should this be in the .md file?

nservicebus/azure-service-bus/performance-tuning.md
```diff
@@ -1,23 +1,11 @@
 ---
 title: Performance Tuning
+component: ASB
 tags:
  - Azure
  - Cloud
 reviewed: 2016-09-21
```
SimonCropp: should review date be bumped?

Snippets/ASB/ASB_7/Performance.cs
```diff
@@ -0,0 +1,43 @@
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
```
SimonCropp: redundant usings

Snippets/ASB/ASB_7/Performance.cs
```diff
@@ -0,0 +1,43 @@
...
+        var endpointConfiguration = new EndpointConfiguration("whatever");
+        var endpointInstance = await Endpoint.Start(endpointConfiguration).ConfigureAwait(false);
+
+        #region asb-slow-send
+
+        for (var i = 0; i < 100; i++)
+        {
+            // by awaiting each individual send, no client side batching can take place
+            // latency is incurred for each send and thus lowest performance possible
+            await endpointInstance.Send(new SomeMessage());
```
SimonCropp: missing ConfigureAwait

Snippets/ASB/ASB_7/Performance.cs
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        #endregion
+
+        #region asb-fast-send
+        var tasks = new List<Task>();
+        for (var i = 0; i < 10000; i++)
+        {
+            var task = endpointInstance.Send(new SomeMessage());
```
SimonCropp: do we need a ConfigureAwait here?

nservicebus/azure-service-bus/performance-tuning_body_asb_[7,).partial.md
```diff
@@ -0,0 +1,63 @@
...
+  * All send operations that occur in the scope of a single receive (inside a handler), must fit into the transaction. Therefore a maximum of 100 messages can be sent this way. The good news is that doing so has little additional implications on performance tuning as 1 send operation or 100 is actually the same thing except for the size of the operation.
+
+NOTE: Send operations inside a handler are only executed when the handler operation returns, unless the send options request for immediate dispatch. Therefore using `await` on a Send inside a handler will probably not seem to impact performance negatively, this is because no send is actually occurring, yet it is still advised to return the task instead of awaiting.
+
+### Receive only
+
+This mode has a lot less constraints and will work better with high numbers for concurrency and prefetch count, which should result in a pretty high throughput (likely over 1000/s even for remote connections)
```
SimonCropp: "a lot less" => "fewer

nservicebus/azure-service-bus/performance-tuning_body_asb_[7,).partial.md
```diff
@@ -0,0 +1,63 @@
...
+	  - Concurrency: Optimal values are 2 or 4 per client
+	  - PrefetchCount: Optimal value is 20
+  * Setting up the transactional connection takes up quite some time, this causes the first operational minutes to perform very 'bumpy', but after a while the throughput becomes consistent.
+  * All send operations that occur in the scope of a single receive (inside a handler), must fit into the transaction. Therefore a maximum of 100 messages can be sent this way. The good news is that doing so has little additional implications on performance tuning as 1 send operation or 100 is actually the same thing except for the size of the operation.
+
+NOTE: Send operations inside a handler are only executed when the handler operation returns, unless the send options request for immediate dispatch. Therefore using `await` on a Send inside a handler will probably not seem to impact performance negatively, this is because no send is actually occurring, yet it is still advised to return the task instead of awaiting.
```
SimonCropp: > only executed when the handler operation returns

isnt this not strictly correct? @Particular/nservicebus-maintainers

nservicebus/azure-service-bus/performance-tuning_body_asb_[7,).partial.md
SimonCropp: "should "? are we not certain?

nservicebus/azure-service-bus/performance-tuning_body_asb_[7,).partial.md
```diff
@@ -0,0 +1,63 @@
...
+  * Concurrency: Optimal values in the range of 16 per client in this scenario.
+  * PrefetchCount: Set this to 1x or 2x the per receiver concurrency, so 16 or 32.
+
+## Performance Tuning Samples
+
+Check out the [performance tuning samples](/samples/azure/performance-tuning-asb/) to experiment with each of the settings in each scenario mentioned.
+
+
+
+
```
SimonCropp: kill the trailing newlines

samples/azure/performance-tuning-asb/ASB_7/FastAtomicReceiver/Program.cs
```diff
@@ -0,0 +1,72 @@
+﻿using System;
+using System.Threading.Tasks;
+using NServiceBus;
+using NServiceBus.Logging;
+
+class Program
+{
+    private static ILog _logger = LogManager.GetLogger<Program>();
```
SimonCropp: no `private`.

`_logger ` => `logger `

samples/azure/performance-tuning-asb/ASB_7/FastAtomicReceiver/Program.cs
```diff
@@ -0,0 +1,72 @@
...
+    static void Main()
+    {
+        MainAsync().GetAwaiter().GetResult();
+    }
+
+    static async Task MainAsync()
+    {
+        Console.Title = "Samples.ASB.Performance.FastAtomicReceiver";
+
+        ReceiveCounter.Subscribe(i => _logger.Warn("Process " + i + " messages per second"));
```
SimonCropp: use string interpolation

samples/azure/performance-tuning-asb/ASB_7/FastAtomicReceiver/Program.cs
```diff
@@ -0,0 +1,72 @@
...
+        endpointConfiguration.EnableInstallers();
+        endpointConfiguration.UsePersistence<InMemoryPersistence>();
+
+        #region fast-atomic-receiver-config
+
+        transportConfiguration.Transactions(TransportTransactionMode.SendsAtomicWithReceive);
+
+        transportConfiguration.Queues().EnablePartitioning(true);
+
+        var perReceiverConcurrency = 2; //values 2 and 4 work best, as tx is serializable it makes no sense to allow many concurrent tasks
```
SimonCropp: put the comments above the variable so it wraps better when rendered

samples/azure/performance-tuning-asb/ASB_7/FastAtomicReceiver/Program.cs
```diff
@@ -0,0 +1,72 @@
...
+        var numberOfReceivers = 16; // increase number of receivers as much as bandwidth allows
+        var globalConcurrency = numberOfReceivers * perReceiverConcurrency;
+
+        endpointConfiguration.LimitMessageProcessingConcurrencyTo(globalConcurrency);
+        transportConfiguration.MessageReceivers().PrefetchCount(20);
+
+        transportConfiguration.MessagingFactories().NumberOfMessagingFactoriesPerNamespace(numberOfReceivers);
```
SimonCropp: add a local variable for `transportConfiguration.MessagingFactories()` and ` transportConfiguration.MessageReceivers()` to improve the code wrapping when rendered

samples/azure/performance-tuning-asb/ASB_7/FastAtomicReceiver/Program.cs
```diff
@@ -0,0 +1,72 @@
...
+        transportConfiguration.NumberOfClientsPerEntity(numberOfReceivers);
+
+        #endregion
+
+
+        var endpointInstance = await Endpoint.Start(endpointConfiguration)
+            .ConfigureAwait(false);
+        try
```
SimonCropp: if we r only dong console writes i think the `try` is redundant

SimonCropp:
remove the prerelease.txt

samples/azure/performance-tuning-asb/ASB_7/FastAtomicSenderReceiver/FastAtomicSenderReceiver.csproj
```diff
@@ -0,0 +1,102 @@
...
+    <Reference Include="System.Xml.Linq" />
+    <Reference Include="System.Data.DataSetExtensions" />
+    <Reference Include="Microsoft.CSharp" />
+    <Reference Include="System.Data" />
+    <Reference Include="System.Xml" />
+    <Reference Include="WindowsBase" />
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Include="Program.cs" />
+    <Compile Include="Properties\AssemblyInfo.cs" />
```
SimonCropp: remove this file

samples/azure/performance-tuning-asb/ASB_7/FastNonAtomicSenderReceiver/FastNonAtomicSenderReceiver.csproj
```diff
@@ -0,0 +1,103 @@
...
+    <Reference Include="System.Xml.Linq" />
+    <Reference Include="System.Data.DataSetExtensions" />
+    <Reference Include="Microsoft.CSharp" />
+    <Reference Include="System.Data" />
+    <Reference Include="System.Xml" />
+    <Reference Include="WindowsBase" />
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Include="Program.cs" />
+    <Compile Include="Properties\AssemblyInfo.cs" />
```
SimonCropp: remove this file

samples/azure/performance-tuning-asb/ASB_7/FastNonAtomicSenderReceiver/App.config
```diff
@@ -0,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<configuration>
+    <startup> 
+        <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.5.2" />
```
SimonCropp: remove this file

nservicebus/azure-service-bus/performance-tuning_body_asb_[,7).partial.md
SeanFeldman: @ramonsmits are you referring to v6 or v7 of the transport?

nservicebus/azure-service-bus/performance-tuning_body_asb_[7,).partial.md
danielmarbach: All outgoing operations are batched by default unless explicitly requested otherwise. Should we just link to batched dispatch here as well as to the async handler documentation? Therefore remove the note?

Snippets/ASB/ASB_7/Performance.cs
danielmarbach: @SeanFeldman is correct

nservicebus/azure-service-bus/performance-tuning_body_asb_[,7).partial.md
```diff
@@ -0,0 +1,14 @@
...
+
+In scenario where handlers are IO intense, it is advised to set the number of threads  to 12 threads per logical core using `MaximumConcurrencyLevel` setting, and set the `BatchSize` to a number of messages that takes to process. Take into account the expected (or measured) processing time and IO latency of a single message. Start with a small `BatchSize` and through adjustment and measurement gradually increase it, while adjusting accordingly `LockDuration` and `MaxDeliveryCount`.
```
yvesgoeleven: In older versions of the CLR it would default to 25 threads per logical core, 12 is about half of that (12 for receiving, 12 for sending, 1 for GC) is the logic. In recent versions of the CLR this logic doesn't hold though as it now dynamically adjusts

nservicebus/azure-service-bus/performance-tuning_body_asb_[7,).partial.md
```diff
@@ -0,0 +1,63 @@
...
+Snippet: asb-fast-send
+
+## Tuning receive performance
+
+Tuning receive performance depends on quite a few things that vary per scenario, but the following settings bring improvements for all scenarios.
+
+  * `MessagingFactories().NumberOfMessagingFactoriesPerNamespace()`: Each factory manages a TCP connection to a front end node in the broker, this connection has throughput limits. Opening more factories significantly improves receive performance. Optimal value ares between 16 and 40 for receiving.
+  * `NumberOfClientsPerEntity()`: Keep this value equal to the number of factories, this ensures there is 1 internal receive client per factory.
+
+Other settings that impact performance a lot, but their values differ quite a bit per scenario, are:
```
yvesgoeleven: I'll go for 'Changing the following settings will also improve application performance, but their values depend on the nature of the application'

nservicebus/azure-service-bus/performance-tuning_body_asb_[7,).partial.md
yvesgoeleven: The best I can come up with is  that what the official doco says '20 times the maximum processing rates of all receivers of the factory.' Where we have only 1 receiver per factory and execution is mostly serial.

yvesgoeleven: the offical doco seems to suggest that using more receivers per factory works best, but this does not seem to be the case in our testing for this mode. and the prefetch value is also very exagerated it resulted in message lock lost exceptions for me. Now I don't know what mode was used by ASB team to come up with these values, they don't mention that.

yvesgoeleven: The concurrency can be explained though, 1 receive while allowing 1 complete, so 2 is a great place to start.

yvesgoeleven: I'll do my best to word it, but it is a bit of a black art.

nservicebus/azure-service-bus/performance-tuning_body_asb_[7,).partial.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Sends Atomic With Receive
+
+In order to guarantee that sends occur as an atomic operation with the receive confirmation, the ASB SDK requires that both operations are enclosed in a `Serializable` transaction. This implies that:
+
+  * All atomic operations are executed in a 'Serial' fashion, i.e. 1 by 1, therefore it does not make sense to allow many concurrent operations per client, nor does it make much sense to prefetch a lot of messages.
+	  - Concurrency: Optimal values are 2 or 4 per client
+	  - PrefetchCount: Optimal value is 20
+  * Setting up the transactional connection takes up quite some time, this causes the first operational minutes to perform very 'bumpy', but after a while the throughput becomes consistent.
+  * All send operations that occur in the scope of a single receive (inside a handler), must fit into the transaction. Therefore a maximum of 100 messages can be sent this way. The good news is that doing so has little additional implications on performance tuning as 1 send operation or 100 is actually the same thing except for the size of the operation.
```
yvesgoeleven: ASB SDK enforces this, if you try 101 it throws on you, but only if inside a scope

nservicebus/azure-service-bus/performance-tuning_body_asb_[7,).partial.md
yvesgoeleven: it depends on the sendoptions. if you specify to send immediatly core will do just that and call transport, in that case the await would kill perf. If you choose default, then core will just add the operation to a list for later execution, in that case the await would not have a visible effect. To not trip you up later, it's advised to not use await but instead return tasks

yvesgoeleven: I'll change it to: When send operations occur inside a handler the default behavior is to add the operation to a batch, which will be executed after the handler function returns. Unless the send options request for immediate dispatch, then execution is immediate. Using an `await` on the Send operation therefore has a negative impact when used on an immediate dispatch, but not on the default. It is advised to return the task instead of awaiting the task to never be impacted by this difference in behavior.

nservicebus/azure-service-bus/performance-tuning_body_asb_[7,).partial.md
```diff
@@ -0,0 +1,63 @@
...
+NOTE: Send operations inside a handler are only executed when the handler operation returns, unless the send options request for immediate dispatch. Therefore using `await` on a Send inside a handler will probably not seem to impact performance negatively, this is because no send is actually occurring, yet it is still advised to return the task instead of awaiting.
+
+### Receive only
+
+This mode has a lot less constraints and will work better with high numbers for concurrency and prefetch count, which should result in a pretty high throughput (likely over 1000/s even for remote connections)
+
+  * Concurrency: Optimal values are in the range of 128 per client.
+  * PrefetchCount: Set this to 1x or 2x the per receiver concurrency, so 128 or 256.
+
+But, sending messages out at the same time does influence this mode more. If the average handler sends out an equal amount of data then it receives, then the above values still hold true. However if the handler sends out more messages (or larger messages), let's say 10x or 100x total data size per receive, then concurrency and prefetchcount should be dialed down to reduce incoming bandwidth and allow more bandwidth to the senders.
```
yvesgoeleven: it's not the number or individual size, it's the compound size. Receive 1msg of 100Kb and send 10 of 10Kb is the same as send of 1msg 100Kb

nservicebus/azure-service-bus/performance-tuning_body_asb_[7,).partial.md
```diff
@@ -0,0 +1,63 @@
...
+
+  * Concurrency: Optimal values in the range of 16 per client in this scenario.
+  * PrefetchCount: Set this to 1x or 2x the per receiver concurrency, so 16 or 32.
+
```
yvesgoeleven: Added the explanation above. 16 connections because of the 16 partitions in an entity, multiples of that for concurrency and prefetch.

samples/azure/performance-tuning-asb/sample.md
yvesgoeleven: No, it are various settings, number of factories, number of clients, batchflushinterval, ... as well as using await vs just combining the task objects.

yvesgoeleven: the explanation is also limited to 'outside a handler scope', different rules apply inside a handler

samples/azure/performance-tuning-asb/sample.md
yvesgoeleven: I'll remove combines

nservicebus/azure-service-bus/performance-tuning_body_asb_[7,).partial.md
ramonsmits: Should that really be part of the ASB specific tuning? I think this applies to all transports so should be moved to generic document about such tweaks.

samples/azure/performance-tuning-asb/sample.md
yvesgoeleven: That really depends on the code inside the handler, a db connection may not like 128 concurrent operations, while an in memory operation on a concurrentdictionary may

samples/azure/performance-tuning-asb/ASB_7/FastAtomicReceiver/Program.cs
```diff
@@ -0,0 +1,72 @@
...
+        }
+        transportConfiguration.ConnectionString(connectionString);
+        var topology = transportConfiguration.UseTopology<ForwardingTopology>();
+
+        endpointConfiguration.SendFailedMessagesTo("error");
+        endpointConfiguration.UseSerialization<JsonSerializer>();
+        endpointConfiguration.EnableInstallers();
+        endpointConfiguration.UsePersistence<InMemoryPersistence>();
+
+        #region fast-atomic-receiver-config
```
yvesgoeleven: I'll put it in text, otherwise need to adjust all the visual studio project names, assembly names etc...

samples/azure/performance-tuning-asb/ASB_7/FastNonAtomicReceiver/Program.cs
yvesgoeleven: I can add it, but as the samples are not (yet?) covering publishes it doesn't matter

nservicebus/azure-service-bus/performance-tuning_body_asb_[,7).partial.md
yvesgoeleven: In V6 it definitly were threads

nservicebus/azure-service-bus/performance-tuning_body_asb_[,7).partial.md
```diff
@@ -0,0 +1,14 @@
+### CPU vs IO bound processing
+
```
yvesgoeleven: TBH you can push it, to around 400/s. But you need to do the sends in parallel yourself

nservicebus/azure-service-bus/performance-tuning.md
yvesgoeleven: should this reflect when I started, or when I ended :trollface: (just kidding, I'll update it)

nservicebus/azure-service-bus/performance-tuning_body_asb_[7,).partial.md
yvesgoeleven: Will link, but the note should stay imo as it matters a lot for asb

samples/azure/performance-tuning-asb/ASB_7/FastAtomicSenderReceiver/SomeMessageHandler.cs
```diff
@@ -0,0 +1,20 @@
...
+    // maximum 100 for atomic sends
+    public const int NumberOfMessagesToSend = 1;
+
+    public async Task Handle(SomeMessage message, IMessageHandlerContext context)
+    {
+        Program.ReceiveCounter.OnNext(message);
+
+        for (var i = 0; i < NumberOfMessagesToSend; i++)
+        {
+            await context.Send(new SomeMessage());
```
yvesgoeleven: Should be list of tasks

yvesgoeleven:
A very big thank you for the thorough review @weralabaj @SimonCropp @danielmarbach @SeanFeldman  @ramonsmits ! All feedback is incorporated

Issue #1865, opened 2016-09-06
ASQ - Multiple storage accounts mixes 2 concerns


SeanFeldman:
[Multiple storage accounts](http://docs.particular.net/nservicebus/azure-storage-queues/multi-storageaccount-support) is mixing two topics

1. Azure Storage Scalability and Performance
1. Using aliases for connection strings

Scaling out was possible with previous versions of ASQ using "targeted" entities (entity with a specified connection string. F.e. "queue@connecitonString"). That is the focus of the document.

Aliases are new to ASQ and are only applicable to v7 of the transport.

## Suggestion

ASQ documentation to be slightly re-structure in regards to the above.

1. Rename "Multiple storage accounts" to "Cross account routing"
1. Move "Using aliases for connection strings" to a separate document called "Securing Connection Strings To Storage Accounts"

cc @Particular/azure-storage-queues-maintainers

Scooletz:
I don't agree. I can see the reason behind having a separate section in the `Multiple storage accounts`, which is the usage of aliases changes when you're in a `Scale Out` scenario. Effectively, I prefer to read a single `Scale Out` recipe that shows how to leverage multiple accounts rather than read this page and go to another one for more information. The section in `Multiple storage accounts` is written specifically for this scenario. The default configuration page covers a standard usage of this feature.

If we moved this section though, I'd preserve its name as it is "Using aliases for connection strings". We agreed that bringing the `security` notion can be seen as being not secure by default or bringing complexity to other elements of the platform (ServiceControl).

Thoughts? @Particular/azure-storage-queues-maintainers

SeanFeldman:
Preserving the old name for a new document is absolutely fine. I only suggested to change the name, not mandated that :wink: 

As for the document as it is right now - I still think connection strings aliases is **feature**. And as such, should have a dedicated document. Current document can reference standalone "Using aliases for connection strings" w/o a problem.

Additionally, currently the document makes it hard to understand what is available in what version. E.g: aliases are strictly v7+, where cross account routing is 7 _and_ below.

Scooletz:
Let's sync.

SeanFeldman:
Synced. Summary: use partials for sure and review if Aliases should reside in a separate document. 
/cc @Scooletz

weralabaj:
@SeanFeldman @Scooletz What partials have to do with all of this? :)

[...]

PR #1926, opened 2016-09-20
ASQ Multi Storage Accounts partialized

Scooletz:
Connects to #1865

Multi storage accounts gets partialized in regards to aliases. Based on the following https://github.com/Particular/docs.particular.net/issues/1865#issuecomment-248010424

PR #1951, opened 2016-09-26
[WIP] ASQ aliases moved to a separate page

Scooletz:
Connects to #1865

This PR extracts aliases for multiple accounts to a separate `aliases.include.md`. The include is used in the scale out page as well as in the newly created separate page for aliases. 

My personal doubt: this menu item unfortunately is visible for all versions of NSB as there's no menu versioning. I can't provide a general article as this page describes a feature that has been introduced in the ASQ 7.

### DIFF
--- a/menu/menu.yaml
+++ b/menu/menu.yaml
@@ -415,6 +415,8 @@
       Title: Avoiding Transactions
     - Url: nservicebus/azure-storage-queues/configuration
       Title: Configuration
+    - Url: nservicebus/azure-storage-queues/aliases
+      Title: Aliases for connection strings
     - Url: nservicebus/azure-storage-queues/performance-tuning
       Title: Performance Tuning
     - Url: nservicebus/azure-storage-queues/multi-storageaccount-support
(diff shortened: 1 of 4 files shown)


Scooletz:
Could you take a look @SeanFeldman ? 
I'm raising my doubt in this PR description `This unfortunately is visible for all versions of NSB as there's no menu versioning.`

weralabaj:
ping @SimonCropp thoughts on this approach?

SeanFeldman:
This is not the only page that is targeting vNext and will show up on the menu. And at the same time, vNext will become shortly vCurrent.

@weralabaj this: https://github.com/Particular/docs.particular.net/pull/1926#issuecomment-248976184

weralabaj:
But in Google group you can always link to paragraph, and we do that quite often in fact.

@SimonCropp thoughts?

SimonCropp:
i dont mind splitting a page out if the topic is deep/complex enough to justify its own page

but i dont like including the same large body of the include in multiple locations. why not link to the new page?

Scooletz:
@SimonCropp Do I understand you correctly:
1. move the body to the new page which is included in the menu
1. put a link in the original page navigating to the newly created
1. remove the include
?

SimonCropp:
@Scooletz yes

SimonCropp:
@Scooletz note that i said above though

> i dont mind splitting a page out if the topic is deep/complex enough to justify its own page

i prefer to only split out pages when is absolutely clear that is it necessary.

Scooletz:
@SeanFeldman could you chime in?

SeanFeldman:
I think we're discussing this for way too long. My point on this has not changed, if customers don't know what to do and can't [easily locate the feature at the doco site](https://groups.google.com/forum/#!topic/particularsoftware/ANBSuy69oLw) then I'd like to see it listed out just like we've [done for ASB](https://docs.particular.net/nservicebus/azure-service-bus/securing-connection-strings). Saying that, if people disagree, well, this is what democracy is for :wink:

Scooletz:
I'm closing this without merging as no consensus has been reached.

PR #1956, opened 2016-09-27
WIP: MSMQ TTBR using immediate dispatch

ramonsmits:


ramonsmits:
@Particular/docs-maintainers I want to edit

https://docs.particular.net/nservicebus/messaging/send-a-message?#dispatching-a-message-immediately

This is weird, for V6 it shows the TransactionScope Suppress first. I think it's better to only show the SendOptions/PublishOptions approach. Maybe mention that similar behavior could be achieve using transaction scope suppress on SQL and MSMQ transport but without an example.

indualagarsamy:
@ramonsmits - is this good for review? or is it still WIP?

nservicebus/messaging/discard-old-messages.md
```diff
@@ -40,8 +40,9 @@ TimeToBeReceived relies on underlying functionality in the transport infrastruct
 
 ### MSMQ transport
 
-MSMQ continuously checks the TimeToBeReceived attribute of all queued messages. As soon as the message has expired, it is removed from the queue, and disk space reclaimed. MSMQ will however only allow a single TimeToBeReceived for all messages in a transaction. It will silently copy the TimeToBeReceived from the first message enlisted to all other messages in the transaction, leading to a potential message loss scenario. This issue, unfortunately, make it impossible to support setting TimeToBeReceived for messages sent from a transactional MSMQ endpoint.
+MSMQ continuously checks the TimeToBeReceived attribute of all queued messages. As soon as the message has expired, it is removed from the queue, and disk space reclaimed. MSMQ will however only allow a single TimeToBeReceived for all messages in a transaction. It will silently copy the TimeToBeReceived from the first message enlisted to all other messages in the transaction, leading to a potential message loss scenario.
```
indualagarsamy: Any reason, this was removed?

>This issue, unfortunately, make it impossible to support setting TimeToBeReceived for messages sent from a transactional MSMQ endpoint.

If you're planning to add it, I suggest slightly tweaking it to:
This behavior, unfortunately, makes it impossible to support setting TimeToBeReceived for messages sent from a transactional MSMQ endpoint.

nservicebus/messaging/discard-old-messages.md
```diff
@@ -40,8 +40,9 @@ TimeToBeReceived relies on underlying functionality in the transport infrastruct
...
 
+If TTBR is required than [immediate message dispatch](send-a-message.md#dispatching-a-message-immediately) can be used with the caveat that this message will be send even though an error occurs after sending.
```
indualagarsamy: So, if we use immediate dispatch, we can the TTBR attribute can be used? 

>If TTBR is required than [immediate message dispatch](send-a-message.md#dispatching-a-message-immediately) can be used with the caveat that this message will be send even though an error occurs after sending.

Minor Typo.  Also I believe the warning needs to be much more explicit. For example:

If TTBR is required then, [immediate message dispatch](send-a-message.md#dispatching-a-message-immediately) can be used. However, when using immediate dispatch, messages will get sent or published, even if the message handling resulted in an error. Therefore, based on the number of retries that are configured on the endpoint, the same message could be sent or published multiple times. This behavior needs to be taken into consideration when choosing to use immediate dispatch.

nservicebus/messaging/discard-old-messages.md
ramonsmits: @indualagarsamy It was removed because it isn't true hence the added section.

nservicebus/messaging/discard-old-messages.md
ramonsmits: Wouldn't we then be repeating the same content already present behind the link?

ramonsmits:
@indualagarsamy its still WIP because of my comment https://github.com/Particular/docs.particular.net/pull/1956#issuecomment-249927618 I need guidance on how to change that content but that could also be done in an additional PR.

SzymonPobiega:
@ramonsmits how about moving the `immediate-dispatch` partial around so that it comes first for the V6, before the transaction scope thing. I'd rather have both of them in snippets but you are right that the preferred approach should come first.

SimonCropp:
This is not really a viable approach for any real world business app. Given the following

 - multiple handlers and saga that are sending messages in the same transaction.
 - handlers and saga possibly being written by different people
 - TTBR being controlled by a convention that is not apparent in the context of a handler or a saga

So with the above how would u know, in the context of writing a handler or a saga, that you should use immediate dispatch instead of a normal send.

The only way to be sure would be to always use immediate just in case. and then u would always loose all transactional guarantees

So while this "is technically a workaround" it is not one that anyone could use. 

I think it would be better if the core would re-order the messages so the one at the front has the highest ttbr. cc @Particular/nservicebus-maintainers

andreasohlund:
>I think it would be better if the core would re-order the messages so the one at the front has the highest ttbr. cc @Particular/nservicebus-maintainers 

That is doable but for most users this would mean that TTBR becomes pointless since audit is on by default. This means that we always "send" at least one message that has infinite TTBR => TTBR on the messages the users sends from a handler would never to honored?

ramonsmits:
The real fix is that TTBR should just work with ReceiveOnly but this is not about fixing but a workaround. Your ordering is also a workaround but that would be I think a feature request?

indualagarsamy:
The problem I have with specifying the immediate dispatch as a workaround is that:
- When using immediate dispatch, messages will get sent or published, even if the message handling resulted in an error. Therefore, based on the number of retries that are configured on the endpoint, the same message could be sent or published multiple times. 
- A user may not realize the implicit danger involved in using immediate dispatch. 
- What are the use cases where TTBRs need to be specified differently?
- Being able to set different TTBRs for messages in the same transaction scope was NEVER supported in MSMQ. See https://msdn.microsoft.com/en-us/library/system.messaging.message.timetobereceived(v=vs.110).aspx
See: When several messages are sent in a single transaction, Message Queuing uses the TimeToBeReceived property of the first message.

In my opinion, using immediate dispatch is not a valid workaround. I am against documenting this. If anything, let the core support it in a different way or throw an exception (as @SimonCropp suggested during my discussion with him) when different TTBRs are attempted to be set in the same message handler for the outgoing messages.

SimonCropp:
@andreasohlund 

> That is doable but for most users this would mean that TTBR becomes pointless since audit is on by default

In that case i would say TTBR in the context of MSMQ should be an "all messages setting" and we do not support it as a "per message convention"

andreasohlund:
>In that case i would say TTBR in the context of MSMQ should be an "all messages setting" and we do not support it as a "per message convention"

Not sure, that would we want audit messages to share the same TTBR?

Also we should be able to support it with txmode.none|receiveonly (yes a bug prevents this atm).

Perhaps we shouldn't support TTBR for txmode > txmode.receiveonly for msmq?

ramonsmits:
Edited the guidance as @andreasohlund mentioned in https://github.com/Particular/NServiceBus/issues/3562 that the outbox can also be used as a workaround.

andreasohlund:
@ramonsmits I think my workaround suggestion is wrong, see https://github.com/Particular/NServiceBus/issues/3562#issuecomment-250718092 ?

indualagarsamy:
I am against merging this PR in for all the reasons I mentioned earlier: https://github.com/Particular/docs.particular.net/pull/1956#issuecomment-250528916

I don't think immediate dispatch is a workaround for TTBR for MSMQ. Also based on @andreasohlund most recent comment, Outbox isn't either. 
Any improvements or suggestions that need to be addressed in the core needs to be raised as separate issues on the core. As far as documentation is concerned, this PR is a no go for me.

cc / @Particular/docs-maintainers

danielmarbach:
Closing this on behalf of core maintainers

SimonCropp:
@danielmarbach can use add reasons/as to why

danielmarbach:
I think enough has been said by indu and andreas in direkt or indirect (linked) statements

ramonsmits:
Did you even look at the latest revision @indualagarsamy / @Particular/docs-maintainers  ??

What is wrong with the outbox 'workaround'? I haven't seen @andreasohlund mentioning any of that. He only responded in the context of immediate dispatch AFAIK.

I also reordered the content for immediate dispatch to favor sendoptions instead of using transaction scope. That is now also lost!

indualagarsamy:
@ramonsmits - I have raised an issue in the core relating setting different TTBRs for messages when using MSMQ. See: https://github.com/Particular/NServiceBus/issues/4185.

Yes, I looked at the changes. We don't document bugs in docs. This needs to be fixed in the core. I apologize for not having raised the issue sooner. Regarding the re-ordering content, the deleted branch is simple to restore in GH in case you don't already have your local branch. Let us know and we can help.

danielmarbach:
I closed this based on https://github.com/Particular/docs.particular.net/pull/1956#issuecomment-250790913 the branch has been deleted by @simoncropp as a cleanup afterwards

ramonsmits:
@indualagarsamy Using outbox isnt a workaround, it is a perfectly valid feature to use on MSMQ. That is not a bug neither a workaround.

Only thing is that TTBR wont work when not using Outbox. That is a limitation and I think a perfect fit for doco. The immediate dispatch was actually a workaround and can understand somewhat that we don't want to put bad practices in our doco.

PR #1957, opened 2016-09-27
MSMQ subscription storage is suitable for scale-out scenarios

timbussmann:
Remove the warning for the v6 doc on MSMQ subscription storage.

Relates to https://github.com/Particular/NServiceBus/pull/4155


nservicebus/msmq/subscription-persistence_config_core_[,4].partial.md
```diff
@@ -1,3 +1,5 @@
+WARNING: Storing subscriptions in MSMQ is not suitable when endpoints are scaled-out, because the subscription queue cannot be shared among multiple endpoints.s
```
Comment (indualagarsamy) [inline nservicebus/msmq/subscription-persistence_config_core_[,4].partial.md]: typo

> endpoints.s

nservicebus/msmq/subscription-persistence_config_core_[5,6].partial.md
```diff
@@ -0,0 +1,11 @@
+WARNING: Storing subscriptions in MSMQ is not suitable when endpoints are scaled-out, because the subscription queue cannot be shared among multiple endpoints.
```
Comment (indualagarsamy) [inline nservicebus/msmq/subscription-persistence_config_core_[5,6].partial.md]: Should it be:

WARNING: Storing subscriptions in MSMQ is not suitable when endpoints are scaled-out, because the subscription queue cannot be shared among multiple endpoints, especially if the endpoints are running on different machines.

because they could run on the same machine in a competing consumer pattern on the same machine.

nservicebus/msmq/subscription-persistence_config_core_[5,6].partial.md
```diff
@@ -0,0 +1,11 @@
...
+
+To configure MSMQ as the subscription persistence:
+
+snippet:ConfiguringMsmqPersistence
+
+By default NServiceBus uses a queue called `NServiceBus.Subscriptions`. If not specified otherwise, all endpoints will share that queue to store subscriptions. In order to specify a different queue, add the following `configSections` and subsequent config entry:
+
+snippet:MsmqSubscriptionAppConfig
+
+WARNING: When using MSMQ Subscription Persistence on multiple endpoints running on the same machine, every endpoint requires a dedicated subscription storage queue.
```
Comment (indualagarsamy) [inline nservicebus/msmq/subscription-persistence_config_core_[5,6].partial.md]: Can we remove this warning and make it as part of the explanation in the previous paragraph.

nservicebus/msmq/subscription-persistence_config_core_[5,6].partial.md
```diff
@@ -0,0 +1,11 @@
+WARNING: Storing subscriptions in MSMQ is not suitable when endpoints are scaled-out, because the subscription queue cannot be shared among multiple endpoints.
+
+To configure MSMQ as the subscription persistence:
+
+snippet:ConfiguringMsmqPersistence
+
+By default NServiceBus uses a queue called `NServiceBus.Subscriptions`. If not specified otherwise, all endpoints will share that queue to store subscriptions. In order to specify a different queue, add the following `configSections` and subsequent config entry:
```
Comment (indualagarsamy) [inline nservicebus/msmq/subscription-persistence_config_core_[5,6].partial.md]: > By default NServiceBus uses a queue called `NServiceBus.Subscriptions`. If not specified otherwise, all endpoints will share that queue to store subscriptions. In order to specify a different queue, add the following `configSections` and subsequent config entry:

Can we combine this and the warning to make it read better? For example, how about:

When using MSMQ Subscription persistence, the default queue to store the subscription messages is `NServiceBus.Subscriptions`. It is important to override this queue name for each endpoint to avoid sharing the same subscription queue for different endpoints running on the same machine. To specify a different subscriptions queue, add the following configuration as shown below:

PR #1960, opened 2016-09-28
[WIP]Remove versioning section

andreasohlund:
@Particular/docs-maintainers is this the way to go or do we add partials? (we'll backport the removal to 5.2.X as well)

### WIP

* Upgrade guide

### DIFF
--- a/nservicebus/messaging/publish-subscribe/index.md
+++ b/nservicebus/messaging/publish-subscribe/index.md
@@ -140,7 +140,3 @@ Transport->Subscriber2: Send Message1
 
 ![](mechanics-native-publish.svg)
 
-
-## Versioning subscriptions
-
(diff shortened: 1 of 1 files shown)


weralabaj:
yes, seems good

weralabaj:
Instead of removing that section should we say the right way is using interfaces and oint to https://docs.particular.net/samples/versioning/?

weralabaj:
> do we add partials

that still will be a bug in the old versions right? so I would remove it completely

andreasohlund:
>Instead of removing that section should we say the right way is using interfaces and oint to

Not sure our guidance around contract versioning is correct. Will talk to @SimonCropp tomorrow, wan't to join @weralabaj ?

weralabaj:
yes please ping me

weralabaj:
As agreed in the meeting with @andreasohlund and @SimonCropp, merging as it is. New guidance will be handled in the https://github.com/Particular/docs.particular.net/pull/1971

PR #1961, opened 2016-09-28
WCF to messaging menu

danielmarbach:


indualagarsamy:
@danielmarbach - I think under Endpoints in Hosting will be a better location for this. 
https://docs.particular.net/nservicebus/endpoints/

Thoughts?

danielmarbach:
I thought about that too but actually it is no longer hosting related. You
could argue the feature hosts WCF services but actually it is about sending
messages from WCF to the bus isn't it?

indualagarsamy:
@danielmarbach - true. However the title of the doco is `Exposing Endpoints via WCF`.

danielmarbach:
Nothing we can't change ;)

PR #1966, opened 2016-09-29
Mention removed satellite pipeline performance counters

timbussmann:
We also use screenshots of performance counters in the related sample here: https://docs.particular.net/samples/performance-counters/

is there value in providing update screenshots for v6 which do not contain the satellite instances? @Particular/docs-maintainers

### DIFF
--- a/nservicebus/upgrades/5to6/extension-seams.md
+++ b/nservicebus/upgrades/5to6/extension-seams.md
@@ -39,9 +39,17 @@ Feature Dependencies, using the string API, now need to be declared using the ta
 snippet: 5to6-DependentFeature
 
 
+## Satellites
+
+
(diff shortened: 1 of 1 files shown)


weralabaj:
@timbussmann I wouldn't include the screenshot here, but from the guide you can link to the sample for more details if you think that's useful (IMO yes).

timbussmann:
@weralabaj I didn't mean to include screenshots here, but whether we should update the screenshots for the sample (not directly related to this PR, but to the linked change)

weralabaj:
ah I see, yes, I think it's a good idea, might be confusing for somebody new, let's open an issue and handle it after RTM? unless you have some time to spare right now, than that'd be more than awesome :D

PR #1971, opened 2016-09-30
Evolving contracts guidance

weralabaj:
As per discussion with @andreasohlund and @SimonCropp, things to cover:

Versioning:
- evolving message contracts
  - serializers API (better), message mutators (not so great, rare use cases?) 
  - add-drain-delete flow => SemVer breaking change
  - default values - beware!
  - put your contracts on a diet
  - events vs commands
  - when to use version numbers in contracts (in names)
  - how to structure your contracts => 1 vs multiple assemblies, commands vs events
- update sample => interfaces to classes
- look at evolving web API guidance || SemVer in web apis => research, reference 
- related: interfaces vs classes => separately handled


weralabaj:
@andreasohlund @SimonCropp seems like up to this point we said the exact opposite of what we discussed, e.g. in https://docs.particular.net/nservicebus/messaging/messages-as-interfaces: "Use interfaces for events"

I don't think it's enough to quietly update docs.

andreasohlund:
About to raise a platdev issue, just wanted to digg around in the code before going ahead:

https://github.com/Particular/NServiceBus/pull/4187

nservicebus/messaging/evolving-contracts.md
```diff
@@ -0,0 +1,56 @@
...
+
+TODO: specific example?
+
+
+## Handling breaking changes in contracts
+
+
+### Handling both versions of messages
+
+Serializers API => not sure what that was about? Using custom serialization rules e.g. to ignore extra properties or to have safe defaults for missing information?
```
weralabaj: @SimonCropp help!

weralabaj:
@andreasohlund @SimonCropp have a look at the outline? does it make sense?

If so, I'll try to fill each section with a few sentences. Should be good enough for a start?

Scooletz:
@SimonCropp Could you chime in? It's vital for the https://github.com/Particular/PlatformDevelopment/issues/960

nservicebus/messaging/evolving-contracts.md
```diff
@@ -0,0 +1,56 @@
...
+
+This article presents basic guidelines for choosing contracts evolution strategy, avoiding common mistakes and ensuring the contracts will be easy to evolve over time.
+
+
+## Designing contracts
+
+
+### Messages
+
+Messages should be:
```
SimonCropp: perhaps "should ideally be"

nservicebus/messaging/evolving-contracts.md
```diff
@@ -0,0 +1,56 @@
...
+
+- **POCO classes**.
```
SimonCropp: link https://en.wikipedia.org/wiki/Plain_Old_CLR_Object and spell out the acronym

nservicebus/messaging/evolving-contracts.md
```diff
@@ -0,0 +1,56 @@
...
+- **Small and focused**, carrying only the minimum amount of information, should be split if they get too big.
```
SimonCropp: "only" is redundant

nservicebus/messaging/evolving-contracts.md
```diff
@@ -0,0 +1,56 @@
...
+- **Server a single purpose**, they shouldn't be used for anything else than carrying data, in particular shouldn't be used as data access objects, etc.
```
SimonCropp: list some any patterns. eg "dont use messages as UI binding models"

nservicebus/messaging/evolving-contracts.md
```diff
@@ -0,0 +1,56 @@
...
+
+
+### Assemblies
+
+The assembly containing messages becomes part of the contract between two endpoints, as both need to refer it.
+
+The messages can be defined in more than one assembly and it's recommended for systems with multiple message types to keep them separately. Especially commands and events tend to evolve at different pace and be independent from each other, so they can be kept in separate assemblies.
```
SimonCropp: > Especially commands and events tend to evolve at different pace and be independent from each other, so they can be kept in separate assemblies. 

not sure that is true. careful evolution of message contracts should mean in most cases commands and events can be in the same assembly

SimonCropp:
looks good as a start

nservicebus/messaging/evolving-contracts.md
andreasohlund: >careful evolution of message contracts should mean in most cases commands and events can be in the same assembly

Agreed from a technical standpoint. That's said (not sure if it was worth the gain) I've tried to separate the event contracts that other "services" should care about from the internal events, commands and messages that "my service" used. Kind of like https://en.wikipedia.org/wiki/Principle_of_least_privilege ie to only provide the contracts that the consumer should care about. Lumping all messages into a single assembly would break that. Is this something that our guidance should include?

Arguably you could fix this with making the internal contracts `internal` and use `InternalsVisibleTo` for endpoints within the same service but I've never seen that in the wild and it seems like a maintenance pain.

Thoughts?

weralabaj: Hm sounds a bit complicated, I'll simplify. If somebody needs more complex approach, then hopefully they can think for themselves :)

weralabaj:
@SimonCropp @andreasohlund would that be good enough for v1 or you'd rather cover some more topics?

nservicebus/messaging/evolving-contracts.md
```diff
@@ -0,0 +1,54 @@
...
+
+In some advanced scenarios it might be necessary to introduce breaking changes in serialization formats. The guidance is provided in the [Transition serialization formats](/samples/serializers/transitioning-formats/) formats.
+
+
+#### Upgrading endpoints
+
+When there are significant changes in a message type, such as removing property, changing the property type, etc. the upgrade process should consist of the following steps:
+
+ * Add the upgraded endpoint that will handle new incoming messages.
+ * Drain the incoming queue, i.e. keep the old endpoint running until all messages in the old format are processed.
```
SimonCropp: and most real world systems "Drain the incoming queue" is not possible

nservicebus/messaging/evolving-contracts.md
```diff
@@ -0,0 +1,54 @@
...
+### Handling breaking changes in contracts
+
+There is a number of approaches when it comes to handling breaking changes in messages. One of the simplest is adding a version to the type name.
+
+In some advanced scenarios it might be necessary to introduce breaking changes in serialization formats. The guidance is provided in the [Transition serialization formats](/samples/serializers/transitioning-formats/) formats.
+
+
+#### Upgrading endpoints
+
+When there are significant changes in a message type, such as removing property, changing the property type, etc. the upgrade process should consist of the following steps:
```
SimonCropp: i prefer this approach https://docs.particular.net/samples/serializers/transitioning-formats/

nservicebus/messaging/evolving-contracts.md
```diff
@@ -0,0 +1,54 @@
...
+
+ * Add the upgraded endpoint that will handle new incoming messages.
```
andreasohlund: >how would the new endpoint only handle the new format of messages

It would handle both for some period of time

weralabaj:
Updated, see now

weralabaj:
@SimonCropp @andreasohlund ☝️

nservicebus/messaging/evolving-contracts.md
```diff
@@ -0,0 +1,53 @@
...
+### Messages
+
+Messages should ideally be:
+
+ * [**Plain Old CLR Objects** (POCO)](https://en.wikipedia.org/wiki/Plain_Old_CLR_Object).
+ * **Small and focused**. Messages should carry the minimum amount of information and should be split if they get too big.
+ * **Serve a single purpose**. Messages shouldn't be used for anything else than carrying data. For example, messages shouldn't be used as data access objects or as UI binding models.
+
+
+### Assemblies
```
andreasohlund: I've seen projects that just share the C# files should we mention that as well? (they used source only packages)

The upside of that is that you won't need to worry about the NServiceBus version the assembly is compiled against and the mess of assembly redirect that ensues

nservicebus/messaging/evolving-contracts.md
```diff
@@ -0,0 +1,53 @@
...
+
+This article presents basic guidelines for choosing contracts evolution strategy, avoiding common mistakes and ensuring the contracts will be easy to evolve over time.
+
+
+## Designing contracts
+
+
+### Messages
+
+Messages should ideally be:
```
SimonCropp: shouldnt this advice exist on this page https://docs.particular.net/nservicebus/messaging/messages-events-commands and linked to from here ?

nservicebus/messaging/evolving-contracts.md
```diff
@@ -0,0 +1,53 @@
...
+
+ * [**Plain Old CLR Objects** (POCO)](https://en.wikipedia.org/wiki/Plain_Old_CLR_Object).
+ * **Small and focused**. Messages should carry the minimum amount of information and should be split if they get too big.
+ * **Serve a single purpose**. Messages shouldn't be used for anything else than carrying data. For example, messages shouldn't be used as data access objects or as UI binding models.
+
+
+### Assemblies
+
+The assembly containing messages becomes part of the contract between two endpoints, as both need to refer it. It's recommended to use a dedicated assembly for messages definitions. By keeping messages in a separate assembly, the amount of information and dependencies shared between services is minimized.
```
SimonCropp: "refer it" > "refererence it"

nservicebus/messaging/evolving-contracts.md
```diff
@@ -0,0 +1,53 @@
...
+
+
+### Default values
+
+When a message is extended with additional properties it's necessary to carefully examine what will be the default values for those properties, especially if endpoints running in other versions don't recognize them. In particular, it's important to consider how clients might interpret the default value and provide appropriate guidelines for them.
+
+
+### Handling breaking changes in contracts
+
+There is a number of approaches when it comes to handling breaking changes in messages. One of the simplest is adding a version to the type name and creating handlers for the new contract. The handlers for the old contract might still be running, to handle in-flight messages and messages generated by old endpoints.
```
SimonCropp: and when the old endpoint consumes one of the new messages what happens?

nservicebus/messaging/evolving-contracts.md
```diff
@@ -0,0 +1,53 @@
...
+
+When there are significant changes in a message type, such as adding or removing property, changing the property type, etc. the upgrade process should consist of the following steps:
+
+ * Update contract to the new version.
+ * Update senders to use the new contract version. Ensure changes are visible for receivers, such as:
+	 * Decorate the existing property with `Obsolete` attribute with a `WARN`, when removing or renaming properties.
+ * Update receivers to handle the new contract version. Make sure the new properties are handled correctly, e.g. instead of relying on .NET to set the default value for `int Age = 1`, it's better to use nullable types and represent missing values as `null`.
+ * When all senders and receivers are updated and in-flight messages in the old format have been handled, obsolete the properties with `ERR` or simply remove it.
```
SimonCropp: ERR?

nservicebus/messaging/evolving-contracts.md
weralabaj: I guess that depends on how you handle it, but generally it'll go to the error queue unless you change the type name, right?

weralabaj: Not sure if I got it right, so would that be to clarify that adding a new property is not a breaking change, but when you remove it then it is and should be handled as described?

weralabaj: BTW Suggestion of how would you say that would be welcome ;)

nservicebus/messaging/evolving-contracts.md
```diff
@@ -0,0 +1,53 @@
...
+
+Messages definitions can be divided between multiple assemblies, which might be useful in more complex systems, e.g. to narrow down the number of contracts exposed to different services.
+
+
+## Common challenges
+
+
+### Default values
+
+When a message is extended with additional properties it's necessary to carefully examine what will be the default values for those properties, especially if endpoints running in other versions don't recognize them. In particular, it's important to consider how clients might interpret the default value and provide appropriate guidelines for them.
```
andreasohlund: I was thinking to suggest that `Nullables` could be used to highlight that the value might not be there.

This turns out to be a bad idea since if you eg do

`public int? Age`

while this helps consumers to know when age is not set vs null it means that you would have to do a breaking changes aging at some later point to switch it to its "final" signature

`public int Age`

In short: Ignore my bad idea :)

nservicebus/messaging/evolving-contracts.md
andreasohlund: Perhaps a phrase in the intro about "The best way to handle breaking changes is to avoid them, but if you really have to you need to do it in a staged fashion"?

weralabaj: @andreasohlund could you make a commit with your proposed changes? TBH I know about the subject only as much as I learned during our conversation with Simon. It's hard for me to write on a subject I know so little about so more help will be appreciated :D

andreasohlund: Rereading this I changed my mind, since this doc is about evolving message contracts I don't think we need to spell out that adding a property isn't a breaking change. 

I'm 👍 to go with this as a first version

PR #1983, opened 2016-10-04
List DTC support as advantage

timbussmann:
shouldn't we include this on the advantages?


MarcinHoppe:
I would also add that when combined with NHibernate persistence, we can support lightweight distributed transactions without relying on the DTC.

timbussmann:
@MarcinHoppe since I don't explicitly mention DTC, I'd say this is going to deep?

MarcinHoppe:
I'd err on the side of spelling out the benefits explicitly, if possible.

SimonCropp:
@MarcinHoppe 

> I'd err on the side of spelling out the benefits explicitly, if possible.

i gave the link to the official doco for MSDTC. i think it is MSs job to "sell DTC" as a good thing

PR #1990, opened 2016-10-04
Sql multi-instance mode depricated but for now will not be removed.

tmasternak:
/cc: @Particular/sqlserver-transport-maintainers


MarcinHoppe:
LGTM

Issue #2002, opened 2016-10-06
Documentation around bus.Dispose


danielmarbach:
https://docs.particular.net/nservicebus/containers/#using-an-existing-container-cleanup

Should we now remove the note about diposing the bus? With the new container releases it is now safe to dispose the bus when you have an external container since the external container will not be disposed. Thoughts? @Particular/docs-maintainers @Particular/container-maintainers

SimonCropp:
@danielmarbach agreed

PR #2024, opened 2016-10-10
Clarify v5 dispose sequence

danielmarbach:
Closes https://github.com/Particular/docs.particular.net/issues/2002

### DIFF
--- a/nservicebus/containers/index_content_core_[,5].partial.md
+++ b/nservicebus/containers/index_content_core_[,5].partial.md
@@ -6,7 +6,8 @@ Note that the instance of `IBus` is scoped for the lifetime of the container. He
 ### Cleanup
 
 When using an external container, the bus instance is not automatically disposed. To dispose of the resources properly:
-1. Call IBus.Dispose() to dispose of the bus.
-2. And then call container.Dispose() to dispose of the container.
 
-NOTE: Even when an external container is not used, the bus instance still needs to be properly disposed of by calling IBus.Dispose().
+ 1. Call IBus.Dispose() to dispose of the bus.
+ 1. And then call container.Dispose() to dispose of the container.
+
+NOTE: When an external container is not used, the bus instance still needs to be properly disposed of by calling `IBus.Dispose()`.
--- a/nservicebus/containers/index_content_core_[,5].partial.md
+++ b/nservicebus/containers/index_content_core_[,5].partial.md
@@ -5,7 +5,6 @@ Note that the instance of `IBus` is scoped for the lifetime of the container. He
 
 ### Cleanup
 
-When using an external container normally the bus instance is not disposed of automatically. The following dispose sequence should be applied if latest container adapter packages are used:
-
-1. Dispose the bus by calling `IBus.Dispose()`.
-2. Dispose the external container with `container.Dispose()`.
+When using an external container, the bus instance is not automatically disposed. In order to dispose the resources properly:
+1. Dispose the bus by calling IBus.Dispose().
+2. Dispose the external container with container.Dispose()
--- a/nservicebus/containers/index_content_core_[,5].partial.md
+++ b/nservicebus/containers/index_content_core_[,5].partial.md
@@ -5,6 +5,7 @@ Note that the instance of `IBus` is scoped for the lifetime of the container. He
 
 ### Cleanup
 
-When using an external container normally the bus instance is not disposed of manually. If `IBus.Dispose()` is called that would indirectly trigger the container to dispose lifetime scope.
+When using an external container normally the bus instance is not disposed of automatically. The following dispose sequence should be applied if latest container adapter packages are used:
 
-NOTE: Do NOT call `IBus.Dispose` when using an external container, instead call dispose in the container during shutdown.
+1. Dispose the bus by calling `IBus.Dispose()`.
+2. Dispose the external container with `container.Dispose()`.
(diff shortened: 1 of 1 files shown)


nservicebus/containers/index_content_core_[,5].partial.md
```diff
@@ -5,6 +5,7 @@ Note that the instance of `IBus` is scoped for the lifetime of the container. He
 
 ### Cleanup
 
-When using an external container normally the bus instance is not disposed of manually. If `IBus.Dispose()` is called that would indirectly trigger the container to dispose lifetime scope.
+When using an external container normally the bus instance is not disposed of automatically. The following dispose sequence should be applied if latest container adapter packages are used:
```
Comment (indualagarsamy) [inline nservicebus/containers/index_content_core_[,5].partial.md]: @danielmarbach - the only thing that's confusing is the context of "latest container adapter packages"
What's considered latest vs latest 4 months from now. Can we add a version specific language? Can we anchor it to the version of the NServiceBus core version that the containers use that exhibit this behavior?

Comment (danielmarbach) [inline nservicebus/containers/index_content_core_[,5].partial.md]: I had a hard time when I wrote this. First I didn't mention it at all, then I figured it might be helpful because we can't assume people run the latest bugfix. We would need to specify the version which contains the change for all container packages. Should we do it?

Comment (indualagarsamy) [inline nservicebus/containers/index_content_core_[,5].partial.md]: @danielmarbach - do the changes pertain to Version 6 of NServiceBus? or just different versions of the container packages?

Comment (danielmarbach) [inline nservicebus/containers/index_content_core_[,5].partial.md]: Nope we backported it to v5 supporting containers

Comment (indualagarsamy) [inline nservicebus/containers/index_content_core_[,5].partial.md]: @danielmarbach - yeah, i get it. it is tricky! how about the below? 

When using an external container, the bus instance is not automatically disposed. In order to dispose the resources properly:
1. Dispose the bus by calling `IBus.Dispose()`.
2. Dispose the external container with `container.Dispose()`

This behavior was introduced from Versions 6 and above of NServiceBus and also back-ported to Version 5 of NServiceBus. Therefore, ensure that the version of NServiceBus used in the project supports this behavior when making the suggested code change.

Comment (indualagarsamy) [inline nservicebus/containers/index_content_core_[,5].partial.md]: @danielmarbach - 

After trying to write this in a meaningful way to the user, I failed. As @SimonCropp pointed out, there's more ambiguity to be addressed. For example, he mentioned that this wasn't back ported to all versions of Version 5. For example, while it was back ported to it was back-ported to NServiceBus Unity 6.2.x, it was not back-ported to NServiceBus Unity 6.1.x.

Given that, I think it's best to include a table of versions that support this change, For example:
```
When using an external container, the bus instance is not automatically disposed. In order to dispose the resources properly:
1. Dispose the bus by calling IBus.Dispose().
2. Dispose the external container with container.Dispose()

This new behavior was introduced in the versions of the container that depend on NServiceBus  Versions 6 and above. It was also back-ported to some container versions that depend on NServiceBus Version 5 as a patch release. The code change is applicable to all the versions listed in the table below:

[Insert table of containers and versions that support this change]
```

As a side note, @Particular/nservicebus-maintainers - trying to document this points out the flaw in couple of things. This should not have been back ported in the first place. If it was back-ported, it should've been back-ported to not just partial versions. I believe having a docs PR at the time of making this change would have highlighted this problem.

Comment (danielmarbach) [inline nservicebus/containers/index_content_core_[,5].partial.md]: @indualagarsamy we did the triaging here and concluded we only backport this to latest minors of the containers one version back. If we now need a version table, so be it. I'll update the PR

nservicebus/containers/index_content_core_[,5].partial.md
```diff
@@ -5,6 +5,7 @@ Note that the instance of `IBus` is scoped for the lifetime of the container. He
...
 
-NOTE: Do NOT call `IBus.Dispose` when using an external container, instead call dispose in the container during shutdown.
+1. Dispose the bus by calling `IBus.Dispose()`.
+2. Dispose the external container with `container.Dispose()`.
```
Comment (danielmarbach) [inline nservicebus/containers/index_content_core_[,5].partial.md]: @johnsimons we had a fundamental flaw in how we disposed containers. we changed it now to no longer dispose externally owned containers. So if you pass in an existing container into the BusConfiguration then that container is no longer disposed and you have to dispose it yourself.

Comment (danielmarbach) [inline nservicebus/containers/index_content_core_[,5].partial.md]: @johnsimons https://github.com/Particular/ServiceControl/pull/799

nservicebus/containers/index_content_core_[,5].partial.md
Comment (danielmarbach) [inline nservicebus/containers/index_content_core_[,5].partial.md]: While rubber ducking with @SimonCropp I came to the conclusion that we don't need a version table. The sequence of disposing the bus and then disposing the container is a valid sequence under all circumstances. If you pass in a container into the bus configuration you have to create the container first. All those containers are disposable. when you create the container the code doing it becomes the composition root of the container and also the bus. Leaving aside the implementation details of UnicastBus the valid sequence of disposing is

1. Dispose the bus by calling IBus.Dispose().
2. Dispose the external container with container.Dispose()

in all scenarios independent from the backported changes. So the documentation in that sense holds up and we don't need a version table. @indualagarsamy ping me when you are online

indualagarsamy:
Discussed with @danielmarbach and he felt that it was still not clear that the bus needs to be disposed when not using an external container. So I added an extra note.

PR #2032, opened 2016-10-11
docs updated to sqlserver 3.0.0

tmasternak:


timbussmann:
(merging this so we can run the update script again to verify all pushed packages were updated)

Issue #1902, opened 2016-09-13
Remove reference to pre-release nuget steps


mikeminutillo:
Currently users installing V6 require the Prerelease flag in NuGet.

After the RTM this flag is no longer required (or desired). We need to review the upgrade guide and remove all references to this flag.

e.g. https://github.com/Particular/docs.particular.net/pull/1878#discussion_r78493928

PR #1878, opened 2016-09-08
[WIP] Upgrade Guide Homepage

mikeminutillo:
Creating a lighter-weight homepage for the upgrade guide to guide readers through the process at a high level and link off to additional detail where necessary.

PR #2038, opened 2016-10-11
Removed the note to include the -pre release option.

indualagarsamy:
Fixes https://github.com/Particular/docs.particular.net/issues/1902

@SimonCropp @mikeminutillo - any more references?


indualagarsamy:
After going through the issue and the related comment, I believe this is it. https://github.com/Particular/docs.particular.net/pull/1878#r78493928

PR #2043, opened 2016-10-13
Clarify ReceiveOnly support in v6

timbussmann:
Relates to https://github.com/Particular/NServiceBus/pull/4218

@SimonCropp I realized that the full section is "limited" to v6, as it mentions the `TransportTransactionMode`s introduced with v6. So that part doesn't seem to be working for older versions anyway.


SimonCropp:
replaced by https://github.com/Particular/docs.particular.net/pull/2048

PR #2091, opened 2016-10-19
second review of SI sequence diagram page

adamralph:
This time largely focused on tweaking the language used.

@Particular/docs-maintainers @Particular/serviceinsight-maintainers please review.

### DIFF
--- a/serviceinsight/sequence-diagram/index.md
+++ b/serviceinsight/sequence-diagram/index.md
@@ -73,20 +73,20 @@ Events are represented similarly to other messages but with dashed lines and a d
 NOTE: Each event published will appear once for each subscriber, as if individual messages were sent to each subscriber by the sender, regardless of whether Unicast or Multicast routing is used. [Learn more about Message Routing](/nservicebus/messaging/routing.md).
 
 
-### Loopbacks
+### Loopback Messages
 
-![Loopback](loopback.png)
+![Loopback message](loopback.png)
 
-A loopback is a message that an endpoint sends to itself and is represented by a short arrow with a special icon, that does not connect to another endpoint lifeline. As with any other message, hovering over or selecting the message will highlight the processing for that message in the lifeline.
+A loopback message is a message that an endpoint sends to itself and is represented by a short arrow with a special icon, that does not connect to another endpoint lifeline. As with any other message, hovering over or selecting the message will highlight the processing for that message in the lifeline.
 
 
-### Timeouts
+### Timeout messages
 
-![Timeout](timeout.png)
+![Timeout message](timeout.png)
 
-A timeout is a special type of loopback where handling is deferred until later. It is represented as a loopback with a clock icon.
+A timeout message is a special type of loopback message where handling is deferred until later. It is represented as a loopback message with a clock icon.
 
-NOTE: The time of processing may not correspond to the time at which a timeout was sent back for processing by the timeout scheduler. The sequence diagram does not currently support visualization of the time at which the timeout was sent back and will only indicate when it was processed.
+NOTE: The time of processing may not correspond to the time at which a timeout message was sent back for processing by the timeout message scheduler. The sequence diagram does not currently support visualization of the time at which the timeout message was sent back and will only indicate when it was processed.
 
 
 ### Differences with UML sequence diagrams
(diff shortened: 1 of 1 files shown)


serviceinsight/sequence-diagram/index.md
```diff
@@ -1,114 +1,115 @@
...
 
 ![Sequence Diagram](sequence-diagram.png 'width=500')
 
 ## How the diagram is generated
 
-When NServiceBus sends or publishes a message it adds specific headers such as `NServiceBus.MessageId`, a unique ID for the message. The first message to be sent in a conversation will also get a unique `ConversationId` header.
+When NServiceBus sends or publishes a message it adds specific headers such as `NServiceBus.MessageId`, which is a unique ID for that message. The first message to be sent in a conversation will also get a unique `ConversationId` header.
 
-If the handling of an incoming message causes more messages to be sent or published, those outgoing messages are given a `NServiceBus.RelatedTo` header equal to the `MessageId` of the incoming message and are also given the same `ConversationId` as the incoming message.
+If the handling of an incoming message causes more messages to be sent or published, those outgoing messages are given an `NServiceBus.RelatedTo` header equal to the `MessageId` of the incoming message and are given the same `ConversationId` as the incoming message.
```
HEskandari: Should we clarify this bit?

>...causes more messages to be sent or published

Something like:

>...causes more commands to be sent or events published

serviceinsight/sequence-diagram/index.md
```diff
@@ -1,114 +1,115 @@
...
+Each conversation starts with a single command or event. This message is often triggered by some action external to the system such as a user clicking a Submit Order button on a website. The metadata used to generate the diagram does not include the trigger such cases but it does include the endpoint that sent or published the message, represented by a Start of Conversation marker on the endpoint lifeline.
 
-NOTE: Not every sequence diagram will have the start of conversation marker. This can happen if the conversation started a long time ago and the initiating message has expired. It can also happen if the number of messages in the conversation is very large. In this case, ServiceInsight will only get 50 messages from the conversation and this may not include the initiating message.
+NOTE: Not every sequence diagram will have a Start of Conversation marker. This can happen if the conversation started a long time ago and the initiating message has expired. It can also happen if the number of messages in the conversation is very large. The diagram will only show get the last 50 messages from the conversation and this may not include the initiating message.
```
HEskandari: The first part of the paragraph seems a bit misleading. Could we rephrase? Something like:

>Start of Conversation marker might not be always visible in a sequence diagram. This can happen...

serviceinsight/sequence-diagram/index.md
```diff
@@ -1,114 +1,115 @@
...
+NOTE: Each event published will appear once for each subscriber, as if individual messages were sent to each subscriber by the sender, regardless of whether Unicast or Multicast routing is used. [Learn more about Message Routing](/nservicebus/messaging/routing.md).
 
 
 ### Loopback messages
 
 ![Loopback](loopback.png)
 
-When an endpoint sends a message to itself this is called a loopback message. On the sequence diagram this is represented by a short arrow that does not connect to another endpoint lifeline and a special icon. As with any other type of message, hovering over or selecting the message will highlight the processing for that message in the lifeline.
+A loopback message is a message that an endpoint sends to itself and is represented by a short arrow with a special icon, that does not connect to another endpoint lifeline. As with any other message, hovering over or selecting the message will highlight the processing for that message in the lifeline.
```
HEskandari: Can we go with one 'message? e.g.

>A loopback is a message that...

serviceinsight/sequence-diagram/index.md
adamralph: 👍 I've changed this to

> If the handling of an incoming message results in outgoing messages, those messages are given...

serviceinsight/sequence-diagram/index.md
adamralph: 👍 I've changed this to

> NOTE: The Start of Conversation marker may not always be visible. This can happen...

serviceinsight/sequence-diagram/index.md
adamralph: It depends whether the concept is called a "loopback" or a "loopback message". If the former, we should change the section header to `### Loopbacks`. If the latter, then this sentence needs to stay as is.

adamralph:
Back to you @HEskandari. The changes I made are slightly different to what you suggested. Please let me know what you think.

serviceinsight/sequence-diagram/index.md
HEskandari: This is just how the self-sent messages are named in ServiceInsight. A search in docs website didn't reveal any concept named 'loopback' so definitely not a ubiquitous word.

HEskandari:
I'm still not happy with the term "Loopback". I have looked at the ServiceInsight code base there's no reference there. There's also no reference to this term across the API. Funny enough, in the table below, it is called as:

>Send to self messages / loopbacks

I'm happy to do that separately though, as we also need to update the images.

adamralph:
@HEskandari

> If we agree on the name, I'm happy to do that separately, as we also need to update the images.

👍 agreed.

In that case, in order to get this PR merged, we need the document to say "loopback messages" and "timeout messages". I've add a commit which does that.

HEskandari:
@adamralph sounds good to me.

adamralph:
@HEskandari great. For some reason the CI builds aren't kicking in after my latest commits. Once they've run, I'll merge.

PR #2096, opened 2016-10-19
Added glossary

dvdstelt:
Was discussing terms not being clear in a document and decided perhaps we could use a glossary. Reading up on terms like `consumer` and the difference between `logical endpoint` and `endpoint instance` is missing at the moment, whereas we are using them throughout the documentation. This makes it available and perhaps others/other-documents could make use of it as well. 

It's just a first kick off of this document to see if it's usable, before spending too much time on it. If we agree this could be usefull, we can extend it before merging it.

### DIFF
--- a/glossary/glossary.md
+++ b/glossary/glossary.md
@@ -1,0 +1,55 @@
+---
+title: Glossary
+summary: Definitions of terms applicable when using NServiceBus and/or the Particular Platform
+tags:
+- Glossary
+---
+
+# Glossary
+
+### Message
+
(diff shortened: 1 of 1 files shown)


SimonCropp:
Doesn't this heavily overlap with https://docs.particular.net/nservicebus/concept-overview ?

dvdstelt:
@SimonCropp Very much so, I have seen this before but wasn't aware of its existence when thinking about this.
Let's close this and I'll have a look if the other document might need some additions.

Issue #2078, opened 2016-10-18
The page 'Installation' requires review


SimonCropp:
The page [Installation](https://docs.particular.net/nservicebus/hosting/nservicebus-host/installation) has never been reviewed.

 * [Page review instructions](https://github.com/Particular/docs.particular.net#reviewing-a-page)
 * [Location in GitHub](https://github.com/Particular/docs.particular.net/blob/master/nservicebus/hosting/nservicebus-host/installation.md)

SimonCropp:
ping @Particular/nservicebus-maintainers

danielmarbach:
Talked to @timbussmann and he is going to tackle this together with @dbelcham

PR #2099, opened 2016-10-20
Review installation.md

timbussmann:
relates to https://github.com/Particular/docs.particular.net/issues/2078

* Split the installation and uninstallation parts to make the specific sections easier to read.
* added a section for each installation argument
* removed `/installInfrastructure` since it doesn't seem to be around in supported versions
* I didn't find the `instance` parameter in the code of the host at a quick glance. Maybe I didn't look close enough, but can somebody help me to understand whether this is still supported and how to document it? /cc @ramonsmits @sfarmar

### DIFF
--- a/nservicebus/hosting/nservicebus-host/installation.md
+++ b/nservicebus/hosting/nservicebus-host/installation.md
@@ -86,7 +86,7 @@ Configures NServiceBus to scan only the specified assemblies.
 
 ### DependsOn
 
-Specifies the names of services or groups which must start before this service. Multiple services/groups are separated by a comma.
+Specifies the names of services or groups which must start before this service, e.g. `/dependsOn:"MSMQ"`. Multiple services/groups are separated by a comma.
 
 
 ### SideBySide
--- a/nservicebus/hosting/nservicebus-host/installation.md
+++ b/nservicebus/hosting/nservicebus-host/installation.md
@@ -45,7 +45,7 @@ NServiceBus.Host.exe /install
 /description:"Service for publishing event messages"
 /endpointConfigurationType:"QualifiedNameSpace.EndpointConfigType, AssemblyName"
 /username:"corp\serviceuser"
-/password:"p@ssw0rd!" NServiceBus.Production
+/password:"p@ssw0rd!"
 ```
 
 
--- a/nservicebus/hosting/nservicebus-host/installation.md
+++ b/nservicebus/hosting/nservicebus-host/installation.md
@@ -21,19 +21,19 @@ To install the process as a Windows Service, include `/install` as a command lin
 
 ```dos
 NServiceBus.Host.exe /install
-[/serviceName]
-[/displayName]
-[/description]
-[/endpointConfigurationType]
-[/endpointName]
+[/serviceName:<string>]
+[/displayName:<string>]
+[/description:<string>]
+[/endpointConfigurationType:<string>]
+[/endpointName:<string>]
 [/installInfrastructure]
-[/scannedAssemblies]
-[/dependsOn]
+[/scannedAssemblies:<string>]
+[/dependsOn:<string>]
 [/sideBySide]
 [/startManually]
-[/username]
-[/password]
-[profile]
+[/username:<string>]
+[/password:<string>]
+[<profile>]
 ```
 
 Here is an example of the `/install` command line:
--- a/nservicebus/hosting/nservicebus-host/installation.md
+++ b/nservicebus/hosting/nservicebus-host/installation.md
@@ -33,6 +33,7 @@ NServiceBus.Host.exe /install
 [/startManually]
 [/username]
 [/password]
+[profile]
 ```
 
 Here is an example of the `/install` command line:
@@ -111,6 +112,11 @@ Version 7 and above supports [Group Managed Service Accounts (GMSA)](http://blog
 ### Password
 
 If the specified account which runs the Windows Services requires a password, set it using the `/password:"<password>"` parameter.
+
+
+### Profile
+
+A [host profile](profiles.md) can be specified as the last parameter, e.g. `NServiceBus.Host.exe /install NServiceBus.Lite`. By default, the `NServiceBus.Production` profile is applied.
 
 
 ## Uninstalling a Windows Service
--- a/nservicebus/hosting/nservicebus-host/installation.md
+++ b/nservicebus/hosting/nservicebus-host/installation.md
@@ -6,7 +6,7 @@ tags:
 reviewed: 2016-10-20
 ---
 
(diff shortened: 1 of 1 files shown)


nservicebus/hosting/nservicebus-host/installation.md
```diff
@@ -20,83 +32,103 @@ NServiceBus.Host.exe [/install [/serviceName]
...
 
 ```dos
-NServiceBus.Host.exe /?
+NServiceBus.Host.exe /install 
+/serviceName:"MyPublisher"
+/displayName:"My Publisher Service"
+/description:"Service for publishing event messages"
+/endpointConfigurationType:"QualifiedNameSpace.EndpointConfigType, AssemblyName"
+/username:"corp\serviceuser"
+/password:"p@ssw0rd!" NServiceBus.Production
```
indualagarsamy: @timbussmann - In older versions of NServiceBus (maybe version 3 and below), you had to specify the profile. In the current versions, NServiceBus.Production is the default profile and therefore I don't think needs to be specified.

nservicebus/hosting/nservicebus-host/installation.md
```diff
@@ -20,83 +32,103 @@ NServiceBus.Host.exe [/install [/serviceName]
...
 ```
 
-To set the actual name of the Windows Services in the registry, specify `/serviceName:ServiceName`. This is different from what is displayed in the Windows Service Manager.
+
+### ServiceName
+
+To set the actual name of the Windows Services in the registry, specify `/serviceName:ServiceName`. By default, the name of the service is the name of the endpoint and the endpoint name is the namespace of the endpoint configuration class.
```
indualagarsamy: >This name  is the namespace of the endpoint configuration class

Now that endpoint name needs to be explicitly provided in the constructor, does the namespace name of the class matter anymore?

nservicebus/hosting/nservicebus-host/installation.md
```diff
@@ -20,83 +32,103 @@ NServiceBus.Host.exe [/install [/serviceName]
...
+
+Note: The Windows Service name is different from what is displayed in the Windows Service Manager. The Windows Service Manager shows the *DisplayName*.
+
+
+### DisplayName
 
 To set the name of the Windows Service as it is displayed in the Windows Service Manager, specify `/displayName:ServiceDisplayName`.
 
-If `/displayName` is not specified, but `/serviceName` is, the display name does not become what was passed in the `/serviceName`, but rather the default described above.
+If `/displayName` is not specified, but `/serviceName` is, the display name does not become what was passed in the `/serviceName`, but the default *ServiceName* as described above is used.
```
indualagarsamy: >If `/displayName` is not specified, but `/serviceName` is, the display name does not become what was passed in the `/serviceName`, but the default *ServiceName* as described above is used.

How about:
If the /displayName is not specified, then the default *ServiceName* will be used as the display name and not the provided service name as specified in the /serviceName parameter.

nservicebus/hosting/nservicebus-host/installation.md
```diff
@@ -20,83 +32,103 @@ NServiceBus.Host.exe [/install [/serviceName]
...
+
+
+### EndpointName
+
+Configures the name of the endpoint. By default, the endpoint name is the namespace of the endpoint configuration class.
+
+
+### ScannedAssemblies
+
+Configures NServiceBus to scan only the specified assemblies.
```
indualagarsamy: Can we provide an example for this?

nservicebus/hosting/nservicebus-host/installation.md
```diff
@@ -20,83 +32,103 @@ NServiceBus.Host.exe [/install [/serviceName]
...
+
+
+### DependsOn
+
+Specifies the names of services or groups which must start before this service. Multiple services/groups are separated by a comma.
```
indualagarsamy: Can we provide an example for this?

nservicebus/hosting/nservicebus-host/installation.md
```diff
@@ -20,83 +32,103 @@ NServiceBus.Host.exe [/install [/serviceName]
...
 
 By default, Windows Services start automatically when the operating system starts. To change that, add
 `/startManually` to the `/install` command.
 
-To specify under which account the service runs, pass in the username and password of that account.
 
-Here is an example of the `/install` command line:
+### Username
+
+To specify under which account the service runs, pass in the username of that account.
```
indualagarsamy: What will the service account default be? LocalUser? LocalService? LocalSystem? can you specify that? 

> To specify under which account the service runs, pass in the username of that account.

How about:
Provide the username that the service should execute in the context of, instead of running as [whatever the default is]

nservicebus/hosting/nservicebus-host/installation.md
```diff
@@ -20,83 +32,103 @@ NServiceBus.Host.exe [/install [/serviceName]
...
+
+NOTE: When installing the Host using a custom user account, the user account is added to the `Performance Monitor Users` and is granted `run as a service` privileges. If, at a later time, the user needs to be changed it is suggested to uninstall the Host and re-install it in order to guarantee that the new user is correctly setup.
```
indualagarsamy: I think it will be good to mention that it is also up to the administrator's responsibility to clean up i.e. remove the user from the Performance Monitor user group and remove the privilege because we don't do it.

nservicebus/hosting/nservicebus-host/installation.md
timbussmann: ah good to know. I was wondering about that argument anyway. What do you think about including the profile on the install argument list and link to the profile docs?

nservicebus/hosting/nservicebus-host/installation.md
timbussmann: I think the behavior is still as described since the host provides the parameter to the `EndpointConfiguration`

it seems to use:

```
endpointName = specifier.GetType().Namespace ?? specifier.GetType().Assembly.GetName().Name;
var configuration = new EndpointConfiguration(endpointName);
```

nservicebus/hosting/nservicebus-host/installation.md
timbussmann: need to dig deeper into that argument since from a first glance at the code it looks like this parameter can be used repeatedly.

nservicebus/hosting/nservicebus-host/installation.md
indualagarsamy: @timbussmann - yeah, I'd take out NServiceBus.Production from the example and  👍 to provide it in the argument list like you've done and add a link.

nservicebus/hosting/nservicebus-host/installation.md
indualagarsamy: Thanks @timbussmann - that explains.

indualagarsamy:
@timbussmann - can you ping me when this is complete? Thanks.

timbussmann:
@indualagarsamy should be good now

PR #2100, opened 2016-10-20
add deprecation section to release policy

adamralph:


nservicebus/upgrades/release-policy.md
```diff
@@ -42,11 +42,16 @@ Some examples:
...
 
 [Contact support](https://particular.net/support) if there are any bugfixes that needs to be back-ported to a specific version.
 
 
+### Deprecation
+
+Public API methods and properties are deprecated using the [ObsoleteAttribute messages](https://msdn.microsoft.com/en-us/library/system.obsoleteattribute.aspx). Continued usage of a deprecated method or property will result either in a compiler warning or error. A warning message indicates the version in which the deprecation will start to generate a compiler error (usually the next major version). The message will also indicate the version in which the deprecated method or property will be removed. Deprecations in minor releases will always generate a compiler warning and not an error.
```
adamralph: Also "are deprecated using the [ObsoleteAttribute messages]" this is a good change, but "the" does not belong here. I believe "messages" should also be outside the link, since it points to the attribute page and not a specific section about the messages.

SimonCropp: @adamralph  in an effort to keep things moving quickly in docs PRs are merged ASAP. If you would prefer this not to happen then you need to add a WIP.

Tweaking of PRs is something that happens in the docs repo regularly. For trivial things it greatly reduces the friction for PRs. 

Since we care less about history in docs, if ever something is not correctly fixed, then it can always be fixed after the fact. 

Also a significant number of small changes are made by the docs maintainers with no PR. So the change Indu did could have been done after the PR.

If you want to further tweak the md then submit a another PR and mark it as WIP so we can have an extended conversation about it

adamralph: Thanks @SimonCropp. To clarify, I believe the PR was ready to merge after my changes which is why I didn't mark it WIP. I've raised https://github.com/Particular/docs.particular.net/pull/2120 to fix this up.

PR #2110, opened 2016-10-21
[WIP] Replace Rx

SeanFeldman:
RX was throwing NRE exception.
@yvesgoeleven have a look at this approach.

### DIFF
--- a/samples/azure/performance-tuning-asb/ASB_7/Shared/ReceiveCounter.cs
+++ b/samples/azure/performance-tuning-asb/ASB_7/Shared/ReceiveCounter.cs
@@ -1,23 +1,34 @@
 ﻿using System;
-using System.Reactive.Linq;
-using System.Reactive.Subjects;
+using System.Threading;
+using System.Timers;
+using Timer = System.Timers.Timer;
 
 public class ReceiveCounter
 {
-    Subject<SomeMessage> messages = new Subject<SomeMessage>();
+    static int count;
+    Timer timer;
+    Action<int> action;
 
     public void Subscribe(Action<int> action)
     {
-        var observable = from e in messages
-            group e by "" into c
-            from v in c.Buffer(TimeSpan.FromSeconds(1), TimeSpan.FromSeconds(1))
-            select v.Count;
-
-        observable.Subscribe(action);
+        this.action = action;
+        timer = new Timer
+        {
+            Interval = 1000,
+            AutoReset = true
+        };
+        timer.Elapsed += TimerOnElapsed;
+        timer.Start();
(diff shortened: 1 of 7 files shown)


samples/azure/performance-tuning-asb/ASB_7/Shared/ReceiveCounter.cs
```diff
@@ -1,23 +1,34 @@
 ﻿using System;
-using System.Reactive.Linq;
-using System.Reactive.Subjects;
+using System.Threading;
+using System.Timers;
```
indualagarsamy: Why System.Timers and not System.Threading.Timers?

SeanFeldman: I recall a difference between the two, where Threading.Timers was not advised.

SimonCropp: my understanding is System.Timers.Timer is primarily for a design surface. so in this case we should use System.Threading.Timers. System.Threading.Timers is what we use in the core

SeanFeldman: Ah, it's the other way around. Thanks @SimonCropp, fixed.

SeanFeldman:
Long ago had the same question and was quite surprised:
http://stackoverflow.com/a/1416809/59563

indualagarsamy:
@seanfeldman:

System.Threading.Timer is a plain timer. It calls you back on a thread pool thread (from the worker pool).

System.Timers.Timer is a System.ComponentModel.Component that wraps a System.Threading.Timer, and provides some additional features used for dispatching on a particular thread.

System.Windows.Forms.Timer instead wraps a native message-only-HWND and uses Window Timers to raise events in that HWNDs message loop.

If your app has no UI, and you want the most light-weight and general-purpose .Net timer possible, (because you are happy figuring out your own threading/dispatching) then System.Threading.Timer is as good as it gets in the framework.



From: http://stackoverflow.com/questions/1416803/system-timers-timer-vs-system-threading-timer

SeanFeldman:
@indualagarsamy it has been already updated :)

SimonCropp:
@SeanFeldman should this still have a WIP?

SeanFeldman:
@SimonCropp I was hoping @yvesgoeleven could have a look since I'm taking out Rx he has implemented originally. If you're happy with implementation, please merge.

SeanFeldman:
Superseded by #2128

PR #2112, opened 2016-10-24
fixes to delayed delivery support

tmasternak:
After investigation made with @Scooletz we verified that:
 1. Delayed delivery works with TransportTransaction.None if endpoint is not send-only
 2. For send-only endpoints it works if transport supports delayed delivery natively.

/cc: @timbussmann


SimonCropp:
merged manually here https://github.com/Particular/docs.particular.net/commit/4775fa486fdcecd332aa8dc6f76c3c86662a9a9d

PR #2117, opened 2016-10-26
update to NSB.RabbitMQ 4.1.0

adamralph:


Snippets/Rabbit/Rabbit_4/Usage.cs
```diff
@@ -92,7 +92,7 @@ void UseRoutingTopology(EndpointConfiguration endpointConfiguration)
         #region rabbitmq-config-useroutingtopology
 
         var transport = endpointConfiguration.UseTransport<RabbitMQTransport>();
-        transport.UseRoutingTopology<MyRoutingTopology>();
+        transport.UseRoutingTopology(durableExchangesAndQueues => new MyRoutingTopology());
```
adamralph: @Particular/rabbitmq-transport-maintainers this is the only non-trivial change. Do you think the expression parameter name is OK?

Also, the snippet will no longer work for 4.0, only 4.1. Does that matter?

SimonCropp: Isnt this effectively a new api? In which case we should document both?

SimonCropp: > Can we version snippets/partials by minor?

Not easily. Prob just easier to document both inside the v4.

adamralph: OK, I can do that. The only problem is that the compilation of snippets throws a compiler error on the obsolete, since warnings are reported as errors. How should I handle that? Should I use a local `#pragma warning disable CS0618`?

SimonCropp: Yep. Pragma it

SimonCropp:
Will there be n ulgrade guide fir this cgsnge?

adamralph:
> Will there be n ulgrade guide fir this cgsnge?

No, but I will provide "an upgrade guide for the change" 😉

adamralph:
@SimonCropp please see latest commit. I hope I did the partials and snippets correctly...

bording:
Pushed up some changes that I think gets all of the versioning for things sorted.

adamralph:
@bording thanks for the fix ups.

weralabaj:
@Particular/rabbitmq-transport-maintainers is that ready to merge or you're still working on it?

menu/menu.yaml
```diff
@@ -111,6 +111,8 @@
       Articles:
       - Url: nservicebus/upgrades/rabbitmq-3to4
         Title: Version 3 to 4
+      - Url: nservicebus/upgrades/rabbitmq-4.0to4.1
+        Title: Version 4.0 to 4.1
```
bording: Everything looks fine to me, not sure what you're referring to?

bording:
@weralabaj I believe this should be ready if you don't see anything that needs to be fixed.

menu/menu.yaml
adamralph: I'm not sure what I was referring to either. It looks fine.

PR #2126, opened 2016-10-28
Added info on worker capacity.

SzymonPobiega:


samples/scaleout/distributor-upgrade/sample.md
```diff
@@ -70,6 +70,10 @@ In Versions 6 and above there is a new API for enlisting with the distributor. T
 
 snippet:Enlisting
 
+The last parameter, `capacity` is the maximum number of messages in-flight between the distributor and the worker. The worker sends an acknowledgement (ACK) message (also called `ready` message) for each message processed. The distributor keeps track of these ACKs to ensure no worker is flooded with work that it can't process
```
SimonCropp: given this is a public API in the distributor shouldnt it be documented there and linked to from this page?

ramonsmits: So:  **capacity = max concurrency level + prefetch count**

SimonCropp: ie shouldnt this api be documented https://github.com/Particular/NServiceBus.Distributor.Msmq/blob/develop/src/NServiceBus.Distributor.MSMQ/IWorkerAvailabilityManager.cs#L20  ?

SzymonPobiega: @SimonCropp well, it should because it is public but I haven't heard anybody using a custom availability manager.

Re the `capacity` thing I guess ideally it should be documented somewhere in the distributor doco but it was only used in the upgrade guide. How about we keep it where it is for now and create a new issue to add a proper distributor API doco that would include both pieces?

samples/scaleout/distributor-upgrade/Core_6/Worker1/Program.cs
```diff
@@ -23,7 +24,7 @@ static void Main()
         endpointConfiguration.EnlistWithLegacyMSMQDistributor(
             masterNodeAddress: ConfigurationManager.AppSettings["DistributorAddress"],
             masterNodeControlAddress: ConfigurationManager.AppSettings["DistributorControlAddress"],
-            capacity: 10);
+            capacity: PushRuntimeSettings.Default.MaxConcurrency);
```
ramonsmits: I wondered what `PushRuntimeSettings` is, I searched docs but it is only used in a few samples.

SzymonPobiega: I think it normally is only used in the satellite registration code. Not sure if it is the best choice but it is also responsible for defining the default max concurrency.

samples/scaleout/distributor-upgrade/sample.md
```diff
@@ -70,6 +70,10 @@ In Versions 6 and above there is a new API for enlisting with the distributor. T
...
+
+Prior to Version 6 the capacity was always equal to the maximum concurrency setting. That caused the worker queue to be empty most of the time because there could never be more messages in-flight than threads ready to process them on the worker side. In Version 6 these settings are decoupled. The snippet above shows how to achieve exactly same behavior as in previous versions.
```
weralabaj: >  That caused the worker queue to be empty most of the time because there could never be more messages in-flight than threads ready to process them on the worker side.

Do we really want to say that? Sounds like a bad thing the first time you read it :) Also, it's not obvious for me why that cuases worker queue to be empty?

> Prior to Version 6 the capacity was always equal to the maximum concurrency setting. That caused the worker queue to be empty most of the time because there could never be more messages in-flight than threads ready to process them on the worker side. In Version 6 these settings are decoupled. The snippet above shows how to achieve exactly same behavior as in previous versions.

In Versions 6 and below the capacity was always set to the same value as the maximum concurrency setting, which means there were never more messages in-flight than threads ready to process them on the worker side.  (clarify the empty thing here)

A new setting was introduced in Version 6, which allows to explicitly control the capacity value. The snippet above shows how to set it to the same value as in previous versions.

SzymonPobiega: ❤️

samples/scaleout/distributor-upgrade/sample.md
SimonCropp: > How about we keep it where it is for now and create a new issue to add a proper distributor API doco that would include both pieces

i am ok with that

SzymonPobiega: Added issue https://github.com/Particular/docs.particular.net/issues/2127

samples/scaleout/distributor-upgrade/Core_6/Worker1/Program.cs
SzymonPobiega: I'll add this to the #2127 PoA

PR #2140, opened 2016-11-01
[WIP] Update Pipeline Stages

mikeminutillo:
There is no `IFaultContext` any more.

Also there are some other contexts that are not mentioned here. Should some/all of these be added? Were they left out intentionally?

- **Transport Receive** comes before Incoming Physical Message. It is executed even before the outbox and the batch sending catcher. I don't think anyone would ever need to drop a behavior in here so leaving it out is probably fine.
- **Routing** comes after Outgoing Physical. Is currently used to force immediate dispatch, attach sender info, and route deferred messages to the timeout manager. Also catches messages in the batch dispatch bucket.
- **Dispatch** comes after routing. It currently just contains a terminator for handing off to the transport
- **Batch Dispatch** is invoked for outgoing messages that are sent from within a handler. All outgoing messages are trapped by the Routing-to-Dispatch connector. When the handler is finished, these messages are all passed off to the **Batch Dispatch** context which passes them on to the Dispatch context.
- **Forwarding** runs parallel to audit but it's a different context with different behaviors.

### DIFF
--- a/nservicebus/pipeline/steps-stages-connectors_extra_core_[6.0,).partial.md
+++ b/nservicebus/pipeline/steps-stages-connectors_extra_core_[6.0,).partial.md
@@ -25,9 +25,12 @@ The following lists describe some of the common stages that behaviors can be bui
 
  * Operation specific processing: There is a dedicated stage for each bus operation (e.g. Send, Publish, Subscribe, ...). Behaviors can use one of the following contexts: `IOutgoingSendContext`, `IOutgoingPublishContext`, `IOutgoingReplyContext`, `ISubscribeContext`, `IUnsubscribeContext`. Subscribe and Unsubsscribe are not shown on the diagram above.
  * Outgoing Logical Message: Behaviors on this stage have access to the message which should be sent. Use `IOutgoingLogicalMessageContext` in a behavior to enlist in this stage.
- * Audit: In here behaviors have access to the message to be audited/sent to the audit queue and audit address. Behaviors should use `IAuditContext` to enlist in this stage.
  * Outgoing Physical Message: Enables to access the serialized message. This stage provides `IOutgoingPhysicalMessageContext` to it's behaviors.
 
+
+### Other Pipeline Stages
+
+ * Audit: Behaviors in the Audit stage have access to the message to be audited/sent to the audit queue and audit address. Behaviors should use `IAuditContext` to enlist in this stage.
 
 
 ## Stage Connectors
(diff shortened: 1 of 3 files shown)


mikeminutillo:
Should we include an SVG version of this map on this page?

pipeline-contexts-v4

mikeminutillo:
So it turns out you can export from Google Slides as SVG but it exports everything as paths (even the text) so it ends up being nearly 200K in size. I tried redoing it using graphviz but it's really hard to have any control over the layout. I'll see how difficult it is to replicate it using something like InkScape.

SimonCropp:
@mikeminutillo need any help with this ?

mikeminutillo:
I would appreciate a review by the @Particular/nservicebus-maintainers to see if I've captured everything correctly. Are you happy to document all of the behavior stages? I've included Transport Receive, Dispatch, and Batch Dispatch and implied that you probably don't want to write or replace behaviors in there if you don't know what you're doing. Do any of our downstreams use these contexts?

Also, do we want to document `StageForkConnector`? We don't use it any more. Should it be obsoleted immediately?

mikeminutillo:
I am wrong about `StageForkConnector`. We *do* kind of use it. The Transport Receive to Incoming Physical Message connector implements the interface version. It does not rely on the base class which is public deliberately there to enable customers to more easily implement the interface.

SimonCropp:
ok in the interest of "this is incrementally better" i am going to merge this. if anyone sees anything outstanding please raise a new PR

nservicebus/pipeline/steps-stages-connectors_extra_core_[6.0,).partial.md
```diff
@@ -3,24 +3,45 @@ NServiceBus Version 6 further splits the existing pipelines into smaller composa
...
+
+### Optional Pipeline Stages
 
+ * Audit: Behaviors in the Audit stage have access to the message to be audited/sent to the audit queue and audit address. Behaviors should use `IAuditContext` to enlist in this stage. This stage is only entered if [Message Auditing](/nservicebus/operations/auditing.md) is enabled. 
+ * Forwarding: Behaviors in the Forwarding stage have access to the message to be sent to the forwarding queue and the address of the forwarding queue. Behaviors should use `IForwardingContext` to enlist in this stage. This stage is only entered if [Message Forwarding](/nservicebus/messaging/forwarding.md) is enabled.
+
+
+### Other Pipeline Stages
+
+NOTE: These stages are documented for the sake of completeness. Replacing behaviors in these stages is not recommended without a throrough understanding of the entire pipeline.
```
danielmarbach: Spelling in throrough

PR #2142, opened 2016-11-01
Command/event definitions and comparison

DavidBoike:
This PR expands the "short" definitions of commands/events a bit, and moves them to includes so that they can be included in tutorials or elsewhere.

Also, a table of command/event comparisons developed for the intro tutorial.

### DIFF
--- a/nservicebus/definition-command.include.md
+++ b/nservicebus/definition-command.include.md
@@ -1,0 +1,1 @@
+A [**command**](/nservicebus/messaging/messages-events-commands.md#command) is a message that is used to request that an action should be taken. It can be sent from one or more senders, and is processed by a single logical receiver.
--- a/nservicebus/definition-command.include.md
+++ b/nservicebus/definition-command.include.md
@@ -1,1 +1,1 @@
-A [**command**](/nservicebus/messaging/messages-events-commands.md#command) is a message that is used to request that an action should be taken. It can be sent from one or more senders, and is processed by a single logical receiver.
+A [**command**](/nservicebus/messaging/messages-events-commands.md#command) is a message that is used to request that an action should be taken. It can be sent from one or more endpoints, and is processed by a single logical endpoint.
--- a/nservicebus/definition-event.include.md
+++ b/nservicebus/definition-event.include.md
@@ -1,0 +1,1 @@
+An [**event**](/nservicebus/messaging/messages-events-commands.md#event) is a message that is used to communicate that some action has taken place. It is published from a single logical sender, and is processed by (potentially) many receivers.
--- a/nservicebus/definition-event.include.md
+++ b/nservicebus/definition-event.include.md
@@ -1,1 +1,1 @@
-An [**event**](/nservicebus/messaging/messages-events-commands.md#event) is a message that is used to communicate that some action has taken place. It is published from a single logical sender, and is processed by (potentially) many receivers.
+An [**event**](/nservicebus/messaging/messages-events-commands.md#event) is a message that is used to communicate that some action has taken place. It is published from a single logical endpoint, and is processed by potentially many subscribing endpoints.
--- a/nservicebus/messaging/command-event-comparison-table.include.md
+++ b/nservicebus/messaging/command-event-comparison-table.include.md
@@ -1,0 +1,9 @@
+|   | Commands | Events |
+|---|:--------:|:------:|
+| Marker Interface | `ICommand` | `IEvent` |
+| Logical Senders | Many | 1 |
+| Logical Receivers | 1 | Many (or none) |
+| Purpose | "Please do something" | "Something has happened" |
+| Naming (Tense) | Imperative | Past |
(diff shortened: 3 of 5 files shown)


nservicebus/definition-event.include.md
```diff
@@ -0,0 +1 @@
+An [**event**](/nservicebus/messaging/messages-events-commands.md#event) is a message that is used to communicate that some action has taken place. It is published from a single logical sender, and is processed by (potentially) many receivers.
```
indualagarsamy: >It is published from a single logical sender

events can be published from receivers too.  It'd be better to say single logical endpoint instead of saying a sender or a receiver.

indualagarsamy: >and is processed by (potentially) many receivers.

remove the parenthesis for potentially. adds no value. harder to read.  Also change receivers to endpoints.

nservicebus/messaging/messages-events-commands.md
```diff
@@ -44,6 +44,12 @@ Used to communicate that some action has taken place. An *Event* should be _publ
 
 Note: For reply messages in a request and response pattern, use `IMessage` since these replies are neither a Command nor an Event.
 
+### Command/Event Comparison
+
+Here is a quick side-by-side comparison of some of the differences between commands and events:
```
SimonCropp: Here is a comparison of some of the differences between commands and events:

nservicebus/messaging/command-event-comparison-table.include.md
```diff
@@ -0,0 +1,9 @@
+|   | Commands | Events |
+|---|:--------:|:------:|
+| Marker Interface | `ICommand` | `IEvent` |
+| Logical Senders | Many | 1 |
```
indualagarsamy: Are logical senders and receivers new defined terms? 
A logical sender or a logical receiver can publish an event. From that perspective, this is confusing.

nservicebus/messaging/messages-events-commands.md
indualagarsamy: The differences laid out as a table is hard to understand and will be harder still for a person that is new to these terms.  Rather than having a comparison table, these must be included as part of the command and event definition where it is much easier to read. And also gives a whole lot of more context into what a command is or what an event is.

nservicebus/definition-command.include.md
```diff
@@ -0,0 +1 @@
+A [**command**](/nservicebus/messaging/messages-events-commands.md#command) is a message that is used to request that an action should be taken. It can be sent from one or more senders, and is processed by a single logical receiver.
```
indualagarsamy: >It can be sent from one or more senders, and is processed by a single logical receiver.

should be:
It can be sent from one or more endpoints, and is processed by a single logical endpoint.

nservicebus/messaging/command-event-comparison-table.include.md
DavidBoike: @indualagarsamy I think what you're saying about receivers publishing is confusing. :wink: What I'm really trying to get at here is the rules for a *single* event. So yeah, when an endpoint receives a message, it can turn around and publish something else. But if I'm talking about the OrderPlaced event, it has one logical sender and multiple logical receivers. Can you help me make that more clear?

DavidBoike:
My pending commit is going to lose a comment from @indualagarsamy:

> The differences laid out as a table is hard to understand and will be harder still for a person that is new to these terms. Rather than having a comparison table, these must be included as part of the command and event definition where it is much easier to read. And also gives a whole lot of more context into what a command is or what an event is.

The text that probably makes the table make sense comes from [this section of the NSB101 course](https://github.com/Particular/docs.particular.net/blob/nsb101-with-renderer/tutorials/nservicebus-101/lesson-4/tutorial.md#what-is-an-event). I need help to refactor this if they're going to share includes.

My concern is that the tutorial needs to be very basic, only covering the concepts that make sense for someone just starting to learn NServiceBus. The [bullet lists here](https://docs.particular.net/nservicebus/messaging/messages-events-commands#command) hint at too many advanced topics (e.g. best practice enforcement, Gateway) that have no place in an intro tutorial.

Would it make sense to take the [two big paragraphs above the table in the tutorial](https://github.com/Particular/docs.particular.net/blob/nsb101-with-renderer/tutorials/nservicebus-101/lesson-4/tutorial.md#what-is-an-event) as paragraphs in [Messages/Events/Commands](https://docs.particular.net/nservicebus/messaging/messages-events-commands) that go above the bullet lists, and then introduce the bullet lists with something like "Additionally, commands: (bullet, bullet bullet)"?

DavidBoike:
The more I think about this I'm not entirely sure it's a good idea to have this stuff as snippets. See https://github.com/Particular/OptimizeTrialExtensions/issues/133.

weralabaj:
@DavidBoike my 2 cents, I wouldn't bother with tweaking content to have identical definitions in both places. It's not something that's likely to change, so getting out of sync is very improbable. On the other hand, as you said context matters. IMO it's better to have duplicated definitions that fit their contexts.

DavidBoike:
@weralabaj That's where my views have coalescing to as well. I'm going to close this PR for now.

PR #2143, opened 2016-11-02
[WIP] Azure cloud showcase

mikeminutillo:
As a part of the V6 Launch we want to have an [Azure Cloud Showcase Sample](https://github.com/Particular/V6Launch/issues/102) (mirroring the existing On-Premise Showcase). 

## Notes
- Had to drop the array from the persisted saga data as the [Azure Persistence doesn't allow for it](https://github.com/Particular/NServiceBus.Persistence.AzureStorage/blob/develop/src/NServiceBus.Persistence.AzureStorage/SagaPersisters/DictionaryTableEntityExtensions.cs#L112).
- Code can now run completely offline


samples/show-case/cloud-azure/Core_6/DefaultStartupProjects.txt
```diff
@@ -0,0 +1,5 @@
+Store.ContentManagement\Store.ContentManagement.csproj
```
SimonCropp: why is this file required?

mikeminutillo: DefaultStartupProjects.txt or that particular csproj?

mikeminutillo:
@SeanFeldman this runs on my local Storage Emulator. Is the next step to write some deployment instructions for how to get it running in Azure using ASQ, Web sites, and Web Jobs?

SeanFeldman:
@mikeminutillo on my list of TODOs for tomorrow.

SeanFeldman:
@mikeminutillo a bit early - there are no webjobs at this point. I'll add those.

SeanFeldman:
So webjobs v2 has a better story for a local development (ie using emulator than a real storage account). Drawback - it's in beta right now. Should the sample use beta webjobs package to make local development easier, or not?

SeanFeldman:
Follow up on the previous comment: currently WebJobs library is at version 2.0.0-beta2 and will probably have a final version very soon. I suggest we take 2.0.0-beta2 as a dependency unless people disagree. Please chime in.

SeanFeldman:
@mikeminutillo to run "locally", you'll need an environment variable `AzureStorageQueue.ConnectionString` set to a real storage account (limitation of webjobs v1).
To run on Azure, `AzureWebJobsStorage` and `AzureWebJobsDashboard` have to be defined for a webapp and point to the storage account intended to be used.

SeanFeldman:
Updated the general description with items left to do.
@mikeminutillo we should sync.

Deployed project on Azure: http://azure-show-case.azurewebsites.net/

SeanFeldman:
Updated the sample to work with the local emulator. Real storage account is no longer needed.

SeanFeldman:
In the original issue we had the following items:
- Add "instance stamp" onto the generated web pages to show scaling out
- Introduce SignalR Backplane (SQL Server) to enable proper scaling
- Write instructions from running locally and deploying to azure

Adding "instance stamp" is challenging. With F5 / on-premises this makes no sense. Only in Azure were scale is possible. What do we do about this? I feel this sample has a bit of an identity crisis. It's designed to be able to run locally, but scale and deployment are not possible with F5. Is this a local sample, or a working sample **to be deployed into azure**?

SeanFeldman:
@mikeminutillo ping https://github.com/Particular/docs.particular.net/pull/2143#issuecomment-264559723

mikeminutillo:
@SeanFeldman The intent of this sample is to have something that is deployable to Azure. 

One of the goals is to minimize the amount of effort required to get it to run locally in order to get started. Is it possible to get a reasonably simple F5 experience that can be deployed to Azure and then scaled?

The "instance stamp" thing won't work locally because we only have one instance. The SignalR backplane is the same. I think the sample should include them to show the extra effort that scaling out is going to require even if they add no value for the F5 experience. 

Does that help to clarify the perspective of this sample?

weralabaj:
As discussed with @SeanFeldman last week, my opinion is that this sample should be as simple as possible (so giving F5 experience with emulator is probably the way to go). Since Sean said scaling out is a bit more involved, I think the scaling out should be a separate sample that is focused on just this one thing. Similar to https://docs.particular.net/nservicebus/msmq/sender-side-distribution and https://docs.particular.net/nservicebus/msmq/distributor/.

@Particular/docs-maintainers thoughts ☝️

SimonCropp:
> so giving F5 experience with emulator is probably the way to go  Then from this showcase sample we can link to scaling out sample, saying this is the way to go, if you're interested.

i think that is reasonable

SeanFeldman:
If F5 is all we need for this one, then it can be wrapped up. Otherwise, it requires more work.
@mikeminutillo if you don't have capacity, I can work on this one after Wednesday and wrap it up. Does that work for you?

SeanFeldman:
yay/nay?

SimonCropp:
> If F5 is all we need for this one, then it can be wrapped up. 

+1

mikeminutillo:
@SeanFeldman I'm happy to go for an F5 experience as a first step. What else needs to be done in that case?

mikeminutillo:
Actually, thinking more about this, I don't like the idea of creating a Cloud Showcase that won't run on the cloud without modification :(

SeanFeldman:
I'd be able to get back to this next year. This week is busy and next week I'm off.

samples/show-case/cloud-azure/Core_6/DefaultStartupProjects.txt
SimonCropp: DefaultStartupProjects.txt

SeanFeldman: Probably a copy from the original show case code.

mikeminutillo:
Updated the readme to include a list of Azure features and provide some guidance for making the sample deployable to Azure. I'm more or less happy with this in it's current form. A couple of quick questions for the @Particular/docs-maintainers:

1. Rather than repeat the list of features from the main sample should we just link back to it? Should that be a shared component of some kind?
2. The core showcase sample has been upgraded to include a more detailed walkthough. Should that appear on this page or should we link off to it?

@SeanFeldman if you get a chance, please check the content I added to make sure it makes sense from an Azure standpoint.

SimonCropp:
@mike i would imagine some of the shared content can be places in `.include.md` files

samples/show-case/cloud-azure/sample.md
```diff
@@ -14,6 +14,23 @@ This sample implements a fictional store that can be deployed to Azure. It is di
...
+### Azure Storage Queues Transport
+
+All endpoints in the solution communicate using the [Azure Storage Queues transport](/nservicebus/azure-storage-queues/index.md). The sample has been configured to use the Azure Storage Emulator. 
+
+
+### Azure Storage Persistence
+
+The endpoints in the solution persist data using the [Azure Storage persistence](/nservicebus/azure-storage-persistence/index.md). This persistence is used to store subscription, timeout, and saga data.
+
+NOTE: The Azure Storage persistence does not support collection types. The `ProcessOrderSaga` avoids this issue by combining a collection of `ProductIds` into a single string for persistence.
```
SeanFeldman: This does not add up with what the persistence claims to do: https://github.com/Particular/NServiceBus.Persistence.AzureStorage/pull/99

SeanFeldman:
@mikeminutillo aside from the comment for the persistence and Saga with collection, the rest looks good.

samples/show-case/cloud-azure/sample.md
mikeminutillo: Oh. It threw an exception when I first ported the sample but that was before that PR was merged. I will make sure I am running the latest persistence code and revert to the original Saga code to see if it now works.

mikeminutillo:
@SeanFeldman can you also check that the links to the Azure resources I put in are the most canonical for those resources? I really don't know the Azure doco all that well.

samples/show-case/cloud-azure/sample.md
mikeminutillo: Unfortunately it still throws an exception saying 

> The property type `String[]` is not supported in windows azure table storage

I did some investigation and it appears that the change in https://github.com/Particular/NServiceBus.Persistence.AzureStorage/pull/99 was [reverted](https://github.com/Particular/NServiceBus.Persistence.AzureStorage/commit/88d1b2b3792b1b179ecab92e76fb1e11f4da25bd) by @Scooletz

Perhaps he can provide some more context?

SeanFeldman: Ping @Scooletz

SeanFeldman:
All 4 links look good @mikeminutillo 

>  I really don't know the Azure doco all that well.

Find someone who does. Including MSFT folks.

SeanFeldman:
If there's nothing else, :shipit:

SeanFeldman:
🎆 WOOHOO! 🎉 
I swear I'll get drunk tonight celebrating the end of this work 🍾

PR #2147, opened 2016-11-02
Add component to ServiceControl and Multi-Instance Mode page

MarcinHoppe:


weralabaj:
@MarcinHoppe seems that for this page to work, the snippets should be moved from service control solution to sqlserver or you'll need to use a different component in the header (but we only split content based on major versions)

MarcinHoppe:
@weralabaj This is the way it was set up initially. The downside is that ServiceControl snippets get SQL related heading (i.e. would be labeled as `3.x - N NServiceBus.SqlServer`) which is misleading.

weralabaj:
Then it'd be better to move the snippets to the SQL Server solution. I think that makes more sense.

MarcinHoppe:
Of all available solutions, I think what we already have looks and works the best. Closing.

PR #2149, opened 2016-11-02
Add named parameters to improve readability

MarcinHoppe:
Two string parameters can be ambiguous. Explicit parameter names help
distinguish which parameter denotes endpoint, queue or schema.

/cc @DavidBoike


MarcinHoppe:
@Particular/docs-maintainers This is ready for review and merge.

PR #2151, opened 2016-11-02
Updating host profile doco to reflect host v7

andreasohlund:


andreasohlund:
@Particular/docs-maintainers need some help to figure out how to best split the info into v6/v7 sections since v7 differs quite a bit

weralabaj:
@andreasohlund use partials? they can cover even a few sections of the article, e.g. see https://docs.particular.net/nservicebus/scalability-and-ha/scale-out and how partials for v5 and v6 look like

andreasohlund:
@weralabaj page is now using partials. Can you give this a once over to see that I'm on the right track?

weralabaj:
@andreasohlund you missed one part of the partial files name - component :) fixed now, I'm reviewing content :)

PR #2205, opened 2016-11-04
Simplified NumberOfRetries method a bit

Aetherix:


SimonCropp:
@Aetherix i am not certain this change makes it any simpler. it also loses some of the intent. since the before snippet shows that the `int.tryparse` does not need to be called if the header does not exist

Aetherix:
Thank you Simon for the review. Very much appreciated. Does this also mean that I can assume that if the 'retries' header exists, it will always contain a valid number and never "" or null?

SimonCropp:
>  I can assume that if the 'retries' header exists, it will always contain a valid number and never "" or null?

yes this is correct

SimonCropp:
@Aetherix i added a note to clarify that https://docs.particular.net/nservicebus/messaging/headers#retries-handling-headers

Closing this. please raise a new PR if u think it needs more clarification. 

And thanks for contributing

PR #2207, opened 2016-11-06
Scheduling review

johnsimons:
Indented warning to be part of bullet point
Removed "Converting a scheduled task into a saga" section, we already point out when to move to a saga, no need to add a special section.


SimonCropp:
@indualagarsamy did u add the "to saga code"? thoughts on this PR?

nservicebus/scheduling/index.md
```diff
@@ -29,49 +30,20 @@ snippet:ScheduleTask
...
+* The scheduler API does not support scaling out the endpoint or doing a side-by-side deployment of an endpoint. When there are multiple instances of the endpoint that are running on the same machine, while using a non-broker transport such as MSMQ, or when there are scaling out the endpoint instances while using a broker transport such as RabbitMQ, these endpoint instances share the same input queue. Since each endpoint maintains its own created tasks in memory, when the specified time is up and the task is queued at the endpoint, any of the endpoint instances that are currently running can dequeue that message. If an endpoint that did not originally create this task happened to dequeue this message in order to execute it, it will not find the task in its list.  
 WARNING: This will result in the task not being executed but also not being rescheduled.
 
-
-## Converting a scheduled task into a saga
```
indualagarsamy: @johnsimons - why are you removing this section? I believe this was added when we made an announcement on the limitations of the scheduler and how users who are using the scheduler can convert to a saga instead.

indualagarsamy:
@johnsimons - For more information regarding the scheduler and why it was added see:
https://github.com/Particular/PlatformDevelopment/issues/188

nservicebus/scheduling/index.md
johnsimons: @indualagarsamy I was asked by @SimonCropp to review this page, and as part of my review I don't think we need a special section for convert scheduled task into a saga.
First, when I read the article, I found strange that we were mixing 2 different subjects.
Second, when the "Converting a scheduled task into a saga" section is about the same length as the main section, it tells me that info does not belong in there.
Third we already provide a link to sagas in "When not to use it" section.

So in summary, it is **my opinion** that we don't need that section, if you disagree, do not merge this change.

indualagarsamy: @Particular/docs-maintainers @johnsimons - I am 👎 for remove the content we have, especially given the information documented in issue: https://github.com/Particular/PlatformDevelopment/issues/188

timbussmann: I'm a bit on the fence. Showing how to write a saga replacing a task seems to be not that bad given the limitations of the scheduling. But I agree with john that this section is a bit too long for the context it stays in. Maybe we should extract this to a dedicated page?

SimonCropp:
@johnsimons @indualagarsamy @timbussmann can u get on a call and find a way forward for this one?

indualagarsamy:
@johnsimons - I am going to merge this PR. Add the relevant scheduler parts elsewhere and create a different PR for it. Thanks for your review.

Issue #2186, opened 2016-11-04
The page 'Deploying in a Windows Failover Cluster' requires a review


SimonCropp:
The page [Deploying in a Windows Failover Cluster](https://docs.particular.net/nservicebus/scalability-and-ha/deploying-to-a-windows-failover-cluster) has never been reviewed.

 * [Page review instructions](https://github.com/Particular/docs.particular.net#reviewing-a-page)
 * [Location in GitHub](https://github.com/Particular/docs.particular.net/blob/master/nservicebus/scalability-and-ha/deploying-to-a-windows-failover-cluster.md)

SimonCropp:
so this page is problematic for a few reasons

 - refers to NSB host without and mention of self hosting
 - it is confusing if the content is specific to the Distributor  or not.
 - for the content that does talk about the Distributor how would that work with the new routing
 - talks about win server 2008. does 2016 or 2008r2 change any of this. are the screen shots still correct?

@SzymonPobiega @mikeminutillo @gbiellem @janovesk @ramonsmits  all of u have experience in the various part of this. Can i ask u to meet up and form plan-of-attack to work out how to clean-up/update this page?

I suspect at the very least it should be split into more manageable pages. 

Also do a first pass to see if there is any redundant content or content we should not own. so that can be removed and make the problem smaller.

SimonCropp:
in the above context can u also have a look at this issue  https://github.com/Particular/docs.particular.net/issues/2187  ?

SzymonPobiega:
https://github.com/Particular/docs.particular.net/pull/2212 moves that content under msmq directly (not under distributor) because I can imagine a situation where a HA clustered queue is used for a non scaled-out endpoint.

Regarding the actual content, I don't have enough expertise to review/change it :-(

ramonsmits:
@SimonCropp Just before the weekend I started rebuilding my lab to Windows Server 2016. If I finish that probably this or next week that I'll know if clustering has any changes.I'll assign myself to verify this.

weralabaj:
from @ramonsmits on another issue

> Good decision to move the distributor bits under MSMQ but the fail-over guidance isn't specific to MSMQ except for the MSMQ dependency configuration although we mention MSMQ later. Should maybe this page be split in two documents? One that explains how to achieve high availability using Windows Clustering and the other about the MSMQ peculiarities?

weralabaj:
@ramonsmits I know you've been busy with Service Control lately, do you have time now to look into it?

[...]

PR #2212, opened 2016-11-07
Moved scaling out content under msmq

SzymonPobiega:
Moves the whole "Scaling out and HA" section under the MSMQ transport.


ramonsmits:
Good decision to move the distributor bits under MSMQ but the fail-over guidance isn't specific to MSMQ except for the MSMQ dependency configuration although we mention MSMQ later. Should maybe this page be split in two documents? One that explains how to achieve high availability using Windows Clustering and the other about the MSMQ peculiarities?

weralabaj:
@ramonsmits can you address the concern you raised in https://github.com/Particular/docs.particular.net/issues/2186?

ramonsmits:
@weralabaj Yes, I already assigned myself to verify if it needs an update regarding Windows Server 2016

PR #2214, opened 2016-11-07
Explicit reminder to use nuget aliases instead of components in partials

Scooletz:
Ping @SimonCropp 
Is it ok or repetitive to much?

### DIFF
--- a/README.md
+++ b/README.md
@@ -474,7 +474,7 @@ There are only rendered in the target page when the version filter matches the c
 
 Partial Convention: `filePrefix_key_nugetAlias_version.partial.md`
 
-Usually the NuGet alias will be the same as the component name. When a component has multiple NuGet aliases, a proper alias should be used, not the component name. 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -474,6 +474,8 @@ There are only rendered in the target page when the version filter matches the c
 
 Partial Convention: `filePrefix_key_nugetAlias_version.partial.md`
 
+Usually the NuGet alias will be the same as the component name. When a component has multiple NuGet aliases, a proper alias should be used, not the component name.
```
Scooletz: Yes. I think @weralabaj that I can't up with a better description that makes this differentiation explicit :(

weralabaj: How about

> Make sure to use component alias (as defined in components.yaml file) in the partial name. For most components component alias will be identical to NuGet alias, however it's not always the case, e.g. the Callbacks feature has been moved out of core package to the dedicated NServiceBus.Callbacks package.

Scooletz:
How about the updated version @weralabaj ? :P

Issue #2172, opened 2016-11-04
The page 'Non-Durable Messaging' requires a review


SimonCropp:
The page [Non-Durable Messaging](https://docs.particular.net/nservicebus/messaging/non-durable-messaging) has never been reviewed.

 * [Page review instructions](https://github.com/Particular/docs.particular.net#reviewing-a-page)
 * [Location in GitHub](https://github.com/Particular/docs.particular.net/blob/master/nservicebus/messaging/non-durable-messaging.md)

SimonCropp:
transport maintainer: can u review and mark of the various part of this

 - [ ] SQL @Particular/sqlserver-transport-maintainers https://docs.particular.net/nservicebus/messaging/non-durable-messaging#effect-on-transports-sql-server
 - [ ] MSMQ @Particular/nservicebus-maintainers https://docs.particular.net/nservicebus/messaging/non-durable-messaging#effect-on-transports-msmq
 - [ ] Azure @Particular/azure-service-bus-maintainers @Particular/azure-storage-queues-maintainers  https://docs.particular.net/nservicebus/messaging/non-durable-messaging#effect-on-transports-azure If this is ignored by azure should we log an error or throw an exception when it is detected?
 - [ ] Rabbit @Particular/rabbitmq-transport-maintainers https://docs.particular.net/nservicebus/messaging/non-durable-messaging#effect-on-transports-rabbitmq

SeanFeldman:
Updated for ASB & ASQ - info is correct

SimonCropp:
@SeanFeldman thanks. And the error/throw question I had?

SeanFeldman:
@SimonCropp throwing probably not, logging an error sounds like a good idea, 👍

SimonCropp:
@SeanFeldman want me to raise an issue?

SeanFeldman:
@SimonCropp https://github.com/Particular/NServiceBus.AzureServiceBus/issues/416

SimonCropp:
@SeanFeldman same case for storage queues?

SeanFeldman:
Would make sense. Could you please raise it @SimonCropp ? (need to sign off 😃)

andreasohlund:
We recently debated this in the core

https://github.com/Particular/NServiceBus/issues/4162

and the action point decided was to

[Move support for Express/NonDurable messages down to the transport level](https://github.com/Particular/NServiceBus/issues/4252)

Is there still value to update ASB/ASQ in the interim?

[...]

PR #2218, opened 2016-11-07
Clarify limitations of msmq express messages

andreasohlund:
Related to #2172


andreasohlund:
@ramonsmits can you give this a once over?

PR #2228, opened 2016-11-10
UOW sample using the pipeline

andreasohlund:
Needs a review and also a discussion about how to dispose the session (now leaning on the container)

Related to https://github.com/Particular/NServiceBus/issues/4281

### DIFF
--- a/samples/pipeline/unit-of-work/Core_6/Endpoint/MySession.cs
+++ b/samples/pipeline/unit-of-work/Core_6/Endpoint/MySession.cs
@@ -3,9 +3,13 @@ using System.Collections.Generic;
 using System.Threading.Tasks;
 using NServiceBus.Logging;
 
-class MySession : IMySession, IDisposable
+class MySession :
+    IMySession,
+    IDisposable
 {
-    readonly string tennant;
+    string tennant;
+    ILog log = LogManager.GetLogger<MySession>();
+    List<object> entities = new List<object>();
 
     public MySession(string tennant)
     {
@@ -25,7 +29,7 @@ class MySession : IMySession, IDisposable
     public Task Commit()
     {
         var entitiesStored = string.Join(",", entities);
-        Logger.Info($"{entitiesStored} stored in tennant database: {tennant}DB by session {GetHashCode()}");
+        log.Info($"{entitiesStored} stored in tennant database: {tennant}DB by session {GetHashCode()}");
         return Task.FromResult(0);
     }
 
@@ -36,8 +40,4 @@ class MySession : IMySession, IDisposable
 
     public override string ToString() => $"Session for {tennant}(instance: {GetHashCode()})";
 
-    List<object> entities = new List<object>();
-
-
-    ILog Logger = LogManager.GetLogger<MySession>();
 }
--- a/samples/pipeline/unit-of-work/Core_6/Endpoint/MySession.cs
+++ b/samples/pipeline/unit-of-work/Core_6/Endpoint/MySession.cs
@@ -1,9 +1,9 @@
 using System;
 using System.Collections.Generic;
 using System.Threading.Tasks;
+using NServiceBus.Logging;
 
-class MySession : IMySession,
-    IDisposable
+class MySession : IMySession, IDisposable
 {
     readonly string tennant;
 
@@ -25,7 +25,7 @@ class MySession : IMySession,
     public Task Commit()
     {
         var entitiesStored = string.Join(",", entities);
-        Console.Out.WriteLine($"{entitiesStored} stored in tennant database: {tennant}DB by session {GetHashCode()}");
+        Logger.Info($"{entitiesStored} stored in tennant database: {tennant}DB by session {GetHashCode()}");
         return Task.FromResult(0);
     }
 
@@ -37,4 +37,7 @@ class MySession : IMySession,
     public override string ToString() => $"Session for {tennant}(instance: {GetHashCode()})";
 
     List<object> entities = new List<object>();
+
+
+    ILog Logger = LogManager.GetLogger<MySession>();
 }
--- a/samples/pipeline/unit-of-work/Core_6/Endpoint/MySession.cs
+++ b/samples/pipeline/unit-of-work/Core_6/Endpoint/MySession.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Threading.Tasks;
 
 class MySession : IMySession,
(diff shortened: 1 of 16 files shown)


andreasohlund:
@DavidBoike @SimonCropp @timbussmann can you give this a first pass

samples/pipeline/unit-of-work/Core_6/Endpoint/Program.cs
```diff
@@ -0,0 +1,137 @@
...
+                await endpointInstance.SendLocal(new MyMessage());
+            }
+        }
+
+        await endpointInstance.Stop()
+            .ConfigureAwait(false);
+    }
+}
+
+class MyUowBehavior : Behavior<IIncomingPhysicalMessageContext>
```
timbussmann: isn't this scenario something which you can already achieve with the current UoW interface? Do we need to show how to access data from the context?

andreasohlund: Duh, indeed, I'll update to make use of header/message to create the session

andreasohlund:
@timbussmann pushed https://github.com/Particular/docs.particular.net/pull/2228/commits/25efc81407aebd4eaea9fc60380f5a9fff28fd71 that makes use of headers when opening the uow.

The open methods looks kind of funny, do you have better ideas?

samples/pipeline/unit-of-work/Core_6/Endpoint/Program.cs
```diff
@@ -0,0 +1,154 @@
...
+        return Task.FromResult(0);
+    }
+
+    public Task Open(string tennant)
+    {
+        return Task.FromResult(0);
+    }
+}
+
+class SessionFactory
```
DavidBoike: Not using this currently. Probably should register this as a SingleInstance. Then the registration for session has a choice whether to use the builder to build the SessionFactory, or use it directly in the closure. Not sure how to illustrate that or which approach is better?

samples/pipeline/unit-of-work/Core_6/Endpoint/Program.cs
DavidBoike: Shouldn't the behavior be stateless and not use DI? Then within the Invoke method, use context.Builder to build the UoW-scoped session.

samples/pipeline/unit-of-work/Core_6/Endpoint/Program.cs
andreasohlund: I'll remove for now

andreasohlund:
@danielmarbach how do we handle `IDisposable` in this context? (asp. given how async makes it much trickier)

andreasohlund:
@DavidBoike addressed your comment. Still don't like the `Open` hack but I don't see another way to connect the message/headers with the session for the given pipeline instance?

DavidBoike:
A lot is balled up in what constitutes multi-tenancy right? Is multi-tenancy a SiteId on all queries? Or does it mean a different database? If the latter, and you were using NHibernate, you'd probably have a dictionary of session factories right?

samples/pipeline/unit-of-work/Core_6/Endpoint/Program.cs
```diff
@@ -0,0 +1,142 @@
...
+}
+
+class MyUowBehavior : Behavior<IIncomingPhysicalMessageContext>
+{
+
+    public override async Task Invoke(IIncomingPhysicalMessageContext context, Func<Task> next)
+    {
+
+        var tennant = context.MessageHeaders["tennant"];
+        var session = context.Builder.Build<IMySession>();
```
danielmarbach: Agree with @DavidBoike 

I'd say the only dependency it should have is a dependency to a session factory which is a singleton. Then it explicitly creates a sessions and adds it to the extension part of the context.

samples/pipeline/unit-of-work/Core_6/Endpoint/Program.cs
```diff
@@ -0,0 +1,142 @@
...
+    {
+        return Task.FromResult(0);
+    }
+}
+
+class MyMessageHandler1 : IHandleMessages<MyMessage>
+{
+    readonly IMySession session;
+
+    public MyMessageHandler1(IMySession session)
```
danielmarbach: Remove the ctor injection and access it via custom extension method over the message handler context

andreasohlund: I added it on purpose since the user asked for it but I guess in v6 using the context is the way we recommend?

danielmarbach: Well who owns the state? The handler or the pipeline execution? In my opinion it is the pipeline execution. Therefore the state should be exposed only via context.

andreasohlund:
Pushed an update that exposes the session on the context instead

andreasohlund:
@SimonCropp are you happy with the general direction of this? (I'll complete the .md and make it ready to pull in)

andreasohlund:
Added sample description. This one is now ready for review

samples/pipeline/unit-of-work/Core_6/Endpoint/MySession.cs
```diff
@@ -0,0 +1,40 @@
...
+    public Task Store<T>(T entity)
+    {
+        entities.Add(entity);
+        return Task.FromResult(0);
+    }
+
+    public Task Commit()
+    {
+        var entitiesStored = string.Join(",", entities);
+        Console.Out.WriteLine($"{entitiesStored} stored in tennant database: {tennant}DB by session {GetHashCode()}");
```
SimonCropp: write to a logger

samples/pipeline/unit-of-work/Core_6/Endpoint/Program.cs
```diff
@@ -0,0 +1,49 @@
...
+    {
+        Console.Title = "Samples.UnitOfWork.Endpoint";
+
+        var endpointConfiguration = new EndpointConfiguration("Samples.UnitOfWork.Endpoint");
+        endpointConfiguration.UsePersistence<InMemoryPersistence>();
+        endpointConfiguration.EnableInstallers();
+        endpointConfiguration.SendFailedMessagesTo("error");
+
+        #region configuration
+        endpointConfiguration.RegisterComponents(r => r.ConfigureComponent(b => new MySessionProvider(), DependencyLifecycle.SingleInstance));
```
andreasohlund: I could skip DI here and just pass the provider to the behavior?

PR #2231, opened 2016-11-10
[WIP] ASP configuration clarified

Scooletz:
This removes a few versioned related notes and makes the configuration.md less version dependent (there's one note at the end, but I'm not going to push a partial for one config value).


Scooletz:
@tmasternak could you take a look before passing to @Particular/docs-maintainers ?

tmasternak:
LGTM

PR #2235, opened 2016-11-11
Remove frequency from release policy

timbussmann:
It seems to me that the release frequency is not something we are following as described, so it probably makes sense to remove that section from the table?


SimonCropp:
@andreasohlund @udidahan given both of u worked on this, can u comment

andreasohlund:
What value do we see for our user by providing the frequency? 

I don't see much value esp. since the frequency might be different for downsteams(thinking about majors)

I'm good with removing

PR #2236, opened 2016-11-11
Mention behavior lifetime

timbussmann:
@andreasohlund @Particular/docs-maintainers thoughts? Should the snippet maybe have a sample of to use the "DI registration" to have more context on that warning?

### DIFF
--- a/nservicebus/pipeline/manipulate-with-behaviors_registernew_core_[6.0,).partial.md
+++ b/nservicebus/pipeline/manipulate-with-behaviors_registernew_core_[6.0,).partial.md
@@ -7,4 +7,4 @@ Behaviors can also be registered from a `Feature` as shown below:
 
 snippet:RegisterBehaviorFromFeature
 
-WARNING: Behaviors are instantiated once and that instance is then reused on every invocation of the pipeline, even when registering the behavior to be created by a Dependency Injection Container. Never use dependency injection for dependencies configured with an *InstancePerUnitOfWork* scope.
+WARNING: Behaviors are only created once and the same instance is reused on every invocation of the pipeline. Consequently, every behavior dependency will also behave as a singleton, even if a different option was specified when registering it in the DI container.
--- a/nservicebus/pipeline/manipulate-with-behaviors_registernew_core_[6.0,).partial.md
+++ b/nservicebus/pipeline/manipulate-with-behaviors_registernew_core_[6.0,).partial.md
@@ -7,4 +7,4 @@ Behaviors can also be registered from a `Feature` as shown below:
 
 snippet:RegisterBehaviorFromFeature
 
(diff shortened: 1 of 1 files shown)


nservicebus/pipeline/manipulate-with-behaviors_registernew_core_[6.0,).partial.md
```diff
@@ -5,4 +5,6 @@ snippet:RegisterBehaviorEndpointConfiguration
 
 Behaviors can also be registered from a `Feature` as shown below:
 
-snippet:RegisterBehaviorFromFeature
\ No newline at end of file
+snippet:RegisterBehaviorFromFeature
+
+WARNING: Behaviors are only created once and the same instance is reused on every invocation of the pipeline, even when registering the behavior to be created by a Dependency Injection Container. Never use dependency injection for dependencies configured with an *InstancePerUnitOfWork* scope.
```
andreasohlund: Would it be better to say:

Behaviors MUST only depend on components configured as singletons (link to DI doco)  since they are created once .....

?

timbussmann: Hm I don't think that statement holds up. E.g. you can define a stateless dependency in transient/instancePerCall scope and that works just fine.

andreasohlund: Wouldn't the same be true for InstancePerUnitOfWork?

andreasohlund: >but the main risk is probably with unitofwork scope since that is mostly used to share state across 

Agreed, but doesn't the inverse apply as well? Ie that you would use InstancePerCall to avoid sharing state?

So anything other than singletons would be a potential bug?

andreasohlund: >Every behavior dependency will behave as a singleton, regardless of the registered scope.

:shipit:

weralabaj: IMO it's still not clear to understand. How about:

WARNING: Behaviors are only created once and the same instance is reused on every invocation of the pipeline. Consequently, every behaviour dependency will also behave as a singleton, even if a different option was specified when registering it in the DI container.

I'm still not sure why you say this though:
>  Never use dependency injection for dependencies configured with an *InstancePerUnitOfWork* scope.
 
What should they use in such situation?

SimonCropp:
can we simplify this PR in some way by linking here https://docstest.particular.net/nservicebus/containers/#dependency-lifecycle-singleinstance  ?

timbussmann:
used @weralabaj proposal on the warning text.
@SimonCropp I don't think that this link provides much value related to that warning.

PR #2250, opened 2016-11-16
[WIP] test mermeid diagrams

weralabaj:
Requires deploying https://github.com/Particular/DocsEngine/commit/dc8e7060506e35419057a564c42bcb33c87845cc.


weralabaj:
Examples:

ex1

weralabaj:
For more complex diagrams it might not be ideal, I could style nodes, but the diagram is of one type (left->right, top-> down, etc.). I'm not sure how to make complex diagrams like pipeline stages or if it's even possible (seems not).

@SzymonPobiega @SimonCropp ☝️ thoughts?

Should I look for another tool or would that be good enough for simple diagrams? But then what about consistency?

weralabaj:
Personally I think it might be better to stick to one tool (shared LucidChart account?) and not use those diagrams.

SimonCropp:
@weralabaj i am no so sure u should rule it out.  I think that pipeline diagram is difficult to express since it is too complex. if we split it into incoming and outgoing we get something like this 

### outgoing

```
graph LR
UC[Initiating<br>User Code]



subgraph Outgoing Pipeline
Outgoing{Outgoing}
OP[Outgoing<br>Publish]
OS[Outgoing<br>Send]
OR[Outgoing<br>Reply]
OLM[Outgoing<br>Local<br>Message]
OPM[Outgoing<br>Physical<br>Message]
Routing[Routing]
end


subgraph Dispatch
Dispatch{Dispatch}
Transport[Transport]
BD[Batch<br>Dispatch]
ID[Immediate<br>Dispatch]
end

UC --> Outgoing
Outgoing --> OP
Outgoing --> OS
Outgoing --> OR
OP --> OLM
OS --> OLM
OR --> OLM
OLM --> OPM
OPM --> Routing
Routing --> Dispatch

ID--> Transport
Dispatch --> BD
Dispatch --> ID
BD --> Transport
```

image

### Incoming

```
graph LR

Transport[Transport]

subgraph Incoming Pipeline
TR[Transport<br>Receive]
IH[Invoke<br>Handler]
ILM[Incoming<br>Logical<br>Message]
IPM[Incoming<br>Physical<br>Message]

subgraph Ancillary Actions
Forwarding[Forwarding]
Audit[Audit]
Dispatch
end
end

AncillaryTransport[Transport]


RUC[Receiving<br>User Code]


Transport --> TR
IPM -.-> Forwarding
IPM -.-> Audit
ILM --> IH
IPM --> ILM
TR --> IPM
IH --> RUC

Dispatch --> AncillaryTransport
Audit--> Dispatch
Forwarding --> Dispatch
```

image


also mermaid give us the ability to make the nodes links to actual doco

@mikeminutillo thoughts seeing as u did the initial diagram?

SimonCropp:
@weralabaj also lucid chart (or other external service) has several other problems

 - more complex workflow. export + copy + paste, or some kind of deep linking into diagram shared on the external service
 - dependency on a 3rd party service (possibly paid) to create+edit diagrams
 - difficulty in doing diffs. either we are diff'ing a binary or we need to look at the change in an external service
 - mismatched version control. if we point to external service to get the image we now have one version history of the content in GH and another in the external system. so reverts and viewing history now become much more difficult 
- if we add more binaries into the docs repo we increase the size of the repo at a much higher velocity.

mikeminutillo:
> also mermaid give us the ability to make the nodes links to actual doco

The SVG diagram can do that too.

I love a good DSL. You could get a similar effect using something like Graphviz at compile time to generate SVGs. I wasn't able to do that with the Pipeline Map because controlling the layout can be difficult. Doing the split as suggested by @SimonCropp allows you to do this:

```dot
digraph G {
  rankdir = LR;
  node [shape=rectangle;style=filled;color=lightgrey];


  UC [label = "Initiating\nUser Code"]

  subgraph cluster_Outgoing {
      label = "Outgoing Pipeline";
      style=filled;
      color="#fdfd96";
      Outgoing [shape=diamond];
      OP [label = "Outgoing\nPbublish"];
      OS [label = "Outgoing\nSend"];
      OR [label = "Outgoing\nReply"];
      OLM [label = "Outgoing\n\Local\nMessage"];
      OPM [label = "Outgoing\n\Physical\nMessage"];
      Routing;
  }

  subgraph cluster_Dispatch {
    label="Dispatch";
      style=filled;
      color="#fdfd96";
    Dispatch [shape=diamond];
    ID [label = "Immediate\nDispatch"];
    BD [label = "Batch\nDispatch"];
    Transport;
  }

  UC -> Outgoing -> OP, OS, OR -> OLM -> OPM -> Routing -> Dispatch -> ID, BD -> Transport
}
```

which generates an SVG that looks much like the example given above:

image

NOTE: Github won't let me upload an SVG directly into a comment so this is a screenshot of it.

An SVG is rendered natively by the browser without a special JS file. It also scales up and down much nicer than a regular image (although it's possible that mermaid will just generate the image at the appropriate size so this is less relevant). 

It does require a bit more styling info.

SzymonPobiega:
I prefer DSLs embedded in the doco for the reasons @SimonCropp outlined over any external graphical editor. I like the simplicity of Mermaid.

weralabaj:
Hm... and also we can always make contributions to mermeid :P

@mikeminutillo Mermeid works basically the same. Plus we can have custom styles by providing .css. It uses https://d3js.org/, so I guess it should be possible to extend as D3 is pretty powerful.

I think overall that gives more options than GraphViz, while is very similar in how it works and syntax (see here https://knsv.github.io/mermaid/#flowcharts-basic-syntax and https://knsv.github.io/mermaid/#sequence-diagrams).

So are we going with it, at least for now?

weralabaj:
Btw in such a case, would it be better to stick to just a single library ultimately?

weralabaj:
OK, so I'll merge Szymon's diagram for now only and keep the pipeline diagram as is. I'll open a separate issue for this, as we'll have more diagrams to rework.

PR #2252, opened 2016-11-16
warn about interfaces with direct routing topology

adamralph:
Fixes https://github.com/Particular/NServiceBus.RabbitMQ/issues/268

### DIFF
--- a/nservicebus/rabbitmq/routing-topology.md
+++ b/nservicebus/rabbitmq/routing-topology.md
@@ -54,7 +54,7 @@ Adjust the conventions for exchange name and routing key by using the overload:
 
 snippet:rabbitmq-config-usedirectroutingtopologywithcustomconventions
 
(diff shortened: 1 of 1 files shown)


adamralph:
@Particular/docs-maintainers there seems to be something strange about the deep link.

In markdown it works: https://github.com/Particular/docs.particular.net/blob/master/nservicebus/messaging/messages-events-commands.md#marker-interfaces

but in the rendered site it doesn't: https://docs.particular.net/nservicebus/messaging/messages-events-commands#marker-interfaces

For some reason the anchor in the rendered site is https://docs.particular.net/nservicebus/messaging/messages-events-commands#defining-messages-marker-interfaces

Is this because the heading is third level (`###`) instead of second level (`##`)? Should we link to the higher level heading instead?

SimonCropp:
@adamralph since u can have the same lower level heading text  multiple headings the heading anchor is the concatenation of the heading tree. So u can link to the lower. Level heading but it needs to be fully qualified

weralabaj:
@adamralph To ensure the anchor links are unique within a page we combine text from parent headers with the current header. The headers don't have to be unique within a page, but after combining with parent headers they should. E.g. it's very useful here https://docs.particular.net/nservicebus/msmq/operations-scripting, where we have multiple "C#" headers, yet the path is always unique.

So either you use the link from the rendered site or link to the higher level, both are good :)

weralabaj:
BTW that "nested" link is generated when we render HTML, so that might be the reason why it works for .md, because Github renders HTML in a standard way, without our smart customization ;)

adamralph:
@SimonCropp @weralabaj thanks for the explanations. It makes sense now. I'll send another PR to fix the link.

PR #2253, opened 2016-11-16
NSB101 snippet fixes

DavidBoike:


tutorials/nservicebus-101/lesson-2/Solution/ClientUI/Program.cs
```diff
@@ -39,6 +37,8 @@ await endpointInstance.Stop()
 
         static async Task RunLoop(IEndpointInstance endpointInstance)
         {
+            var logger = LogManager.GetLogger<Program>();
```
SimonCropp: @DavidBoike this goes against our advice that loggers should be static. People should never do this in a production system

DavidBoike:
Sorry, I had merged thinking these were very minor fixes. In this case they are essentially single instances within that class, but you're right, people could get that idea. I'll have to prepare a new PR that goes back to static loggers and covers creating it within the tutorial. (That's what was missing before.)

SimonCropp:
@DavidBoike thanks. u can link here in the new PR https://docs.particular.net/nservicebus/logging/#when-to-configure-logging

SimonCropp:
sorry wrong link https://docs.particular.net/nservicebus/logging/usage

PR #2254, opened 2016-11-16
Restore static loggers, within snippet, add note

DavidBoike:
By keeping the static logger as part of the existing snippet, it reduces the overall number of steps, which should make it easier to follow. The only downside is the static logger winds up in the middle of the file, which makes my OCD twitch just a little bit, but I think the tradeoff might be worth it.

@SimonCropp you OK with that?

Added the link to https://docs.particular.net/nservicebus/logging/usage where the first instance of a logger appears.


SimonCropp:
LGTM

Issue #2263, opened 2016-11-21
broken link in 5to6/distributor


SimonCropp:
so the text "Configure it to enlist it with the distributor" here
https://docs.particular.net/nservicebus/upgrades/5to6/distributor#general-upgrade-scenario

links to https://docs.particular.net/nservicebus/upgrades/5to6/distributor#enlist-version-6-endpoints-with-a-distributor which does not exist on that page. 

The only similar reference is here https://docs.particular.net/samples/scaleout/distributor-upgrade/#upgrade-enlisting-with-the-distributor.

so was that a document splitting problem?

SimonCropp:
ping @Particular/nservicebus-maintainers @SzymonPobiega

weralabaj:
seems so

weralabaj:
I'll work on anchor links checking today, that should help :)

SzymonPobiega:
@timbussmann can you take a look at that link? I am not sure what it should be linking to (if anything).

PR #2264, opened 2016-11-22
Fix enlisting with the distributor link

timbussmann:
fixes https://github.com/Particular/docs.particular.net/issues/2263


weralabaj:
Link fix is ok, but seems the link generation is broken - there shouldn't be "1" at the end. I'll investigate and fix the link afterwards.

weralabaj:
proposed fix https://github.com/Particular/DocsEngine/commit/91734d428db9134fecc786d8f4b919bee5f354a3, adjust the link after it gets deployed (if that happens) ;)

PR #2280, opened 2016-11-28
Mermaid

weralabaj:
Replace other diagrams with mermaid


weralabaj:
@Particular/docs-maintainers ping, doesn't look as good as old ones, though :/

PR #2288, opened 2016-11-30
[SPIKE] Updated automatic routing for the new core routing proposal.

SzymonPobiega:
Shows the new way automatic routing can be implemented by taking advantage of subscriber table approach https://github.com/Particular/NServiceBus/pull/4316. The endpoints now publish also the list of handled events allowing others to configure the publishing without using MDPS.

It also takes advantage of the change in `UnicastRoutingTable` that allows having multiple routes (provided they all share the same logical endpoint).


SzymonPobiega:
@Scooletz @timbussmann please take a look. As shown in the line counters, it simplifies the whole thing a lot, making it very similar to the file based approach with the only difference being the fact that the endpoint themselves publish the information about who handles what (instead of somebody writing the XML file).

SimonCropp:
@SzymonPobiega is this spike doco still valid?

PR #2290, opened 2016-11-30
[WIP]Multiple inheritance sample

andreasohlund:
@SimonCropp what do you think, should we pursue this?


SimonCropp:
@andreasohlund even if this only helps inform the decision about "should we support this moving forward" i think it is useful

SimonCropp:
i moved it over to xml serializer since json broke for me

andreasohlund:
Yea, json is broken

SimonCropp:
@andreasohlund need any help from @Particular/docs-maintainers to get this done?

indualagarsamy:
@andreasohlund - I have a PR to simplify Pub/Sub. Let me know how I can help you. The goal of my PR was to simplify pub/sub so I removed multiple inheritance completely. 
If we want to show message inheritance, can we do it using a separate sample that's specifically targeted? The reason is, when someone tries to evaluate our framework, almost the very first sample they tend to play with is pub-sub and I think it's important to keep that sample simple and clean. 
Also, if we separate out the multiple inheritance separately, then we can highlight the pitfalls of it. For example the serializers. Thoughts?

indualagarsamy:
@andreasohlund - here's the PR I have that's related: https://github.com/Particular/docs.particular.net/pull/2325

indualagarsamy:
@SimonCropp - thoughts on this?

SimonCropp:
@indualagarsamy this isnt related to the pub sub sample. @andreasohlund is attempting to illustrate publish by contract. not that the contracts are defined in the  subscribers.

samples/multiple-inheritance/Core_6/USQLApplication1/USQLApplication1.usqlproj
```diff
@@ -0,0 +1,38 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
SimonCropp: what is this file?

indualagarsamy:
@andreasohlund - I apologize for the confusion. @SimonCropp explained that this is a brand new sample. :) I just didn't pay attention to the folder and assumed it was the "same" pub sub sample. Either way the offer stands. Let me know if you need help in getting this done :)

andreasohlund:
@Particular/docs-maintainers pushed a draft sample description, does it make any sense?

andreasohlund:
@Particular/nservicebus-maintainers fyi this is to my understanding the reason we support interface messages

samples/multiple-inheritance/sample.md
```diff
@@ -0,0 +1,20 @@
+## Introduction to Consumer Driven Contracts
+This sample shows a [consumer driven contracts](http://martinfowler.com/articles/consumerDrivenContracts.html)(CDD) approach to messaging. The essense of consumer driven contracts is that the owner ship of the contract is inversed. Instead of the producer defining the contract consumers are now the ones defining the contract they expect and it's up to the producer to fullfill it.
+
+In the NServiceBus world message contracts is plain C# types so to honor a consumer contract the producer would just make the relevant message contract inherit from the contract type.
```
indualagarsamy: Again, for our users, Publishers of events might be better known than Producer of contract.

samples/multiple-inheritance/sample.md
```diff
@@ -0,0 +1,20 @@
+## Introduction to Consumer Driven Contracts
+This sample shows a [consumer driven contracts](http://martinfowler.com/articles/consumerDrivenContracts.html)(CDD) approach to messaging. The essense of consumer driven contracts is that the owner ship of the contract is inversed. Instead of the producer defining the contract consumers are now the ones defining the contract they expect and it's up to the producer to fullfill it.
+
+In the NServiceBus world message contracts is plain C# types so to honor a consumer contract the producer would just make the relevant message contract inherit from the contract type. 
+
+### Contracts as interfaces
+
+CDD is the main reason that `interface` messages are supported in addition to classes and the reason is that more than one consumer can provide a contract that would be satisfied by the same publisher message. If only classes where used this would not be possible since multiple inheritance isn't supported by C#. The solution is to use interfaces instead since with interfaces the publisher could implement all relevant contract types on the same message type.
```
indualagarsamy: What's CDD?

samples/multiple-inheritance/sample.md
```diff
@@ -0,0 +1,20 @@
...
+
+Assuming that the producers `MyEvent` would satisfy both Consumer1Contract and Consumer2Contract this would look like this:
```
indualagarsamy: Can we use different event names, instead of `MyEvent`. It would be great if we can name the events after a well known event, eg. in the retail vertical?

samples/multiple-inheritance/sample.md
```diff
@@ -0,0 +1,20 @@
...
+
+snippet: publisher-contracts
+
+NOTE: The limitation of this approach is two or more consumers requiring a property with the same name but different types.
+
+## Running the sample
+
+Hit F5 to run the sample and notice how both consumers recives its contract when the producer publishes the `MyEvent`.
+
+NOTE: Sharing contract types between enpoints is a larger topic and this sample is using linked files for simplicity. See the [message contacts documentation](/nservicebus/messaging/evolving-contracts.md) for more details.
```
indualagarsamy: - typo in endpoints.
- typo in contracts `message contracts documentation`

samples/multiple-inheritance/Core_6/Consumer2/Contracts/Consumer2Contract.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace Subscriber2.Contracts
+{
+    using NServiceBus;
+    public interface Consumer2Contract :
```
indualagarsamy: Can we use different event names?

samples/multiple-inheritance/Core_6/Producer/Contracts/MyEvent.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace Publisher.Contracts
+{
+    using Subscriber1.Contracts;
+    using Subscriber2.Contracts;
+
+    #region publisher-contracts
+    class MyEvent :
+        Consumer1Contract,
+        Consumer2Contract
+    {
```
indualagarsamy: A more real world event that illustrates this would be super helpful.

indualagarsamy:
@andreasohlund - i reviewed the sample draft and i feel like it doesn't answer, where this would make sense and how does it compare to the other way around. While the article points to Fowler's article, I think it makes sense to expand on it using our own terminology to try and focus on what problem it's supposed to solve and how to go about it and then what the limitations are.

samples/multiple-inheritance/Core_6/Producer/Contracts/MyEvent.cs
andreasohlund: I tried but couldn't come up with a real life use case, when we agree on https://github.com/Particular/docs.particular.net/pull/2290#issuecomment-266915497 we might find one

andreasohlund:
@indualagarsamy this is where I need some help, possibly from @udidahan since I'm not really sure.

Some kind of integration scenario perhaps where multiple services defines different contracts that "Operations" satisfy via data from the same integration perhaps?

samples/multiple-inheritance/Core_6/Consumer2/Contracts/Consumer2Contract.cs
andreasohlund: I did this on purpose since I don't have a business use case yet

samples/multiple-inheritance/sample.md
andreasohlund: See below, I couldn't come up with a good use case for this

samples/multiple-inheritance/sample.md
andreasohlund: Defined on the first line of the doc?

samples/multiple-inheritance/sample.md
andreasohlund: I chose producer/consumer since that's the lingo used in the definition of CDD. I'll add some text to explain that producer = publisher, consumer = subscriber in the NServiceBus world

andreasohlund:
@mauroservienti see https://github.com/Particular/docs.particular.net/pull/2290#issuecomment-266961872 , this might be relevant for the "IReactTo<Operations.SomeThing>" discussion we had last week?

udidahan:
Multiple inheritance comes into play for when the business comes up with a new requirement that we didn't imagine early on. 

For example, if we're modeling a zoo, we may have logic for how we handle animals that lay eggs, other logic for those that swim, and something else focused on mammals.

If we model each of those as an interface (ILayEggs, ISwim, IAmWarmBlooded), when the zoo brings in a platypus, we can have it inherit from all three and have all the existing logic just continue working without doing any special content-based routing that would otherwise have been needed:

> The platypus (Ornithorhynchus anatinus), also known as the duck-billed platypus, is a semiaquatic egg-laying mammal endemic to eastern Australia

andreasohlund:
>Multiple inheritance comes into play for when the business comes up with a new requirement that we didn't imagine early on.

@udidahan  Would you say that a sample and wording that focus on "Consumer driven contracts" is valuable or is it to generic? (or does it just muddle the water)

Given your example from above 

>Multiple inheritance comes into play for when the business comes up with a new requirement that we didn't imagine early on.

A new consumer has requirements that we didn't imagine earlier on

>For example, if we're modeling a zoo, we may have logic for how we handle animals that lay eggs, other logic for those that swim, and something else focused on mammals.If we model each of those as an interface (ILayEggs, ISwim, IAmWarmBlooded), when the zoo brings in a platypus, we can have it inherit from all three and have all the existing logic just continue working without doing any special content-based routing that would otherwise have been needed:

What would be a real life example? 

Perhaps:

We have a bank integration for credit checks and Operations exposes this as a message that the business services can decide to react on

```
class CreditCheckResultAvailable
    public string CustomerId;
    public string CreditRating;
    ... //lots of extra stuff including the users known address
```
Now some time later CustomerCare get wind of this and say hey, if the RiskService put up the dollars to perform the credit check we want in to get the address. (but we won't pay! :)  )

Fine says operations, just add a handler for `CreditCheckResultAvailable` and we're off to the races.

CustomerCare say fine, but we don't want to repeat our selves and we already have that address integration that gives us 

```
class AddressCheckResult
   public string CustomerId;
   public AddressDetails Address;
```
what can we do?

Operations says: Fine just add a handler for `IAddressDetailsAvailable` and we'll have you covered from this point on but you owe us a beer for this, ok?

`CreditCheckResultAvailable: IAddressDetailsAvailable`
`AddressCheckResult: IAddressDetailsAvailable`

```
class IAddressDetailsAvailable
   public string CustomerId;
   public AddressDetails Address;
```

Would this be correct/fit the use case for multiple inheritance?

udidahan:
I suggest we put this on-hold for now. Most users don't leverage multiple inheritance. This is complex. There are probably many higher impact things that could be done instead.

andreasohlund:
Given that the sample is functioning and exercises the feature I'd say we pull it?

andreasohlund:
And given that we haven't agreed on a business scenario yet I think the "consumer/producer" lingo is ok until we reach consensus?

PR #2291, opened 2016-12-01
Forwarding Queues only created as needed

gbiellem:
Related to https://github.com/Particular/ServiceControl/pull/852

Merge when ServiceControl v1.29 ships

### DIFF
--- a/servicecontrol/creating-config-file.md
+++ b/servicecontrol/creating-config-file.md
@@ -259,7 +259,7 @@ Default: `<ErrorQueue>.log`
 
 Version 1.28 and below created the queue specified by this setting when the service instance is installed regardless of the value of `ServiceControl/ForwardErrorMessages`.  From V1.29 the queue is only created if `ServiceControl/ForwardErrorMessages` is enabled. 
 
-NOTE: Changing the configuration file directly will not result in the queue being created. Please use the ServiceControl Management tools. 
+NOTE: Changing the configuration file directly will not result in the queue being created. Use ServiceControl Management to add or alter the forwarding queue.
 
 #### ServiceBus/AuditLogQueue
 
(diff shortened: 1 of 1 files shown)


servicecontrol/creating-config-file.md
```diff
@@ -257,6 +257,9 @@ Type: string
 
 Default: `<ErrorQueue>.log`
 
+Version 1.28 and below created the queue specified by this setting when the service instance is installed regardless of the value of `ServiceControl/ForwardErrorMessages`.  From V1.29 the queue is only created if `ServiceControl/ForwardErrorMessages` is enabled.
```
indualagarsamy: >From V1.29 the queue is only created if `ServiceControl/ForwardErrorMessages` is enabled. 

Should be `When using Versions 1.29 and above, ...`
See Version language here: 
https://github.com/particular/docs.particular.net/#version-language

servicecontrol/creating-config-file.md
```diff
@@ -266,6 +269,9 @@ Type: string
 
 Default: `<AuditQueue>.log`
 
+Version 1.28 and below created the queue specified by this setting when the service instance is installed regardless of the value of  `ServiceControl/ForwardAuditMessages`. From V1.29 the queue is only created if `ServiceControl/ForwardAuditMessages` is enabled.
```
indualagarsamy: Same as above.

servicecontrol/creating-config-file.md
gbiellem: I don't think the current text is ambiguous which is what version language guideline is trying to prevent.  Happy to change it though if you think it is

SimonCropp: @gbiellem i know it is 50/50. both "Version 1.28 and below" and "Versions 1.28 and below" are correct and unambiguous. but when we threw it out for a vote most people preferred "Versions 1.28 and below". so that is what we push for in the interest of consistency

gbiellem: No problem @SimonCropp  and @indualagarsamy 
I will change them to read "Versions 1.28 and below" and "Versions 1.29 and above"

PR #2305, opened 2016-12-06
Custom checks update

mikeminutillo:
Update sample to use 2.0.1 version of Custom Checks plugin.

@Particular/docs-maintainers version 2 of the plugin requires explicit configuration of the ServiceControl queue. Should that be called out in the [text of the sample](https://docs.particular.net/samples/custom-checks/monitoring3rdparty/)?


weralabaj:
I see that the error they get if the queue is not specified is very clear. So not critical, but I think it'll be better to include the snippet when talking about Sample project.

BTW since we're talking, I'm not sure what is Service Control queue https://docs.particular.net/servicecontrol/plugins/custom-checks#configuration-servicecontrol-queue. How do I find out what should I specify there? Could we also clarify that part of the doco?

I see this https://docs.particular.net/servicecontrol/plugins/#connecting-to-servicecontrol, but couldn't find anything about ServiceQueue specifically (what it is, etc.). Shouldn't it be mentioned here https://docs.particular.net/servicecontrol/installation?

weralabaj:
Maybe the text explaining what the queue is could be include for all the plugins, and each could have its own snippets? Seems it's very similar.

SimonCropp:
> version 2 of the plugin requires explicit configuration of the ServiceControl queue

wounldnt that justify a upgrade guide?

mikeminutillo:
> wounldnt that justify a upgrade guide?

Yep https://github.com/Particular/docs.particular.net/pull/2312

mikeminutillo:
> Maybe the text explaining what the queue is could be include for all the plugins, and each could have its own snippets? Seems it's very similar.

Yeah I think this is the way to go for the moment. I had a chat with @gbiellem about displaying it in the SC Management UI somewhere but I don't think we're ready for that just yet.

gbiellem:
@mikeminutillo  raised an issue following our conversation - https://github.com/Particular/ServiceControl/issues/868

mikeminutillo:
Crud. This relies on a snippet I just added in https://github.com/Particular/docs.particular.net/pull/2312

Once it's merged I'll rebase

SimonCropp:
#2312 is merged

mikeminutillo:
Rebased. As there are no new commits, this won't kick off the build again, right?

PR #2322, opened 2016-12-11
Changed "Go straight to runnable code" to "Samples"

udidahan:
The purpose of this PR is to put "Samples" on the same level as "Tutorials".

The "Go straight to runnable code" language pushes the reader much more strongly to the samples than to the tutorials.

### DIFF
--- a/get-started.md
+++ b/get-started.md
@@ -22,7 +22,7 @@ These articles are recommended to gain an understanding of the key concepts and 
 The [Messaging Basics Tutorial](/tutorials/nservicebus-101/) gives a guided walk through of building an NServiceBus system and introduces the basic concepts of messaging. Each lesson includes a downloadable solution with the completed exercise.
 
 
-## Go straight to runnable code
+## Samples
 
 Samples are fully functional solutions that can be run from Visual Studio. The samples allow to learn by exploring code. Each includes an explanation of the scenario, technology or concept that it illustrates.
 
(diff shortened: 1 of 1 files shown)


SimonCropp:
@udidahan i tried to think of an equivalent phrasing for tutorials that conveys the value, and failed. any ideas?

get-started.md
```diff
@@ -22,7 +22,7 @@ These articles are recommended to gain an understanding of the key concepts and
 The [Messaging Basics Tutorial](/tutorials/nservicebus-101/) gives a guided walk through of building an NServiceBus system and introduces the basic concepts of messaging. Each lesson includes a downloadable solution with the completed exercise.
 
 
-## Go straight to runnable code
+## Samples
 
 Samples are fully functional solutions that can be run from Visual Studio. The samples allow to learn by exploring code. Each includes an explanation of the scenario, technology or concept that it illustrates.
```
indualagarsamy: @SimonCropp @udidahan - How about keeping the title as "Samples", but changing the paragraph a little that explains what you're after? This way, you can keep the heading as "Samples", but follow it up with:

Samples are a great way to learn and explore NServiceBus by looking at runnable code. They are fully functional solutions that can be run from Visual Studio. Learn by exploring code. Each sample includes an explanation of the scenario and the technology or concept that it illustrates.

Thoughts?

udidahan: 👍

PR #2327, opened 2016-12-13
Make handler output visible.

timbussmann:
This sample uses `log.Info` to write the output which is stated in the docs page but the program is configured to use log level `Warn` so it will actually never show the documented output as described here: https://docs.particular.net/samples/errorhandling/#the-output

I switched to use `Console.WriteLine` because it makes the output easily visible.
* changing to log level info will drown the input in logged exceptions from immediate retries
* changing the log statements to use warn instead of info will make the logged lines hardly visible as they are the same color as the warnings from delayed retries which dumps the full stacktrace.

thoughts @Particular/docs-maintainers ?

### DIFF
--- a/samples/errorhandling/Core_4/WithDelayedRetries/MyMessageHandler.cs
+++ b/samples/errorhandling/Core_4/WithDelayedRetries/MyMessageHandler.cs
@@ -1,12 +1,10 @@
 ﻿using System;
 using System.Collections.Concurrent;
 using NServiceBus;
-using NServiceBus.Logging;
 
 public class MyMessageHandler :
     IHandleMessages<MyMessage>
 {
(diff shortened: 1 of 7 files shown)


SimonCropp:
@timbussmann LGTM. but do the version 4 and 5 not need the same change?

timbussmann:
@SimonCropp haven't had the time to check. Will try to do that. Assigning this to me.

indualagarsamy:
@SimonCropp - wasn't there a decision in docs specifically NOT to use Console.Writeline and to use logging? @timbussmann - any specific reason to move to Console.Writelines instead of logging?

indualagarsamy:
@timbussmann - my bad. note to self: read pr description :)

timbussmann:
thanks for helping out @indualagarsamy 👏

PR #2331, opened 2016-12-14
Link to Rabbit connection settings article

DavidBoike:
To avoid confusion, as the connection settings page is a little buried currently, and the existing text makes it sound like UseTransport is all that's required.

### DIFF
--- a/nservicebus/rabbitmq/index.md
+++ b/nservicebus/rabbitmq/index.md
@@ -21,6 +21,8 @@ To use RabbitMQ as the underlying transport:
 
 snippet:rabbitmq-config-basic
 
+The RabbitMQ transport requires a connection string to connect to the RabbitMQ broker. See [Connection settings](/nservicebus/rabbitmq/) for options on how to provide the connection string.
+
 
 ## Advantages and Disadvantages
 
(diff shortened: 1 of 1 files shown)


nservicebus/rabbitmq/index.md
```diff
@@ -21,6 +21,8 @@ To use RabbitMQ as the underlying transport:
 
 snippet:rabbitmq-config-basic
 
+The RabbitMQ transport requires a connection string to connect to the RabbitMQ broker. See [Connection settings](/nservicebus/rabbitmq/) for options on how to provide the connection string.
```
bording: Feels like there is one too many "RabbitMQ"s in the sentence.

bording: The link url should be `connection-settings.md`.

DavidBoike:
@bording Oops, nice catch on the self-link. Better now?

bording:
@DavidBoike I think the link could be be just "connection-settings.md" and not need the rest of the path, but I'll leave that to the @Particular/docs-maintainers to decide how they prefer it.

SimonCropp:
Indifferent to link depth. Both will work and we verify them. The relative is a little more concise, but up to you

DavidBoike:
I'll remove WIP and merge tonight during my talk.

DavidBoike:
Didn't end up making it that far in my talk. :-( Oh well.

PR #2341, opened 2016-12-19
alternate diagrams

SimonCropp:
@mikeminutillo thoughts?


mikeminutillo:
I like the sequence diagram for the provisioning part. Is it OK to switch representations between diagrams like this? i.e. Graph, Graph, Sequence, Graph

I'm not sold on having the events inside the box for the first one. I liked having them outside as it implies that they aren't as contained as commands. Anyone can subscribe to these events and take action based on it. It also provides a consistent view once you start adding other subscribers in the later diagrams

SimonCropp:
> Is it OK to switch representations between diagrams like this? i.e. Graph, Graph, Sequence, Graph

yeah not sure

Issue #1889, opened 2016-09-09
Sample for self hosting how to only run installer once


danielmarbach:
Related https://github.com/Particular/NServiceBus/issues/1703

// cc @timbussmann

SimonCropp:
dont think we need a full sample for this. should doco be enough?

SimonCropp:
@danielmarbach thoughts?

danielmarbach:
Somewhere here? http://docs.particular.net/nservicebus/operations/installers

SimonCropp:
@danielmarbach LGTM

SimonCropp:
@danielmarbach you interested in doing this doco?

danielmarbach:
@SimonCropp should I go with the approach outlined in https://github.com/Particular/NServiceBus/issues/1703#issuecomment-245869012 ?

SimonCropp:
i think show multiple approaches. one per endpoint

 * switch by command line
 * switch by machine name convention. eg many orgs have naming conventions for machines. eg all production machines are "machine1-PROD"
 * switch by computer AD membership (if a corp environment)

PR #2344, opened 2016-12-19
Installers

danielmarbach:
@SimonCropp Some spike code which shows the v6 part of the sample for https://github.com/Particular/docs.particular.net/issues/1889 

I'm very unsure about the AD part because I've never written AD queries before.

### DIFF
--- a/Snippets/Core/Core_6/Installers.cs
+++ b/Snippets/Core/Core_6/Installers.cs
@@ -2,6 +2,7 @@
 {
     using System.Threading.Tasks;
     using NServiceBus;
+    using NServiceBus.Support;
 
     class ForInstallationOnReplacement
     {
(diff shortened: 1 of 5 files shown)


danielmarbach:
Haven't received feedback so far. Will pick this up again after my leave

SimonCropp:
@gbiellem can u review this?

danielmarbach:
Rebased @gbiellem @SimonCropp

gbiellem:
@danielmarbach I'm not sure in what scenario I'd ever use SwitchInstallersByADMemberShip.  How do you see this working in practice?

Snippets/Core/Core_6/Installers.cs
```diff
@@ -17,6 +21,65 @@ await Endpoint.Start(endpointConfiguration)
...
+            #endregion
+        }
+
+        #region InstallersRunWhenNecessaryADMembershipQuery
+        static List<string> GetGroupNames(string userName, string domain)
+        {
+            var pc = new PrincipalContext(ContextType.Domain, domain);
+            var src = UserPrincipal.FindByIdentity(pc, userName).GetGroups(pc);
+            var result = new List<string>();
+            src.ToList().ForEach(sr => result.Add(sr.UserPrincipalName));
```
gbiellem: 'UserPrincipalName" will return null for groups.  So this won't work.  The attribute you want is just 'Name'

PR #2348, opened 2016-12-20
Fix TestSaga() to reflect proper usage

ksigmund:
ExpectPublish() needs to be called before WhenSagaTimesOut(), otherwise it doesn't appear to be evaluated.


andreasohlund:
Thanks @ksigmund !

I've also [raised an issue to consider changing the assert call](https://github.com/Particular/NServiceBus.Testing/issues/100) since its quite confusing when you need to mix `ExpectX` and `AssertY` in the same test.

Issue #2361, opened 2016-12-28
Feedback: Step by Step Sample - NServiceBus Samples


wyldebill:
Page: https://docs.particular.net/samples/step-by-step/?

For version 6.x of this page, under the Server section, there appears to be some copy from the version 5.x documentation.  The line that contains "The dependency injection system (which supports constructor or property injection) injects the IBus instance" is incorrect as version 6.x uses the new IMessageHandlerContext instead of the IBus property.

ramonsmits:
Thanks for the feedback @wyldebill

PR #2362, opened 2016-12-29
Resolves Issue #2361

ramonsmits:
Resolves Issue #2361


weralabaj:
@ramonsmits can you have a look? seems there's missing version of the snippet? or you need to specify versions for this article explicitly?

PR #2363, opened 2016-12-30
A sample showing how to gradually migrate saga persistence

SzymonPobiega:
Shows how to migrate saga persistences without an off-line migration procedure/script by using a pair of endpoints where existing sagas are continued to be handled by an "old" version of the endpoint while new sagas are started in the "new" version of the endpoint, using the new persistence.

It covers:
 * Routing of messages with correlation property value
 * Routing of messages with saga ID header
 * Re-routing of timeouts after the temp endpoint is decomissioned

### DIFF
--- a/samples/saga/migration/Core_6/Client/Program.cs
+++ b/samples/saga/migration/Core_6/Client/Program.cs
@@ -21,11 +21,11 @@ class Program
 
         while (true)
         {
-            Console.WriteLine("Type 'start <SagaId>' or 'ping <SagaId>'.");
+            Console.WriteLine("Type 'start <SagaId>', 'complete <SagaId>' or hit enter to exit.");
             var line = Console.ReadLine();
-            if (line == null)
+            if (string.IsNullOrEmpty(line))
             {
-                continue;
+                break;
             }
             var parts = line.Split(' ');
             if (parts.Length != 2)
@@ -43,7 +43,7 @@ class Program
                     .ConfigureAwait(false);
                 Console.WriteLine($"Sent a starting message to {sagaId}");
             }
-            else if (string.Equals(parts[0], "ping", StringComparison.OrdinalIgnoreCase))
+            else if (string.Equals(parts[0], "complete", StringComparison.OrdinalIgnoreCase))
             {
                 var correlatedMessage = new CorrelatedMessage
                 {
--- a/samples/saga/migration/Core_6/Client/Program.cs
+++ b/samples/saga/migration/Core_6/Client/Program.cs
@@ -1,0 +1,58 @@
+﻿using System;
+using System.Threading.Tasks;
+using NServiceBus;
+
+class Program
+{
+    static void Main()
+    {
+        Start().GetAwaiter().GetResult();
+    }
+
+    static async Task Start()
+    {
+        Console.Title = "Samples.SagaMigration.Client";
+
+        var config = new EndpointConfiguration("Samples.SagaMigration.Client");
+        config.UsePersistence<InMemoryPersistence>();
+        config.SendFailedMessagesTo("error");
+
+        var endpoint = await Endpoint.Start(config).ConfigureAwait(false);
+
+        while (true)
+        {
+            Console.WriteLine("Type 'start <SagaId>' or 'ping <SagaId>'.");
+            var line = Console.ReadLine();
+            if (line == null)
+            {
+                continue;
+            }
+            var parts = line.Split(' ');
+            if (parts.Length != 2)
+            {
+                continue;
+            }
+            var sagaId = parts[1];
+            if (string.Equals(parts[0], "start", StringComparison.OrdinalIgnoreCase))
+            {
+                var startingMessage = new StartingMessage
+                {
+                    SomeId = sagaId
+                };
+                await endpoint.Send("Samples.SagaMigration.Server", startingMessage)
+                    .ConfigureAwait(false);
+                Console.WriteLine($"Sent a starting message to {sagaId}");
+            }
+            else if (string.Equals(parts[0], "ping", StringComparison.OrdinalIgnoreCase))
+            {
+                var correlatedMessage = new CorrelatedMessage
+                {
+                    SomeId = sagaId
+                };
+                await endpoint.Send("Samples.SagaMigration.Server", correlatedMessage)
+                    .ConfigureAwait(false);
+                Console.WriteLine($"Sent a correlated message to {sagaId}");
+            }
+        }
+    }
+}
--- a/samples/saga/migration/Core_6/Client/Program.cs
+++ b/samples/saga/migration/Core_6/Client/Program.cs
@@ -27,13 +27,13 @@ class Program
             {
                 break;
             }
-            var parts = line.Split(' ');
+            var parts = line.ToLowerInvariant().Split(' ');
             if (parts.Length != 2)
             {
                 continue;
             }
             var sagaId = parts[1];
-            if (string.Equals(parts[0], "start", StringComparison.OrdinalIgnoreCase))
+            if (parts[0] == "start")
             {
                 var startingMessage = new StartingMessage
                 {
@@ -43,7 +43,7 @@ class Program
                     .ConfigureAwait(false);
                 Console.WriteLine($"Sent a starting message to {sagaId}");
             }
-            else if (string.Equals(parts[0], "complete", StringComparison.OrdinalIgnoreCase))
+            else if (parts[0] == "complete")
             {
                 var correlatedMessage = new CorrelatedMessage
                 {
--- a/samples/saga/migration/Core_6/Server.New/Program.cs
+++ b/samples/saga/migration/Core_6/Server.New/Program.cs
@@ -1,0 +1,92 @@
+﻿//#define POST_MIGRATION
+
+using System;
+using System.Data.SqlClient;
+using System.Threading.Tasks;
+using NServiceBus;
+using NServiceBus.Persistence.Sql;
+// ReSharper disable RedundantUsingDirective
+using NServiceBus.Features;
+using NServiceBus.Routing;
+using NServiceBus.Transport;
+// ReSharper restore RedundantUsingDirective
+
+class Program
+{
+    static void Main()
+    {
+        Start().GetAwaiter().GetResult();
+    }
+
+    static async Task Start()
+    {
+        Console.Title = "Samples.SagaMigration.Server.New";
+        var endpointConfiguration = new EndpointConfiguration("Samples.SagaMigration.Server");
+
+#if !POST_MIGRATION
+        endpointConfiguration.OverrideLocalAddress("Samples.SagaMigration.Server.New");
+#endif
+
+        var persistence = endpointConfiguration.UsePersistence<SqlPersistence>();
+        persistence.ConnectionBuilder(
+            connectionBuilder: () =>
+        {
+            return new SqlConnection(@"Data Source=.\SQLEXPRESS;Initial Catalog=nservicebus;Integrated Security=True;");
+        });
+        persistence.TablePrefix("New");
+        endpointConfiguration.SendFailedMessagesTo("error");
+
+        var endpoint = await Endpoint.Start(endpointConfiguration)
+            .ConfigureAwait(false);
+
+        Console.WriteLine("Press <enter> to exit.");
+        Console.ReadLine();
+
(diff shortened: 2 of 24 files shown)


SzymonPobiega:
@SimonCropp I think this can be a starting point. The offline migration via an installer might be an option but I haven't yet figured out how to pass an argument to the installer. Ideally we would start an endpoint just like when running the installers from the host but, instead of creating tables etc. we would call some custom code for each saga instance found in the database.

SzymonPobiega:
I've added `sample.md`. This is now ready for review.

SimonCropp:
Let's get this finished in Lisbon

SimonCropp:
Also ping @Particular/ravendb-persistence-maintainers @Particular/nhibernate-persistence-maintainers @Particular/azure-maintainers @Particular/nservicebus-maintainers 

Any thoughts?

indualagarsamy:
@SzymonPobiega - @SimonCropp and I talked to all the relevant maintainer groups. A member of each maintainer group has been assigned to this issue. They will ping you either on the issue or in person in Lisbon.

yvesgoeleven:
Using the unicast routing infrastructure will only work for unicast transports (like msmq & asq), or the command side of brokers. But in order to route events to the correct saga instance will be a custom implementation as each broker uses a different strategy under the hood to deliver those

SimonCropp:
@SzymonPobiega i did a bit of a cleanup

SimonCropp:
@yvesgoeleven @SzymonPobiega so what was the final status on azure? do we need custom doco or sample for that scenario?

yvesgoeleven:
@SimonCropp for ASB we would indeed need additional doco

SimonCropp:
So new note on this page? Or new sample? Or extra variant in the same sample?

SeanFeldman:
For ASB it would need to be a scripted change or guidance how to achieve it.

yvesgoeleven:
@SimonCropp I'd say an addendum to the sample as some actions related to subsccriptions need to be performed manually (using powershell)

samples/saga/migration/Core_6/Client/Program.cs
```diff
@@ -0,0 +1,58 @@
...
+            {
+                var correlatedMessage = new CorrelatedMessage
+                {
+                    SomeId = sagaId
+                };
+                await endpoint.Send("Samples.SagaMigration.Server", correlatedMessage)
+                    .ConfigureAwait(false);
+                Console.WriteLine($"Sent a correlated message to {sagaId}");
+            }
+        }
```
timbussmann: nitpick: do we need an exit condition?

samples/saga/migration/sample.md
```diff
@@ -0,0 +1,119 @@
...
+ 1. Verify that the messages were handled by "Server" endpoint.
+ 1. Ping one of the sagas (e.g. `ping 1`). Observe the message flow. It can take up to 10 seconds to complete as the flow involves a saga timeout. The result should be completion of a saga (verify by checking that a corresponding row is removed from the saga table by running again the previous query)
+ 1. Stop the solution
+ 1. Uncomment the `#define MIGRATION` line in `TestSaga.cs`
+ 1. Start the solution
+ 1. Start and ping some new sagas (e.g. `start A`, `start B` and `start C`)
+ 1. Verify sagas are started by running `SELECT * FROM [nservicebus].[dbo].[NewTestSaga]`
+ 1. Verify that the messages were handled by "Server.New" endpoint.
+ 1. Notice that "Server" console shows information indicating that the not-found handler has been used
+ 1. Ping the previously created sagas (`ping 2` or `ping 3`) to drain the saga store
```
timbussmann: should ping be renamed to complete?

samples/saga/migration/sample.md
```diff
@@ -0,0 +1,119 @@
...
+ 1. Verify the messages are handled by the old "Server" endpoint, not the "Server.New"
+ 1. Ping one of the new sagas (e.g. `ping A`) to verify it is handled properly "Server.New"
```
timbussmann: > to verify it is handled properly "Server.New"

at/by "Server.New" ?

samples/saga/migration/sample.md
```diff
@@ -0,0 +1,119 @@
...
+ 1. Notice that the timeout message that completes the saga does not go through "Server". The destination as stored in the timeout store is set to "Server.New"
```
timbussmann: how is this noticed? is this valuable information?

samples/saga/migration/sample.md
```diff
@@ -0,0 +1,119 @@
...
+ 1. Ping another saga (e.g. `ping B`) and stop the solution as soon as `Got a follow-up message.` is shown in the console
+ 1. Run `SELECT [Destination], [SagaId] FROM [nservicebus].[dbo].[NewTimeoutData]` to verify the timeout is stored in the database and the destination is the "Server.New" queue
+ 1. Uncomment the `#define POST_MIGRATION` in `Program.cs` of "Server.New". This changes the input queue of "Server.New" back to the well-known `Samples.SagaMigration.Server` and enables an additional receiver that drains the temporary queue. 
+ 1. Start only the "Server.New" project by right-clicking the project in Solution Explorer and selecting "Debug -> Start new instance"
+ 1. Notice "Server.New" prints `Moving message from Samples.SagaMigration.Server.New@<machine> to Samples.SagaMigration.Server@<machine>` and then `Got timeout. Completing.` which means the timeout has been successfully redirected from the temporary queue.
```
timbussmann: this is only the case if there are still messages in the satellite queue which might not be the case?

samples/saga/migration/sample.md
```diff
@@ -0,0 +1,119 @@
...
+
+The message flow is designed to demonstrate the correctness of migration logic:
+ * The `StartingMessage` (sent via `start` command) starts the saga and assigns the correlation property.
+ * The `CorrelatedMessage` (sent via `ping` command) contains the correlation property value of an already started saga. Handling of this message results in sending back a `ReplyMessage`
+ * The `ReplyMessage` sent by the saga contains the saga ID header containing the storage ID (not the correlation property) of the saga instance that sent it
+ * The `ReplyFollowUpMessage` send as a response to `ReplyMessage` contains the mentioned saga ID header by which the target saga is being looked up
+
+snippet:Handlers
+
+To summarise, sagas can be either looked up by their correlation property value or the storage ID.
```
timbussmann: summarize

samples/routing/custom/Core_6/Shared/AutomaticRouting/RoutingInfoSubscriber.cs
```diff
@@ -129,7 +129,7 @@ public Task OnChanged(Entry entry)
         publishers.AddOrReplacePublishers("AutomaticRouting", newPublisherMap.Select(
             x => new PublisherTableEntry(x.Key, PublisherAddress.CreateFromEndpointName(x.Value))).ToList());
 
-        endpointInstances.AddOrReplaceInstances("AutomaticRouting", newInstanceMap.SelectMany(x => x.Value).ToList());
+        endpointInstances.AddOrReplaceInstances("AutomaticRouting", newInstanceMap.SelectMany(x => x.Value).Where(IsActive).ToList());
```
timbussmann: this file already exists and seems to not directly belong to this sample? Are these changes necessary/related?

samples/saga/migration/Core_6/Server.New/Program.cs
```diff
@@ -0,0 +1,92 @@
...
+using System;
+using System.Data.SqlClient;
+using System.Threading.Tasks;
+using NServiceBus;
+using NServiceBus.Persistence.Sql;
+// ReSharper disable RedundantUsingDirective
+using NServiceBus.Features;
+using NServiceBus.Routing;
+using NServiceBus.Transport;
+// ReSharper restore RedundantUsingDirective
```
timbussmann: ordering seems a bit messy

samples/saga/migration/Core_6/Server/TestSaga.cs
```diff
@@ -0,0 +1,92 @@
...
+
+#endregion
+
+using System;
+using System.Threading.Tasks;
+using NServiceBus;
+
+#region Header
+
+#if NEW
```
timbussmann: maybe make a comment that `NEW` is defined by Server.New?

timbussmann: or even better, would `Server.New`  be valid instead?

samples/saga/migration/Core_6/Server/TestSaga.cs
```diff
@@ -0,0 +1,92 @@
...
+    IAmStartedByMessages<DummyMessage>
+#else
+        IAmStartedByMessages<StartingMessage>
+#endif
+
+    #endregion
+
+{
+#if MIGRATION && !NEW
+    public Task Handle(DummyMessage message, IMessageHandlerContext context)
```
timbussmann: let's add the comment about this being required for validation reasons comment from the doco page here as well

samples/saga/migration/sample.md
```diff
@@ -0,0 +1,119 @@
...
+
+The sample shows how to gradually migrate from one saga persister to another without requiring an off-line migration procedure/script. In this example NHibernate and SQL persisters as source and target respectively but any persister can be used in any role i.e. the same method can be used to migrate e.g. from RavenDB to NHibernate persister.
+
+
+### Message flow
+
+The message flow is designed to demonstrate the correctness of migration logic:
+ * The `StartingMessage` (sent via `start` command) starts the saga and assigns the correlation property.
+ * The `CorrelatedMessage` (sent via `ping` command) contains the correlation property value of an already started saga. Handling of this message results in sending back a `ReplyMessage`
+ * The `ReplyMessage` sent by the saga contains the saga ID header containing the storage ID (not the correlation property) of the saga instance that sent it
```
timbussmann: I'm not sure I understand the part about the storage ID? I seem to miss anything related to the storage on the returned correlation property?

samples/saga/migration/Core_6/Client/Program.cs
SzymonPobiega: Added.

samples/saga/migration/Core_6/Server.New/Program.cs
SzymonPobiega: what do you mean?

samples/saga/migration/sample.md
SzymonPobiega: Removed.

samples/saga/migration/sample.md
SzymonPobiega: I wanted to undersline that the saga id header does not contain the correlation property value but the saga id.

samples/saga/migration/Core_6/Server.New/Program.cs
timbussmann: ignore my comment :) Although it's kinda weird that our samples contain resharper specific comments but I guess that's a different topic.

mauroservienti:
What happens from the events publisher perspective?

User wants to migrate Saga A from Persistence X to Y. There are instances of Saga A running, the above is put in place and new instances of A will be created on a different endpoint with a different persistence. At a certain point A instances will be running onto 2 different endpoints. If an A instance needs to publish an event now we have 2 possible sources of truth. Subscribers can only subscribe to a single publisher but they have no way to know which one.

Should the legacy instance, the one we want to migrate away from, subscribe to all events published by new instance and re-route them?

Thoughts?

SzymonPobiega:
@mauroservienti when using native pub sub (e.g. RabbitMQ or ASB new topology) it is not a problem because pub-sub is handled by the broker.

In message-driven pub sub transports it is also not a problem because the new version of the endpoint uses the same endpoint name (but different queue):

```
endpointConfiguration.OverrideLocalAddress("Samples.SagaMigration.Server.New");
```

this way it can access the same underlying subscription persistence. The subscribers need only to know about the "Samples.SagaMigration.Server" endpoint and it is enough to subscribe there. The new version will fetch the subscriptions from there.

That said, I believe there is an issue with the ASB legacy topology but as we discussed with @yvesgoeleven  in Lisbon, this issue can be solved by proper (scripted) configuration.

mauroservienti:
Cool! thanks for the details @SzymonPobiega, all clear now.

SzymonPobiega:
@yvesgoeleven are you fine with merging this one and creating an issue for documenting the ASB endpoint-oriented-topology variant of it?

PR #2365, opened 2017-01-02
MSMQ permissions security advisory

MarcinHoppe:
Connects to Particular/PlatformDevelopment#1059

### DIFF
--- a/security-advisories/msmq-permissions.md
+++ b/security-advisories/msmq-permissions.md
@@ -9,17 +9,17 @@ tags:
 ---
 
 This advisory discloses a security vulnerability that has been found in NServiceBus and fixed in a recently released hotfix.
-* You should upgrade to the latest version of the NServiceBus package to fix this vulnerability if you are:
+* All endpoints should be upgraded to the latest version of the NServiceBus package to fix this vulnerability if:
   * using the NServiceBus MSMQ transport,
-  * running the NServiceBus [installers](https://docs.particular.net/nservicebus/operations/installers) to provision your endpoints,
+  * running the NServiceBus [installers](https://docs.particular.net/nservicebus/operations/installers) to provision endpoints,
   * customized the queue permissions by revoking access for Anonymous and/or Everyone account(s) .  
 
 This advisory affects all versions of the NServiceBus up to and including 5.2.19 and 6.0.2.
 
-If you have any questions or concerns regarding this advisory, please send an email to security@particular.net
+If there are any questions or concerns regarding this advisory, send an email to security@particular.net
 
 ## NServiceBus vulnerability: installers change permissions on existing MSMQ queues
-We have fixed a vulnerability in NServiceBus that allows attackers to grant Anonymous and Everyone accounts access to existing MSMQ queues by enforcing an endpoint restart.
+A vulnerability has been fixed in NServiceBus that allows attackers to grant Anonymous and Everyone accounts access to existing MSMQ queues by enforcing an endpoint restart.
 
 ### Impact
 Attackers can use this vulnerability to:
@@ -35,7 +35,7 @@ The exploitation of this vulnerability requires that all of the conditions below
 All versions of NServiceBus, up to and including 5.2.19 and 6.0.2, are affected by this vulnerability. The issue is tracked in https://github.com/Particular/NServiceBus/issues/4266.
 
 ### Risk Mitigation
-If you are unable to upgrade your endpoints that are using the affected version of NServiceBus, the following can be used as a **temporary workaround**:
+If it is not possible to upgrade all endpoints that are using the affected version of NServiceBus, the following can be used as a **temporary workaround**:
 
 * Schedule a task that revokes Anonymous and Everyone accounts access to MSMQ queues managed by NServiceBus after each endpoint restart.
 
@@ -57,4 +57,4 @@ Update the NuGet package using `Update-Package NServiceBus -Version 6.0.3`, re-c
 Download the NuGet package using `Update-Package NServiceBus -Version 6.0.3`, stop affected endpoints, copy the new `NServiceBus.Core.dll` assembly, overwriting the one(s) currently deployed for the endpoints, and restart the affected endpoints.
 
 ### Contact Info
-If you have any questions or concerns regarding this advisory, please send an email to security@particular.net
+If there are any questions or concerns regarding this advisory, send an email to security@particular.net
--- a/security-advisories/msmq-permissions.md
+++ b/security-advisories/msmq-permissions.md
@@ -11,7 +11,7 @@ tags:
 This advisory discloses a security vulnerability that has been found in NServiceBus and fixed in a recently released hotfix.
 * All endpoints should be upgraded to the latest version of the NServiceBus package to fix this vulnerability if:
   * using the NServiceBus MSMQ transport,
-  * running the NServiceBus [installers](https://docs.particular.net/nservicebus/operations/installers) to provision endpoints,
+  * running the NServiceBus [installers](/nservicebus/operations/installers.md) to provision endpoints,
   * customized the queue permissions by revoking access for Anonymous and/or Everyone account(s) .  
 
 This advisory affects all versions of the NServiceBus up to and including 5.2.19 and 6.0.2.
--- a/security-advisories/msmq-permissions.md
+++ b/security-advisories/msmq-permissions.md
@@ -23,7 +23,7 @@ This advisory affects all versions of the NServiceBus up to and including 5.2.19
 A vulnerability has been fixed that allows attackers to grant Anonymous and Everyone accounts access to existing MSMQ queues by enforcing an endpoint restart.
 
 
-### Impact
+## Impact
 
 Attackers can use this vulnerability to:
 
@@ -31,7 +31,7 @@ Attackers can use this vulnerability to:
  * peek and receive messages from MSMQ queues managed by NServiceBus.
 
 
-### Exploitability
+## Exploitability
 
 The exploitation of this vulnerability requires that all of the conditions below are met at the same time the attacker must be able to:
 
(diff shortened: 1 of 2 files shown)


MarcinHoppe:
@Particular/docs-maintainers @dbelcham @tmasternak Please review the security advisory for the MSMQ permissions issue.

security-advisories/msmq-permissions.md
```diff
@@ -0,0 +1,60 @@
...
+ - MSMQ
+ - Transport
+ - Security
+---
+
+This advisory discloses a security vulnerability that has been found in NServiceBus and fixed in a recently released hotfix.
+* All endpoints should be upgraded to the latest version of the NServiceBus package to fix this vulnerability if:
+  * using the NServiceBus MSMQ transport,
+  * running the NServiceBus [installers](/nservicebus/operations/installers.md) to provision endpoints,
+  * customized the queue permissions by revoking access for Anonymous and/or Everyone account(s) .
```
tmasternak: This and line 12 does not read well. I think it should be something like 

>**user** customized the queue ...

or

> the queue permissions have been customized ...

security-advisories/msmq-permissions.md
```diff
@@ -0,0 +1,60 @@
...
+
+### Impact
+Attackers can use this vulnerability to:
+ * send unauthorized messages to MSMQ queues managed by NServiceBus,
+ * peek and receive messages from MSMQ queues managed by NServiceBus.
+
+### Exploitability
+The exploitation of this vulnerability requires that all of the conditions below are met at the same time:
+ 1. Attacker must be able to enforce an endpoint restart,
+ 1. Attacker must be able to access MSMQ queues managed by the restarted endpoint using an Anonymous or Everyone account.
```
tmasternak: Should be "2."

security-advisories/msmq-permissions.md
MarcinHoppe: Fixed.

security-advisories/msmq-permissions.md
MarcinHoppe: Expanding the list should be handled by the Markdown engine.

MarcinHoppe:
@Particular/docs-maintainers Any chance for a review and possibly merge?

security-advisories/msmq-permissions.md
```diff
@@ -0,0 +1,85 @@
...
+### Affected versions
+
+All versions of NServiceBus, up to and including 5.2.19 and 6.0.2, are affected by this vulnerability. The issue is tracked in https://github.com/Particular/NServiceBus/issues/4266.
+
+
+### Risk Mitigation
+
+If it is not possible to upgrade all endpoints that are using the affected version of NServiceBus, the following can be used as a **temporary workaround**:
+
+ * Schedule a task that revokes Anonymous and Everyone accounts access to MSMQ queues managed by NServiceBus after each endpoint restart.
```
SimonCropp: how does a scheduled task mitigate this ? is it still vulnerable between the time a scheduled task executes?

security-advisories/msmq-permissions.md
```diff
@@ -0,0 +1,85 @@
...
+
+
+### Fix
+
+This vulnerability can be fixed by upgrading the NServiceBus package that is being used. Upgrades should be performed as follows:
```
SimonCropp: "that is being used" is redundant

security-advisories/msmq-permissions.md
```diff
@@ -0,0 +1,85 @@
...
+
+
+#### Version .x users should upgrade to Version 5.2.20 or higher.
```
SimonCropp: .x users?

security-advisories/msmq-permissions.md
```diff
@@ -0,0 +1,85 @@
...
+
+**Option 1: Full update**
+
+Update the NuGet package using `Update-Package NServiceBus -Version 5.2.20`, re-compile the endpoint/application, and redeploy the endpoint/application.
+
+**Option 2: In-place update**
+
+Download the NuGet package using `Update-Package NServiceBus -Version 5.2.20`, stop affected endpoints, copy the new `NServiceBus.Core.dll` assembly, overwriting the one(s) currently deployed for the endpoints, and restart the affected endpoints.
```
SimonCropp: Update-Package does not download the nuget

SzymonPobiega:
Looks good. One question, though. It says it requires using installers. Would I be affected if I used NSB host to install the queues one time? As I understood it, it requires "EnableInstallers" to be hard-coded into the configuration to cause the endpoint to be affected.

security-advisories/msmq-permissions.md
```diff
@@ -0,0 +1,85 @@
...
+
+
+#### Version 6.x users should upgrade to Version 6.0.3 or higher
+
+**Option 1 Full update**
+
+Update the NuGet package using `Update-Package NServiceBus -Version 6.0.3`, re-compile the endpoint/application, and redeploy the endpoint/application.
```
SimonCropp: shouldnt they just Update-Package NServiceBus so if higher than  6.0.3 they get that?

security-advisories/msmq-permissions.md
```diff
@@ -0,0 +1,85 @@
...
+tags:
+ - MSMQ
+ - Transport
+ - Security
+---
+
+This advisory discloses a security vulnerability that has been found in NServiceBus and fixed in a recently released hotfix.
+
+ * All endpoints should be upgraded to the latest version of the NServiceBus package to fix this vulnerability if:
+   * using the NServiceBus MSMQ transport,
```
SimonCropp: link to msmq transport

security-advisories/msmq-permissions.md
```diff
@@ -0,0 +1,85 @@
...
+   * running the NServiceBus [installers](/nservicebus/operations/installers.md) to provision endpoints,
+   * queue permissions for Anonymous and Everyone accounts have been revoked.
```
SimonCropp: can u qualify this with more info?

security-advisories/msmq-permissions.md
```diff
@@ -0,0 +1,85 @@
...
+
+This advisory affects all versions of the NServiceBus up to and including 5.2.19 and 6.0.2.
+
+If there are any questions or concerns regarding this advisory, send an email to security@particular.net
```
SimonCropp: this is duplicate below? perhaps remove this one

security-advisories/msmq-permissions.md
```diff
@@ -0,0 +1,85 @@
...
+ * send unauthorized messages to MSMQ queues managed by NServiceBus,
+ * peek and receive messages from MSMQ queues managed by NServiceBus.
+
+
+### Exploitability
+
+The exploitation of this vulnerability requires that all of the conditions below are met at the same time the attacker must be able to:
+
+ 1. Enforce an endpoint restart,
+ 1. Access MSMQ queues managed by the restarted endpoint using an Anonymous or Everyone account.
```
SimonCropp: is this really try. if they know when an endpoint is restarted wouldnt that suffice?

security-advisories/msmq-permissions.md
```diff
@@ -0,0 +1,85 @@
...
+### Fix
+
+This vulnerability can be fixed by upgrading the NServiceBus package that is being used. Upgrades should be performed as follows:
+
+
+#### Version .x users should upgrade to Version 5.2.20 or higher.
+
+**Option 1: Full update**
+
+Update the NuGet package using `Update-Package NServiceBus -Version 5.2.20`, re-compile the endpoint/application, and redeploy the endpoint/application.
```
SimonCropp: `Update-Package NServiceBus -Version 5.2.20` . perhaps link to the release notes instead. since they could update using the UI as well

security-advisories/msmq-permissions.md
```diff
@@ -0,0 +1,85 @@
...
+Update the NuGet package using `Update-Package NServiceBus -Version 6.0.3`, re-compile the endpoint/application, and redeploy the endpoint/application.
+
+**Option 2  In-place update**
+
+Download the NuGet package using `Update-Package NServiceBus -Version 6.0.3`, stop affected endpoints, copy the new `NServiceBus.Core.dll` assembly, overwriting the one(s) currently deployed for the endpoints, and restart the affected endpoints.
+
+
+### Contact Info
+
+If there are any questions or concerns regarding this advisory, send an email to security@particular.net
```
SimonCropp: email should be `[security@particular.net](mailto://security@particular.net)`

security-advisories/msmq-permissions.md
MarcinHoppe: The period between the service starts (vulnerability is introduced) and the scheduled task is executed (vulnerability fixed) will typically be quite short.

This is not meant to be a complete solution but something that merely reduces the risk.

Does this make sense?

MarcinHoppe:
@SimonCropp I improved the advisory as suggested. Thanks!

@SzymonPobiega If you used the host to run the installers once, then locked down queue permissions and then installers would not be invoked again (no `cfg.EnableInstallers()`), then you should be fine.

SimonCropp:
From my perspective the scheduled task in no way mitigates the vulnerability. We already state the attack vector assumes restart permissions. So any time an attacker wants to manipulate the queue they simply restart and perform the queue action. The scheduled task would need to be repeated in the milliseconds range to even have a chance of preventing the action

MarcinHoppe:
It is not meant to be a full fix for the issue but a means of decreasing the risk as much as possible for those customers that cannot update their endpoints for some reason.

Do you suggest we remove this information? Is there a better way to mitigate risk that we didn't think of?

SimonCropp:
yes i think it should be removed since it does not mitigate the security risks.  no i dont think there is a better way

SimonCropp:
@marcinhoppe if u think the is urgent then you should feel free to merge it yourself. Just be aware that if in the future it is decided it is not a valid mitigation strategy how will you go about communicating that to people who have used it and are thinking they were secure because of it

mikeminutillo:
I'm with @SimonCropp. The proposed risk mitigation strategy shortens the window in which this can be exploited but it could also quietly close the door after it's already been exploited. Rather than propose a solution, should we simply suggest careful monitoring of all queues until the endpoint has been patched?

If timeliness is important here (and it is for a security advisory), should we just drop the risk mitigation section, strongly recommend updating, and publish?

MarcinHoppe:
@SimonCropp @mikeminutillo Thanks for stopping me from merging this advisory and publishing as is!

I read up on MSMQ auditing and it seems to be a much better solution to set up MSMQ auditing and monitor the security event log. I will improve the advisory.

MarcinHoppe:
@SimonCropp @mikeminutillo I changed the risk mitigation to auditing + monitoring, looks better to me now.

PR #2366, opened 2017-01-03
Add new rabbit certificate API

bording:
This adds documentation for a new API that will be included in the 4.3.0 rabbit release.

We don't have an ETA for the 4.3.0 release, so this is WIP for now because there isn't a package to build against.


SimonCropp:
Out of curiosity what is the eta for 4.3 ?

nservicebus/rabbitmq/connection-settings_tls-details_rabbit_[4,].partial.md
```diff
@@ -0,0 +1,13 @@
+## Transport Layer Security support
+
+Secure connections to the broker using Transport Layer Security (TLS) are supported. For information on how to configure TLS on the broker, refer to the [RabbitMQ documentation](http://www.rabbitmq.com/ssl.html). To enable TLS support, set the `UseTls` setting to `true` in the connection string. If the broker has been configured to require client authentication, a client certificate can be specified in the `CertPath` setting. If that certificate requires a password, it can be specified in the `CertPassphrase` setting.
```
Comment (SimonCropp) [inline nservicebus/rabbitmq/connection-settings_tls-details_rabbit_[4,].partial.md]: I think make the text "transport layer security (tls)" link to the rabbit doco and drop the "for information" sentence

nservicebus/rabbitmq/connection-settings_tls-details_rabbit_[4,].partial.md
```diff
@@ -0,0 +1,13 @@
...
+
+An example connection string using these settings:
+
+snippet:rabbitmq-connection-tls
+
+In Versions 4.3 and above, client certificates can be specified via code instead:
+
+snippet:rabbitmq-config-client-certificates
+
+NOTE: TLS 1.2 is required to establish a secure connection, so the broker must have TLS 1.2 enabled.
```
Comment (SimonCropp) [inline nservicebus/rabbitmq/connection-settings_tls-details_rabbit_[4,].partial.md]: Perhaps: tls 1.2 must be enabled on the broker to establish a secure connection

nservicebus/rabbitmq/connection-settings_tls-settings_rabbit_[4,].partial.md
```diff
@@ -0,0 +1,15 @@
+### UseTls
+
+Indicates if the connection to the broker should be secured with [TLS](#transport-layer-security-support).
+
+Default: `false`
+
+
+### CertPath
```
Comment (SimonCropp) [inline nservicebus/rabbitmq/connection-settings_tls-settings_rabbit_[4,].partial.md]: Are the Cert* settings required if tls is enough? If so should that be stated?

nservicebus/rabbitmq/connection-settings_tls-details_rabbit_[4,].partial.md
Comment (SimonCropp) [inline nservicebus/rabbitmq/connection-settings_tls-details_rabbit_[4,].partial.md]: Is it "Can be specified" or "must be specified"?

Comment (adamralph) [inline nservicebus/rabbitmq/connection-settings_tls-details_rabbit_[4,].partial.md]: Probably "must be".

@bording?

adamralph:
> Out of curiosity what is the eta for 4.3 ?

Currently a hot topic for us. We have features and bugfixes ready to go, and we want to release now. On the other hand, we already have 9 supported versions of the transport in production (next most package has 3) and we're reluctant to add to that burden. Best answer we have right now is to restrict ourselves to no more than one release per quarter. Our last release was 30th Nov 2016.

Snippets/Rabbit/Rabbit_4/Usage.cs
```diff
@@ -140,4 +141,14 @@ void PrefetchCount(EndpointConfiguration endpointConfiguration)
         #endregion
     }
 
+    void SetClientCertificates(EndpointConfiguration endpointConfiguration)
+    {
+        #region rabbitmq-config-client-certificates
+
+        var transport = endpointConfiguration.UseTransport<RabbitMQTransport>();
+        transport.SetClientCertificates(new X509CertificateCollection());
```
adamralph: Discussed this with @SimonCropp: perhaps we need to show something other than the setting of an empty collection.

nservicebus/rabbitmq/connection-settings_tls-settings_rabbit_[4,].partial.md
Comment (adamralph) [inline nservicebus/rabbitmq/connection-settings_tls-settings_rabbit_[4,].partial.md]: I discussed this with @SimonCropp: we may need to flesh out the relationship of these properties a little. E.g. if this is set, then X must also be set. If Y is not set then setting this will fail, or will be redundant etc.

Snippets/Rabbit/Rabbit_4/Usage.cs
bording: @adamralph What do you think would be better? Creating a new `X509Certificate2`, adding it to a new collection, and then passing in the collection?

I'm not sure if that would be better or not, because if you're wanting to use this API, you're likely already getting your certificate instances from somewhere else, not creating them just to pass them in.

adamralph: @bording it's a good question. I wonder if it would be OK to declare a collection before the snippet region, and use it within the snippet region.

@SimonCropp could you advise?

bording:
For the most part, this PR is just taking the existing content and splitting them into separate v3 and v4 partials:

details

settings

Given that, I think we should get this merged in as is (once the feedback around the new snippet is addressed), and then look at improving the other things mentioned here in a separate PR.

SimonCropp:
why are we surfacing more rabbit connection APIs rather then just giving users an explicit api to control the connection construction?

adamralph:
> why are we surfacing more rabbit connection APIs rather then just giving users an explicit api to control the connection construction?

Allowing control over connection construction is indeed something we could consider. Perhaps we should hold off on the 4.3 release while we spike something.

bording:
@SimonCropp I'm not sure I see the benefit of that vs. our current approach of enabling certain features as it makes sense. There are certain values on the client's `ConnectionFactory` class that we have to retain control over, so we'd likely have to override values on the passed in instance anyway. That would then lead us to have to document something like "you can set the following values but not these others".

SimonCropp:
What values do we need to retain control over?

bording:
For example, `AutomaticRecoveryEnabled` and `UseBackgroundThreadsForIO` have to be set to `true`, and we add a lot of diagnostic info into `ClientProperties`.

On the flip side, there are a number of properties in there that we intentionally don't touch, for example a number of different timeout values and the ability to pass in more than one hostname for a connection, that may negatively impact the transport working properly if they've been modified.

It seems like if we opened this up, we'd basically have to say "enter at your own risk, we can't support it" and I'm not sure I see the benefit of doing that.

SimonCropp:
> For example, AutomaticRecoveryEnabled and UseBackgroundThreadsForIO have to be set to true

we could detect false and throw

> and we add a lot of diagnostic info into ClientProperties.

i assume we can add them to the instance passed to us

>  a number of different timeout values and the ability to pass in more than one hostname for a connection, that may negatively impact the transport working properly if they've been modified.

we could throw or log an error on those

why i ask is that ATM we are essentially wrapping the rabbit API. so re-implementing + test + re-documenting it. so every time there is a new api to the connection we will need to duplicate and doco. also if there is a breaking change we may need to release a new major of the transport.

Snippets/Rabbit/Rabbit_4/Usage.cs
SimonCropp: if it is assumed knowledge that people will know how to instantiate that collection then that is fine

SimonCropp:
@bording just to be clear my above comments are not meant to block this PR or the new rabbit transport version. it was just a convenient place to ask the questions

adamralph:
@SimonCropp @bording I propose we go ahead with the 4.3 release (and the content in this PR) and spin off a separate issue for surfacing the connection creation.

@SimonCropp assuming you agree with that proposal, are there any further changes you'd like to see in this PR. The [comment by @bording](https://github.com/Particular/docs.particular.net/pull/2366#issuecomment-273271607) shows the *actual* changes in this PR. Your comment regarding the current content which has been copied from the v3 partials to the v4 partials are valid, but we can deal with those in a separate PR for both versions.

SimonCropp:
@adamralph yeah happy for the other stuff to be fixed in a diff PR

and yes my comments were not intended to stall 4.3

indualagarsamy:
@bording is this still WIP?

bording:
Yes, we haven't released 4.3 yet, and still don't have an ETA on it, so the samples and snippets won't build until we release it.

bording:
FYI I've rebased this and I've pushed up an alpha package of the 4.3.0 release, which will be coming out "soon."

I'd say this is ready for merging. I know there are some comments here around content, but they are all for things that preexisted this PR. I'd like to get this merged, and then we can address the other content changes in a separate PR.

indualagarsamy:
@bording - can you update the PR description?

indualagarsamy:
@bording - just so we don't lose the changes that need to be done, can you create a docs issue for the text cleanup that needs to be done, so we can track it better. Thanks.

PR #2368, opened 2017-01-04
Critical time perf counter - confusing description

SeanFeldman:
Have removed the following sentence:
> Monitors the age of the oldest message in the queue. 

The entire paragraph now might need to be re-written.
Could someone with a bit more of b/g knowledge (or perhaps historical knowledge of how this perf counter used to be exercised) chime in?

@Particular/nservicebus-maintainers 

PS: I heard @andreasohlund ❤️  perf counters 😺

### DIFF
--- a/nservicebus/operations/performance-counters.md
+++ b/nservicebus/operations/performance-counters.md
@@ -33,7 +33,7 @@ Since all performance counters in Windows are exposed via Windows Management Ins
 
 **Added in:** Version 3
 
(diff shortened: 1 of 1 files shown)


andreasohlund:
Perhaps

> Critical Time is the time from a message being sent until successfully processed. This metric is useful for monitoring message SLA's similar to "All orders should be processed within X seconds/minutes/hours"

Thoughts?

SeanFeldman:
👍 much better. The original description was confusing.

SeanFeldman:

PR #2374, opened 2017-01-13
Update licensing order selection to v5 and above

WilliamBZA:
The order of license detection is no longer correct for V5 and V6. V5 and V6 no longer look at the `NServiceBus` key at all, looking instead directly at the `ParticularSoftware` key.

Additionally: the table listing the order of locations was all encompassing of all versions of NServicebus. Given that we have a dropdown to change the version of docs that we are looking at, we can make the license order check a lot neater by making it specific to each version. This PR addresses that for versions 5 and 6.

Ping @Particular/nservicebus-maintainers please check the validity of this
Ping @Particular/docs-maintainers please check that I've done the partial files correctly.


andreasohlund:
The keys seems correct but the order wrong, v6 seems to check the license keys in `SOFTWARE\ParticularSoftware` first:

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Licensing/LicenseManager.cs#L86

and then check the other locations:

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Licensing/LicenseLocationConventions.cs#L11

WilliamBZA:
Thanks @andreasohlund, I've updated to match now.

SimonCropp:
@WilliamBZA thanks

PR #2387, opened 2017-01-17
NSB.ASB 7.1.1 upgrade

SeanFeldman:


Snippets/SqlPersistence/SqlPersistence_1/CorrelationPropertyTypeConverterWriter.cs
```diff
@@ -0,0 +1,51 @@
+﻿
```
SimonCropp: i think there is a git issue here

SeanFeldman: I could swear we didn't touch it...

SeanFeldman:
@SimonCropp fixed. You had `master` updated after we branched :)

PR #2391, opened 2017-01-24
[WIP] Wantqueue created

### LINKED ISSUE #2355: There's no migration guide for IWantQueuesCreated
This interface was obsoleted in v6 but there's no migration guide for it.

danielmarbach:
Closes #2355

@HEskandari would that be enough to address #2355 ?

@Particular/docs-maintainers what do you think?


HEskandari:
@danielmarbach Looks good. One thing to add maybe, would be a note + link to the [transport-specific queue creation page](https://docs.particular.net/nservicebus/msmq/operations-scripting#create-queues)?

PR #2392, opened 2017-01-24
Document that immediate retries may happen even if turned off

andreasohlund:
@Particular/nservicebus-maintainers @SimonCropp @ramonsmits thoughts?

### DIFF
--- a/nservicebus/recoverability/index.md
+++ b/nservicebus/recoverability/index.md
@@ -39,6 +39,8 @@ By default up to five immediate retries are performed if the message processing 
 
 Note: The configured value describes the minimum number of times a message will be retried if its processing consistently fails. Especially in environments with competing consumers on the same queue, there is an increased chance of retrying a failing message more times across the endpoints.
 
+Note: Depending on the concurrency and transactionallity settings of the endpoint immediate retries might happen even if configured to be off. To ensure that no retries is performed the endpoints needs to be configured to be either [non transactional](/nservicebus/transports/transactions.md) or the [concurrency retricted to 1](/nservicebus/operations/tuning.md).
+
 
 ### Transport transaction requirements
 
--- a/nservicebus/recoverability/index.md
+++ b/nservicebus/recoverability/index.md
@@ -39,7 +39,7 @@ By default up to five immediate retries are performed if the message processing 
 
 Note: The configured value describes the minimum number of times a message will be retried if its processing consistently fails. Especially in environments with competing consumers on the same queue, there is an increased chance of retrying a failing message more times across the endpoints.
 
-Note: Depending on the concurrency and transactionallity settings of the endpoint immediate retries might happen even if configured to be off. To ensure that no retries is performed the endpoints needs to be configured to be either [non transactional](/nservicebus/transports/transactions.md) or the [concurrency retricted to 1](/nservicebus/operations/tuning.md).
+Note: Depending on the concurrency and transactionality settings of the endpoint immediate retries might happen even if configured to be off. To ensure that no retries is performed the endpoints needs to be configured to be either [non transactional](/nservicebus/transports/transactions.md) or the [concurrency retricted to 1](/nservicebus/operations/tuning.md).
 
 
 ### Transport transaction requirements


nservicebus/recoverability/index.md
```diff
@@ -39,6 +39,8 @@ By default up to five immediate retries are performed if the message processing
 
 Note: The configured value describes the minimum number of times a message will be retried if its processing consistently fails. Especially in environments with competing consumers on the same queue, there is an increased chance of retrying a failing message more times across the endpoints.
 
+Note: Depending on the concurrency and transactionallity settings of the endpoint immediate retries might happen even if configured to be off. To ensure that no retries is performed the endpoints needs to be configured to be either [non transactional](/nservicebus/transports/transactions.md) or the [concurrency retricted to 1](/nservicebus/operations/tuning.md).
```
indualagarsamy: @andreasohlund - can you give an example of when immediate retries might happen?. It is confusing at the moment. For example, using transactional is safe as default to prevent message loss in MSMQ. So, are you saying that retries will happen even if it is turned off? In that case is this is a bug?

andreasohlund: > can you give an example of when immediate retries might happen?

It will happen if running v6 with MSMQ or SqlServer transports and a exception happens during transaction scope dispose/commit, msmq transaction dispose, sql connection dispose etc.

>So, are you saying that retries will happen even if it is turned off? 

Yes

>In that case is this is a bug?

It could be argued that it is yes but the decision was to start by documenting this since there is a workaround? (turn transactions off)

andreasohlund: Also noting that the same applies when you are scaled out since we keep the "immediate retries counter" in memory for the mentioned transports

andreasohlund: So we can't really safely guarantee no retries unless you turn the TX off == one shot at processing the message

indualagarsamy: @andreasohlund - We typically don't document bugs or workarounds in docs. If we identify a behavior as a bug, we capture it in a issue and wait for it to get addressed. The issue usually captures enough details on the behavior and the workaround.

indualagarsamy: @andreasohlund - do you think we can document this in a issue with a "Wont Fix" label? I think it will be much better than docs. In the main docs, there won't be any confusion for the readers that won't ever be affected by this, assuming our configuration is just not reliable. Thoughts? cc / @SimonCropp

andreasohlund: >do you think we can document this in a issue in the NServiceBus repo with a "Wont Fix" label?

Given that this is "by design" for all the transports without a native retries counter I'm not sure that is the way to go. Do we need a "caveats" section for each transport that have this "issue"?

indualagarsamy: @andreasohlund - I think that makes sense. i.e. caveats for the transport section.

andreasohlund: So in essence changing the note to: "The accuracy of the immediate retries depends varies per transport, see the caveats section of the relevant transport for more details"

And then add a caveats section to MSMQ +  sql describing how immediate retries isn't accurate when scaled out + running with concurrency > 1 (cc @Particular/sqlserver-transport-maintainers )

@Particular/docs-maintainers @Particular/nservicebus-maintainers does this make sense?

tmasternak: I thik that caveats are more caused by transport implementations than the transactionality. Here is my take on them:
 * noTx: 
   * 0 retries - caveat is we can have 0 processing attempts
 * receive counter stored in memory cache (MSMQ, SQL) and retires configured to x:
   * inf. retries - the cache for counter is invalidated or endpoint re-starts
   * more than x retries but finite 
      * transaction timeouts with too long running handlers
      * scaled-out instances and counter duplicates (this applies only to SQL)
  * receive counter manged by the broker (ASB, ASP) and retries configured to x:
      *  anything between 0 and x - endpoint re-start with non empty prefetch buffers
  * in-memory/per-message (RabbitMQ)
      * anything from 0 to inf. - endpoint re-starts

andreasohlund: @tmasternak so to summarize: If you need to be 100% sure that we only call your handler once== you have to turn the transactions off and be ok with the "at most once" semantics. This is true for all transports we currently support. Correct?

tmasternak: @andreasohlund yes I think it sums it up.

PR #2399, opened 2017-01-26
[WIP] exclude downstream minors from the last year from the support policy

adamralph:


nservicebus/upgrades/support-policy.md
```diff
@@ -9,7 +9,7 @@ related:
 
 ## Supported Versions
 
-Supported versions include any major version released within the last 3 years. Within a supported major version, the latest minor version is supported, or any minor version released within the last year.
+Supported versions include any major version released within the last 3 years. Within a supported major version, the latest minor version is supported. For the NServiceBus package only (i.e. excluding downstream packages such as [transports](/nservicebus/transports/) and [persistence](/nservicebus/persistence/)), any minor version released within the last year is supported.
```
adamralph: I'm not overjoyed with the wording, but I struggled to come up with something better.

indualagarsamy: @adamralph - I am not sure what is being conveyed here. Are you implying this policy applies only for NServiceBus package and not for other downstream packages such as transports, persistence and containers? What about the minor downstream packages then?

adamralph: @indualagarsamy see https://github.com/Particular/Lisbon2017.Unconference/issues/16#issuecomment-275626656

I guess this does need some work to make it clearer. Perhaps a table would be better?

indualagarsamy: @adamralph - what you have is currently not clear. 

I suggest the following. I also discussed this with @kbaley.

NServiceBus:
Particular Software supports all major versions released within the last 3 years, including any minor versions released within that timeframe.  

For other downstream packages:
However, for downstream packages that use NServiceBus, for example, all the [transports](/nservicebus/transports/), [persistences](/nservicebus/persistence/), [containers](/nservicebus/containers/), etc.,  Particular Software will only support the latest minor version of any major versions that were released within the last three years.

bording: @indualagarsamy I like the idea of actually having separate sections like you've suggested, but the wording around minor versions is incorrect. The first paragraph is saying that all minors are supported in the 3 year period, which is incorrect.

Here's my suggested revision:

>NServiceBus:
Particular Software supports all major versions released within the last 3 years, For each supported major version, the latest minor version is supported. In addition, any minor versions released in the past year are also supported.

>Downstream packages:
For downstream packages that use NServiceBus, for example, all the transports, persistences, containers, etc., Particular Software supports all major versions released within the last 3 years. For each supported major version, only the latest minor version is supported.

I also think we should add something to make it explicit about which patch versions are supported. Something like 

> NOTE: For both cases, for each supported minor version, only the latest patch version is supported.

indualagarsamy: @bording - regarding:

>The first paragraph is saying that all minors are supported in the 3 year period, which is incorrect.

Where ? In NServiceBus? I think what @kbaley and I had were more clear. Also for NServiceBus, what i understood was that the current policy is correct. i.e. not just the latest supported minor, but all minors for supported majors. In that sense, not sure what you have under NServiceBus is right.

bording: >. i.e. not just the latest supported minor, but all minors for supported majors. In that sense, not sure what you have under NServiceBus is right.

Sorry, but that isn't true, and not what the current text of the policy says. Its the latest minor, plus any that have been released in the past year.

So, for a major that has been released 2 years ago and hasn't had any minors released in 2 years, only the latest minor is supported.

SimonCropp:
i think should should be split into two different headings

indualagarsamy:
@adamralph - I think this PR in its current state is irrelevant until we have decided what exactly to do. i.e. the comments in this PR as well. How about we close this PR and once the policy is determined in the internal meetings, open a PR then? That will accurately represent the PR as well.  I don't see why we need to have this open until then. This has already been open for 11 days. cc / @Particular/docs-maintainers

adamralph:
@indualagarsamy sure, makes sense. I'll leave the branch in my fork in case it's of any value later.

PR #2403, opened 2017-02-01
Improvements on critical error callback

ramonsmits:
Based on https://particularsoftware.slack.com/files/ramon/F3ZJCG747/double_finaly_on_critical_error_to_flush_configured_custom_logger_before_failfast.md

### DIFF
--- a/Snippets/Core/Core_3/CriticalErrors.cs
+++ b/Snippets/Core/Core_3/CriticalErrors.cs
@@ -26,16 +26,20 @@ namespace Core3.Host
 
         void OnCriticalError()
         {
-            // Write log entry in version 3 since this is not done by default.
-            log.Fatal("CRITICAL Error");
-
-            // To leave the process active, dispose the bus.
-            // When the bus is disposed sending messages will throw an ObjectDisposedException.
-            ((IDisposable)bus).Dispose();
-
-            // To kill the process, raise a fail fast error as shown below.
-            // var failMessage = "Critical error shutting.";
-            // Environment.FailFast(failMessage);
+            try
+            {
+                // To leave the process active, dispose the bus.
+                // When the bus is disposed, the attempt to send message will cause an ObjectDisposedException.
+                ((IDisposable)bus).Dispose();
+                // Perform custom actions here, e.g.
+                // NLog.LogManager.Shutdown();
+            }
+            finally
+            {
+                // Write log entry in version 3 since this is not done by default.
+                log.Fatal("CRITICAL Error");
+                Environment.FailFast("Critical error shutting down.");
+            }
         }
 
         #endregion
--- a/Snippets/Core/Core_3/CriticalErrors.cs
+++ b/Snippets/Core/Core_3/CriticalErrors.cs
@@ -28,6 +28,8 @@ namespace Core3.Host
         {
             try
             {
+                // Write log entry in version 3 since this is not done by default.
+                log.Fatal("CRITICAL Error");
                 // To leave the process active, dispose the bus.
                 // When the bus is disposed, the attempt to send message will cause an ObjectDisposedException.
                 ((IDisposable)bus).Dispose();
@@ -36,8 +38,6 @@ namespace Core3.Host
             }
             finally
             {
-                // Write log entry in version 3 since this is not done by default.
-                log.Fatal("CRITICAL Error");
                 Environment.FailFast("Critical error shutting down.");
             }
         }
--- a/Snippets/Core/Core_4/CriticalErrors.cs
+++ b/Snippets/Core/Core_4/CriticalErrors.cs
@@ -23,13 +23,19 @@ namespace Core4
 
         void OnCriticalError(string errorMessage, Exception exception)
         {
-            // To leave the process active, dispose the bus.
-            // When the bus is disposed sending messages will throw an ObjectDisposedException.
-            bus.Dispose();
-
-            // To kill the process, raise a fail fast error as shown below.
-            // var failMessage = $"Critical error shutting down:'{errorMessage}'.";
-            // Environment.FailFast(failMessage, exception);
+            try
+            {
+                // To leave the process active, dispose the bus.
+                // When the bus is disposed, the attempt to send message will cause an ObjectDisposedException.
+                bus.Dispose();
+                // Perform custom actions here, e.g.
+                // NLog.LogManager.Shutdown();
+            }
+            finally
+            {
+                var failMessage = $"Critical error shutting down:'{errorMessage}'.";
+                Environment.FailFast(failMessage, exception);
+            }
         }
 
         #endregion
--- a/Snippets/Core/Core_6/CriticalErrors.cs
+++ b/Snippets/Core/Core_6/CriticalErrors.cs
@@ -20,15 +20,21 @@ namespace Core6
 
         #region CustomHostErrorHandlingAction
 
-        Task OnCriticalError(ICriticalErrorContext context)
+        async Task OnCriticalError(ICriticalErrorContext context)
         {
-            // To leave the process active, stop the endpoint.
-            return context.Stop();
-
-            // To kill the process, await the above,
-            // then raise a fail fast error as shown below.
-            // var failMessage = $"Critical error shutting down:'{context.Error}'.";
-            // Environment.FailFast(failMessage, context.Exception);
+            try
+            {
+                // To leave the process active, dispose the bus.
+                // When the bus is disposed, the attempt to send message will cause an ObjectDisposedException.
+                await context.Stop();
+                // Perform custom actions here, e.g.
+                // NLog.LogManager.Shutdown();
+            }
+            finally
+            {
+                var failMessage = $"Critical error shutting down:'{context.Error}'.";
+                Environment.FailFast(failMessage, context.Exception);
+            }
         }
 
         #endregion
--- a/Snippets/Core/Core_6/CriticalErrors.cs
+++ b/Snippets/Core/Core_6/CriticalErrors.cs
@@ -26,7 +26,7 @@ namespace Core6
             {
                 // To leave the process active, dispose the bus.
                 // When the bus is disposed, the attempt to send message will cause an ObjectDisposedException.
-                await context.Stop();
+                await context.Stop().ConfigureAwait(false);
                 // Perform custom actions here, e.g.
                 // NLog.LogManager.Shutdown();
             }
--- a/nservicebus/hosting/critical-errors.md
+++ b/nservicebus/hosting/critical-errors.md
@@ -52,7 +52,7 @@ Consider the following things when implementing a critical error callback:
 - Wrap any tasks in a `try..finally`, making sure the process will exit if any tasks fail.
 - Flush any loggers used which makes sure any unwritten log state is written/pushed to its target(s) making sure the tail of the log is not lost as [Environment.FailFast](https://msdn.microsoft.com/en-us/library/dd289240.aspx) will immediately exit the process.
  - Flush NLog or log4net state by calling `LogManager.Shutdown();`, this still isn't 100% guaranteed to not loose any log entries. Check each logging framework its document.
-- Flush any state/caches used if you normally would flush such data structure in a certain interval.
+- Flush any state/caches used if normally such data structures are would persisted in a certain interval or at graceful shutdown.
 - It is easiest to call [Environment.FailFast](https://msdn.microsoft.com/en-us/library/dd289240.aspx)
 
 Rely on the environment that the process will be automatically restarted. When hosting in IIS the IIS host will automatically spawn a new instance and when hosting as a Windows Service the OS will restart the service after 1 minute if [Windows Service Recovery](/nservicebus/hosting/windows-service.md#installation-restart-recovery) is enabled.
--- a/nservicebus/hosting/critical-errors.md
+++ b/nservicebus/hosting/critical-errors.md
@@ -58,7 +58,7 @@ When implementing a custom critical error callback:
 
 - To exit the process use the [Environment.FailFast](https://msdn.microsoft.com/en-us/library/dd289240.aspx) method. In case the environment has threads running that should be completed before shutdown (e.g. non transactional operations), one may also use the [Environment.Exit](https://msdn.microsoft.com/en-us/library/system.environment.exit(v=vs.110).aspx) method.
 - The code should be wrapped in a `try...finally` clause. In the `try` block perform any custom operations, in the `finally` block call the method that exits the process.
-- The custom operations should include flushing any in-memory state and cached data, if normally its persisted at a certain interval or during graceful shutdown. For example, flush appenders if you are using buffering or asynchrounous appanders for [NLog](http://nlog-project.org/documentation/v4.3.0/html/M_NLog_LogManager_Shutdown.htm) or [log4net](https://logging.apache.org/log4net/log4net-1.2.11/release/sdk/log4net.LogManager.Shutdown.html) state by calling `LogManager.Shutdown();`.
+- The custom operations should include flushing any in-memory state and cached data, if normally its persisted at a certain interval or during graceful shutdown. For example, flush appenders when using buffering or asynchrounous appanders for [NLog](http://nlog-project.org/documentation/v4.3.0/html/M_NLog_LogManager_Shutdown.htm) or [log4net](https://logging.apache.org/log4net/log4net-1.2.11/release/sdk/log4net.LogManager.Shutdown.html) state by calling `LogManager.Shutdown();`.
 
 Whenever possible rely on the environment hosting the endpoint process to automatically restart it:
 - IIS: The IIS host will automatically spawn a new instance.
--- a/nservicebus/hosting/critical-errors.md
+++ b/nservicebus/hosting/critical-errors.md
@@ -55,7 +55,7 @@ Consider the following when implementing a critical error callback:
 - Flush any loggers used which makes sure any unwritten log state is written/pushed to its target(s) making sure the tail of the log is not lost as [Environment.FailFast](https://msdn.microsoft.com/en-us/library/dd289240.aspx) will immediately exit the process.
  - Flush appenders for [NLog](http://nlog-project.org/documentation/v4.3.0/html/M_NLog_LogManager_Shutdown.htm) or [log4net](https://logging.apache.org/log4net/log4net-1.2.11/release/sdk/log4net.LogManager.Shutdown.html) state by calling `LogManager.Shutdown();`.
 - Flush any state/caches used if such data structures would be written/persisted at a certain interval or regular graceful shutdown.
-- It is easiest to call [Environment.FailFast](https://msdn.microsoft.com/en-us/library/dd289240.aspx)
+- It is easiest to call [Environment.FailFast](https://msdn.microsoft.com/en-us/library/dd289240.aspx) or [Environment.Exit](https://msdn.microsoft.com/en-us/library/system.environment.exit(v=vs.110).aspx).
 
 Rely on the environment that the process will be automatically restarted. When hosting in IIS the IIS host will automatically spawn a new instance and when hosting as a Windows Service the OS will restart the service after 1 minute if [Windows Service Recovery](/nservicebus/hosting/windows-service.md#installation-restart-recovery) is enabled.
 
--- a/nservicebus/hosting/critical-errors.md
+++ b/nservicebus/hosting/critical-errors.md
@@ -45,7 +45,11 @@ snippet:CustomHostErrorHandlingAction
 
 ## When to override the default action
 
-When self hosting the should be overriden in order to exit the current process unless you are using the [NServiceBus Host](/nservicebus/hosting/nservicebus-host) which already has a implementation that exits the process.
+When self hosting the critical error callback must be overriden in order to exit the current process. Not overriding the critical error callback will result in the endpoint instance to stop processing messages without the ability to recover from this state.
+
+NOTE: If not killing the process and just disposing the bus, be aware that any `Send` operations will result in [ObjectDisposedException](https://msdn.microsoft.com/en-us/library/system.objectdisposedexception.aspx) being thrown.
+
+When the [NServiceBus Host](/nservicebus/hosting/nservicebus-host) is used it is not required to override this callback as the host has a default implementation that exits the process. However, if it is required to execute tasks just before shutdown then it can be overriden in the same way as when self hosting.
 
 Consider the following things when implementing a critical error callback:
 
@@ -57,7 +61,6 @@ Consider the following things when implementing a critical error callback:
 
 Rely on the environment that the process will be automatically restarted. When hosting in IIS the IIS host will automatically spawn a new instance and when hosting as a Windows Service the OS will restart the service after 1 minute if [Windows Service Recovery](/nservicebus/hosting/windows-service.md#installation-restart-recovery) is enabled.
 
-NOTE: If not killing the process and just dispose the bus, be aware that any `Send` operations will result in [ObjectDisposedException](https://msdn.microsoft.com/en-us/library/system.objectdisposedexception.aspx) being thrown.
 
 
 ## Raising Critical error
--- a/nservicebus/hosting/critical-errors.md
+++ b/nservicebus/hosting/critical-errors.md
@@ -45,10 +45,16 @@ snippet:CustomHostErrorHandlingAction
 
 ## When to override the default action
 
-The default action should be overridden whenever that default does not meet the specific hosting requirements. For example
+When self hosting the should be overriden in order to exit the current process unless you are using the [NServiceBus Host](/nservicebus/hosting/nservicebus-host) which already has a implementation that exits the process.
 
- * If using [NServiceBus Host](/nservicebus/hosting/nservicebus-host), and wish to take a custom action before the endpoint process is killed.
- * If self hosting the process can be shut down the process via [Environment.FailFast](https://msdn.microsoft.com/en-us/library/dd289240.aspx) and re-start the process once the root cause has been diagnosed.
+Consider the following things when implementing a critical error callback:
+
+- Wrap any tasks in a `try..finally`, making sure the process will exit if any tasks fail.
+- Flush any loggers used which makes sure any unwritten tail of the log is persisted as [Environment.FailFast](https://msdn.microsoft.com/en-us/library/dd289240.aspx) will immediately exit the process
+- Flush any state/caches used if you normally would flush such data structure in a certain interval.
+- It is easiest to call [Environment.FailFast](https://msdn.microsoft.com/en-us/library/dd289240.aspx)
+
+Rely on the environemnt that the process will be automatically restarted. When hosting in IIS the IIS host will automatically spawn a new instance and when hosting as a Windows Service the OS will restart the service after 1 minute if [Windows Service Recovery](/nservicebus/hosting/windows-service.md#installation-restart-recovery) is enabled.
 
 NOTE: If not killing the process and just dispose the bus, be aware that any `Send` operations will result in [ObjectDisposedException](https://msdn.microsoft.com/en-us/library/system.objectdisposedexception.aspx) being thrown.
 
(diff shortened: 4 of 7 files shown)


ramonsmits:
@indualagarsamy As promised

SimonCropp:
@ramonsmits can you fix the build errors

nservicebus/hosting/critical-errors.md
```diff
@@ -46,10 +46,16 @@ snippet:CustomHostErrorHandlingAction
...
 
-The default action should be overridden whenever that default does not meet the specific hosting requirements. For example
+When self hosting the should be overriden in order to exit the current process unless you are using the [NServiceBus Host](/nservicebus/hosting/nservicebus-host) which already has a implementation that exits the process.
 
- * If using [NServiceBus Host](/nservicebus/hosting/nservicebus-host), and wish to take a custom action before the endpoint process is killed.
- * If self hosting the process can be shut down the process via [Environment.FailFast](https://msdn.microsoft.com/en-us/library/dd289240.aspx) and re-start the process once the root cause has been diagnosed.
+Consider the following things when implementing a critical error callback:
+
+- Wrap any tasks in a `try..finally`, making sure the process will exit if any tasks fail.
+- Flush any loggers used which makes sure any unwritten tail of the log is persisted as [Environment.FailFast](https://msdn.microsoft.com/en-us/library/dd289240.aspx) will immediately exit the process
```
indualagarsamy: Can you link to relevant logging framework advice here?

nservicebus/hosting/critical-errors.md
```diff
@@ -46,10 +46,16 @@ snippet:CustomHostErrorHandlingAction
...
+- Flush any state/caches used if you normally would flush such data structure in a certain interval.
+- It is easiest to call [Environment.FailFast](https://msdn.microsoft.com/en-us/library/dd289240.aspx)
```
indualagarsamy: are we recommending we fail fast?

nservicebus/hosting/critical-errors.md
```diff
@@ -46,10 +46,16 @@ snippet:CustomHostErrorHandlingAction
...
+
+Rely on the environemnt that the process will be automatically restarted. When hosting in IIS the IIS host will automatically spawn a new instance and when hosting as a Windows Service the OS will restart the service after 1 minute if [Windows Service Recovery](/nservicebus/hosting/windows-service.md#installation-restart-recovery) is enabled.
```
indualagarsamy: > environment

typo

Can we split this into two, i.e. IIS and non-IIS hosting advice?

nservicebus/hosting/critical-errors.md
ramonsmits: Well, that any unflushed log state is written to storage.

nservicebus/hosting/critical-errors.md
```diff
@@ -46,10 +46,16 @@ snippet:CustomHostErrorHandlingAction
...
-The default action should be overridden whenever that default does not meet the specific hosting requirements. For example
+When self hosting the should be overriden in order to exit the current process unless you are using the [NServiceBus Host](/nservicebus/hosting/nservicebus-host) which already has a implementation that exits the process.
 
- * If using [NServiceBus Host](/nservicebus/hosting/nservicebus-host), and wish to take a custom action before the endpoint process is killed.
- * If self hosting the process can be shut down the process via [Environment.FailFast](https://msdn.microsoft.com/en-us/library/dd289240.aspx) and re-start the process once the root cause has been diagnosed.
+Consider the following things when implementing a critical error callback:
+
+- Wrap any tasks in a `try..finally`, making sure the process will exit if any tasks fail.
+- Flush any loggers used which makes sure any unwritten tail of the log is persisted as [Environment.FailFast](https://msdn.microsoft.com/en-us/library/dd289240.aspx) will immediately exit the process
+- Flush any state/caches used if you normally would flush such data structure in a certain interval.
```
ramonsmits: No, not cleared, that in memory state like a cache is persisted. It isn't applicable. If a customer is doing such a thing at normal application shutdown then they should do it here too.

nservicebus/hosting/critical-errors.md
ramonsmits: That is what was there before. Not sure if there are even alternatives.

nservicebus/hosting/critical-errors.md
```diff
@@ -46,12 +46,22 @@ snippet:CustomHostErrorHandlingAction
...
+When self hosting the critical error callback must be overriden in order to exit the current process. Not overriding the critical error callback will result in the endpoint instance to stop processing messages without the ability to recover from this state.
 
- * If using [NServiceBus Host](/nservicebus/hosting/nservicebus-host), and wish to take a custom action before the endpoint process is killed.
- * If self hosting the process can be shut down the process via [Environment.FailFast](https://msdn.microsoft.com/en-us/library/dd289240.aspx) and re-start the process once the root cause has been diagnosed.
+NOTE: If not killing the process and just disposing the bus, be aware that any `Send` operations will result in [ObjectDisposedException](https://msdn.microsoft.com/en-us/library/system.objectdisposedexception.aspx) being thrown.
+
+When the [NServiceBus Host](/nservicebus/hosting/nservicebus-host) is used it is not required to override this callback as the host has a default implementation that exits the process. However, if it is required to execute tasks just before shutdown then it can be overriden in the same way as when self hosting.
```
SimonCropp: this content should exists here https://docs.particular.net/nservicebus/hosting/nservicebus-host/#default-critical-error-action-handling and only be linked to from this page

nservicebus/hosting/critical-errors.md
```diff
@@ -46,12 +46,22 @@ snippet:CustomHostErrorHandlingAction
 
 ## When to override the default action
 
-The default action should be overridden whenever that default does not meet the specific hosting requirements. For example
+When self hosting the critical error callback must be overriden in order to exit the current process. Not overriding the critical error callback will result in the endpoint instance to stop processing messages without the ability to recover from this state.
```
SimonCropp: overridden not overriden

nservicebus/hosting/critical-errors.md
```diff
@@ -46,12 +46,22 @@ snippet:CustomHostErrorHandlingAction
...
+NOTE: If not killing the process and just disposing the bus, be aware that any `Send` operations will result in [ObjectDisposedException](https://msdn.microsoft.com/en-us/library/system.objectdisposedexception.aspx) being thrown.
+
+When the [NServiceBus Host](/nservicebus/hosting/nservicebus-host) is used it is not required to override this callback as the host has a default implementation that exits the process. However, if it is required to execute tasks just before shutdown then it can be overriden in the same way as when self hosting.
+
+Consider the following things when implementing a critical error callback:
+
+- Wrap any tasks in a `try..finally`, making sure the process will exit if any tasks fail.
+- Flush any loggers used which makes sure any unwritten log state is written/pushed to its target(s) making sure the tail of the log is not lost as [Environment.FailFast](https://msdn.microsoft.com/en-us/library/dd289240.aspx) will immediately exit the process.
+ - Flush NLog or log4net state by calling `LogManager.Shutdown();`, this still isn't 100% guaranteed to not loose any log entries. Check each logging framework its document.
```
SimonCropp: instead link to the doco for the explicit NLog or log4net shutdown doco

SimonCropp: if they shut down the logging what happens if that logging is shared with other components in the same process and they try to write log entries after the shutdown?

SimonCropp: "Check each logging framework its document."  this sentence doesnt make sense

nservicebus/hosting/critical-errors.md
```diff
@@ -46,12 +46,22 @@ snippet:CustomHostErrorHandlingAction
...
+- Flush any state/caches used if normally such data structures are would persisted in a certain interval or at graceful shutdown.
```
SimonCropp: this sentence makes no sense

nservicebus/hosting/critical-errors.md
SimonCropp: link to https://docs.particular.net/nservicebus/logging/?

SimonCropp: you mention NLog and log4net. what about all the other logging lobs that we support?

nservicebus/hosting/critical-errors.md
```diff
@@ -46,12 +46,22 @@ snippet:CustomHostErrorHandlingAction
...
 
- * If using [NServiceBus Host](/nservicebus/hosting/nservicebus-host), and wish to take a custom action before the endpoint process is killed.
- * If self hosting the process can be shut down the process via [Environment.FailFast](https://msdn.microsoft.com/en-us/library/dd289240.aspx) and re-start the process once the root cause has been diagnosed.
+NOTE: If not killing the process and just disposing the bus, be aware that any `Send` operations will result in [ObjectDisposedException](https://msdn.microsoft.com/en-us/library/system.objectdisposedexception.aspx) being thrown.
+
+When the [NServiceBus Host](/nservicebus/hosting/nservicebus-host) is used it is not required to override this callback as the host has a default implementation that exits the process. However, if it is required to execute tasks just before shutdown then it can be overriden in the same way as when self hosting.
+
+Consider the following things when implementing a critical error callback:
+
+- Wrap any tasks in a `try..finally`, making sure the process will exit if any tasks fail.
```
SimonCropp: "tasks" is an overloaded word in .net since we are now in an async world

SimonCropp: note that this is redundant of people follow the `Environment.FailFast` direction below

nservicebus/hosting/critical-errors.md
SimonCropp: tasks is an overloaded term and should be avoided

nservicebus/hosting/critical-errors.md
```diff
@@ -46,12 +46,22 @@ snippet:CustomHostErrorHandlingAction
...
 
- * If using [NServiceBus Host](/nservicebus/hosting/nservicebus-host), and wish to take a custom action before the endpoint process is killed.
- * If self hosting the process can be shut down the process via [Environment.FailFast](https://msdn.microsoft.com/en-us/library/dd289240.aspx) and re-start the process once the root cause has been diagnosed.
+NOTE: If not killing the process and just disposing the bus, be aware that any `Send` operations will result in [ObjectDisposedException](https://msdn.microsoft.com/en-us/library/system.objectdisposedexception.aspx) being thrown.
+
+When the [NServiceBus Host](/nservicebus/hosting/nservicebus-host) is used it is not required to override this callback as the host has a default implementation that exits the process. However, if it is required to execute tasks just before shutdown then it can be overriden in the same way as when self hosting.
+
+Consider the following things when implementing a critical error callback:
```
SimonCropp: things  is redundant

nservicebus/hosting/critical-errors.md
```diff
@@ -46,12 +46,22 @@ snippet:CustomHostErrorHandlingAction
...
+
+- Wrap any tasks in a `try..finally`, making sure the process will exit if any tasks fail.
+- Flush any loggers used which makes sure any unwritten log state is written/pushed to its target(s) making sure the tail of the log is not lost as [Environment.FailFast](https://msdn.microsoft.com/en-us/library/dd289240.aspx) will immediately exit the process.
+ - Flush NLog or log4net state by calling `LogManager.Shutdown();`, this still isn't 100% guaranteed to not loose any log entries. Check each logging framework its document.
+- Flush any state/caches used if normally such data structures are would persisted in a certain interval or at graceful shutdown.
+- It is easiest to call [Environment.FailFast](https://msdn.microsoft.com/en-us/library/dd289240.aspx)
```
gbiellem: Do we really want to recommending calling FailFast witthout listing the ramifications?  Given it bypasses any finalisation and also triggers Window Error Reporting

SimonCropp:
@Particular/nservicebus-maintainers given this is tweaking recommendations about the shut down of an endpopint would like a review please

nservicebus/hosting/critical-errors.md
```diff
@@ -46,12 +46,22 @@ snippet:CustomHostErrorHandlingAction
...
+
+Rely on the environment that the process will be automatically restarted. When hosting in IIS the IIS host will automatically spawn a new instance and when hosting as a Windows Service the OS will restart the service after 1 minute if [Windows Service Recovery](/nservicebus/hosting/windows-service.md#installation-restart-recovery) is enabled.
```
gbiellem: I think what your trying to say here is that where there is a option to rely on an Windows capability to restart the process then make use of it.  Needs rewording to get that across.  At the moment it reads like dot points and does not flow

nservicebus/hosting/critical-errors.md
timbussmann: can this be reformulated into something like "the default behavior is to stop the endpoint, not the process. When self hosting endpoints, it's recommended override the CE action to exit the process"?

timbussmann: should this section rather be in the self hosting docs?

SimonCropp: is this content correct for all versions of NSB?

ramonsmits: AFAIK yes, all versions have this issue.

nservicebus/hosting/critical-errors.md
ramonsmits: @SimonCropp Our default implementation is writing immediately. Why link to our logging page/

ramonsmits: @SimonCropp Not sure what happens the but I guess nothing happens but that isn't really an issue when doing an Environment.FailFast right after the shutdown.

ramonsmits: Good question, I only really worked with log4net and NLog and only experimented with Serilog. The main bullet says what to do, the sub item giving a hint for these frameworks. We can remove the hint if you want.

nservicebus/hosting/critical-errors.md
ramonsmits: @SimonCropp Updated that line.

nservicebus/hosting/critical-errors.md
ramonsmits: @gbiellem I believe so, as it is the only way to make sure we exit as NServiceBus as we are FUBAR hence the reason mentioning that if the process requires any additional shutdown code to be invoked to do that before it but yes, any thread running are terminated too. Added the regular .Exit too.

nservicebus/hosting/critical-errors.md
ramonsmits: @timbussmann added note

ramonsmits:
@timbussmann @SimonCropp @gbiellem @indualagarsamy thanks, I think I addressed your feedback.

nservicebus/hosting/critical-errors.md
gbiellem: @ramonsmits I get that failfast can makes sense, my concern was that the recommendation doesn't spell out the ramifications of using it.  Although the link you've provided does so perhaps I'm being overly pedantic

nservicebus/hosting/critical-errors.md
```diff
@@ -44,14 +44,29 @@ Next define what action to take when this scenario occurs:
 snippet:CustomHostErrorHandlingAction
 
 
-## When to override the default action
+## When to override the default critical error action
 
-The default action should be overridden whenever that default does not meet the specific hosting requirements. For example
+The default behavior is to stop the endpoint and not the process. It is recommended to exit the process when a critical error occurs. Not overriding the critical error callback will result in the endpoint instance to stop processing messages without the ability to recover from this state.
```
weralabaj: > The default behavior is to stop the endpoint and not the process. It is recommended to exit the process when a critical error occurs. Not overriding the critical error callback will result in the endpoint instance to stop processing messages without the ability to recover from this state.

To be honest I'm not sure if I understand this paragraph. Are we essentially saying that the default implementation is broken and you should always override it, otherwise you won't be able to recover? Is that true for all hosts?

nservicebus/hosting/critical-errors.md
```diff
@@ -44,14 +44,29 @@ Next define what action to take when this scenario occurs:
...
 
- * If using [NServiceBus Host](/nservicebus/hosting/nservicebus-host), and wish to take a custom action before the endpoint process is killed.
- * If self hosting the process can be shut down the process via [Environment.FailFast](https://msdn.microsoft.com/en-us/library/dd289240.aspx) and re-start the process once the root cause has been diagnosed.
+NOTE: If not killing the process and just disposing the bus, be aware that any `Send` operations will result in [ObjectDisposedException](https://msdn.microsoft.com/en-us/library/system.objectdisposedexception.aspx) being thrown.
```
weralabaj: > If not killing the process and just disposing the bus

I know it's been here before, but I don't get this info. Is that the default behaviour? Is that a potential custom implementation? If the latter, should we show snippet here to explain what do we mean by that?

nservicebus/hosting/critical-errors.md
```diff
@@ -44,14 +44,29 @@ Next define what action to take when this scenario occurs:
...
+
+Consider the following when implementing a critical error callback:
+
+- If calling any code wrap it in a `try..finally`, making sure the process will invoke exit code in the `finally` like `Environment.FailFast` if any code fails.
+- Flush any loggers used which makes sure any unwritten log state is written/pushed to its target(s) making sure the tail of the log is not lost as [Environment.FailFast](https://msdn.microsoft.com/en-us/library/dd289240.aspx) will immediately exit the process.
+ - Flush appenders for [NLog](http://nlog-project.org/documentation/v4.3.0/html/M_NLog_LogManager_Shutdown.htm) or [log4net](https://logging.apache.org/log4net/log4net-1.2.11/release/sdk/log4net.LogManager.Shutdown.html) state by calling `LogManager.Shutdown();`.
+- Flush any state/caches used if such data structures would be written/persisted at a certain interval or regular graceful shutdown.
+- It is easiest to call [Environment.FailFast](https://msdn.microsoft.com/en-us/library/dd289240.aspx) or [Environment.Exit](https://msdn.microsoft.com/en-us/library/system.environment.exit(v=vs.110).aspx).
+
```
weralabaj: When implementing custom critical error:
- Wrap code in a `try..finally` block. The process responsible for exiting the process, such as [`Environment.FailFast`](https://msdn.microsoft.com/en-us/library/dd289240.aspx) or [`Environment.Exit`](https://msdn.microsoft.com/en-us/library/system.environment.exit(v=vs.110).aspx) should be placed in the `finally` block, which will be executed upon critical error.
- Flush any loggers used, to ensure the unwritten log state is pushed to its target(s).
- Flush appenders for [NLog](http://nlog-project.org/documentation/v4.3.0/html/M_NLog_LogManager_Shutdown.htm) or [log4net](https://logging.apache.org/log4net/log4net-1.2.11/release/sdk/log4net.LogManager.Shutdown.html) state by calling `LogManager.Shutdown();`.

BTW seems here https://docs.particular.net/nservicebus/hosting/critical-errors#a-possible-custom-implementation we don't follow our own recommendations. Should we also update this snippet?

weralabaj: > Flush any state/caches used if such data structures would be written/persisted at a certain interval or regular graceful shutdown.

Example? From the user perspective I'm not sure what I should look out for...

> Flush any loggers used which makes sure any unwritten log state... 
> Flush appenders for [NLog](http://nlog-project.org/documentation/v4.3.0/html/M_NLog_LogManager_Shutdown.htm) or [log4net](https://logging.apache.org/log4net/log4net-1.2.11/release/sdk/log4net.LogManager.Shutdown.html) ...
> Flush any state/caches used if such data structures would be written/persisted...

Should that all be done in the finally block? It's not clear.

weralabaj: Overall, I think we should add a snippet showing all this in practice.

nservicebus/hosting/critical-errors.md
```diff
@@ -44,14 +44,29 @@ Next define what action to take when this scenario occurs:
...
+
+Rely on the environment hosting the endpoint process for it to be automatically restarted.
```
weralabaj: > Rely on the environment hosting the endpoint process for it to be automatically restarted

What are we trying to say here? That we should try to avoid having to restart processes manually, when they crash? What about other hosting options (other than IIS, Windows Service with Windows Service Recovery enabled), should we in other cases manually restart it?

nservicebus/hosting/index.md
```diff
@@ -27,6 +27,8 @@ There are several approaches to hosting.
  * [Endpoint Lifecycle](/nservicebus/lifecycle/)
  * [Critical Error handling](critical-errors.md)
 
+NOTE: Be aware of critical errors, not configuring a critical error callback makes an endpoints unresponsive after a critical error without terminating the process.
```
weralabaj: Are we documenting a bug here? Should we force users to always provide the custom implementation?

nservicebus/hosting/nservicebus-host/index.md
```diff
@@ -94,7 +94,7 @@ The default [Critical Error Action](/nservicebus/hosting/critical-errors.md) for
 
 snippet:DefaultHostCriticalErrorAction
 
-WARNING: It is important to consider the effect these defaults will have on other things hosted in the same process. For example if co-hosting NServiceBus with a web-service or website.
+It is not required to override this callback. However, if it is required to execute code just before shutdown then it can be overridden in the same way as when self hosting. See [When to override the default critical error action](/nservicebus/hosting/critical-errors.md#when-to-override-the-default-critical-error-action) for guidance.
```
weralabaj: > It is not required to override this callback.

So in this case we won't have the problem with stopping the endpoint without stopping the process? If so, we should make clear in the hosting article which hosting options have this problem, as it's not clear.

nservicebus/hosting/critical-errors.md
```diff
@@ -44,14 +44,29 @@ Next define what action to take when this scenario occurs:
...
+
+- IIS: When hosting in IIS the IIS host will automatically spawn a new instance
+- Windows Service: The OS can restart the service after 1 minute if [Windows Service Recovery](/nservicebus/hosting/windows-service.md#installation-restart-recovery) is enabled.
+
+
+WARNING: It is important to consider the effect these defaults will have on other things hosted in the same process. For example if co-hosting NServiceBus with a web-service or website.
```
weralabaj: Do we have any resources that give more information on those considerations (e.g. on MSDN)?

weralabaj:
@ramonsmits I'm coming without much context, so if we need to talk it over the call just ping me, I had a few comments

nservicebus/hosting/critical-errors.md
weralabaj: Why remove the old explanation with examples? Is it somehow incorrect?

nservicebus/hosting/index.md
ramonsmits: @weralabaj well.... https://github.com/Particular/NServiceBus/issues/3612

> IMHO NServiceBus should be stable by default and not exit / fail-fast due to a few errors and even if we do so, this behavior should not be hidden. It should be an explicit opt-in/out as you would not expect that a framework just 'kills' the application.

weralabaj: I looked at the issue, so that won't be changed, correct? We're documenting broken behaviour? Just trying to understand :)

weralabaj: Either way we need to be more explicit in doco, if that's broken by default then shouldn't be custom implementation obligatory?

nservicebus/hosting/nservicebus-host/index.md
weralabaj: By default you don't have to override it with NSB Host, but in the situations described in critical errors page you should provide your own implementation for graceful shutdown

nservicebus/hosting/critical-errors.md
weralabaj: This is the default implementation => check it out

Snippets/Core/Core_4/CriticalErrors.cs
```diff
@@ -23,13 +23,19 @@ class CriticalErrors
...
+            {
+                // To leave the process active, dispose the bus.
+                // When the bus is disposed, the attempt to send message will cause an ObjectDisposedException.
+                bus.Dispose();
+                // Perform custom actions here, e.g.
+                // NLog.LogManager.Shutdown();
+            }
+            finally
+            {
+                var failMessage = $"Critical error shutting down:'{errorMessage}'.";
```
ramonsmits: Log statement is missing here compared to the V3 snippet.

Snippets/Core/Core_3/CriticalErrors.cs
```diff
@@ -26,16 +26,20 @@ class CriticalErrors
...
+                // To leave the process active, dispose the bus.
+                // When the bus is disposed, the attempt to send message will cause an ObjectDisposedException.
+                ((IDisposable)bus).Dispose();
+                // Perform custom actions here, e.g.
+                // NLog.LogManager.Shutdown();
+            }
+            finally
+            {
+                // Write log entry in version 3 since this is not done by default.
+                log.Fatal("CRITICAL Error");
```
ramonsmits: Can't log after suggested logger shutdown on line 35

Snippets/Core/Core_6/CriticalErrors.cs
```diff
@@ -20,15 +20,21 @@ class CriticalErrors
...
-
-            // To kill the process, await the above,
-            // then raise a fail fast error as shown below.
-            // var failMessage = $"Critical error shutting down:'{context.Error}'.";
-            // Environment.FailFast(failMessage, context.Exception);
+            try
+            {
+                // To leave the process active, dispose the bus.
+                // When the bus is disposed, the attempt to send message will cause an ObjectDisposedException.
+                await context.Stop();
```
ramonsmits: Maybe add `.ConfigureAwait(false)`?

nservicebus/hosting/critical-errors.md
```diff
@@ -43,14 +43,28 @@ Next define what action to take when this scenario occurs:
...
+ * When self hosting.
 
-NOTE: If not killing the process and just dispose the bus, be aware that any `Send` operations will result in [ObjectDisposedException](https://msdn.microsoft.com/en-us/library/system.objectdisposedexception.aspx) being thrown.
+Warning: The default action should be always overriden when self-hosting NServiceBus, as by default when a critical error occurs NServiceBus will stop the endpoint without exiting the process.
+
+NOTE: If the endpoint is stopped without exiting the process, then any `Send` operations will result in [ObjectDisposedException](https://msdn.microsoft.com/en-us/library/system.objectdisposedexception.aspx) being thrown.
+
+When implementing a custom critical error callback:
+
+- To exit the process use the [Environment.FailFast](https://msdn.microsoft.com/en-us/library/dd289240.aspx) method. In case of send-only endpoints, also [Environment.Exit](https://msdn.microsoft.com/en-us/library/system.environment.exit(v=vs.110).aspx) may be used.
```
ramonsmits: Send only endpoints are often used in web applications. It's not to say send-only endpoints should use `Environment.Exit`.  If an environment has threads running that should be competed first (non transactional stuff) then `.Exit` should be used.

nservicebus/hosting/critical-errors.md
```diff
@@ -43,14 +43,28 @@ Next define what action to take when this scenario occurs:
...
+- The code should be wrapped in a `try...finally` clause. In the `try` block perform any custom operations, in the `finally` block call the method that exits the process.
+- The custom operations should include flushing any in-memory state and cached data, if normally its persisted at a certain interval or during graceful shutdown. For example, flush appenders for [NLog](http://nlog-project.org/documentation/v4.3.0/html/M_NLog_LogManager_Shutdown.htm) or [log4net](https://logging.apache.org/log4net/log4net-1.2.11/release/sdk/log4net.LogManager.Shutdown.html) state by calling `LogManager.Shutdown();`.
```
ramonsmits: > For example, flush appenders *if you are using buffering or asynchrounous appanders*

nservicebus/hosting/index.md
```diff
@@ -27,6 +27,8 @@ There are several approaches to hosting.
  * [Endpoint Lifecycle](/nservicebus/lifecycle/)
  * [Critical Error handling](critical-errors.md)
 
+Note: Override the default critical error callback when self-hosting NServiceBus. By default NServiceBus will stop the endpoint without exiting the process. Refer to the [Critical Errors](/nservicebus/hosting/critical-errors.md) article for more information.
```
ramonsmits: Should be use *endpoint instance* instead of *endpoint*?

Snippets/Core/Core_3/CriticalErrors.cs
ramonsmits: If logging, that should be done within the `try` as that call could fail and then `FailFast` will not be invoked.

PR #2406, opened 2017-02-03
Removed "using in a saga"

udidahan:
Sagas shouldn't be talking to databases - only managing their own state. As such, showing people how to use it in a saga is guiding them down the wrong path.

### DIFF
--- a/nservicebus/sql-persistence/index.md
+++ b/nservicebus/sql-persistence/index.md
@@ -112,7 +112,3 @@ The current [DbConnection](https://msdn.microsoft.com/en-us/library/system.data.
 
 snippet: handler-sqlPersistenceSession
 
-
-### Using in a Saga
-
-snippet: saga-sqlPersistenceSession


SimonCropp:
@udidahan in the interest of "making developers better" i would prefer to leave the snippet and add a [warning](https://github.com/Particular/docs.particular.net#alerts) explaining why it is not a recommended approach. Removing the snippet means they can still discover the API in a saga and use it without know it is an anti pattern

udidahan:
That makes sense. Closing.

udidahan:
Created a PR to describe this in the main saga docs:

SimonCropp:
@udidahan how is this?

https://docs.particular.net/nservicebus/nhibernate/accessing-data#using-in-a-saga
https://docs.particular.net/nservicebus/ravendb/#shared-session-using-in-a-saga
https://docs.particular.net/nservicebus/sql-persistence/#sqlstoragesession-using-in-a-saga

SimonCropp:
the include is here if you want to tweak the text https://github.com/Particular/docs.particular.net/blob/master/nservicebus/sagas/saga-business-data-access.include.md

udidahan:
I've made some additional minor modifications - committed directly to master. Take a look and see if it breaks any of the conventions of docs.

SimonCropp:
@udidahan looks all good

PR #2407, opened 2017-02-03
Data distribution warning, remove related Azure article

### LINKED ISSUE #2375: Using Data Distribution using NServiceBus
In a discussion earlier today with @DavidBoike and @udidahan and others in the unconference session, @udidahan mentioned that Data Distribution is a different style of architecture. 

Step 1: Add a disclaimer in the sample: 
https://docs.particular.net/samples/routing/data-distribution/
https://docs.particular.net/nservicebus/azure/data-distribution

~~Step 2: Come up with a general guidance for the comparison on both.~~ *Covered by #2410*

cc / @SzymonPobiega

DavidBoike:
Fixes #2375 

(At least, the start of a conversation)

* I added the warning we talked about in Lisbon to the [Data Distribution sample](https://docs.particular.net/samples/routing/data-distribution/)
* For the data distribution article under Azure, I was seriously conflicted, but after deliberation decided to remove it - or at least (via this PR) propose doing so.
  * The [V6 version](https://docs.particular.net/nservicebus/azure/data-distribution), which is default, doesn't really say much _at all_ and then links to the other sample, which is not even an Azure sample.
  * The [V5 version](https://docs.particular.net/nservicebus/azure/data-distribution?version=core_5) has two small code snippets, that do not use the Azure NuGet packages at all, but I think it's rather dubious, and we don't support that going forward anyway. It raises a lot of questions, and currently doesn't paint a clear picture of anything.

If we agree on removing that article, I don't know precisely what needs to be done for removed articles that have no replacement.


DavidBoike:
I know a lot of people were present for that discussion, but I know that @indualagarsamy you were present, you raised #2375, and you're a Docs Maintainer, so I'll ask you first. What do you think?

indualagarsamy:
@DavidBoike - Yes, I raised 2375 based on our discussion in Lisbon. The main idea for 2375 was to add a disclaimer note and add a architecture guidance article of some sort for which  Udi wanted a different issue.
Adding a disclaimer is good. 
Regarding removing the related Azure article, I think that we need a conversation with @Particular/azure-maintainers and definitely @yvesgoeleven before we can make that call because I think it was added in the first place to solve something else. Does that make sense? 
@yvesgoeleven @Particular/azure-maintainers - can you please chime in. If necessary, please get on a call with Udi.

samples/routing/data-distribution/sample.md
```diff
@@ -8,6 +8,8 @@ tags:
 - Data distribution
 ---
 
+DANGER: Asynchronous messaging is not an optimal solution for data distribution scenarios. It is usually better to use a dedicated data distribution technology for that purpose, such as a distributed cache or distributed configuration service. This sample aims to serve only as an example of the flexibility in the NServiceBus routing engine.
```
indualagarsamy: @DavidBoike - this disclaimer looks good to me and captures what we wanted in the Lisbon meeting.

indualagarsamy:
@DavidBoike - two things. 

>For the data distribution article under Azure, I was seriously conflicted, but after deliberation decided to remove it - or at least (via this PR) propose doing so.

This would have been better as an issue first, where we could've asked the @Particular/azure-maintainers before doing this. 

If you'd added just the disclaimer, we could've merged the PR. Small PRs always for the win. And issues when you're conflicted :)

yvesgoeleven:
@indualagarsamy of course I feel sad that this content has to go. It reduces the value of nservicebus in azure if we don't want to support this use case. Frontend cache invalidation based on data distribution is quite a common use case as there simply is no decent caching solution in a PAAS world. Resource throttling does not bode well for caches, caches are there to take traffic spikes in frontends or api's, so throttling at a peak moment is the last thing you want your cache to do at that moment. But if it is decided not to support this, then so be it, remove it.

SeanFeldman:
I'm not sure I like the "let's just rip it out" approach. If this is Azure specific only, perhaps move it into Azure section. And add a warning if this is not aligned with the general NSB approach.

BTW, one of the samples we have is all about Data Distribution with ASB. Even though the name does it say so: [Custom ASB Namespace Partitioning](https://docs.particular.net/samples/azure/custom-partitioning-asb/).

indualagarsamy:
@yvesgoeleven @SeanFeldman  - 

>And add a warning if this is not aligned with the general NSB approach.

The issue i raised in docs was meant to do just that. We did not talk about removing the samples in Lisbon. 

@DavidBoike - Based on the input from @SeanFeldman and @yvesgoeleven, I am closing this PR. Please raise a separate PR to add just the disclaimer.

PR #2415, opened 2017-02-08
Add compatibility section to the NSB.Encryption.MessagePropertyEncryption upgrade guide

timbussmann:


nservicebus/upgrades/externalize-encryption.md
```diff
@@ -36,4 +35,23 @@ snippet: SplitDefiningEncryptedPropertiesAs
...
+snippet: SplitEncryptionFromIEncryptionService
+
+
+## Compatibility
+
+The NServiceBus.Encryption.MessageProperty package is partially compatible with endpoints using NServiceBus core's encryption functionality:
+
+
+### Sending and receiving using NServiceBus.Encryption.MessageProperty
+* NServiceBus.Encryption.MessageProperty can de- and encrypt all messages with message properties of type `NServiceBus.WireEncryptedString`.
```
SimonCropp: re `de- and encrypt` i think `decrypt and encrypt` reads better

nservicebus/upgrades/externalize-encryption.md
```diff
@@ -11,7 +11,6 @@ related:
 
 The [Message Property Encryption feature](/nservicebus/security/property-encryption.md), has been moved from the NServiceBus core to the separate NuGet package [NServiceBus.Encryption.MessageProperty](https://www.nuget.org/packages/NServiceBus.Encryption.MessageProperty/). That package should be used in order to encryption message properties when using versions NServiceBus Versions 6.2 and above.
```
adamralph: Suggestion: "from the NServiceBus ~core~*package*"

nservicebus/upgrades/externalize-encryption.md
```diff
@@ -36,4 +35,23 @@ snippet: SplitDefiningEncryptedPropertiesAs
...
 ## Custom encryption Service
 
-snippet: SplitEncryptionFromIEncryptionService
\ No newline at end of file
+snippet: SplitEncryptionFromIEncryptionService
+
+
+## Compatibility
+
+The NServiceBus.Encryption.MessageProperty package is partially compatible with endpoints using NServiceBus core's encryption functionality:
```
adamralph: Suggestion: "using *the* NServiceBus ~core~package's encryption functionality"

nservicebus/upgrades/externalize-encryption.md
```diff
@@ -36,4 +35,23 @@ snippet: SplitDefiningEncryptedPropertiesAs
...
+
+
+### Sending and receiving using NServiceBus.Encryption.MessageProperty
+* NServiceBus.Encryption.MessageProperty can de- and encrypt all messages with message properties of type `NServiceBus.WireEncryptedString`.
+* NServiceBus.Encryption.MessageProperty can de- and encrypt all messages with message properties of type `NServiceBus.Encryption.MessageProperty.WireEncryptedString`.
+* NServiceBus.Encryption.MessageProperty can de- and encrypt all messages using an encrypted property convention.
+
+
+### Sending and receiving using NServiceBus core
```
adamralph: Given the previous section is titled "Sending and receiving using NServiceBus.Encryption.MessageProperty", I think the context is already set for "NServiceBus" to mean the `NServiceBus` package, so I think "core" can probably be omitted here.

nservicebus/upgrades/externalize-encryption.md
timbussmann: good with that, will change 👍

timbussmann:
addressed all comments.

nservicebus/upgrades/externalize-encryption.md
timbussmann: should it actually be `encrypt and decrypt` then?

nservicebus/upgrades/externalize-encryption.md
```diff
@@ -36,4 +35,23 @@ snippet: SplitDefiningEncryptedPropertiesAs
...
\ No newline at end of file
+snippet: SplitEncryptionFromIEncryptionService
+
+
+## Compatibility
+
+The NServiceBus.Encryption.MessageProperty package is partially compatible with endpoints using NServiceBus package's encryption functionality:
+
+
+### Sending and receiving using NServiceBus.Encryption.MessageProperty
```
weralabaj: > Sending and receiving using NServiceBus.Encryption.MessageProperty

Do we actually send and receive messages with this package, or should it be "Sending and receiving messages with properties encrypted using `NServiceBus.Encryption.MessageProperty`" or something like that?

nservicebus/upgrades/externalize-encryption.md
```diff
@@ -36,4 +35,23 @@ snippet: SplitDefiningEncryptedPropertiesAs
...
+* NServiceBus.Encryption.MessageProperty can decrypt and encrypt all messages with message properties of type `NServiceBus.Encryption.MessageProperty.WireEncryptedString`.
+* NServiceBus.Encryption.MessageProperty can decrypt and encrypt all messages using an encrypted property convention.
+
+
+### Sending and receiving using NServiceBus
+* NServiceBus can decrypt and encrypt all messages with message properties of type `NServiceBus.WireEncryptedString`.
+* NServiceBus can decrypt and encrypt all messages using an encrypted property convention.
+* NServiceBus.Encryption.MessageProperty **cannot** decrypt and encrypt messages with message properties of type `NServiceBus.Encryption.MessageProperty.WireEncryptedString`.
+
+When migrating endpoints to use the NServiceBus.Encryption.MessageProperty package, message contracts should only be updated to use the `NServiceBus.Encryption.MessageProperty.WireEncryptedString` property type when all endpoints are migrated.
```
weralabaj: I'd rephrase this section like this:
1. Upgrade all the endpoints to use the new package. The endpoints will continue to process all messages the same way as they've done until then.
2. Upgrade message contracts to use the new package.

weralabaj: Any comments regarding what to do when you can't upgrade all endpoints at once?
What about in-flight messages? Should we make an explicit note to allow some time before steps 1 and 2?

nservicebus/upgrades/externalize-encryption.md
weralabaj: Also I wonder if it'd be simpler to either only say this "NServiceBus.Encryption.MessageProperty **cannot** decrypt and encrypt messages with message properties of type `NServiceBus.Encryption.MessageProperty.WireEncryptedString`" or have sections "Supported" & "not supported". It's quite easy to miss the bolded **cannot** word right now.

nservicebus/upgrades/externalize-encryption.md
```diff
@@ -9,8 +9,7 @@ related:
  - nservicebus/security/property-encryption
 ---
 
-The [Message Property Encryption feature](/nservicebus/security/property-encryption.md), has been moved from the NServiceBus core to the separate NuGet package [NServiceBus.Encryption.MessageProperty](https://www.nuget.org/packages/NServiceBus.Encryption.MessageProperty/). That package should be used in order to encryption message properties when using versions NServiceBus Versions 6.2 and above.
-
+The [Message Property Encryption feature](/nservicebus/security/property-encryption.md), has been moved from the NServiceBus package to the separate [NServiceBus.Encryption.MessageProperty](https://www.nuget.org/packages/NServiceBus.Encryption.MessageProperty/) NuGet package. That package should be used in order to encryption message properties when using versions NServiceBus Versions 6.2 and above.
```
kbaley: > That package should be used in order to encryption message properties

Minor nit. I'd also remove "in order"

nservicebus/upgrades/externalize-encryption.md
kbaley: I'd consider something like "Encrypting and decrypting with NServiceBus.Encryption.MessageProperty" for the title since the section doesn't have anything to do with sending/receiving. Maybe even "Supported types when encrypting and decrypting with NServiceBus.Encryption.MessageProperty"

timbussmann: 👍 to the title changes

> It's quite easy to miss the bolded cannot word right now.

hm not sure I agree with that. Would it be good enough to move it on top of the bullet points?

nservicebus/upgrades/externalize-encryption.md
timbussmann: > Any comments regarding what to do when you can't upgrade all endpoints at once?

then you should just stick with the existing message contracts which are compatible with the plugin.

> What about in-flight messages? 

nothing specific to this plugin. The endpoints using the plugin are compatible. It could be a general problem in case the message contract was adjusted instead of creating a new message contract though?

nservicebus/upgrades/externalize-encryption.md
```diff
@@ -9,14 +9,21 @@ related:
...
 
 ## Removed APIs
 
-Configuring encryption [Via app.config](/nservicebus/security/property-encryption.md#configuration-via-app-config) and [Via IProvideConfiguration](/nservicebus/security/property-encryption.md#configuration-via-iprovideconfiguration) have been removed. Instead use [Configuration via code](/nservicebus/security/property-encryption.md#configuration-via-code).
+Configuring encryption [via app.config](/nservicebus/security/property-encryption.md#configuration-via-app-config) and [via IProvideConfiguration](/nservicebus/security/property-encryption.md#configuration-via-iprovideconfiguration) have been removed. Instead use [configuration via code](/nservicebus/security/property-encryption.md#configuration-via-code).
+
+
+## Migration steps
+
+When migrating endpoints to use the `NServiceBus.Encryption.MessageProperty` package:
```
kbaley: @timbussmann Should we make it clear that the user needs to add the nuget package, update the configuration, and deploy the updated endpoints as part of step 1?

DavidBoike:
So the compatibility section spells out the behavior in each individual permutation, but doesn't lay out the guiding principle (the why) that causes it and would make it possible for you to remember which case is which. Basically...

Because the core implementation is unaware of the existence of the external package, it is unable to decrypt messages using the external package's implementation of WiteEncryptedString. Therefore:

<individual cases>

timbussmann:
will add some background to the compatibility reasons @DavidBoike. Currently await a PR which slightly changes some public APIs and this PR needs to be changed because of that anyway.

nservicebus/upgrades/externalize-encryption.md
```diff
@@ -9,14 +9,27 @@ related:
...
+
+
+## Migration example
+
+For a system with two or more endpoints, these are the steps to migrate to the `NServiceBus.Encryption.MessageProperty` package:
+
+ 1. Install the `NServiceBus.Encryption.MessageProperty` NuGet package into all endpoints.
+ 1. Update the configuration for all endpoints to use either [RijndaelEncryptionService](#enabling-rijndaelencryptionservice) or [a custom encryption service](#custom-encryption-service).
+ 1. Deploy all endpoints.
+ 1. Update message contracts in all endpoints to use the `NServiceBus.Encryption.MessageProperty.WireEncryptedString` property type.
```
indualagarsamy: Shouldn't this step come in before you deploy all endpoints?

nservicebus/upgrades/externalize-encryption.md
```diff
@@ -9,14 +9,27 @@ related:
...
+ 1. Deploy all endpoints.
```
indualagarsamy: is this a duplicate?

nservicebus/upgrades/externalize-encryption.md
```diff
@@ -29,11 +42,31 @@ snippet: SplitEncryptionFromCode
...
-## Custom encryption Service
+## Custom encryption service
+
+snippet: SplitEncryptionFromIEncryptionService
+
+
+
+## Compatibility
+
+The NServiceBus.Encryption.MessageProperty package is partially compatible with endpoints using NServiceBus package's encryption functionality:
```
indualagarsamy: What does partially compatible with endpoints mean? Can you add more here?

kbaley:
@DavidBoike @indualagarsamy Added some clarification on the compatibility

/cc @timbussmann

PR #2424, opened 2017-02-10
Tidy up MVC Send sample

mikeminutillo:
The endpoint no longer needs a custom container because we aren't updating it.

The MVC application container was still called updated based on when we were updating the Endpoint container. If we are no longer doing that, it should be called something else.

### DIFF
--- a/samples/web/send-from-mvc-controller/Core_6/WebApplication/Global.asax.cs
+++ b/samples/web/send-from-mvc-controller/Core_6/WebApplication/Global.asax.cs
@@ -34,15 +34,15 @@ public class MvcApplication :
 
         endpoint = Endpoint.Start(endpointConfiguration).GetAwaiter().GetResult();
 
-        var updater = new ContainerBuilder();
-        updater.RegisterInstance(endpoint);
+        var mvcContainerBuilder = new ContainerBuilder();
+        mvcContainerBuilder.RegisterInstance(endpoint);
 
         // Register MVC controllers.
-        updater.RegisterControllers(typeof(MvcApplication).Assembly);
+        mvcContainerBuilder.RegisterControllers(typeof(MvcApplication).Assembly);
 
-        var updated = updater.Build();
+        var mvcContainer = mvcContainerBuilder.Build();
 
-        DependencyResolver.SetResolver(new AutofacDependencyResolver(updated));
+        DependencyResolver.SetResolver(new AutofacDependencyResolver(mvcContainer));
 
         AreaRegistration.RegisterAllAreas();
         RegisterRoutes(RouteTable.Routes);
--- a/samples/web/send-from-mvc-controller/Core_6/WebApplication/Global.asax.cs
+++ b/samples/web/send-from-mvc-controller/Core_6/WebApplication/Global.asax.cs
@@ -15,18 +15,8 @@ public class MvcApplication :
     {
         #region ApplicationStart
 
-        var builder = new ContainerBuilder();
-
-        // Set the dependency resolver to be Autofac.
-        var container = builder.Build();
-
         var endpointConfiguration = new EndpointConfiguration("Samples.MvcInjection.WebApplication");
-        // instruct NServiceBus to use a custom AutoFac configuration
-        endpointConfiguration.UseContainer<AutofacBuilder>(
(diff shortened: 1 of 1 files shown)


mikeminutillo:
I'm not sure if we want that 2nd commit (which is why it's separate). Effectively, this sample had two containers and one of them we didn't do anything with. It we do nothing, NSB will handle that for you so that code can be removed.

It might make sense to keep it in and talk about the two containers, why there are two, and what each one is for. 

It still doesn't solve the case where I have dependencies I want to share between handlers and controllers. Arguably, you shouldn't be doing that anyway.

indualagarsamy:
@mikeminutillo - I think having 2 containers is better. NSB should be using its own container to register all the internal stuff it needs. IMO, the only thing that should be registered in the user's container is the IEndpointInstance in V6 and IBus in V5.  Thoughts? I had originally discussed this with @danielmarbach

mikeminutillo:
@indualagarsamy I agree. 

My point was that the sample don't actually *need* to create two containers because NSB will transparently create and manage it's own.

indualagarsamy:
@mikeminutillo - although a lot of the times, to do anything useful, controllers might need access to the IEndpointInstance to send other messages? So I think its beneficial to show the 2 containers, perhaps add a comment as to why we are doing it? Thoughts?

mikeminutillo:
I'm not sure. I think if we're going to have the second custom container we should put some dummy services into each container and show where getting them back out will work and where it won't.

If we put `ISomeService` in the NSB container then we can get it in handlers but not in controllers.

If we put `ISomeService` in the MVC container then we can get it in controllers but not in handlers.

If we put `ISomeService` in BOTH containers then we can get it in both but they are going to be different objects with different lifetimes.

If we register in both containers as a singleton then we STILL get two instances, one out of each container. 

If we really need/want to share access to a single instance then we're better off creating it outside of the containers and registering that single instance in each container with an external lifetime scope!

At that point, are we outside the scope of the original sample?

danielmarbach:
@mikeminutillo is right, if we have two containers the dependencies that need to be shared need to be created by the composition root and passed into the two containers or explicitly managed by one container while the other container registration code resolves from the first one (although not really preferred way to do it since you get weird coupling).

mikeminutillo:
Yeah. You get into weirdness like this:

```csharp
containerBuilder2.Register(b => container1.Resolve<ISomeService>()).ExternallyOwned();
```

WilliamBZA:
Urgh. The more I think about the more I think we need to leave the sample as is. 😞  Removing the call to `Update` means we should direct the users down a path of multiple containers, which at this stage adds a lot of confusion.

weralabaj:
@WilliamBZA you mean "as is" in this PR with Mike's changes, correct?
So @mikeminutillo is that ready to merge? I think we could also add a note that covers https://github.com/Particular/docs.particular.net/pull/2424#issuecomment-278872635, to stress how it works.

WilliamBZA:
@weralabaj No, I mean that the original code with the 2 containers is probably more useful to customers than stripping it down.

If we present the sample with only one container, the customers will have the same questions as us: "How do we share registrations between containers?"

weralabaj:
I'm confused now, but all Mike did was remove some unused code, correct? So should we update the sample to some earlier version where the 2nd container was actually used?

WilliamBZA:
@weralabaj This PR removes a chunk of code which shows customers how they can pass us an existing container, and still register the `IEndpointInstance` in their own container.

indualagarsamy:
@weralabaj - I think what @mikeminutillo has done here is good. For this sample, this doesn't need 2 containers. Doesn't need autofac explicitly built. All its doing is registering IEndpointInstance, so that it can be later used in the controllers. 
I think however, adding a paragraph about how people should do this in the scenarios we explored might be good. I will discuss later in the day with @mikeminutillo.

weralabaj:
@indualagarsamy did you discuss it? how can we proceed?

SimonCropp:
This seems abandoned. Shall we close?

mikeminutillo:
Merged after conversation with @SimonCropp 

I will create another issue to capture the other points in doco (and not just in a sample)

PR #2427, opened 2017-02-10
Automate support dates

mikeminutillo:
- Extracts the supported versions into an include
- Automates the generation of the include using data from NuGet

### DIFF
--- a/tools/supportedVersions.linq
+++ b/tools/supportedVersions.linq
@@ -1,0 +1,71 @@
+<Query Kind="Program">
+  <NuGetReference>NuGet.Core</NuGetReference>
+  <Namespace>NuGet</Namespace>
+  <Namespace>System.Collections.Concurrent</Namespace>
+  <Namespace>System.Threading.Tasks</Namespace>
+</Query>
+
+void Main()
+{
+	var corePackageName = "NServiceBus";
+	var nuGet = PackageRepositoryFactory.Default.CreateRepository("https://www.nuget.org/api/v2/");
+	var minVersion = new SemanticVersion("3.3");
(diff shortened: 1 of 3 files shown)


tools/supportedVersions.linq
```diff
@@ -0,0 +1,71 @@
...
+			package.Published, 
+			VersionString = package.Version.ToString()
+		 }
+		 group result by result.Version.Version.Major into g
+		 select g).Dump("All Versions by Major", 2);
+		 
+	var majorDates = 
+		(from majorRelease in releasesByMajor
+		 let initialRelease = majorRelease.Min(x => x.Published)
+		 let latestMinor = majorRelease.Max(r => r.Published)
```
mikeminutillo: Not sure if this is right. What is this date meant to represent in the table?

Is it the most recent date that a minor was released for this major?

Is it the date that the latest minor was first released (irrespective of patch/build)?

I'm really not sure why it's helpful in the table at all.

mikeminutillo:
Note that the dates do not match up with what we already had exactly. 

I'm not sure why. I'm not sure the legal consequences of changing them.

tools/supportedVersions.linq
andreasohlund: I'd say if we only show majors the minor data isn't adding any value.

Wouldn't it be relevant to show the minor(s) that we support currently on this page as well?

mikeminutillo: We can do that. In that case would it just be:

| Version | Released | Support Ends |
|    6.1     |  x             | y + 3 years     |
|    6.0     |  y             | y + 1 years     |

PR #2431, opened 2017-02-12
Sagas should not access databases

udidahan:


nservicebus/sagas/index.md
```diff
@@ -106,6 +106,13 @@ A typical scenario is a saga controlling the process of billing a customer throu
...
 
 
+## Accessing databases and other resources from a saga
+
+WARNING: Other than interacting with its own internal state, a saga should **not** access a database, call out to web services, or access other resources - neither directly nor indirectly by having such dependencies injected into it.
+
+Instead, a saga should send messages to other endpoints in which a regular message handler would access databases, call web services, or access other resources. Usually those message handlers will send response messages back to the saga, as described in the previous section.
```
SimonCropp: "send messages"  perhaps "send or publish"?

why explicitly "other endpoints"?

udidahan: I consider it less likely that a `publish` would be used for those purposes.

With regards to other endpoints, if you mean that those handlers could be hosted in the same endpoint as the saga - yes, I guess that could be fine.

indualagarsamy: >Instead, a saga should send messages to other endpoints in which a regular message handler would access databases, call web services, or access other resources. Usually those message handlers will send response messages back to the saga, as described in the previous section.

How about:

Instead, a saga should send messages to perform those actions. The corresponding message handler would access databases, call web services, or access other resources. Usually those message handlers will send response messages back to the saga, as described in the previous section. These message handlers can be hosted in separate endpoints or in the same endpoint depending on other architectural decisions.

udidahan: @indualagarsamy sounds good. Updating.

PR #2445, opened 2017-02-14
WIP Routing strategy snippet 6.2

danielmarbach:
Relates to https://github.com/Particular/NServiceBus/pull/4461

@Particular/docs-maintainers wasn't exactly sure if we need more explanation. I went with the approach to have the same snippet versioned.

Since the old way still works and is supported would be still need an upgrade guide?

// cc @Particular/nservicebus-maintainers

### DIFF
--- a/Snippets/Core/Core_6/Routing/Extensibility.cs
+++ b/Snippets/Core/Core_6/Routing/Extensibility.cs
@@ -157,7 +157,7 @@
         class RandomStrategy :
             DistributionStrategy
         {
-            static Random r = new Random();
+            static Random random = new Random();
 
             public RandomStrategy(string endpoint, DistributionStrategyScope scope) : base(endpoint, scope)
             {
@@ -165,7 +165,7 @@
 
             public override string SelectReceiver(string[] receiverAddresses)
             {
-                return receiverAddresses[r.Next(receiverAddresses.Length)];
+                return receiverAddresses[random.Next(receiverAddresses.Length)];
             }
         }
 
@@ -184,12 +184,12 @@
 
     public class Extensibility_62
     {
-        #region RoutingExtensibility-DistributionStrategy [6.2,7.0)
+        #region RoutingExtensibility-DistributionStrategy [6.2,)
 
         class RandomStrategy :
             DistributionStrategy
         {
-            static Random r = new Random();
+            static Random random = new Random();
 
             public RandomStrategy(string endpoint, DistributionStrategyScope scope) : base(endpoint, scope)
             {
@@ -204,7 +204,7 @@
             public override string SelectDestination(DistributionContext context)
             {
                 // access to headers, payload...
-                return context.ReceiverAddresses[r.Next(context.ReceiverAddresses.Length)];
+                return context.ReceiverAddresses[andomr.Next(context.ReceiverAddresses.Length)];
             }
         }
 
--- a/Snippets/Core/Core_6/Routing/Extensibility.cs
+++ b/Snippets/Core/Core_6/Routing/Extensibility.cs
@@ -204,7 +204,7 @@
             public override string SelectDestination(DistributionContext context)
             {
                 // access to headers, payload...
-                return context.ReceiverAddresses[andomr.Next(context.ReceiverAddresses.Length)];
+                return context.ReceiverAddresses[random.Next(context.ReceiverAddresses.Length)];
             }
         }
 
--- a/Snippets/Core/Core_6/Routing/Extensibility.cs
+++ b/Snippets/Core/Core_6/Routing/Extensibility.cs
@@ -152,7 +152,7 @@
             #endregion
         }
 
-        #region RoutingExtensibility-DistributionStrategy [6.0,6.2)
+        #region RoutingExtensibility-DistributionStrategy
 
         class RandomStrategy :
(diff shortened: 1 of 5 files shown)


SzymonPobiega:
I think it requires more explanation than just a single comment line. As a user I would not expect to see `NotImplementedException` in the working code.

danielmarbach:
@SzymonPobiega so actually split the partial then? Fine for me

SzymonPobiega:
@danielmarbach yeah. Maybe provide snippets for both cases in 6.2 partial: the simple one and the context-aware one.

danielmarbach:
@SzymonPobiega pushed a change. Would that be what you had in mind?

nservicebus/messaging/routing-extensibility_distribution_core_[6,6.2).partial.md
```diff
@@ -0,0 +1,7 @@
+include: routing-extensibility.distribution
```
SimonCropp: partials must match the version numbers of the snippets dirs. so unless u want to add a 6.2 here https://github.com/Particular/docs.particular.net/tree/master/Snippets/Core this partial will be ignored

weralabaj:
@danielmarbach feel free to continue editing text, etc.
as discussed, I'll make sure we'll figure out the best way for moving forward from the engine/editing perspective

danielmarbach:
@Particular/docs-maintainers @Particular/nservicebus-maintainers please review. I used a local nuget package to write this sample. The PR is merge into develop. We can always push an unstable package

nservicebus/messaging/routing-extensibility_content_core_[6,).partial.md
```diff
@@ -64,4 +64,6 @@ snippet:RoutingExtensibility-Distribution
 
 snippet:RoutingExtensibility-DistributionStrategy
 
+As of NServiceBus version 6.2 and later it is possible to override the virtual method `SelectDestination`. The method provides access to the `DistributionStrategyContext` that enables more advanced distribution scenarios like distributing based on the headers of the message. When `SelectDestination` is overridden, it is advised not to call `base.SelectDestination` since the base method calls into `SelectReceiver` for backward compatibility reasons. `SelectReceiver` can throw a `NotImplementedException` since it will never be called in this case.
```
SeanFeldman: > The method provides access to the `DistributionStrategyContext` that enables more advanced distribution scenarios like distributing based on the headers of the message.

replace "like" with "such as". Possibly:

The method provides access to the `DistributionStrategyContext` enabling more advanced distribution scenarios such as message header based distribution.

weralabaj:
@danielmarbach 

> since the base method calls into `SelectReceiver` for backward compatibility reasons. `SelectReceiver` can throw a `NotImplementedException` since it will never be called in this case.

This part is a bit unclear to me. Is `SelectReceiver` part of the public API? Are we getting too deep into implementation details here? Is there actually any situation when you should call the base method? Or could we just say:

> When overridning `SelectDestination` do not call `base.SelectDestination`. For more details go to the sample.

weralabaj:
see updated comment with propositions

danielmarbach:
@weralabaj good suggestion! I think I dove into too much details, you are right. Updating based on your decision

nservicebus/messaging/routing-extensibility_content_core_[6,).partial.md
```diff
@@ -64,4 +64,6 @@ snippet:RoutingExtensibility-Distribution
 
 snippet:RoutingExtensibility-DistributionStrategy
 
-To learn more about creating custom distribution strategies see the [fair distribution sample](/samples/routing/fair-distribution/).
\ No newline at end of file
+In NServiceBus version 6.2 and above it is possible to override the virtual method `SelectDestination`. The method provides access to the `DistributionStrategyContext` that enables implementing more advanced distribution scenarios, such as distributing based on the headers of the message. When `SelectDestination` is overridden, do not call `base.SelectDestination` since the base method calls `SelectReceiver` for backward compatibility reasons. `SelectReceiver` can throw a `NotImplementedException`.
```
SimonCropp: "NServiceBus" is assumed. should be removed

weralabaj:
I think we need to deploy the 6.2 package to MyGet to get it working @danielmarbach

Snippets/Core/Core_6/Routing/Extensibility.cs
```diff
@@ -181,4 +181,33 @@ class MyEvent :
         {
         }
     }
-}
\ No newline at end of file
+
+    public class Extensibility_62
+    {
+        #region RoutingExtensibility-DistributionStrategy [6.2,7.0)
```
SimonCropp: shouldnt this be an unbound upper?

Snippets/Core/Core_6/Routing/Extensibility.cs
```diff
@@ -181,4 +181,33 @@ class MyEvent :
...
+
+        class RandomStrategy :
+            DistributionStrategy
+        {
+            static Random r = new Random();
```
SimonCropp: can we have a better variable name

danielmarbach:
@weralabaj @SimonCropp addressed comments

PR #2450, opened 2017-02-16
WIP Unrecoverable exceptions

danielmarbach:
Documentation for https://github.com/Particular/NServiceBus/pull/4479

@Particular/nservicebus-maintainers @Particular/docs-maintainers please review

### DIFF
--- a/Snippets/Core/Core_6/Recoverability/Delayed/CustomPolicies/FullyCustomizedPolicy.cs
+++ b/Snippets/Core/Core_6/Recoverability/Delayed/CustomPolicies/FullyCustomizedPolicy.cs
@@ -57,4 +57,62 @@
         {
         }
     }
+
+    class FullyCustomizedPolicy62
+    {
+        FullyCustomizedPolicy62(EndpointConfiguration endpointConfiguration)
+        {
+            #region FullyCustomizedPolicyRecoverabilityConfiguration [6.2,)
+
+            var recoverability = endpointConfiguration.Recoverability();
+            recoverability.AddUnrecoverableException<MyBusinessException>();
+            recoverability.CustomPolicy(MyCustomRetryPolicy);
+            // configuration can still be tweaked on this level if desired, data will be passed into the policy
+            recoverability.Immediate(
+                immediate =>
+                {
+                    immediate.NumberOfRetries(3);
+                });
+            recoverability.Delayed(
+                delayed =>
+                {
+                    var retries = delayed.NumberOfRetries(3);
+                    retries.TimeIncrease(TimeSpan.FromSeconds(2));
+                });
+
+            #endregion
+        }
+
+        #region FullyCustomizedPolicy [6.2,)
+
+        RecoverabilityAction MyCustomRetryPolicy(RecoverabilityConfig config, ErrorContext context)
+        {
+            // early decisions and return before custom policy is invoked
+            // i.e. all unrecoverable exceptions should always go to customized error queue
+            foreach (var unrecoverableExceptionType in config.Failed.UnrecoverableExceptionTypes)
+            {
+                if (unrecoverableExceptionType.IsInstanceOfType(context.Exception))
+                {
+                    return RecoverabilityAction.MoveToError("customErrorQueue");
+                }
+            }
+
+            // override delayed retry decision for custom exception
+            // i.e. MyOtherBusinessException should do fixed backoff of 5 seconds
+            if (context.Exception is MyOtherBusinessException && context.DelayedDeliveriesPerformed <= config.Delayed.MaxNumberOfRetries)
+            {
+                return RecoverabilityAction.DelayedRetry(TimeSpan.FromSeconds(5));
+            }
+
+            // in all other cases No Immediate or Delayed Retries, go to default error queue
+            return RecoverabilityAction.MoveToError(config.Failed.ErrorQueue);
+        }
+
+        #endregion
+
+        class MyOtherBusinessException :
+            Exception
+        {
+        }
+    }
 }
--- a/nservicebus/recoverability/custom-recoverability-policy.md
+++ b/nservicebus/recoverability/custom-recoverability-policy.md
@@ -18,7 +18,7 @@ related:
 
 The default Recoverability Policy is implemented in `DefaultRecoverabilityPolicy` and publicly exposed for use cases when the default recoverability behavior needs to be reused as part of a custom recoverability policy. The default policy takes into account the settings provided for Immediate Retries, Delayed Retries and the configured error queue. The default policy works like the following:
 
- 1. When an exception of type `MessageDeserializedException` was raised, the `MoveToError` recoverability action is returned with the default error queue.
+ 1. When an unrecoverable exception was raised, by default exceptions of type `MessageDeserializedException`, the `MoveToError` recoverability action is returned with the default error queue.
  1. If the `ImmediateProcessingFailures` is less or equal to the configured `MaxNumberOfRetries` for Immediate Retries, the `ImmediateRetry` recoverability action is returned.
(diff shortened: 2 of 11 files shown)


danielmarbach:
// cc @tmasternak

nservicebus/recoverability/index_unrecoverableexceptions_core_[6.0,).partial.md
```diff
@@ -0,0 +1,11 @@
...
+
+snippet: UnrecoverableExceptions
+
+The above snippets defines `ValidationException` and `ArgumentException` as unrecoverable exceptions. For example if a `ArgumentNullException` is raised during message handling, the message will be automatically moved to the error queue since `ArgumentNullException` is assignable to `ArgumentException`.
+
+snippet: UnrecoverableExceptionsSettings
+
+It is also possible to define certain exceptions as unrecoverable in plugins such as persisters, transports and features. Everywhere there is access to `SettingsHolder` the non-generic `AddUnrecoverableException` can be used.
```
andreasohlund: If a user has a custom recoverability policy defined will that get involed for exceptions that are unrecoverable?(I think yes based on reading the code)

danielmarbach: @andreasohlund only if they do a partial override and call the default policy. If they implement it fully they have to do that manually since they are in full control

andreasohlund: Should we add that as a note 

> This has no effect if you define a custom recoverability policy

Or similar?

That brings up another question, should this really be exposed to end users since they have full control via a custom policy anyway? (And this feature is only needed by infrastructure extensions)

danielmarbach: I still think it makes sense to expose it. A full custom policy is non-trivial. Being able to define unrecoverable exceptions without needing to fall to a custom policy makes sense I think. @timbussmann brought that up on the core PR and we discussed it. I just realised I need to change the doco on https://docs.particular.net/nservicebus/recoverability/custom-recoverability-policy anyway and I can make a note there that a fully customized policy can take this into account

andreasohlund: 👍

nservicebus/recoverability/index_unrecoverableexceptions_core_[6.0,).partial.md
```diff
@@ -0,0 +1,11 @@
+## Unrecoverable exceptions
+
+By default every exception, unless it is NServiceBus' built-in `MessageDeserializationException`, will go through the recoverability as defined in the endpoint configuration. Depending on the business domain some exceptions might indicate that is never useful to go through Recoverability since all subsequent retries would lead to the same exception over and over again. An example is message validation. When a validator would raise a `ValidationExceptions`, or derivatives of that exception base type, subsequent retries of the same message would fail the validation again. In such scenarios, it is possible to mark certain exception types as unrecoverable. When an exception is marked as unrecoverable, then message handling that raises such an exception type will immediately move the message to the error queue and not even try to attempt to do [Immediate](/nservicebus/recoverability/configure-immediate-retries) and [Delayed Retries](/nservicebus/messaging/delayed-delivery).
```
tmasternak: How about we write it this way:

Recoverability enables declaring an exception type as unrecoverable. When a message processing fails due to unrecoverable exception being thrown it does not go through retry process but is moved to the error queue instead. By default `MessageDeserializationException` is the only unrecoverable exception however based on the business domain users can declare additional. For example user might declare `ValidationExceptions`  thrown when message fails validation as unrecoverable. Assuming that validation is a deterministic, processing will always fail so additional retires will not bring any benefit.

NOTE: Declaring exception type as unrecoverable declares the whole inheritance tree as unrecoverable i.e. any direct or indirect sub-class.

danielmarbach:
@tmasternak @andreasohlund pushed an update with more snippets

nservicebus/recoverability/custom-recoverability-policy.md
```diff
@@ -84,6 +85,6 @@ Fully customizing the Recoverability Policy basically means not calling `Default
 
 snippet:FullyCustomizedPolicyRecoverabilityConfiguration
 
-The configuration will be passed into the custom policy. Below is a policy which moves all exceptions of type `MyBusinessException` to a custom error queue, does Delayed Retries with a constant time increase of five seconds for `MyOtherBusinessException` and for all other cases moves messages to the configured standard error queue.
+The configuration will be passed into the custom policy. Below is a policy which moves all unrecoverable exceptions such as `MyBusinessException` to a custom error queue, does Delayed Retries with a constant time increase of five seconds for `MyOtherBusinessException` and for all other cases moves messages to the configured standard error queue.
```
SeanFeldman: Slight modification to this paragraph:
(does -> executes, and comma)
> The configuration will be passed into the custom policy. Below is a policy which moves all unrecoverable exceptions such as `MyBusinessException` to a custom error queue, executes Delayed Retries with a constant time increase of five seconds for `MyOtherBusinessException`, and for all other cases moves messages to the configured standard error queue.

nservicebus/recoverability/index_unrecoverableexceptions_core_[6.0,).partial.md
```diff
@@ -0,0 +1,13 @@
+## Unrecoverable exceptions
+
+Recoverability enables declaring an exception type as unrecoverable. When a message processing fails due to a unrecoverable exception being thrown, it does not go through retry process but is moved to the error queue instead. By default MessageDeserializationException is the only unrecoverable exception however based on the business domain users can declare additional. For example, a user might declare ValidationExceptions thrown when a message fails validation as unrecoverable. Assuming that validation is deterministic, processing will always fail so extra retries will not bring any benefit.
```
udidahan: > it does not go through retry process but is moved to the error queue instead

Slightly better as: 

> it does not go through retry process but is moved to the error queue directly instead

Snippets/Core/Core_6/Recoverability/Delayed/CustomPolicies/FullyCustomizedPolicy.cs
```diff
@@ -57,4 +57,62 @@ class MyOtherBusinessException :
         {
         }
     }
+
+    class FullyCustomizedPolicy62
```
SimonCropp: why a different class

nservicebus/recoverability/index_unrecoverableexceptions_core_[6.0,).partial.md
```diff
@@ -0,0 +1,13 @@
+## Unrecoverable exceptions
+
+Recoverability enables declaring an exception type as unrecoverable. When a message processing fails due to a unrecoverable exception being thrown, it does not go through retry process but is moved to the error queue directly instead. By default MessageDeserializationException is the only unrecoverable exception however based on the business domain users can declare additional. For example, a user might declare ValidationExceptions thrown when a message fails validation as unrecoverable. Assuming that validation is deterministic, processing will always fail so extra retries will not bring any benefit.
```
SimonCropp: ticks around MessageDeserializationException

nservicebus/recoverability/custom-recoverability-policy.md
SimonCropp: this sentence needs to be split

Snippets/Core/Core_6/Recoverability/Delayed/CustomPolicies/FullyCustomizedPolicy.cs
danielmarbach: Does it matter? happy to change if you don't like it.

PR #2451, opened 2017-02-16
Add RabbitMQ native delayed delivery pages

bording:


SeanFeldman:
Will need an upgrade guide as well. It requires to be in a format of `rabbitmq-4.2to4.??.md` so perhaps just to add a placeholder `rabbitmq-4.2to4.TBD.md`?

nservicebus/rabbitmq/delayed-delivery.md
```diff
@@ -0,0 +1,43 @@
...
+component: Rabbit
+versions: '[4,]'
+---
+
+Starting with Version 4.3, the transport no longer relies on the [timeout manager](/nservicebus/messaging/timeout-manager.md) to implement [delayed delivery](/nservicebus/messaging/delayed-delivery.md). The transport now creates an infrastructure inside the broker to natively handle delaying messages.
+
+
+## How it works
+
+The transport sets up a series of exchanges, queues, and bindings that work together to provide the necessary infrastructure to support delayed messages. These exchange & queues are grouped into a series of levels, 0 - 27. There is one final "delivery" exchange in addition to the level.
```
SeanFeldman: > in addition to the level.

Should it be "in addition to the level exchanges"?

nservicebus/rabbitmq/delayed-delivery.md
```diff
@@ -0,0 +1,43 @@
...
+
+
+### Delay levels
+
+Each level represents a delay amount corresponding to a number of seconds equal to 2 ^ level seconds. This is achieved by declaring the queue for each level with an `x-message-ttl` value corresponding to level's delay amount. It is also declared with an `x-dead-letter-exchange` value corresponding to the next lowest level. This means a message will sit in the queue for the amount of time specified in  `x-message-ttl` and then be routed to the next level.
```
SeanFeldman: > 2 ^ level

no need for spaces? I.e. "**2^level**" or better "2^level (two in power of level) to be very clear)?

SeanFeldman: > level's delay amount

"level's delay amount **in seconds**"?

SeanFeldman: > It is also declared with an `x-dead-letter-exchange` value corresponding to the next lowest level.

Is that right? It's always points to the "the next low**er** level, not lowest. I.e. from level X to level X-1

nservicebus/rabbitmq/delayed-delivery.md
```diff
@@ -0,0 +1,43 @@
...
+
+
+When a message needs to be delayed, the value of the desired delay is first converted to seconds. The binary representation of this value is then used as the routing key to determine which of the delay levels the message needs to
```
SeanFeldman: > needs to

Unfinished sentence? Needs to go?

nservicebus/rabbitmq/delayed-delivery.md
```diff
@@ -0,0 +1,43 @@
...
+snippet: rabbitmq-delay-disable-timeout-manager
+
+Once this has been called, the .Timeouts and .TimeoutsDispatcher exchanges and queues for the endpoint can be deleted from the broker.
+
+
+### All endpoints support delayed delivery
+
+When an endpoint supports delayed delivery it is responsible for ensuring that it can receive delayed messages from the delay infrastructure. However, it cannot assume that all other endpoints in the system are also capable of doing the same. Because of this, before each delayed message is sent, the endpoint first makes an additional call to bind the destination to the delay infrastructure.
+
+Once all endpoints have been upgraded to version of the transport that supports delayed delivery, this behavior is no longer needed and can be removed by calling:
```
SeanFeldman: > this behavior is no longer needed and can be removed

While the behavior is removed, from the user perspective should we say "can be **disabled**"?

nservicebus/rabbitmq/delayed-delivery.md
bording: > "level's delay amount in seconds"?

It should actually be milliseconds here.

> Is that right? It's always points to the "the next lower level, not lowest. I.e. from level X to level X-1

Reading up on usage of "next lowest" it seems like my intended usage here is fairly common, but it can be rather ambiguous. I'll reword it to avoid the phrase altogether.

bording:
@adamralph is still planning on reviewing this, but otherwise it should be good to go. The task force has decided that we don't think there is value in having an upgrade guide.

SeanFeldman:
> The task force has decided that we don't think there is value in having an upgrade guide.

Addition: we decided that the release notes will contain the relevant information originally considered for the upgrade guide.

nservicebus/rabbitmq/delayed-delivery.md
```diff
@@ -0,0 +1,129 @@
+---
+title: Delayed Delivery
+summary: Describes the native delayed delivery implementation in the RabbitMQ transport
+component: Rabbit
+versions: '[4,]'
+---
+
+Starting with Version 4.3, the transport no longer relies on the [timeout manager](/nservicebus/messaging/timeout-manager.md) to implement [delayed delivery](/nservicebus/messaging/delayed-delivery.md). The transport now creates an infrastructure inside the broker to natively handle delaying messages.
```
weralabaj: > Starting with Version 4.3

In Versions 4.3 and above, RabbitMQ transport...

nservicebus/rabbitmq/delayed-delivery.md
```diff
@@ -0,0 +1,129 @@
...
+snippet: rabbitmq-delay-disable-timeout-manager
+
+Once this has been called, the .Timeouts and .TimeoutsDispatcher exchanges and queues for the endpoint can be deleted from the broker.
+
+
+### All endpoints support delayed delivery
+
+When an endpoint supports delayed delivery it is responsible for ensuring that it can receive delayed messages from the delay infrastructure. However, it cannot assume that all other endpoints in the system are also capable of doing the same. Because of this, before each delayed message is sent, the endpoint first makes an additional call to bind the destination to the delay infrastructure.
+
+Once all endpoints have been upgraded to version of the transport that supports delayed delivery, this behavior is no longer needed and can be disabled by calling:
```
weralabaj: > to version of

to a version of

nservicebus/rabbitmq/delayed-delivery.md
```diff
@@ -0,0 +1,129 @@
...
+class q0,q2 usedQueue
+
+end
+```
+
+
+## Settings
+
+
+### Disable the timeout manager
```
weralabaj: IMO better as "Disabling the timeout manager"

nservicebus/rabbitmq/delayed-delivery.md
```diff
@@ -0,0 +1,129 @@
...
+class exchange3,exchange2,exchange1,exchange0,exchange-delivery exchangeClass
+
+classDef usedQueue fill:#11ff00;
+class q0,q2 usedQueue
+
+end
+```
+
+
+## Settings
```
weralabaj: IMO that section needs some more work, it's not about settings, it's about upgrade, so how about changing the heading

Settings => Upgrading an endpoint/solution ?

All endpoints support delayed delivery => that sounds like running endpoints in backwards compatible mode or something like that? heading should reflect that

Then providing a list of steps with clear explanations, e.g. now it's not clear to me when can I disable timeout manager? When I upgraded one, some or all endpoints?

nservicebus/rabbitmq/routing-topology_support-delayed-delivery_rabbit_[4,].partial.md
```diff
@@ -0,0 +1,7 @@
+### Delayed delivery
+
+Starting with Version 4.3, it is possible to support [delayed delivery](delayed-delivery.md) without requiring the [timeout manager](/nservicebus/messaging/timeout-manager.md) by implementing `ISupportDelayedDelivery` in addition to `IRoutingTopology` in the routing topology class.
```
Comment (weralabaj) [inline nservicebus/rabbitmq/routing-topology_support-delayed-delivery_rabbit_[4,].partial.md]: > Starting with Version 4.3

In Versions 4.3 and above, 

> it is possible to support [delayed delivery](delayed-delivery.md)

RabbitMQ supports [delayed delivery](delayed-delivery.md) ...

> by implementing `ISupportDelayedDelivery` in addition to `IRoutingTopology` in the routing topology class.

... and further text

Question (not sure what's the answer) - is that level of detail relevant for the reader? or do we go too much into implementation details?

nservicebus/rabbitmq/routing-topology_support-delayed-delivery_rabbit_[4,].partial.md
```diff
@@ -0,0 +1,7 @@
...
+
+When the routing topology implements `ISupportDelayedDelivery`, the transport will call the `BindToDelayInfrastructure` method to allow the routing topology to create the appropriate bindings to the delivery exchange that is part of the native delayed delivery infrastructure. 
+
+`IRoutingTopology` and `ISupportDelayedDelivery` will be combined into a single interface in Version 5.0.
```
Comment (weralabaj) [inline nservicebus/rabbitmq/routing-topology_support-delayed-delivery_rabbit_[4,].partial.md]: We should add something in https://docs.particular.net/nservicebus/messaging/timeout-manager, there we say

> NServiceBus provides a delayed-delivery implementation for transports that don't support it natively.

which doesn't seem to be technically correct?

Also https://docs.particular.net/nservicebus/messaging/delayed-delivery#how-it-works

One thing could be simply adding a sentence, saying that Rabbit in Versions x and below uses TM, in versions X and above has native delayed delivery implementation + link to the new article with details

nservicebus/rabbitmq/delayed-delivery.md
bording: If this wording is preferred, then we need to change it here as well: https://docs.particular.net/nservicebus/rabbitmq/routing-topology#custom-routing-topology "Starting with Version 4.2"

nservicebus/rabbitmq/delayed-delivery.md
```diff
@@ -0,0 +1,129 @@
...
+
+
+### Disable the timeout manager
+
+By default, the timeout manager continues to run to allow any preexisting delayed messages stored in the persistence database to be successfully sent. However, once the persistence database is empty, there is no more need for the timeout manager. It can be disabled by calling:
```
bording: We have Particular/NServiceBus.RabbitMQ/issues/349 to track considering making changes to the default values in the future, but since we haven't actually committed to making a change here, I'd prefer not to mention it in the docs yet.

nservicebus/rabbitmq/routing-topology_support-delayed-delivery_rabbit_[4,].partial.md
Comment (bording) [inline nservicebus/rabbitmq/routing-topology_support-delayed-delivery_rabbit_[4,].partial.md]: 👍 for changing the version wording, but the other change isn't correct. In this overall section of the page, we aren't talking about the RabbitMQ transport, we are talking specifically about an end-user creating a custom routing topology by implementing `IRoutingTopology`. This new section is about augmenting this implementation by also implementing `ISupportDelayedDelivery`.

Based on that, yes we do need to go into this level of detail here, because it is information that someone implementing the interface needs.

Comment (bording) [inline nservicebus/rabbitmq/routing-topology_support-delayed-delivery_rabbit_[4,].partial.md]: Looking at the initial sentence, perhaps it would be better as:

In Versions 4.3 and above, custom routing topologies can support delayed delivery..."

nservicebus/rabbitmq/routing-topology_support-delayed-delivery_rabbit_[4,].partial.md
Comment (bording) [inline nservicebus/rabbitmq/routing-topology_support-delayed-delivery_rabbit_[4,].partial.md]: > We should add something in https://docs.particular.net/nservicebus/messaging/timeout-manager, there we say

>>NServiceBus provides a delayed-delivery implementation for transports that don't support it natively.

>which doesn't seem to be technically correct?

I'm not seeing what's incorrect about that. If a transport doesn't have some sort of native support for delayed delivery, like we're adding for the RabbitMQ transport, then it does have to use the timeout manager.

> Also https://docs.particular.net/nservicebus/messaging/delayed-delivery#how-it-works

> One thing could be simply adding a sentence, saying that Rabbit in Versions x and below uses TM, in versions X and above has native delayed delivery implementation + link to the new article with details

I'll tweak the wording there as well.

nservicebus/rabbitmq/delayed-delivery.md
bording: >IMO that section needs some more work, it's not about settings, it's about upgrade, so how about changing the heading

>Settings => Upgrading an endpoint/solution ?

The settings here are specifically about things you can tweak after you've performed the upgrade, not something you'd use as part of upgrading to the 4.3 release. 

> All endpoints support delayed delivery => that sounds like running endpoints in backwards compatible mode or something like that? heading should reflect that

This setting is about removing a backwards-compatible behavior that isn't needed once all endpoints have been upgraded.

> Then providing a list of steps with clear explanations, e.g. now it's not clear to me when can I disable timeout manager? When I upgraded one, some or all endpoints?

There's text there already about preexisting entries in the persistence database, and how once it's empty you can disable it.

bording: It does seem like the Settings section could use some intro text, perhaps talking about default behavior, and how there are settings that can modify the behavior once the new feature is in use, so I'll add something that covers that, and maybe that will reveal a better name for the section as well.

nservicebus/rabbitmq/routing-topology_support-delayed-delivery_rabbit_[4,].partial.md
Comment (bording) [inline nservicebus/rabbitmq/routing-topology_support-delayed-delivery_rabbit_[4,].partial.md]: Looking at https://docs.particular.net/nservicebus/messaging/delayed-delivery#how-it-works a bit more, I don't think it makes sense to add something about RabbitMQ there. 

Instead, I'd say that entire section needs revising or removing, because despite the section title, it's not really "How it works" at all.

Regardless, I think making those changes are out of scope for this PR.

nservicebus/rabbitmq/delayed-delivery.md
```diff
@@ -108,22 +108,22 @@ end
 ```
 
 
-## Settings
+## Backwards compatibility
+
+When using a version of the transport that supports delayed delivery natively, the endpoint continues to operate in a backwards-compatible manner by default. The timeout manager will continue to be enabled, so any delayed messages already stored in the endpoint's persistence database will be sent when their timeouts expire. New delayed messages will be sent through the delay infrastructure, but the endpoint cannot assume that other endpoints have also been upgraded, so it will incur some additional overhead when sending delayed messages to ensure that they can be delivered successfully.
```
SeanFeldman: > enabled, so any delayed messages already ...

Break into 2 sentences?

"enabled. Any delayed messaged already..."

nservicebus/rabbitmq/delayed-delivery.md
```diff
@@ -108,22 +108,22 @@ end
...
 
 
 ### Disabling the timeout manager
 
-By default, the timeout manager continues to run to allow any preexisting delayed messages stored in the persistence database to be successfully sent. However, once the persistence database is empty, there is no more need for the timeout manager. It can be disabled by calling:
+Once an endpoint has no more delayed messages in its persistence database, there is no more need for the timeout manager. It can be disabled by calling:
```
SeanFeldman: 👍

nservicebus/rabbitmq/delayed-delivery.md
bording: I prefer the sentence how it currently is, so let's see if we get any additional feedback on it first.

bording:
@weralabaj I believe all of my changes address your feedback, except for where I've left followup comments.

bording:
@SimonCropp We were wanting @adamralph to review the content of the PR before it was merged.

SimonCropp:
sorry. @adamralph please submit a new PR

nservicebus/rabbitmq/delayed-delivery.md
weralabaj: Not only preferred, but will be consistent with other documents as well. I believe I mentioned it in both places, or at least that was my intention :)

bording: You did mention it in both places in this PR, but the place I linked to above was for something that was already merged an live and not part of this PR.

nservicebus/rabbitmq/delayed-delivery.md
```diff
@@ -0,0 +1,131 @@
...
+
+
+### Delay levels
+
+Each exchange/queue pair that makes up a level represents one bit of the total delay value. By having 28 of these levels, corresponding to `2^27` through `2^0`, the delay infrastructure can delay a message for any value that can be represented by a 28-bit number. With 28 total levels, the maximum delay value is 268,435,455 seconds.
+
+A level is created by declaring a topic exchange that is bound to a queue with a routing key of `1`, and is also bound to the exchange corresponding to `level - 1` with a routing key of `0`. The queue for the level is declared with an `x-message-ttl` value corresponding to `2^level` seconds. The queue is also declared with an `x-dead-letter-exchange` value corresponding to the `level - 1` exchange, so that when a message in the queue expires, it will be routed to the `level - 1` exchange.
+
+```mermaid
+
```
adamralph: The mermaid graph has a title of "Undef". Is that intentional?

Snippets/Rabbit/Rabbit_4/Usage.cs
```diff
@@ -151,4 +151,25 @@ void SetClientCertificates(EndpointConfiguration endpointConfiguration)
         #endregion
     }
 
-}
\ No newline at end of file
+    void DisableTimeoutManager(EndpointConfiguration endpointConfiguration)
+    {
+        #region rabbitmq-delay-disable-timeout-manager 4.3
```
bording: I was trying to scope the name of the snippet, which some of the others already do as well.

menu/menu.yaml
```diff
@@ -527,22 +527,24 @@
   - Url: nservicebus/rabbitmq
     Title: RabbitMQ Transport
     Articles:
-    - Url: nservicebus/rabbitmq/rabbitmqclient-nuget
-      Title: RabbitMQ.Client NuGet usage
```
bording: When deciding where in the menu the new page needed to go, I decided to reorder things as well, because I was unhappy with the current ordering.

menu/menu.yaml
```diff
@@ -527,22 +527,24 @@
...
     - Url: nservicebus/rabbitmq/connection-settings
       Title: Connection settings
     - Url: nservicebus/rabbitmq/routing-topology
       Title: Routing topology
-    - Url: nservicebus/rabbitmq/callbacks
-      Title: Callback support
```
bording: Same as above

nservicebus/rabbitmq/delayed-delivery.md
bording: No. When viewing it in their live editor, not having a title just meant it was blank, but the docs site seems to be treating it differently. We'll want to add a "delay level" title or something.

adamralph: @bording I see it in the live editor:

image

SeanFeldman: To get rid of "Undef", a space after `subgraph` would do it.

bording: Hmm, not sure what's going on there then. Pretty sure we didn't see that when we were working on it. 

Regardless, we just need to add a title then.

SeanFeldman: @bording correct. I like your suggestion of giving it a proper title, something like "Delay Level Overview"

udidahan:
Maybe a customer communication or story would make sense now that the documentation is done - open an issue?

SeanFeldman:
[An issue](https://github.com/Particular/Stories/issues/493) already exists @udidahan

udidahan:
My bad

PR #2459, opened 2017-02-21
Input box to filter nugets.linq script to certain projects

DavidBoike:
As a Raven maintainer, I commonly edit this tool with some custom code on the fly to limit to only Raven samples, but then don't commit that change. I got tired of doing that.

@Particular/docs-maintainers What do you think?

### DIFF
--- a/tools/nugets.linq
+++ b/tools/nugets.linq
@@ -1,4 +1,6 @@
 <Query Kind="Program">
+  <Reference>&lt;RuntimeDirectory&gt;\Microsoft.VisualBasic.dll</Reference>
+  <Namespace>Microsoft.VisualBasic</Namespace>
   <Namespace>System.Threading.Tasks</Namespace>
 </Query>
 
@@ -11,6 +13,10 @@ void Main()
 	var samplesDirectory = Path.Combine(docsDirectory, "samples");
 	var tutorialsDirectory = Path.Combine(docsDirectory, "tutorials");
 	var nugetConfigFile = Path.Combine(docsDirectory, "nuget.config");
+	
+	var filter = Interaction.InputBox("To filter solutions to update, enter a simple filter value. Leave blank to update all solutions.",
+		"Filter Solutions to Update",
+		string.Empty);
 
 	var solutionFiles = Directory.EnumerateFiles(samplesDirectory, "*.sln", SearchOption.AllDirectories).ToList();
 	solutionFiles.AddRange(Directory.EnumerateFiles(tutorialsDirectory, "*.sln", SearchOption.AllDirectories));
@@ -21,6 +27,11 @@ void Main()
 	new ParallelOptions() { MaxDegreeOfParallelism = 20 },
 	(solutionFile) =>
 		{
+			if (!string.IsNullOrEmpty(filter) && solutionFile.IndexOf(filter, StringComparison.OrdinalIgnoreCase) < 0)
+			{
+				return;
+			}
+			
 			Debug.WriteLine(solutionFile);
 			try
 			{ 


tools/nugets.linq
```diff
@@ -11,6 +13,10 @@ void Main()
 	var samplesDirectory = Path.Combine(docsDirectory, "samples");
 	var tutorialsDirectory = Path.Combine(docsDirectory, "tutorials");
 	var nugetConfigFile = Path.Combine(docsDirectory, "nuget.config");
+	
+	var filter = Interaction.InputBox("To filter solutions to update, enter a simple filter value. Leave blank to update all solutions.",
```
SeanFeldman: Linqpad has a native input option
```c#
Util.ReadLine(prompt: "your value", defaultValue: string.Empty);
```

SeanFeldman:
I like the idea. Myself have updated the script multiple time and every time need to ensure I don't commit the change.

tools/nugets.linq
DavidBoike: I copied from something @adamralph made. 😉

SeanFeldman: And that's fine. I'm just saying there's something built in that linqpad users might be more familiar with.

SimonCropp:
Why would you want to filter. If other nugets show up as needing to be updated then u should smoke test them or contact the appropriate maintainer group to verify they have done it.

tools/nugets.linq
adamralph: I'm not sure if I did the right thing. 😉

DavidBoike:
> Why would you want to filter. If other nugets show up as needing to be updated then u should smoke test them or contact the appropriate maintainer group to verify they have done it.

I realize that from a docs maintenance perspective that would be ideal, but it isn't going to happen. If I'm releasing Raven, I need to smoke test Raven samples. If I run the tool and a bunch of Rabbit or Azure stuff is using an older patch, I'm not going to have any knowledge or context in that. Because I don't know if it's safe, and might not be able to find out quickly, I'm going to have to manually revert any changes that don't immediately apply to my current task so I can do the release I need to get done. Those manual steps will increase the chances that I screw something up and break the Docs build, in which case I have to do more work and waste more time before the PR can be merged.

I do think that an automated reminder to maintainer groups when they're behind on something would be valuable though. It would be cool if a weekly build could attempt the package updates in their entirety, analyze any resulting pending changes, somehow blame them on specific maintainer groups, and then notify them in some way to take action. But I see that as totally separate.

This change is just about an easy automation for something I dare say most of us are doing regularly anyway.

SimonCropp:
But either way this seems wrong. The existence of a text string in the sln path does not capture all possible uses of that nuget.  Or am I missing something?

SimonCropp:
> I realize that from a docs maintenance perspective that would be ideal, but it isn't going to happen. If I'm releasing Raven, I need to smoke test Raven samples.

And this is problematic. The fact that this happens is a sign a maintainer group is not doing their job. When it occurs should should take the time to do due diligence and follow up. Not hack a script and ignore the problem. Because it is likely to affect the next person who does an update.

SeanFeldman:
@DavidBoike raised a valid question which needs to be addressed and this is an attempt.

>  If I'm releasing Raven, I need to smoke test Raven samples. If I run the tool and a bunch of Rabbit or Azure stuff is using an older patch, I'm not going to have any knowledge or context in that. Because I don't know if it's safe, and might not be able to find out quickly, I'm going to have to manually revert any changes that don't immediately apply to my current task so I can do the release I need to get done.

Poking holes in it is fine. Raising additional issues is fine. But leaving it w/o actual resolution of some sort is just going to force to continue manual patch of the script with occasion "human error" of committing modifications to the script.

SimonCropp:
"smoke test" does not imply that it has to be immediate committed to master. if you notice other nuget have been updated, the push it to a branch and do a pr with "i smoke tested x and it was fine, but y was also updated and i dont have time to test it" and the @Particular/docs-maintainers will handle it.

weralabaj:
> And this is problematic. The fact that this happens is a sign a maintainer group is not doing their job. 

I'm with Simon on this. Yes, it's an extra effort. Yes, It's annoying. And **it should be**. If we start adding features like this, then we just encourage even less frequent updates. I don't think this is what we're after.

SimonCropp:
since this has been elevated to a platform debv issue, and the script implementation [is broken](https://github.com/Particular/docs.particular.net/pull/2459#issuecomment-281466973), i am closing this.

PR #2465, opened 2017-02-22
Native MySQL requires at least Server 5.7

DavidBoike:
Discovered that native SQL persister won't work with 5.6 or lower (which is default version right now for an RDS instance, by the way) because the json datatype wasn't added until 5.7.

@SimonCropp is there a better way to call this out in doco?


SimonCropp:
@DavidBoike i think this is fine. can u add a link to the json data type doco. also it says there are conflict, but for a change of this size just make it direct to master.

SimonCropp:
@DavidBoike merged manually here https://github.com/Particular/docs.particular.net/commit/eab6b699f2b487df3a7e8ddda14fa0e5b7cce7e0

thanks

PR #2467, opened 2017-02-23
WIP add support warning for sql persistence

SimonCropp:


SimonCropp:
ping @Particular/sqlpersistence-maintainers thoughts

also @Particular/ravendb-persistence-maintainers, @Particular/azure-maintainers, @Particular/nhibernate-persistence-maintainers should equivalent warnings be in other areas?

@andreasohlund @udidahan given you two mostly wrote the support policy do you have any thoughts?

nservicebus/sql-persistence/index.md
```diff
@@ -16,6 +16,8 @@ The SQL Persistence uses [Json.NET](http://www.newtonsoft.com/json) to serialize
  * [SQL Server](https://www.microsoft.com/en-au/sql-server/) ([Version 2012](https://docs.microsoft.com/en-us/sql/release-notes/sql-server-2012-release-notes) and above due to the use of the [THROW functionality](https://msdn.microsoft.com/en-us/library/ee677615.aspx)).
  * [MySQL](https://www.mysql.com/) ([Version 5.7](https://dev.mysql.com/doc/relnotes/mysql/5.7/en/) and above due to the use of [JSON Data Type](https://dev.mysql.com/doc/refman/5.7/en/json.html)).
 
+WARNING: This persistence will run on the free version of the above engines, i.e. [SQL Server Express](https://www.microsoft.com/en-au/sql-server/sql-server-editions-express) and [MySQL Community Edition](https://www.mysql.com/products/community/). However, due to the benefits of support, it is strongly recommended to purchase a paid version for running any production system. Its is also recommended to ensure that support agreements are in place from either [MySQL support](https://www.mysql.com/support/), [Microsoft Premier Support](https://www.microsoft.com/en-us/microsoftservices/support.aspx) or another third party support provider.
```
SeanFeldman: > Its is

typo

SeanFeldman:
> should equivalent warnings be in other areas?

I'd like to say "this is a common sense", but it might not be. I'd rather not have it on every page that has to do something with a 3rd party technology. Can't we have it under [support policy page](https://docs.particular.net/nservicebus/upgrades/support-policy) ?

weralabaj:
Makes sense, but then do we plan to add similar warnings to SQL transport, and other tools?

andreasohlund:
What @weralabaj said, should we add this to RabbitMQ as well?

This seems to apply all infrastructure the users run on so we could just add a note to

nservicebus/sql-persistence/index.md
udidahan: This is good, but if we rearrange things a bit it can be even stronger:

> Although this persistence will run on the free version of the above engines, i.e. [SQL Server Express](https://www.microsoft.com/en-au/sql-server/sql-server-editions-express) and [MySQL Community Edition](https://www.mysql.com/products/community/), it is strongly recommended to use commercial versions for any production system.

SeanFeldman:
@andreasohlund that's exactly what I suggested [here](https://github.com/Particular/docs.particular.net/pull/2467#issuecomment-281900345)... 😕

andreasohlund:
>that's exactly what I suggested here... 😕

Oops :) Then what @SeanFeldman said!

SimonCropp:
@andreasohlund @SeanFeldman while i am not against putting teething on the support page. i think it needs to be called out explicitly in the persistence doco, otherwise it is too hidden

SimonCropp:
> Makes sense, but then do we plan to add similar warnings to SQL transport, and other tools?

@weralabaj good point. i missed the @Particular/sqlserver-transport-maintainers

PR #2491, opened 2017-02-28
clarify prerequisites

weralabaj:


samples/saga/sql-sagafinder/sample.md
```diff
@@ -18,7 +18,7 @@ include: sagafinder-into
 
 
 include: sql-persistence-prereqs
-
+1. Ensure the SQL Server version is 2016 or higher, as saga finder uses `JSON_VALUE` function that is not available in older versions
```
indualagarsamy: Shouldn't this be part of sql-persistence-prereqs.include.md ?

SimonCropp: @indualagarsamy nope since this is the only sample that uses JSON_VALUE

PR #2499, opened 2017-03-01
Custom routing refresh exception handling

ramonsmits:
Current implementation crashed when there is a temporary error on the database. Also have execution overlap protection added in case the refresh operation duration is larger then the timer interval period.

### DIFF
--- a/samples/routing/custom/Core_6/Shared/AutomaticRouting/AutomaticRoutingFeature.cs
+++ b/samples/routing/custom/Core_6/Shared/AutomaticRouting/AutomaticRoutingFeature.cs
@@ -31,7 +31,9 @@ class AutomaticRoutingFeature :
         // Create the infrastructure
         var dataAccess = new SqlDataAccess(uniqueKey, connectionString);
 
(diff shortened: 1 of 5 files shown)


ramonsmits:
@Particular/nservicebus-maintainers At runtime the startup tasks fail to initialize. I can't seem to be able to do `builder.Build<RoutingInfoCommunicator>(),`.

Any idea why?

danielmarbach:
@ramonsmits yes that is not possible. RegisterStartupTask doesn't register anything on the container. The previous code did it right, it declared the communicator as a local and then added the local to the startup task list and passed the local to the other instances that needed it. 

As I see the problem is here that you need to get access to CriticalError and as of now CriticalError is only exposed on the container but not on the settings. This forces you to use the builder. I'll send it a PR with a suggestion

danielmarbach:
Please see https://github.com/Particular/docs.particular.net/pull/2502

ramonsmits:
@Particular/docs-maintainers This is ready for merging. Based on a customer that had stability issues after adopting the sample in their code base.

PR #2500, opened 2017-03-01
Added doco on location overrides for ravendb index and log files

ramonsmits:
Based on https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/particularsoftware/Wwwi8nMVUek/MTl6OxgUAgAJ

### DIFF
--- a/servicecontrol/creating-config-file.md
+++ b/servicecontrol/creating-config-file.md
@@ -211,6 +211,26 @@ Type: string
 Default: `100`
 
 
+#### Raven/IndexStoragePath
+
+This configures the embedded RavenDB database to use a different location to store the index data. This can help to improve performance if the disk isn't fast enough and you want to spread the load over multiple disks.
+
+See: https://ravendb.net/docs/article-page/2.5/csharp/server/administration/configuration
+
+Default: ~/Data/Indexes
+
+
+#### Raven/Esent/LogsPath
+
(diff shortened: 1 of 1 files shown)


gbiellem:
Aside from the fact that i don't like repeating RavendDB doco in our doco I also have the following problems with handing out this advice as is. 

1, If we are going to support this config then we should add it as an option to the Management Tool
2. The doco doesn't cover what happens when the instance of service control is deleted.  The deletion via the tools would not pick up the new location so the folders would  be orphaned
3. Has this been tested with upgrades?  

I think a better approach would have been to raise this as a issue and get it supported properly

//cc @Particular/servicecontrol-maintainers

gbiellem:
Seems @boblangley  has already added the config entries into https://docs.particular.net/servicecontrol/creating-config-file#host-settings-servicecontroldbpath  so this PR is redundant.

ramonsmits:
Although I agree having support in our product for this would be nice. Why can't we just document it mentioning the caveats as you describe? If we then have support for it in our product we can always label the doco obsolete.

gbiellem:
@ramonsmits  The amount of effort to make this supported is small and I plan to do it in the next few days and get it out. So there will be no need to document the caveats. I think this is actually something a lot of our customer could make use of as we often get queried about performance on VMs,

Issue #2352, opened 2016-12-21
Custom routing sample active/inactive routes not working properly


ramonsmits:
The custom routing sample is not working properly: https://docs.particular.net/samples/routing/custom/

### Inactive route still used

* Start all instances
* Exit sales-1
* Send multiple commands via Client
* Messages are still send round robin

Expected sending only to active route.

### When no active routes sending command fails:

* Start and stop all instances so that subscriptions are configured
* Start Client
* Send command
* Failure because no there are instances

Expected round robin between inactive routes

Active/passive logic must be fixed or removed.

//cc: @SzymonPobiega

vahid8261:
I am suing this sample to use automatic routing. No subscriber is receiving events they subscribe to when using this sample code for automatic routing.  Please let me know how can we fix this ?

weralabaj:
@ramonsmits @SzymonPobiega is that issue still relevant or has it been addressed?

ramonsmits:
@weralabaj Not sure, I see that @SzymonPobiega made this commit 2616eb8309e5ec98f98d050cbdc6c101e66ec7e5 so probably it is.

weralabaj:
@ramonsmits just run the sample now

> When no active routes sending command fails:

After starting Sales-1 and Sales-2 seems the messages are handled in the round-robin fashion. 

> Inactive route still used

Behaves as you described. Some messages are queued for sales-2 and get picked up by it after restarting. However, isn't that the expected behaviour?

SzymonPobiega:
I removed the failure detection to make the sample more clear and focused: https://github.com/Particular/docs.particular.net/pull/2503

PR #2503, opened 2017-03-02
Removed failure detection

SzymonPobiega:
Fixes https://github.com/Particular/docs.particular.net/issues/2352


weralabaj:
smart @SzymonPobiega ;)

SzymonPobiega:
@weralabaj the best code is deleted code. It contains no bugs :-)

weralabaj:
ok i'll smoke test and merge

weralabaj:
@SzymonPobiega do we also need to remove this fragment?

> This information is updated every time the automatic routing feature detects a change in the topology:

After restarting one of the sales processes the other is not really picking up the change. Or do we need to clarify something here?

SzymonPobiega:
@weralabaj I've smoke tested it and works. When I kill `Sales2` the client continue to send every second messages to `Sales2` queue (because the automatic failure detection has been removed). Then when I start it back, it processes all the queued up messages and normally processes the new coming messages.

PR #2513, opened 2017-03-03
[Core 6.2]Doco for the receive pipeline completed event

andreasohlund:
That will ship with 6.2

https://github.com/Particular/NServiceBus/pull/4509


andreasohlund:
Note that we plan to align the syntax for https://docs.particular.net/nservicebus/recoverability/subscribing-to-error-notifications in the near future

Snippets/Core/Core_6/Pipeline/OnReceivePipelineCompleted.cs
```diff
@@ -0,0 +1,25 @@
...
+
+    class OnReceivePipelineCompleted
+    {
+        public OnReceivePipelineCompleted()
+        {
+            var endpointConfiguration = new EndpointConfiguration("MyEndpoint");
+
+            #region ReceivePipelineCompletedSubscription
+
+            endpointConfiguration.Pipeline.OnReceivePipelineCompleted(e =>
```
dvdstelt: When writing a feature, you would use `context.Pipeline.OnReceivePipelineCompleted(e =>`. Would that not be a better example to provide?

andreasohlund: Both are relevant, should I add a snippet showing a feature as well?

dvdstelt: Initially I wasn't sure, but I like to be explicit. Results in less questions from users. And the page doesn't contain the much information yet, so it's probably better to add it.

andreasohlund:
Added snippet to show subscribing from a feature as well

Issue #2483, opened 2017-02-27
Feedback: Unit of Work - NServiceBus


Niklas-Peter:
Page: https://docs.particular.net/nservicebus/pipeline/unit-of-work

> In Version 6 IManageUnitsOfWork does not wrap the Outbox transaction.

Yes, I carefully read the article https://docs.particular.net/nservicebus/outbox/, but I still wonder what exactly the "Outbox transaction" is and which implications not wrapping it has.

Furthermore, in the article about the Outbox, why do you have to suppress the queue transaction for "Dispatch" and "Mark as dispatch"? These 2 are the last steps in the image, so you could simply complete the queue transaction before instead of suppressing it. If you just want to prevent sending be a part of the outer queue transaction, I have no idea, why you would want to do that.
Although I spent quite a lot of time on trying to fully understand the Outbox article, I did not get the transaction handling depicted in the image.

indualagarsamy:
@Niklas-Peter - Thank you for your feedback. Much appreciated. 

ping @Particular/nservicebus-maintainers - Can you please review the documentation and add more clarification to address @Niklas-Peter's questions?

timbussmann:
I agree with @Niklas-Peter , the statement about the outbox transaction is confusing. I tracked it down to the following issue: https://github.com/Particular/NServiceBus/issues/4134

so maybe something like
> In Version 6 IManageUnitsOfWork does not wrap the persistence transactions. The persistence transaction can therefore still fail after the unit of work completed.

thoughts? /cc @tmasternak (you wrote that note I think)

> Furthermore, in the article about the Outbox, why do you have to suppress the queue transaction for "Dispatch" and "Mark as dispatch"? These 2 are the last steps in the image, so you could simply complete the queue transaction before instead of suppressing it. 

if we would complete the queue transaction, the message is consumed. If then the dispatching fails, the consumed message will never be processed again and the outgoing messages are lost. With the wrapping transaction, if the suppressed dispatch fails, any exception will still cause the message to fail (the suppress is required to not cause a rollback of dispatched messages in case the incoming message can't be completed for whatever reason). Through retry mechanism, NServiceBus then tries to dispatch the stored outbox messages again. Does this explanation help you understand the outbox design?

tmasternak:
> In Version 6 IManageUnitsOfWork does not wrap the persistence transactions. The persistence transaction can therefore still fail after the unit of work completed.

[...]

PR #2519, opened 2017-03-06
Clarify transaction wrapping note

timbussmann:
addresses https://github.com/Particular/docs.particular.net/issues/2483

Is there the need to have the partial using an include? @Particular/docs-maintainers

### DIFF
--- a/nservicebus/imanageunitsofwork-outbox.include.md
+++ b/nservicebus/imanageunitsofwork-outbox.include.md
@@ -1,1 +1,0 @@
-NOTE: In Version 6 `IManageUnitsOfWork` does not wrap the Outbox transaction. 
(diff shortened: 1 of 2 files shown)


weralabaj:
@timbussmann it's also used in another place https://github.com/Particular/docs.particular.net/search?utf8=%E2%9C%93&q=imanageunitsofwork-outbox, so if that needs to be kept in sync we need include, if not (and seems to be a minor thing) then we can get rid of the include

timbussmann:
@weralabaj I don't see it used anywhere else? One usage is the partial and the other is the include by the partial which both have the same name, or am I missing something?

weralabaj:
ah... right, super confusing. not necesssary at all :)

PR #2521, opened 2017-03-07
Show how to configure feature settings from EndpointConfiguration

timbussmann:
tries to address a feedback received here: https://github.com/Particular/NServiceBus/issues/4500#issuecomment-284465740

### DIFF
--- a/nservicebus/pipeline/features.md
+++ b/nservicebus/pipeline/features.md
@@ -107,8 +107,8 @@ Note: Features are automatically detected and registered by NServiceBus when the
 
 ## Feature settings
 
+The settings are a good way to share data and configuration between the feature and the endpoint configuration, or various features. Settings can be accessed via the `FeatureConfigurationContext.Settings` property during the *setup phase*. Settings can be configured using *defaults* or `EndpointConfiguration`:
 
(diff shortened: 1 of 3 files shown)


nservicebus/pipeline/features.md
```diff
@@ -114,6 +105,29 @@ snippet:FeatureSetup
...
+The code above configures the key `"Key"` to contain value `"Value"` and the key `"OtherKey"` to contain value `42` as a default value. The querying API allows to distinguish these two types of registrations. Usually, for a given key, a feature registers a default value in this way and also exposes an extension method of the endpoint configuration to allow a user to override this default value.
+
+
+### EndpointConfiguration
+
+The settings can already be accessed during endpoint configuration:
+
+snippet:WriteSettingsFromEndpointConfiguration
+
+Note that *defaults* have not yet been applied at endpoint configuration time.
```
weralabaj: @timbussmann But when it comes to "defaults" and "custom/overrides", the latter always takes precedence, irrelevant of the fact in which order they were applied. Correct?

weralabaj: Also would it work the same way for V5 in https://github.com/Particular/docs.particular.net/pull/2521/files#diff-8e9647fe63d57814e952b878c7fd7534R85? I'll add the snippet.

timbussmann: @weralabaj yes that's correct. But in case you try to do a Get, you won't see the values set by `defaults`. The naming with the settings and feature.defaults is confusing anways, as Feature.Defaults is just a way to access settings before the setup and is independent of settings.set or settings.setdefault

PR #2527, opened 2017-03-08
Added account name validation

gbiellem:
Updated the CreateQueue powershell sample to validate the account name,  previously it would create the queue regardless and them throw a non terminating error  when setting permissions if an invalid account was passed to it


gbiellem:
@SimonCropp  review?

PR #2535, opened 2017-03-09
Adding note about IEvent handlers and automatic subscriptions

WojcikMike:
@timbussmann @SzymonPobiega please have a look

### DIFF
--- a/nservicebus/messaging/publish-subscribe/controlling-what-is-subscribed.md
+++ b/nservicebus/messaging/publish-subscribe/controlling-what-is-subscribed.md
@@ -19,6 +19,8 @@ Messages matching the following criteria will be auto subscribed at startup.
  1. If the selected transport [does not support publish-subscribe natively](/nservicebus/transports/#types-of-transports-unicast-only-transports), the publisher for that message needs to be specified via the [routing](/nservicebus/messaging/routing.md) API.
 
 
+WARNING: Handler of `IEvent` will not cause automatic subscription to be send. The handler will still be executed for every event that is processed by that endpoint.
+
 ### Exclude sagas from auto subscribe
 
 partial: exclude


SzymonPobiega:
I am not convinced we should document this behavior. It looks like a bug to me. 

When I looked up it seems this behavior is caused by `AutoSubscribe` doing

```
.Where(conventions.IsEventType) //only events unless the user asked for all messages
```

and the default convetions are:

```
internal Func<Type, bool> IsEventTypeAction = t => typeof(IEvent).IsAssignableFrom(t) && typeof(IEvent) != t;
```

So adding `|| t == typeof(IEvent)` should fix this. 

@Particular/nservicebus-maintainers or is it a desired behavior and there is reason why it works in such a way?

WojcikMike:
For the record there is a comment in here stating that someone was aware of this behavior: https://github.com/Particular/NServiceBus/blob/b0ec581accd76443480035847f3c8c64fa1a5f89/src/NServiceBus.AcceptanceTests/Routing/When_publishing_with_only_local_messagehandlers.cs#L56

timbussmann:
> I am not convinced we should document this behavior.

me neither.

> It looks like a bug to me.

I tend to disagree with that statement though.

I think the only valid scenario for such a handler would be a "catch all" handler for all events subscribed to by that endpoint via more specific handlers or via manual subscribe calls. 

as for brokered transports, where should IEvent handlers subscribe to? To all events? That sounds like a really bad idea.

bording:
@timbussmann Yes, you'd receive every event published to the broker if you did that.

timbussmann:
I don't think we should support that @bording the behavior also seems to heavily depend on the transport.

SzymonPobiega:
What's the difference between supporting automatic subscriptions to `IEvent` and customer doing 

```
public interface IMyEvent : IEvent {}
```

? I don't see any.

Another thing. We say that you might use handlers and their order to perform some generic actions before a specific handler is invoked. If I do this with `MyGenericHandler<IEvent>` it would not cause subscribing to everything but if I do it with `IMyEvent`, it will. Strange?

timbussmann:
> Another thing. We say that you might use handlers and their order to perform some generic actions before a specific handler is invoked.

if that logic should run before all events, use a behavior.

> If I do this with MyGenericHandler<IEvent> it would not cause subscribing to everything but if I do it with IMyEvent, it will. Strange?

the scenario you were described was also involving a more specific handler for a specific event, that handler would cause a subscription already, so no need for the generic handler to subscribe too?

with MDPS a "subscribe all" IEvent handler would only subscribe to registered publishers. I can't see a reason why you wouldn't have a more specific message handler around in the first class. Also you can always just manually subscribe which seems to make more sense in case you really want to subscribe to several events without having a specific handler for them.
With brokers, "subscribe all" would turn this endpoint essentially in an auditing endpoint, which is not something I think we ever stated as a viable auditing solution  and doesn't sound like a good design for a regular endpoint to me.

SzymonPobiega:
@timbussmann it all makes sense but how do you want to tell it the users? Generally having a handler for event X automatically subscribes you for event Y and Z which inherit from X. But this is not the case with `IEvent`? 

I mean, for the subscription sake we would treat `IEvent` as something else than event (by default). Overriding the conventions add another set of problems to it. While I don't claim it is a good practice to subscribe to `IEvent` it seems to me a lesser evil because it keeps stuff consistent. Less surprises.

timbussmann:
what if you want to use an `IEvent` handler to have a general handler for all incoming events (which I'd argue again should use a behavior instead, but let's ignore that). If we change the behavior you would suddenly receive all sorts of events on a brokered transports which also seems to be quite a surprise.

can we state something like  "`IHandleMessages<IEvent>` will be invoked for all events subscribed by the endpoint, but will not cause additional subscriptions to be issued"?

timbussmann:
ups, hit the wrong button, sorry.

weralabaj:
IMO a docs PR is not a good place to have this discussion. If we don't have clarity internally how that should be handled and how it should work, then how are we going to educate the users?

SimonCropp:
seconding the suggestion from  @weralabaj

PR #2536, opened 2017-03-09
Compile error added to Deprecated Address section to match a search

ramonsmits:


SimonCropp:
Given we do not document the 100s of other obsolete error messages what makes this one special ?

SimonCropp:
@ramonsmits if you think the obsolete is not good enough please discuss with @Particular/nservicebus-maintainers. if you feel we should be documenting all obsoletes then please raise a plat dev issue since it is a significant amount of work.

PR #2548, opened 2017-03-14
Document web application container integration recommendations

### LINKED ISSUE #2479: Document hosting full duplex endpoint and web controllers in same host
Spun out from https://github.com/Particular/docs.particular.net/pull/2424#issuecomment-278872635

> If we put `ISomeService` in the NSB container then we can get it in handlers but not in controllers.
>
> If we put `ISomeService` in the MVC container then we can get it in controllers but not in handlers.
>
> If we put `ISomeService` in BOTH containers then we can get it in both but they are going to be different objects with different lifetimes.
>
> If we register in both containers as a singleton then we STILL get two instances, one out of each container.
>
> If we really need/want to share access to a single instance then we're better off creating it outside of the containers and registering that single instance in each container with an external lifetime scope!

We recommend that customers do not host message handlers/sagas in the same host as http request handlers. Some of the reasoning behind this is documented in [Publishing from Web Applications](https://docs.particular.net/nservicebus/hosting/publishing-from-web-applications).

It is also no longer possible to share a single IoC container between the endpoint and the web host so any solution requiring an IoC container will now require two IoC containers, one for the NSB endpoint and one for the web request handlers. This is OK as long as they don't have to share any resources. As soon as they do, you will run into difficulties getting the lifetimes of the shared resources working properly.

We should move all of this into a guidance document explaining why not to do it, and how to do it properly if you really need to. Perhaps this should be a sub-page of https://docs.particular.net/nservicebus/hosting/web-application

mikeminutillo:
Fixes #2479

### DIFF
--- a/nservicebus/hosting/web-application.md
+++ b/nservicebus/hosting/web-application.md
@@ -22,6 +22,26 @@ NServiceBus can be hosted in any web technology that support .NET. This includes
  * [Web API](https://www.asp.net/web-api)
  * [NancyFX](http://nancyfx.org/)
 
-As most web technologies operate in a scale out manner NServiceBus is **usually** hosted in a "Send-only" manner. In this mode the give web application technology acts as a "forwarder" of messages rather than the "processor". So the handling code (MVC controller, NancyFX module etc) of a given web request simply leverages the  send APIs and no processing is done in the web process. The actual message handling is done in a [Windows Service Endpoint](windows-service.md).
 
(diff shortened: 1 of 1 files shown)


mikeminutillo:
It could probably do with a specific sample but I think I captured the gist of the problem. I didn't make it version specific because I suspect even for V5 customers we want them to work this way so that they can upgrade to V6 cleanly.

@Particular/container-maintainers care to have a review?

SimonCropp:
@boblangley thanks

PR #2553, opened 2017-03-16
fixes and improvements to samples/web/send-from-aspnetcore-webapi/sample.md

adamralph:
Note that a lot of these changes are fixes for obvious typos. I'm surprised the document made it in with so many obvious typos, e.g.

> Howeve it is not required for this sample anc can be remove.


weralabaj:
>  I'm surprised the document made it in with so many obvious typos

Since Indu has left maintainers group, we have a spot open, if you're interested ;)

PR #2555, opened 2017-03-16
Clarify order of pipeline execution

SeanFeldman:
**Changes**:

1. `Ancillary Actions` moved to the bottom to show that message processing is taking the place first
1. Assigned step numbers to highlight the dependency on the message processing before auditing and forwarding can take place

The new rendering looks like the following

image

**Reason for the change**:

While reviewing the [Audit filter pipeline extension sample](https://docs.particular.net/samples/pipeline/audit-filtering/) it was not clear how it would work if `"AuditFilter.Rules"` is connected to the expected next execution of the `"AuditFilter.Filter"` behavior if they are not "connected" in any shape or form on the diagram.


SeanFeldman:

PR #2569, opened 2017-03-22
For callbacks endpoints need to be uniquely addressable

### LINKED ISSUE #2561: Feedback: 'Callback Changes in Version 6'
Feedback for 'Callback Changes in Version 6' https://docs.particular.net/nservicebus/upgrades/5to6/callbacks

Location in GitHub: https://github.com/Particular/docs.particular.net/blob/master/nservicebus/upgrades/5to6/callbacks.md

> In Version 6 the callback handling on broker transports is no longer based on the machine host name. In order to continue using callbacks when using a broker transport, the endpoint needs to be uniquely addressable

When I read this paragraph, I assumed that all this (and the required configuration with `MakeInstanceUniquelyAddressable`) does not apply to MSMQ since MSMQ is not a broker transport. But it seems like this is not the case. If `MakeInstanceUniquelyAddressable` is not called on a MSMQ endpoint, trying to start the endpoint throws an exception with the following message:

> In order to use the callbacks feature you need to specify an endpoint instance ID via EndpointConfiguration.MakeInstanceUniquelyAddressable(string discriminator)

danielmarbach:
Closes #2561

### DIFF
--- a/nservicebus/upgrades/5to6/callbacks.md
+++ b/nservicebus/upgrades/5to6/callbacks.md
@@ -16,7 +16,7 @@ The `NServiceBus.Callbacks` package has to be referenced only by the requesting 
 
 snippet: 5to6-Callbacks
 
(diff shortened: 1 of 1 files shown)


danielmarbach:
@Particular/nservicebus-maintainers please review

PR #2583, opened 2017-03-27
azure/index.md hidden

Scooletz:
Hide the index file that because of the packages being separated now, provides no value as there's no one `Azure` thing.

### DIFF
--- a/nservicebus/azure/index.md
+++ b/nservicebus/azure/index.md
@@ -4,4 +4,5 @@ summary: Using Azure for endpoint hosting and to provide Transports and Persiste
 tags:
 - Azure
 reviewed: 2016-09-21
+hidden: true
 ---


Scooletz:
Can you review the idea of hiding an unneeded `index.md` for azure directory? If ok, pls merge.

Scooletz:
I'll try to propose a better landing page for `Azure` instead of just hiding the index.

SeanFeldman:
We've discussed this give that

> Whoever is searching for "NServiceBus Azure" is not looking into a specific component API, but more of a general info. Even having something that says "you can run on Azure using the following hosting options and the following component (links) could be helpful. Or having in related links section the collection of all the components that users can navigate to.

decided that trying out the approach of a landing azure page that is guiding towards components is what we'll try first.

PR #2591, opened 2017-03-29
Adding a troubleshooting page for ServiceInsight

HEskandari:
This closes https://github.com/Particular/ServiceInsight/issues/541

### DIFF
--- a/menu/menu.yaml
+++ b/menu/menu.yaml
@@ -760,6 +760,9 @@
       Title: Logging
     - Url: serviceinsight/no-data-available
       Title: Data not available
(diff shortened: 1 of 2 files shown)


SimonCropp:
under what circumstances would this happen?

HEskandari:
@SimonCropp this has happened in the past when the config file is missing. As to why it is missing, we don't know!

SimonCropp:
so isnt it good enough that this is a known bug with a work around. that should be enough for people to discover it by google and to point people to in support cases. ie no need for doco

PR #2610, opened 2017-04-03
Minor SF routing update

danielmarbach:


danielmarbach:
@Particular/docs-maintainers we currently use the prerelease.txt to indicate that we are using a pre release version of NServiceBus. But this sample now uses Service Fabric persistence 1.0.0-beta0001. Is it still OK to remove the prerelease.txt since the message at the top is misleading?

PR #2615, opened 2017-04-04
Clarifying additional callback queue

### LINKED ISSUE #2563: Feedback: 'Client side Callbacks'
Feedback for 'Client side Callbacks' https://docs.particular.net/nservicebus/messaging/callbacks

Location in GitHub: https://github.com/Particular/docs.particular.net/blob/master/nservicebus/messaging/callbacks.md

I think this paragraph should be improved:

> The endpoint queue name will include the specified suffix. When upgrading from lower versions the original input queue (without a suffix) will still exist, and the additional queue with the suffix will be created.

I interpreted the "upgrading from lower versions" as "upgrading from NServiceBus versions 5 and below". And then I interpreted the "will still exist" part as "the new version won't create the queue without a suffix". But this doesn't seem correct because if I create a new endpoint with NServiceBus 6 on which `MakeInstanceUniquelyAddressable` was called, both types of queues, with and without a suffix, are created. At least with MSMQ, maybe this is dependend on the transport?

dvdstelt:
Fixes #2563 

User mentioned that difference between shared queue and callback queue wasn't clear.

### DIFF
--- a/nservicebus/messaging/callbacks_route_callbacks_[1,).partial.md
+++ b/nservicebus/messaging/callbacks_route_callbacks_[1,).partial.md
@@ -3,7 +3,7 @@ Callback responses are routed based on the `ReplyTo` header of the request. In o
 
 snippet: Callbacks-InstanceId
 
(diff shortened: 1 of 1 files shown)


dvdstelt:
Not sure who should review this, the @Particular/nservicebus-maintainers ?

PR #2632, opened 2017-04-12
Updating support policy for NServiceBus components

DavidBoike:
Result of https://github.com/Particular/PlatformDevelopment/issues/1133


nservicebus/upgrades/support-policy.md
```diff
@@ -1,15 +1,30 @@
...
  - nservicebus/upgrades/release-policy
+ - nservicebus/upgrades/supported-versions
 ---
 
 ## Supported Versions
 
-Supported versions include any major version released within the last 3 years. Within a supported major version, the latest minor version is supported, or any minor version released within the last year.
+Supported versions of NServiceBus and its associated packages are determined by the release date of the package that replaces them. After a specific version of a package is replaced by a newer version, the older package will remain supported for a fixed period of time afterward. This enables starting a project with the newest versions at the time, with the confidence that those versions will be supported well into the future, regardless of the release schedule for new components.
+
+For the core NServiceBus package:
```
SimonCropp: Perhaps  make this and the bellow nested headings. So they can be linked to

nservicebus/upgrades/all-versions.include.md
```diff
@@ -0,0 +1,2095 @@
+### NServiceBus
+
+<table>
+  <tr>
+    <th>Version</th>
+    <th>Released</th>
+    <th>Supported until</th>
```
SimonCropp: Did u consider making this a markdown table. It would make the file much smaller and concise.

nservicebus/upgrades/support-policy.md
```diff
@@ -33,3 +50,28 @@ NServiceBus versions are wire-compatible, meaning the endpoints using different
...
+
+No, the policy applies only to the supported NServiceBus packages listed on the [NServiceBus Packages Supported Versions](supported-versions.md) page.
+
+**Which version of NServiceBus should be used to start a new project?**
+
+It's preferable to use the newest major version of NServiceBus when starting a new project. The support policy guarantees that the major version will be supported for at least 2 years after it is replaced by a new major version.
+
+**Which version of NServiceBus should be used when adding to an existing project?**
+
+When adding new functionality to an existing project, it's generally easiest to use the same version of NServiceBus already in use by other endpoints, as long as that version of NServiceBus is still supported. It is also possible to use a newer version of NServiceBus in a specific endpoint as a step towards [upgrading](#upgrading) the entire system.
```
SimonCropp: I think "all other endpoints is misleading". If the code for endpoints do not share a common lib that refs nsb, and do not exist in the same sln, then it is fine for them use diff versions of nsb

nservicebus/upgrades/support-policy.md
```diff
@@ -1,15 +1,30 @@
...
+* Minor versions of a supported major version are supported for a period of 6 months after the release of the next minor version.
+* Only the latest patch release of each supported minor version is supported.
+
+For component packages that depend upon NServiceBus:
+
+* Major versions are supported for a period of 1 year after the release of the next major version.
+* Minor versions of a supported major version are supported for a period of 3 months after the release of the next minor version.
+* Only the latest patch release of each supported minor version is supported.
+* A version can only be supported if the NServiceBus package on which it depends is also currently supported.
+* If the version is the last version which works with a given version of NServiceBus, the version will be supported for as long as that version of NServiceBus is supported.
```
SimonCropp: "Last" could be ambiguous. Perhaps "most recent"

SimonCropp:
Should the menu change at all?

SimonCropp:
Btw this is some very good work. Well thought out and much easier for consumers to understand

nservicebus/upgrades/all-versions.include.md
adamralph: I did consider it, but it would mean either padding all the columns to the width of the largest item anywhere in the table:

```markdown
| Version | Released   | Supported until | Explanation                   |
|---------|------------|-----------------|-------------------------------|
| 7.2.x   | 2017-03-24 | -               | -                             |
| 7.0.x   | 2016-10-11 | 2017-06-24      | Superseded by 7.2.x           |
| 6.2.x   | 2015-02-13 | 2018-10-11      | Extended by NServiceBus 5.2.x |
```

or not padding at all:

```markdown
| Version | Released | Supported until | Explanation |
|-|-|-|-|
| 7.2.x | 2017-03-24 | - | - |
| 7.0.x | 2016-10-11 | 2017-06-24 | Superseded by 7.2.x |
| 6.2.x | 2015-02-13 | 2018-10-11 | Extended by NServiceBus 5.2.x |
```

My appetite for both options was limited.

@SimonCropp how strongly do you feel about it?

nservicebus/upgrades/support-policy.md
adamralph: 👍 I agree. I'll commit something.

nservicebus/upgrades/support-policy.md
adamralph: 👍 agreed. Will change.

nservicebus/upgrades/support-policy.md
adamralph: @bording @DavidBoike I'm wondering if we shouldn't remove this FAQ entirely. When adding an endpoint to an existing system, use of the same version should be the natural default that people reach for. Upgrading is covered elsewhere. I'm not sure this FAQ has a place in a support policy doc.

adamralph:
> Should the menu change at all?

Perhaps we should introduce "Supported versions" under "Support policy"? Perhaps "All versions" too?

- Support Policy
  - Supported versions
  - All versions

SimonCropp:
@adamralph happy with that

nservicebus/upgrades/all-versions.include.md
DavidBoike: @adamralph I agreed with that when we discussed it during development, but as I was transitioning the change to the Docs repo, I reran the script. Because NSB.RavenDB 3.2 was released, there was a change, and I noticed that because the table layout is so spread out vertically, I couldn't even tell from the diff hunk which package the change was for without going into a full diff tool.

Also, using MD tables would make it trivially easy to center a column, like "Supported until" which I think needs centering.

So I'm flipping to supporting markdown tables. Padded is nice, but I don't think we need to do anything fancy like figure out what is longest and then dynamically pad. I think it can just be hardcoded widths.

nservicebus/upgrades/support-policy.md
DavidBoike: You mean just this question, right? If so I'm :+1:

nservicebus/upgrades/all-versions.include.md
adamralph: Fair enough. I'll have a go at the script to switch to MD tables.

nservicebus/upgrades/support-policy.md
adamralph: Yep, just this question.

PR #2647, opened 2017-04-18
Update send-a-message_imessagessession-warning_core_[6.0,].partial.md

HS-Zero:
Interface name typo


SimonCropp:
@HS-Zero thanks. good catch

PR #2648, opened 2017-04-18
Update moving-away-from-ibus.md

HS-Zero:
Interface typo


SimonCropp:
@HS-Zero i think i go that one already ? :) 
https://github.com/Particular/docs.particular.net/commit/c6ff6c2d84e9743ba29ed4a66e26dc0c91409eea

SimonCropp:
will merge anyway so u get the Kudos for another PR :)

Issue #2531, opened 2017-03-09
Review setting up SC with SI doco experience


mikeminutillo:
We had a customer go through this and a number of things were not obvious fro reading the existing doco. 

From the [GG post](https://groups.google.com/d/msg/particularsoftware/Nlf9o_8OMlU/MejTASpjDwAJ):

- If audits aren't enabled, SC will show no endpoints because it uses the contents of the queues to populate rather than the existence of the queues.So if the queues are empty, it looks like there are no endpoints. 
- The docs say that every endpoint gets a built in audit log. Apparently that's not true, it's not per endpoint, and it's not by default.
- The docs say that a central audit log is 'also' required. Apparently, this is the same audit queue that every endpoint should be configured to use. 
- The audit queue is actually 'consumed' by ServiceControl which means it will be removed from the audit queue and stored in service control in what seems to be a raven database, and deleted after so much time.

boblangley:
Added a PR to address these observations: #2653, awaiting @mikeminutillo's review/refinement.

weralabaj:
@boblangley since the PR was merged I believe this issues has been addressed. If something is missing, please reopen.

PR #2653, opened 2017-04-19
Clarify SC/SP/SI Audit behavior

boblangley:
Completes #2531 

- [x] _If audits aren't enabled, SC will show no endpoints because it uses the contents of the queues to populate rather than the existence of the queues.So if the queues are empty, it looks like there are no endpoints._
   - [Note added to ServiceInsight index](https://github.com/Particular/docs.particular.net/pull/2653/files#diff-64f51c7ee0e998ad4611a1870370d25cR21)
   - [Note added to ServicePulse index](https://github.com/Particular/docs.particular.net/pull/2653/files#diff-5eea945ad2b91dc2b70ef1475ce4a5feR28)
- [x] _The docs say that every endpoint gets a built in audit log. Apparently that's not true, it's not per endpoint, and it's not by default._
   - [Note added to Audit feature doco](https://github.com/Particular/docs.particular.net/pull/2653/files#diff-13962c69f23c09efc3f7c439de6048f1R16)
- [x] _The docs say that a central audit log is 'also' required. Apparently, this is the same audit queue that every endpoint should be configured to use._
   - [Note added to ServiceControl index](https://github.com/Particular/docs.particular.net/pull/2653/files#diff-6fa3ad603af6bfeabca4ac626169b9d1R13)
- [x] The audit queue is actually 'consumed' by ServiceControl which means it will be removed from the audit queue and stored in service control in what seems to be a raven database, and deleted after so much time
   - [Note added to ServiceControl index](https://github.com/Particular/docs.particular.net/pull/2653/files#diff-6fa3ad603af6bfeabca4ac626169b9d1R15)

### DIFF
--- a/nservicebus/operations/auditing.md
+++ b/nservicebus/operations/auditing.md
@@ -13,7 +13,7 @@ redirects:
 
 The distributed nature of parallel, message-driven systems make them more difficult to debug than their sequential, synchronous and centralized counterparts. For these reasons, NServiceBus provides built-in message auditing for every endpoint. Configure NServiceBus to audit and it will capture a copy of every received message and forward it to a specified audit queue.
 
-NOTE: By default Auditing is not enabled and must be configured.
+NOTE: By default Auditing is not enabled and must be configured for each endpoint.
 
 It is recommended to specify a central auditing queue for all related endpoints (i.e. endpoints that belong to the same system). This gives an overview of the entire system in one place and see how messages correlate. One can look at the audit queue as a central record of everything that is happening in the system. A central audit queue is also required by the Particular Service Platform and especially [ServiceControl](/servicecontrol), which consumes messages from these auditing queues. For more information, see [ServicePulse documentation](/servicepulse/).
 
--- a/nservicebus/operations/auditing.md
+++ b/nservicebus/operations/auditing.md
@@ -13,6 +13,8 @@ redirects:
 
 The distributed nature of parallel, message-driven systems make them more difficult to debug than their sequential, synchronous and centralized counterparts. For these reasons, NServiceBus provides built-in message auditing for every endpoint. Configure NServiceBus to audit and it will capture a copy of every received message and forward it to a specified audit queue.
 
+NOTE: By default Auditing is not enabled and must be configured.
+
 It is recommended to specify a central auditing queue for all related endpoints (i.e. endpoints that belong to the same system). This gives an overview of the entire system in one place and see how messages correlate. One can look at the audit queue as a central record of everything that is happening in the system. A central audit queue is also required by the Particular Service Platform and especially [ServiceControl](/servicecontrol), which consumes messages from these auditing queues. For more information, see [ServicePulse documentation](/servicepulse/).
 
 
--- a/platform/servicecontrol.include.md
+++ b/platform/servicecontrol.include.md
@@ -10,6 +10,8 @@ To enable [ServiceControl](/servicecontrol) to gather this information, configur
  * configure [recoverability](/nservicebus/recoverability) to store information on messages failures;
  * [install plugins on the endpoints](/servicecontrol/plugins/) to monitor their health and sagas and use custom checks.
 
+NOTE: All endpoints should be configured to forward to the same audit, error, and ServiceControl plugin queues unless the system has been sharded between multiple ServiceControl instances.
+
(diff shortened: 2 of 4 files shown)


boblangley:
@mikeminutillo I took a quick `NOTE` heavy stab at addressing these 4 observations. Feel free to modify as you see fit.

nservicebus/operations/auditing.md
```diff
@@ -13,6 +13,8 @@ redirects:
 
 The distributed nature of parallel, message-driven systems make them more difficult to debug than their sequential, synchronous and centralized counterparts. For these reasons, NServiceBus provides built-in message auditing for every endpoint. Configure NServiceBus to audit and it will capture a copy of every received message and forward it to a specified audit queue.
 
+NOTE: By default Auditing is not enabled and must be configured.
```
WilliamBZA: Is it worth being explicit here and saying:

> and must be configured for each endpoint

boblangley: Yes, adding it.

boblangley:
@Particular/serviceinsight-maintainers @Particular/servicecontrol-maintainers  @Particular/docs-maintainers Do you any of you want to give this PR a once over? I will wait until tomorrow and if I don't get any feedback will go ahead and merge as-is.

gbiellem:
@boblangley  LGTM :+1:

nservicebus/operations/auditing.md
```diff
@@ -13,6 +13,8 @@ redirects:
 
 The distributed nature of parallel, message-driven systems make them more difficult to debug than their sequential, synchronous and centralized counterparts. For these reasons, NServiceBus provides built-in message auditing for every endpoint. Configure NServiceBus to audit and it will capture a copy of every received message and forward it to a specified audit queue.
 
+NOTE: By default Auditing is not enabled and must be configured for each endpoint.
```
mikeminutillo: Stylistic comment: I would have gone with

> Auditing is not enabled by default...

I don't mind either way though

platform/servicecontrol.include.md
```diff
@@ -10,6 +10,10 @@ To enable [ServiceControl](/servicecontrol) to gather this information, configur
  * configure [recoverability](/nservicebus/recoverability) to store information on messages failures;
  * [install plugins on the endpoints](/servicecontrol/plugins/) to monitor their health and sagas and use custom checks.
 
+NOTE: All endpoints should be configured to forward to the same audit, error, and ServiceControl plugin queues unless the system has been sharded between multiple ServiceControl instances.
```
mikeminutillo: Is there any doco on sharding? It might be better to just unilaterally tell people to send everything to the same queues. If people make the decision to shard then it is up to them to ignore this rule (or this comment could be moved to that doco)

servicepulse/index.md
```diff
@@ -25,6 +25,8 @@ The Dashboard provides a visual overview of the current state of the monitored N
 
 ServicePulse automatically detects what endpoints exist in the system by analyzing metadata from [audited messages](/nservicebus/operations/auditing.md). All detected endpoints will, by default, be listed in the Endpoints tab, but will not be automatically monitored.
 
+NOTE: Endpoints will not appear unless messages from the endpoint have been successfully processed by ServiceControl.
```
mikeminutillo: If an endpoint sends heartbeats it will appear. I know these are technically messages but it might be worth being explicit

platform/servicecontrol.include.md
boblangley: I am OK with removing the end part from this PR. I thought we talked about splitting your environment in our guidance doco, but apparently not. Since we sometimes provide guidance for large scale customers to split their environment between multiple SC installs, perhaps we should plan another PR to add that option?

servicepulse/index.md
boblangley: >Endpoints will not appear unless messages or heartbeats from the endpoint have been successfully processed by ServiceControl

That work?

nservicebus/operations/auditing.md
boblangley: Neither do I, I am happy to change it.

platform/servicecontrol.include.md
weralabaj: > perhaps we should plan another PR to add that option

that sounds good

I think we can leave as is, even if it doesn't explain how it works it notes there's a possibility, so if somebody is looking for something like this they'll start asking us questions

servicepulse/index.md
weralabaj: soon we'll be also sending metrics as messages, so I don't think it's necessary to mention heartbeats explicitly

weralabaj:
@mikeminutillo I'm tempted to merge as is, have a look at my comments and see if you're ok with it

PR #2656, opened 2017-04-20
Update licensing page

bording:
I made a general review pass over the content, and added a note about the new behavior that will be in 6.3.

@Particular/nservicebus-maintainers What do you think? Should there be more information about the new behavior than what I've added so far?


nservicebus/licensing/index.md
```diff
@@ -32,40 +32,41 @@ partial: validity
...
 
 partial: registry
 
 partial: registry-caveats
 
 
-### Using a License sub-directory
+### Using a license subdirectory
 
 A file located at `[AppDomain.CurrentDomain.BaseDirectory]/License/License.xml` will be automatically detected.
```
timbussmann: should we be more clear that this will only be detected by NServiceBus and not SC/SI as mentioned in the registry approach?

nservicebus/licensing/index_order_core_[3,4).partial.md
```diff
@@ -1,2 +1 @@
-
-File located at `{AppDomain.CurrentDomain.BaseDirectory}\License\License.xml`
+File located at `{AppDomain.CurrentDomain.BaseDirectory}\License\License.xml`
```
timbussmann: should that start with "the" or "a" given that the previous sentence is:
>  In order to find the license, NServiceBus will examine:

nservicebus/licensing/index.md
weralabaj: yes, that would make it clearer

weralabaj: btw what happens for SC/SI then? might be good to also mention/link to more info here

nservicebus/licensing/index_order_core_[4,4].partial.md
```diff
@@ -1,21 +1,17 @@
...
-| `HKEY_CURRENT_USER\Software\Wow6432Node\NServiceBus\{Version}\License`            |  1, 2 |
 | `HKEY_LOCAL_MACHINE\Software\NServiceBus\{Version}\License`                       |   2   |
 | `HKEY_LOCAL_MACHINE\Software\Wow6432Node\NServiceBus\{Version}\License`           |  1, 2 |
 
 **Notes:**
 
  1. The `Wow6432Node` registry keys are only accessed if running a 32-bit host on a 64-bit OS.
- 1. Storing licenses in the registry by NServiceBus version was abandoned in NServiceBus 4.3. For backwards compatibility, newer versions of NServiceBus will still check this pattern for versions `4.0`, `4.1`, `4.2`, and `4.3`.
-
+ 1. Storing licenses in the registry by NServiceBus version was abandoned in NServiceBus 4.3. For backwards compatibility, newer versions of NServiceBus will still check this pattern for versions `4.0`, `4.1`, `4.2`, and `4.3`.
```
Comment (weralabaj) [inline nservicebus/licensing/index_order_core_[4,4].partial.md]: It doesn't explain what it does now though, link to up-to-date info or explain briefly

nservicebus/licensing/index_validity_core_[5,).partial.md
```diff
@@ -1,3 +1,3 @@
-The license is valid if the `ExpirationDate` or the `UpgradeProtectionExpiration` attribute in the license data is greater than or equal to the release date of the `{major}.{minor}.0` version of the NServiceBus assembly used by the endpoint. The release dates for the various versions can be found on [NuGet](https://www.nuget.org/packages/nservicebus).
+The license is valid if the `ExpirationDate` or the `UpgradeProtectionExpiration` attribute in the license data is greater than or equal to the release date of the `{major}.{minor}.0` version of the NServiceBus assembly used by the endpoint. To view the release dates for the various versions, see [NServiceBus Packages Versions](/nservicebus/upgrades/all-versions.md).
```
weralabaj: > The license is valid if the `ExpirationDate` or the `UpgradeProtectionExpiration` attribute in the license data is greater than or equal to the release date of the `{major}.{minor}.0` version of the NServiceBus assembly used by the endpoint. 

I wonder if we should explain such things here, because:
1. The users shouldn't care about such low-level implementation details
2. Having that information they can easily modify the license ;) Aren't we provoking that behavior?

@andreasohlund thoughts ☝️

nservicebus/licensing/index_order_core_[4,4].partial.md
Comment (bording) [inline nservicebus/licensing/index_order_core_[4,4].partial.md]: Not sure I understand this comment. The only change I made here was removing an empty line. This partial is only for v4, so there's no need to specify other behavior here.

nservicebus/licensing/index.md
bording: This is the NServiceBus license page, so I don't want to start adding SC/SI information throughout the page. I can see strengthening up the sentence in the registry section to point out that sharing via the registry is the only way to share across products.

bording:
I've pushed up some tweaks that I think should address @timbussmann's feedback.

@andreasohlund What do you think about @weralabaj's comment about the license validity information?

weralabaj:
Since the detailed info about registries was there before, I'll merge now, no reason to wait for this decision. We can modify later if we want to.

nservicebus/licensing/index_validity_core_[5,).partial.md
andreasohlund: No strong opinion, think we added this for our own internal knowledge as well.

>Having that information they can easily modify the license ;) not that they can't figure it out anyway, but aren't we kind of provoking that behavior?

There is no way to tamper with the license file so I don't think this applies

PR #2661, opened 2017-04-21
Update asb-6to7.md

Oculys-DanL:
Added introductory context around Sanitization differences.


weralabaj:
thanks @Oculys-DanL !

Issue #2638, opened 2017-04-14
Explain error when transport is in TransactionScope mode but database does not handle TransactionScopes


SzymonPobiega:
In SQL persistence, if the database does not support TransactionScopes but the transport has been configured to use TransactionScope mode the syncrhonized storage session adapter will throw an exception from the Enlist method.

This behavior needs to be documented.

weralabaj:
@SzymonPobiega any context (e.g. link to the relevant discussion)? or is that only placeholder for yourself?

SzymonPobiega:
Thanks @weralabaj . Yeah I created it as a reminder for myself but forgot the context already. Fortunately I managed to recall it.

PR #2671, opened 2017-04-25
Clarified case when transport is in transaction scope mode but database does not support ambient transactions.

SzymonPobiega:


SzymonPobiega:
Fixes https://github.com/Particular/docs.particular.net/issues/2638

PR #2680, opened 2017-04-26
Host Roles update

seanfarmar:
Rename index_roles_host_[5,6].partial.md to index_roles_host_[5,6,7] so users see this is obselete


SimonCropp:
@seanfarmar i took the liberty of fixing your file name. the convention is actually a "nuget Version range", not a "list of versions"

SimonCropp:
@Particular/docs-maintainers re-looking at this i think this content and snippet should be moved to a 5 to 6 host upgrade guide

HEskandari:
@SimonCropp This is only fixing the snippet though. Should we merge this as is and open another PR to move the content to upgrade docs?

SimonCropp:
@HEskandari u cant merge this since the file rename means u need to also have the snippet for v7, hence the broken build. the problem with the content is it should not be carried forward for all versions. it is only transitional doco

seanfarmar:
@SimonCropp how can we move this forward then? Can you help?

SimonCropp:
@seanfarmar what approach would u like to take?

weralabaj:
@seanfarmar we need to fix the versions, so probably either add a new snippet for v7/8 or just include that range in snippet here https://github.com/Particular/docs.particular.net/blob/54157f788225dcf7dd81869d0c0f88f278650835/Snippets/Host/Host_6/AsAClientEquivalent.cs.

seanfarmar:
I need to roll back the merge I did

weralabaj:
for some bizarre reason changing the file name leads to problems with link validation on unrelated pages, I can't think of any sensible explanation, need some time to let it rest, maybe then I get illuminated 

"The link /nservicebus/hosting/nservicebus-host/#endpoint-configuration-endpoint-name doesn't exist, but was referenced in page \nservicebus\endpoints\specify-endpoint-name.md?version=Core_3
The link /nservicebus/hosting/nservicebus-host/#endpoint-configuration-endpoint-name doesn't exist, but was referenced in page \nservicebus\endpoints\specify-endpoint-name.md?version=Core_4"

PR #2697, opened 2017-05-02
Async handler improvements

### LINKED ISSUE #2693: Feedback: 'Asynchronous Handlers'
Feedback for 'Asynchronous Handlers' https://docs.particular.net/nservicebus/handlers/async-handlers

Location in GitHub: https://github.com/Particular/docs.particular.net/blob/master/nservicebus/handlers/async-handlers.md

I generally like this documentation page very much. It shows that you not only want to document your code, but also help the customer to use it better and understand the reasons behind it. :thumbsup:

But having said this, I have a few issues/questions about that page:

### (A) Handlers in IO thread pool thread
In the "Introduction" section it says

> Handlers and Sagas will be invoked from an IO thread pool thread.

It this really true? Or to be more precise, is it true for all transports? Because it seems like MSMQ calls the handlers in a `Task.Run` which is - as explained just a few paragraphs down - a worker thread pool thread. Relevant code for MSMQ: https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/Msmq/MessagePump.cs#L175

### (B) Multiple thread pools
Are there even two separate thread pools? I am for sure no expert in this topic, but I've now spend some hours reading documentation and found the following:

From a native (Windows) point of view it seems like there is no distinction any more since Windows Vista. From the MSDN article [Thread Pools](https://msdn.microsoft.com/en-us/library/windows/desktop/ms686760.aspx) you linked under "Further reading":

> The original thread pool has been completely rearchitected in Windows Vista. The new thread pool is improved because it provides a single worker thread type (supports both I/O and non-I/O) [...]

From a managed (.NET) point of view the article [The Managed Thread Pool](https://msdn.microsoft.com/en-us/library/0ka9477y.aspx) says:

> There is only one thread pool per process.

Is there some official documentation that mentions these two thread pools?

### (C) ConfigureAwait
I fully understand why you - as a library developer - need to use `ConfigureAwait(false)` everywhere. But according to the section [Usage of ConfigureAwait](https://docs.particular.net/nservicebus/handlers/async-handlers#usage-of-configureawait) we should also specify `ConfigureAwait(false)` in all message handlers. Could you please explain why this is? Since the message handlers are called from a thread pool thread they shouldn't have a `SynchronizationContext` set anyway, right? So what context is there that will be restored after an `await`?

danielmarbach:
Closes #2693

### DIFF
--- a/nservicebus/handlers/async-handlers.md
+++ b/nservicebus/handlers/async-handlers.md
@@ -11,7 +11,7 @@ versions: '[6.0,)'
 
 WARNING: It is difficult to give generic advice on how asynchronous code should be structured. It is important to understand compute-bound vs. IO-bound operations. Avoid copy-pasting snippets without further analysis if they provide benefit for the involved business scenarios. Don't assume. Measure it.
 
-[Handlers](/nservicebus/handlers/) and [Sagas](/nservicebus/sagas/) will be invoked from from a thread pool thread. Depending on the transport implementation it might use a worker thread pool thread or an IO thread pool thread. Typically message handlers and sagas issue IO bound work like sending or publishing messages, storing information into databases, calling web services and more. In other cases, message handlers are used to schedule compute-bound work. To be able to write efficient message handlers and sagas it is crucial to understand the difference between compute-bound and IO bound work and the thread pools involved.
+[Handlers](/nservicebus/handlers/) and [Sagas](/nservicebus/sagas/) will be invoked from a thread pool thread. Depending on the transport implementation a worker thread pool thread or an IO thread pool thread might be used. Typically message handlers and sagas issue IO bound work like sending or publishing messages, storing information into databases, calling web services and more. In other cases, message handlers are used to schedule compute-bound work. To be able to write efficient message handlers and sagas it is crucial to understand the difference between compute-bound and IO bound work.
 
 
 ### Thread Pool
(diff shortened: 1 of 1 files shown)


nservicebus/handlers/async-handlers.md
```diff
@@ -11,21 +11,23 @@ versions: '[6.0,)'
...
 
-[Handlers](/nservicebus/handlers/) and [Sagas](/nservicebus/sagas/) will be invoked from an IO thread pool thread. Typically message handlers and sagas issue IO bound work like sending or publishing messages, storing information into databases, calling web services and more. In other cases, message handlers are used to schedule compute-bound work. To be able to write efficient message handlers and sagas it is crucial to understand the difference between compute-bound and IO bound work and the thread pools involved.
+[Handlers](/nservicebus/handlers/) and [Sagas](/nservicebus/sagas/) will be invoked from from a thread pool thread. Depending on the transport implementation it might use a worker thread pool thread or an IO thread pool thread. Typically message handlers and sagas issue IO bound work like sending or publishing messages, storing information into databases, calling web services and more. In other cases, message handlers are used to schedule compute-bound work. To be able to write efficient message handlers and sagas it is crucial to understand the difference between compute-bound and IO bound work and the thread pools involved.
```
cremor: Duplicate "from" in first sentence.

I'm not sure if the "it" in the second sentence makes sense. (Yes I know that I wrote that sentence originally :smile:) Also, now that the two thread pools were rewritten, maybe this should be updated too. How about "Depending on the transport implementation this thread might be associated with compute bound work or IO-bound work."

danielmarbach:
@cremor addressed your feedback. Nice catch with the from. My brain completely wiped it out. I was reading the sentence multiple times but couldn't see it

PR #2700, opened 2017-05-03
Learning transport for tutorial

DavidBoike:
Connects to https://github.com/Particular/DeveloperEducation/issues/1028, but also determining viability for learning transport in https://github.com/Particular/DeveloperEducation/issues/910.


SimonCropp:
why is a new linqpad script required?

DavidBoike:
It doesn't, but I'm guessing I may have to update those 10 solutions to new prerelease versions multiple times. After that the standard nugets script will work fine. I can rip out the new one before removing WIP.

SimonCropp:
Once u opt in to pre release nugets the existing nuget script will continually update to new pre release versions

weralabaj:
@DavidBoike since learning transport has been announced, can this be merged?

weralabaj:
I've reviewed, LGTM. One question though - don't we plan on covering persistence and error queue in the intro tutorial? Since it was removed from lesson 1, IMO we should still explain this in other lessons.

DavidBoike:
Nope, still WIP - Do not merge. Lesson 5 text has not been updated at all.

And since Lesson 5 is where we have to change to MSMQ transport in order to use the rest of the platform, I think that's where to address error queue, and some other basic principles around "So you want to use a real transport now".

SimonCropp:
so what is blocking this?

DavidBoike:
Other funnel optimization stuff and me being on vacation until Thurs.

DavidBoike:
OK, I think this might be mergeable now. Removed the WIP. There's probably adjustments we'd want to make to CTAs but would rather handle that outside all the noise of this major change.

The intro tutorial is now all learning transport, so @weralabaj to your earlier question no, not addressing persistence or error queue or serializer at all there.

The platform bits got split out into a new replay tutorial, based on MSMQ, so there's a section in there that kind of describes what's different but still doesn't go into a ton of detail. That stuff _really_ needs to be covered in a "graduating from Learning to other transport" in the main body of doco, I think.

PR #2702, opened 2017-05-04
Bump nuget.exe version to work with msbuild 15

SeanFeldman:
On machines that have VS2017 only, linqpad script to update nugets is failing as it's on version 3.5 which doesn't probe for msbuild15. With nuget.exe v4.x it does.


DavidBoike:
Still works on my 2015-only system.

SeanFeldman:
Thanks for confirming @DavidBoike

SeanFeldman:
@Particular/docs-maintainers good to be merged? 😃

PR #2712, opened 2017-05-08
Remove the documentation about custom conversation ids

timbussmann:
A custom conversation id can only be set when a new message flow is started (messages sent from `IMessageSession`). When the header is set in a message handler it would overwrite an existing conversation id and throw instead.

We currently don't offer a proper API to set a custom conversation id and the removed statement is also not backed by any snippet code so I propose to remove it entirely. If we really think this is a case we should support, we should provide an option on `SendOptions` (which can't differentiate between sends from the message session or the handler context)

@andreasohlund @indualagarsamy thoughts?

### DIFF
--- a/nservicebus/messaging/headers.md
+++ b/nservicebus/messaging/headers.md
@@ -84,8 +84,6 @@ Identifier of the conversation that this message is part of. It enables the trac
 
 The first message that is sent in a new flow is automatically assigned a unique `Conversation Id` that is then propagated to all the messages that are subsequently sent, thus forming a _conversation_. Each message that is sent within a conversation also has a `RelatedTo` value that identifies the originating message that caused it to be sent. 
 
-The `Conversation Id` header can be manipulated to create custom conversations which are meaningful from the domain perspective (e.g. using Order Id as the `Conversation Id` to track the life cycle of the order).
-
 
 ### NServiceBus.RelatedTo
 


weralabaj:
As you can see here https://github.com/Particular/docs.particular.net/commit/540e0f2b8b960b429622322da28126a424f1cb46#diff-9a5afa9daa6ec4056a49f2a19187c8c6, this paragraph was added to clarify differences between correlation and conversation ids. 

As noted by [Daniel](https://github.com/Particular/docs.particular.net/pull/1934#issuecomment-249561838) and [Andreas](https://github.com/Particular/docs.particular.net/pull/1934#issuecomment-249604674) there are some improvements that we should probably make, but I don't think removing this section is a solution, it'll just again make problem less visible.

weralabaj:
Would be the solution raising an issue in core repo describing the original problem?

timbussmann:
> this paragraph was added to clarify differences between correlation and conversation ids.

@weralabaj none of that information has been removed? The only section removed is about customizing this value, which should be avoided in the first place. In case we want to consider this a valid scenario, there should be more documentation around that and why you wanna do that instead of just stating that you can manipulate it.

weralabaj:
Correct, my comment was not very precise

> In case we want to consider this a valid scenario

Based on the discussion around the issue I referenced, I believe this is the case. That's what I wanted to say :)

timbussmann:
@weralabaj I raised https://github.com/Particular/NServiceBus/issues/4683 in the core to provide a better API. I still think we should remove that section from docs until the linked issue is addressed.

weralabaj:
OK

PR #2715, opened 2017-05-10
Remove unneeded config from Simple saga sample

andreasohlund:
Please review @SimonCropp

### DIFF
--- a/samples/saga/simple/Core_6/Sample/Program.cs
+++ b/samples/saga/simple/Core_6/Sample/Program.cs
@@ -14,7 +14,9 @@ class Program
     {
         Console.Title = "Samples.SimpleSaga";
         var endpointConfiguration = new EndpointConfiguration("Samples.SimpleSaga");
-        
+
+        endpointConfiguration.UseSerialization<JsonSerializer>();
+
         #region config
 
         endpointConfiguration.UsePersistence<LearningPersistence>();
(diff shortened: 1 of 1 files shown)


samples/saga/simple/Core_6/Sample/Program.cs
```diff
@@ -16,23 +14,14 @@ static void Main()
     {
         Console.Title = "Samples.SimpleSaga";
         var endpointConfiguration = new EndpointConfiguration("Samples.SimpleSaga");
-        endpointConfiguration.UseSerialization<JsonSerializer>();
```
SimonCropp: this is in keeping with https://docs.particular.net/samples/#technology-choices-serializer

samples/saga/simple/Core_6/Sample/Program.cs
```diff
@@ -16,23 +14,14 @@ static void Main()
...
-        endpointConfiguration.EnableInstallers();
-
-        var recoverability = endpointConfiguration.Recoverability();
-        recoverability.DisableLegacyRetriesSatellite();
```
SimonCropp: without this the learning transport creates a redundant `Retries` directory structure

samples/saga/simple/Core_6/Sample/Program.cs
andreasohlund: Ok I see, will add it back

samples/saga/simple/Core_6/Sample/Program.cs
andreasohlund: I'll pr the core to avoid this for the learning transport so that we don't need this crap here

PR #2718, opened 2017-05-10
improve note on unlisted version support

adamralph:


adamralph:
@Particular/docs-maintainers any objections to merging?

PR #2725, opened 2017-05-15
Clarify UseSchemaFor{Queue, Endpoint} doco

MarcinHoppe:


MarcinHoppe:
@Particular/sqlserver-transport-maintainers @Particular/docs-maintainers Please review, it's a quick change to make it easier for customers to know when they have to use one API vs the other.

Fixes Particular/NServiceBus.SqlServer#350.

PR #2726, opened 2017-05-15
Autofac singleton bug

ramonsmits:
Customer feedback from regarding single instance. Updated V5 sample to be similar to V6 one.


weralabaj:
@ramonsmits what's the bug?

weralabaj:
ping @ramonsmits

ramonsmits:
@weralabaj 

1. Endpoint instance not registered as single instance
2. V5 sample is different in usage compared to V6
3. Container isn't disposed gracefully

Issue #2734, opened 2017-05-16
Improve documentation for ServiceControl queue configuration with remote MSMQ


cremor:
I recently configured a project to use ServiceControl for error, audit and heartbeats. It was well documented how to configure everything for local queues, so I had no problem implementing it on my development machine. Thanks so far 👍 

But then I tried to deploy it on our testing environment. Since all of the **code** configuration samples for [errors](https://docs.particular.net/nservicebus/recoverability/configure-error-handling#configure-the-error-queue-address-using-code), [audit](https://docs.particular.net/nservicebus/operations/auditing#configuring-auditing-using-code) and [heartbeats](https://docs.particular.net/servicecontrol/plugins/heartbeat#configuration-servicecontrol-queue) only showed a queue name, I thought that I could just use the normal [MSMQ physical routing](https://docs.particular.net/nservicebus/msmq/routing) to define the machine name.

As this didn't work as expected, I then searched the documentation some more. The **app.config** configuration samples on the previously linked pages for errors and heartbeats show a queue name in the format "queue@machine" so I tried this using code configuration and it indeed worked. And even though the audit documentation for app.config configuration doesn't show this format, it worked there too.

I think this could be greatly improved.

weralabaj:
@cremor thanks for your feedback!

@Particular/servicecontrol-maintainers can you have a look?

WilliamBZA:
Thanks @cremor.

The Heartbeat documentation already includes the remote queue format but there's definitely a need to be more explicit in the audit and errors queue name sections.

weralabaj:
I believe the concern was addressed in https://github.com/Particular/docs.particular.net/pull/2745, in case more changes are needed please reopen :)

PR #2745, opened 2017-05-17
Add machine name to queue name examples

WilliamBZA:
Connects to https://github.com/Particular/docs.particular.net/issues/2734

This should make it clearer for users that when configuring these features for use with ServiceControl that the queues should be remote queues instead of using the Physical Routing file.

I'm a little unsure of how this looks if the users are using different transports however. We may need a better way to represent this instead of simply assuming MSMQ in documentation.

Ping @Particular/servicecontrol-maintainers please review. I suppose @Particular/nservicebus-maintainers should also review since it touches some of their pages?


bording:
@WilliamBZA I'm 👎 to this change because this is an MSMQ-specific thing that isn't generally true for other transports.

WilliamBZA:
> MSMQ-specific thing that isn't generally true for other transports.

I agree.

Any suggestions on how to make it generic while still conveying the distinction we've created for MSMQ between remote queues and the physical routing file?

SimonCropp:
given historically we actually went through and removed all the msmq assumptions from queue names i would prefer not to put them back :)

WilliamBZA:
Once again: I agree.

Any suggestions on how/where to put it then? We have an arbitrary "use physical routing file here, but not here" issue and we need to make it clear to users when to use each.

weralabaj:
Instead of replacing the existing snippets could we add extra lines to them for MSMQ? Or add separate snippets?

An alternative could be "Configuring MSMQ production environment" or sth like that, and link to it.


@SimonCropp @bording

WilliamBZA:
I was thinking about it some more today, and I think your suggestion is a good one @weralabaj.

This config is mostly specific to configuring ServiceControl, so a dedicated heading in the configuration section of the ServiceControl docs might be a good fit. I'll change the PR.

WilliamBZA:
Still busy on this, just haven't found the time yet.

WilliamBZA:
@Particular/docs-maintainers I'm not sure if I'm doing the snippets and partials correctly here, can somebody please verify that I'm on the correct track?

I'll add heartbeats and auditing next if the error queue config looks correct.

weralabaj:
@WilliamBZA seems like the underlying dependency here is Core, so if I understand correctly you should place those snippets here https://github.com/Particular/docs.particular.net/tree/master/Snippets/Core. Also the component in the doco page should be core. There's nothing depending on SC version here, correct?

If you'd like to sue SC versions then that would be similar to what was done for Heartbeats here https://github.com/Particular/docs.particular.net/tree/master/Snippets/Heartbeats.

SzymonPobiega:
Partials look ok. @weralabaj would snippets in this form work correctly?

WilliamBZA:
@weralabaj so I should move the MSMQ specific snippets into the core sln?

weralabaj:
sorry, I forgot the context of that issue, \docs.particular.net\Snippets\MsmqTransport probably would be even better.

SimonCropp:
i dont think u can use msmq snippets in an SC page

weralabaj:
at the moment there's no component specified, but once we move msmq out of the core it won't work otherwise, will it?

WilliamBZA:
So what's the consensus on where the snippets should live? I'm confused.

weralabaj:
@WilliamBZA I propose this - ensure you have all the content you want, and I'll sync with Simon regarding where it should live and move it accordingly, OK?

weralabaj:
Just let me know if that's ready :D

Snippets/ServiceControl/ServiceControl_All/Core_[,6.0]/ErrorQueueMsmqConfigurationProvider.cs
```diff
@@ -0,0 +1,22 @@
+﻿#pragma warning disable 618
```
Comment (weralabaj) [inline Snippets/ServiceControl/ServiceControl_All/Core_[,6.0]/ErrorQueueMsmqConfigurationProvider.cs]: should that be v6 inclusive or exclusive? hard to figure out which versions it applies to
also, is anything below v5 supported by servicecontrol?

@WilliamBZA ☝️

Snippets/ServiceControl/ServiceControl_All/Core[6.0,)/ConfigErrorQueue.cs
```diff
@@ -0,0 +1,16 @@
...
+
+    class ConfigErrorQueue
+    {
+        Usage(EndpointConfiguration endpointConfiguration)
+        {
+            #region ConfigMsmqErrorWithCode
+
+            endpointConfiguration.SendFailedMessagesTo("targetErrorQueue@machinename");
+
+            #ConfigMsmqErrorWithCode
```
SimonCropp: this wont compile needs to be an end region

Snippets/ServiceControl/ServiceControl_All/Core[6.0,)/ConfigErrorQueue.cs
```diff
@@ -0,0 +1,16 @@
+﻿namespace ServiceControl.ConfigureEndpoints
+{
+    using NServiceBus;
+
+    class ConfigErrorQueue
+    {
+        Usage(EndpointConfiguration endpointConfiguration)
```
SimonCropp: this wont compile as it is missing a return type or needs to be a constructor

servicecontrol/configuring-queues.md
```diff
@@ -0,0 +1,36 @@
...
+ - nservicebus/recoverability/configure-error-handling
+---
+
+ServiceControl reads messages from the input queue specified in the [ServiceControl instance transport configuration](creating-config-file#transport). The queues specified during this stage are populated with messages from each of the production endpoints after which, ServiceControl ingests those messages in order to extract the required information from them. Depending on the transport selected and the deployment scenario, the production endpoints might need to be configured to send the messages to a remote queue.
+
+Of the currently supported transports, only MSMQ requires special configuration to deal with remote queues.
+
+### MSMQ
+
+If the endpoints are on a separate machine to the ServiceControl instance, the endpoints must be configured to send their error and audit messages (if auditing is enabled) to a remote queue as the MSMQ instance mapping file is not used in the case of error and audit queues
```
SimonCropp: all this content and snippets needs to be moved to the MSMQ doco and linked to from here

SimonCropp: if u are talking about audit as well you need to have snippets for audit

SimonCropp:
does anyone want to resurrect this?

weralabaj:
I dropped the ball, will look into this today.

WilliamBZA:
This is still blocked by me. I'm happy to close this PR completely until I can grab it again.

weralabaj:
@WilliamBZA I can help with splitting, I think nothing related to msmq content should surprise me now after splitting some docs ;) Generally the idea was to move the MSMQ-specific content to another page, that will be component: MSMQTransport and there have all the snippets.

That will address the issue of versioning. Mike suggested we should cover only supported versions, i.e. v5 and above.

I think that should be enough for me to progress or is there anything you'd like to add/modify here first?

WilliamBZA:
That'd be great @weralabaj, thanks for jumping on the grenade here!

weralabaj:
@WilliamBZA oI'm not sure if my understanding is correct, so if MSMQ endpoints are on a different machine than SC then you need to configure your endpoints to send errors and audits to the queue on the same machine that SC is running on?

Can you see if snippets make sense?

weralabaj:
@SimonCropp any idea on what I'm doing wrong? that snippet is there for msmq v1

WilliamBZA:
@weralabaj Yes, that's exactly it. The instance mapping file isn't used for error forwarding, so we have to have different ways of configuring queus depending on whether it's a regular "business" message, or if it's a "servicecontrol" message.

Snippets look correct so far.

WilliamBZA:
Thanks for the help @SimonCropp and @weralabaj. I think this can be merged now. I'll raise a separate issue to add the snippets for audits so we don't lose that.

PR #2753, opened 2017-05-18
Update to the cloud show case

SeanFeldman:
**Changes**
- Replacing MEM with Routing API
- Replace hard-coded connection string with an argument to support real world usage
- Connect webjob projects to the webapp to be able to deploy all together
- The sample can be now deployed to Azure using updated deployment notes

With Core 6.3 MEM is obsoleted with warning.
Removing red scary messages from the show case by replacing MEM with Routing API.

Hard-coded connection string is not a good starting point for anyone who'd like to base their tryout (azure deployment) on this sample. Switched to a parameter which can be passed into `ApplyCommonConfiguration()` extension method.

Before the change, deploying webapp would not deploy the webjobs. Now with a publishing settings file, webdeploy pushes webapp and webjobs to Azure hosting plan.

### DIFF
--- a/samples/show-case/cloud-azure/Core_6/Store.ContentManagement/App.config
+++ b/samples/show-case/cloud-azure/Core_6/Store.ContentManagement/App.config
@@ -1,14 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <configuration>
-  <configSections>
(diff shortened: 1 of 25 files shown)


SimonCropp:
@SeanFeldman there were some duplicate imports i removed

SeanFeldman:
Thank you @SimonCropp
If no red flags, then we can push it.

>

SimonCropp:
ok with me

PR #2754, opened 2017-05-19
Add second endpoint as well

andreasohlund:
@SimonCropp I assume both should start?


SimonCropp:
Nope. We want endpoint 1 to start, but messages not be consumed so people can look at them on disk. Read the  .md

PR #2756, opened 2017-05-19
Transports: Queue Creation split

Scooletz:
Queue creation content split to two partials: declaration and creation that are properly versioned.


Scooletz:

PR #2761, opened 2017-05-20
Removing message endpoint mappings from on-premises showcase

SeanFeldman:
Have smoke tested the sample.


SeanFeldman:
@Particular/docs-maintainers tested this and it's working, so went ahead and merged.

PR #2766, opened 2017-05-23
Update Composition.cs

ktheeten:
In the text, you say there should be no trailing slash in the returned path.  Either the text is wrong, or the code.


weralabaj:
thanks @ktheeten for catching it!

@Particular/azure-maintainers can you please clarify which way it is?

SeanFeldman:
👍

PR #2767, opened 2017-05-23
Update Composition.cs

ktheeten:
After testing it out, it seems that the trailing slash IS needed.  So the text is wrong under the code snippet.  Don't know how to edit that though.


weralabaj:
Seems that this page needs to be edited https://github.com/Particular/docs.particular.net/blob/cf00531a7fb161aa0711f88e205ad57db71ae936/nservicebus/azure-service-bus/namespace-hierarchy.md?

weralabaj:
@Particular/azure-maintainers ping :)

SeanFeldman:
I'll update both. Thank you @ktheeten and @weralabaj

PR #2769, opened 2017-05-23
[WIP] Include transport tests in custom transport sample

timbussmann:
As part of https://github.com/Particular/PlatformDevelopment/issues/1256 we were thinking how we could help community packages to use the acceptance/transport test packages. As we only seem to have a sample about how to write a custom transport, should we also consider including at least the transport tests in this sample to show how to include and run these tests on custom transports?

thoughts? I would of course also add a section to the doco page of the sample describing the approach to install & configure the tests if you like this idea. ping @Particular/docs-maintainers


SzymonPobiega:
I like the idea. Can't imagine using a transport that does not pass the acceptance tests.

weralabaj:
I think this is a great idea, do you wnat to add the section in this pr or open a new one and merge this one now?

timbussmann:
@weralabaj I'll update this PR. Marking this as WIP again then.

SimonCropp:
would it be worth converting this to a 2017 csproj format to avoid the need to check in all the acceptance test?

timbussmann:
Sounds like a good idea to me

Simon Cropp <notifications@github.com> schrieb am Fr., 2. Juni 2017, 13:22:

SimonCropp:
merged manually here https://github.com/Particular/docs.particular.net/commit/6f4c6c73f4b10a655883c94760895b0135cc3eec

PR #2776, opened 2017-05-24
Role-based security for ServicePulse

SzymonPobiega:


SzymonPobiega:
@Particular/servicepulse-maintainers please review @Particular/servicecontrol-maintainers please review.

servicepulse/install-servicepulse-in-iis.md
```diff
@@ -100,6 +100,17 @@ WARNING: The default configuration for ServiceControl only allows access to REST
 
 It is also recommended that the IIS web site be configured to use SSL if an authorization provider is used.
 
+### Role-based security
+
+After executing the steps outlined above, ServicePulse requires authentication before accessing any functionality. It does not check any authorization rules though so each authenticated user can do everything. In order to restrict access to specific features, the `IIS URL Authorization` feature can be used. Following snippet can be placed in the `web.config` file in the root of the web site to restrict access based on roles.
```
WilliamBZA: The steps above only recommend that authentication be turned on, unless we make it an explicit step I don't think the `requires authentication before accessing any functionality` part is true (unless reverse proxy by default enforces authentication on?)

Perhaps:

> If IIS Authentication is configured, all authenticated users will have full access to all API calls. In order to restrict access to specific...

WilliamBZA: Nevermind, I see step 12 now - missed that before. I'm not sure if that is strong enough, but it's not part of this review so feel free to ignore my comments aboe 😄

PR #2781, opened 2017-05-26
note on cases sensitivity of endpoint names

tmasternak:
ping @timbussmann

### DIFF
--- a/nservicebus/msmq/routing_content_core_[6,).partial.md
+++ b/nservicebus/msmq/routing_content_core_[6,).partial.md
@@ -11,6 +11,8 @@ NOTE: When running under ASP.NET the `instance-mapping.xml` file may not be loca
 
 snippet: InstanceMappingFile
 
+NOTE: The endpoint names are case sensitive e.g. if endpoint name is used in routing API and the instance mapping file, the file entries will be used only when there is a case sensitive match. 
+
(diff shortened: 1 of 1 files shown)


SimonCropp:
isnt this a bug?

timbussmann:
> isnt this a bug?

I asked this myself too, but as this heavily depends on the transports (e.g. are queue names case-sensitive?) we would have to explicitly implement core & transport logic to be case-insensitive. I think I would classify this as a feature?

nservicebus/msmq/routing_content_core_[6,).partial.md
```diff
@@ -11,6 +11,8 @@ NOTE: When running under ASP.NET the `instance-mapping.xml` file may not be loca
 
 snippet: InstanceMappingFile
 
+NOTE: The endpoint names are case sensitive e.g. if endpoint name is used in routing API and the instance mapping file, the file entries will be used only when there is a case sensitive match.
```
timbussmann: should this be a WARN instead?

timbussmann: also, do we need the whole example part?

tmasternak: I would lean to keeping the example part. This doco [page](https://docs.particular.net/nservicebus/msmq/routing) shows usage of endpoint names only in xml file (which uses each name once) so it might be confusing for readers what is matched with what.

timbussmann: I don't have strong feelings about this. Maybe rephrase to something like `The endpoint names are case sensitive. Endpoint names used in the routing name and the instance mapping file require an exact string match` to avoid the `e.g.` ?

PR #2783, opened 2017-05-26
Azure samples without message endpoint mappings

SeanFeldman:
Removing message endpoint mappings from azure samples


SeanFeldman:

PR #2786, opened 2017-05-28
Removed sentence from opening

udidahan:
Removed this:

> In order to do so, the adapter provides a ServiceControl interface to the business endpoints. 

It isn't clear to me what this means or how this would help me, as a user, understand the sample. Could be just me, though. I'm not as technical as I used to be.


weralabaj:
rebased

PR #2792, opened 2017-05-29
Update disabling legacy retries partial page

timbussmann:
relates to https://github.com/Particular/NServiceBus/pull/4745

the described queue is no longer available in v7, therefore that part can be removed from the v7 documentation.

I noticed there are 2 partials for v6: one using the `[6]` range, the other  using `[6.1, 7]`. The first one doesn't seem to show up as far as I could see, so I removed it. I also adjusted the range of the second one to exclude v7 instead of including v7. @Particular/docs-maintainers


SzymonPobiega:
@timbussmann there is something wrong with one of the snippets: `Some redundant snippets were detected. disablelegacyretriessatellite (>= 7.0.0-pre && < 8.0.0)`

timbussmann:
ah, seems like there is a v7 snippet which no longer is required. I'll fix this.

PR #2798, opened 2017-05-31
Update snippets to external msmq transport

andreasohlund:


weralabaj:
@andreasohlund I think you also need to list the new msmq package here https://github.com/Particular/docs.particular.net/blob/master/components/components.yaml#L389

andreasohlund:
Added

Snippets/Core/Core_7/Headers/Writers/HeaderWriterDefer.cs
```diff
@@ -33,7 +33,7 @@ public void Setup()
             endpointConfiguration.EnableInstallers();
             endpointConfiguration.UsePersistence<InMemoryPersistence>();
             endpointConfiguration.RegisterMessageMutator(new Mutator());
-            var routing = endpointConfiguration.UseTransport<MsmqTransport>().Routing();
+            var routing = endpointConfiguration.UseTransport<LearningTransport>().Routing();
```
SimonCropp: did u run these unit tests?

andreasohlund: no, was assuming the build server would do it, running them now

andreasohlund: do we need to run that on msmq?

andreasohlund: they run fine locally

weralabaj:
This is in small part a fix, in large part just a nasty workaround to get master working. Issues found:
- partials - when working with changes in packages we need to organize them differently, e.g. have core_6 partial and then msmqtransport_1 partial, etc.  next time we need to see if pages where we add new component have any partials and adjust right away. I needed to add a few partials which were basically copies of version 6. Not very happy with that redundancy :/
- seems like versions ranges in headers don't play well with snippets :/ tomorrow I'll investigate how to fix it, for now I just removed problematic snippets/versions, they need to be brought back. most likely we'll need to reorganize some content and possibly some changes in docs engine may be needed as well.

PR #2807, opened 2017-06-01
Warning about MSSQL with mirroring & AlwaysOn configurations

DavidBoike:
@SimonCropp we discussed this last week. You think these are the right locations, and good enough wording to be used in all 3 locations?


SimonCropp:
nice

PR #2815, opened 2017-06-05
SQL Server native delayed delivery

weralabaj:
relates to https://github.com/Particular/PlatformDevelopment/issues/1306


weralabaj:
@SzymonPobiega I think with snippets we need to release to myget?

SzymonPobiega:
I removed the upgrade bits so that it can be merged and then the upgrade PR can be merged after this one (#2937)

SzymonPobiega:
@Particular/docs-maintainers @MarcinHoppe ready for review.

SzymonPobiega:
ping.

transports/sql/native-delayed-delivery.md
```diff
@@ -0,0 +1,49 @@
...
+versions: '[3.1,)'
+---
+
+In Versions 3.1 and above, the SQL Server transport can take advantage of native [delayed delivery](/nservicebus/messaging/delayed-delivery.md) without then need to run the [timeout manager](/nservicebus/messaging/timeout-manager.md). Instead, the transport creates infrastructure which can delay messages using native SQL Server transport features. To enable the native delayed delivery use the following API:
+
+snippet: EnableNativeDelayedDelivery
+
+NOTE: In this mode the timeout manager will still be running in order to process all outstanding delayed messages. Refer to the [Disabling the timeout manager](/transports/sql/native-delayed-delivery.md#disabling-the-timeout-manager) section for details on how to disable the timeout manager entirely.
+
+## How it works
```
weralabaj: Probably better as "configuration" header? I'd also link from here to the upgrade guide, but that can be done in the other PR (let's just make a note to not forget).

weralabaj:
@SzymonPobiega One comment ;)

transports/sql/native-delayed-delivery.md
```diff
@@ -0,0 +1,49 @@
...
+
+In Versions 3.1 and above, the SQL Server transport can take advantage of native [delayed delivery](/nservicebus/messaging/delayed-delivery.md) without then need to run the [timeout manager](/nservicebus/messaging/timeout-manager.md). Instead, the transport creates infrastructure which can delay messages using native SQL Server transport features. To enable the native delayed delivery use the following API:
+
+snippet: EnableNativeDelayedDelivery
+
+NOTE: In this mode the timeout manager will still be running in order to process all outstanding delayed messages. Refer to the [Disabling the timeout manager](/transports/sql/native-delayed-delivery.md#backwards-compatibility-disabling-the-timeout-manager) section for details on how to disable the timeout manager entirely.
+
+## How it works
+
+The transport creates an additional queue that stores delayed messages. The table name has the format `endpoint-name.suffix`, using the suffix specified in the configuration:
```
MarcinHoppe: When reading this, I am not sure if these are mandatory settings. I think we should be explicit about what is required and which settings have a default value.

transports/sql/native-delayed-delivery.md
```diff
@@ -0,0 +1,49 @@
...
+
+snippet: DelayedDeliveryTableSuffix
+
+The timeouts poller checks for expired messages at specified interval:
```
MarcinHoppe: > The timeouts poller checks for expired messages

would be better as

> The transport polls for expired messages...

transports/sql/native-delayed-delivery.md
```diff
@@ -0,0 +1,49 @@
...
+
+snippet: DelayedDeliveryProcessingInterval
+
+The poller picks and dispatches messages in batches of specified size:
```
MarcinHoppe: Ditto. The poller is an implementation detail and it does not have to exist in our user's minds.

transports/sql/native-delayed-delivery.md
```diff
@@ -0,0 +1,49 @@
...
+
+snippet: DelayedDeliveryBatchSize
+
+When the delay time lapses, SQL Server transport moves a batch of messages to the input queue. Note that this means the exact time of delivering delayed message is always approximate.
```
MarcinHoppe: input queue -> destination queue?

transports/sql/native-delayed-delivery.md
```diff
@@ -0,0 +1,49 @@
...
+
+
+## Backwards compatibility
+
+When upgrading to a version of the transport that supports native delayed delivery, it is safe to operate a combination of endpoints using native delayed delivery and endpoints using persistence-based delayed delivery at the same time:
```
MarcinHoppe: persistence-based delayed delivery -> timeout manager?

transports/sql/native-delayed-delivery.md
weralabaj: Given that they have to configure it, I wouldn't say it's just an implementation detail. Also, I'm not sure how we could explain how it works without mentioning the poller.

transports/sql/native-delayed-delivery.md
weralabaj: Seems all are now required, correct @SzymonPobiega ?

transports/sql/native-delayed-delivery.md
MarcinHoppe: Then the poller should be introduced earlier on and an explanation should be given about what it is and how it works. Also, for non-optional params we should provide guidance on how to set them.

How do I know what the right interval and batch size is?

weralabaj: > Then the poller should be introduced earlier on and an explanation should be given about what it is and how it works.

We don't do that in any other place where we do polling, AFAIK. If we say instead "SQL Transport polls messages at intervals specified by..." then would that be acceptable?

> for non-optional params we should provide guidance on how to set them.

OK. Also came to my mind but all I could think about is giving an example, as we don't have any specific "perf tuning" techniques here I think.

MarcinHoppe: I [think that would be better](https://github.com/Particular/docs.particular.net/pull/2815#discussion_r126355399).

weralabaj: OK, misunderstood your comment ;)

MarcinHoppe: If we give an example, then I think we should explain the thinking that went into choosing these values, preferably connecting to some business-level requirements around how accurate we need to be with the timeouts and the throughput maybe.

weralabaj:
@MarcinHoppe please, see now :)

transports/sql/native-delayed-delivery.md
```diff
@@ -0,0 +1,52 @@
...
+In Versions 3.1 and above, the SQL Server transport can take advantage of native [delayed delivery](/nservicebus/messaging/delayed-delivery.md) without then need to run the [timeout manager](/nservicebus/messaging/timeout-manager.md). Instead, the transport creates infrastructure which can delay messages using native SQL Server transport features. To enable the native delayed delivery use the following API:
+
+snippet: EnableNativeDelayedDelivery
+
+NOTE: In this mode the timeout manager will still be running in order to process all outstanding delayed messages. Refer to the [Disabling the timeout manager](/transports/sql/native-delayed-delivery.md#backwards-compatibility-disabling-the-timeout-manager) section for details on how to disable the timeout manager entirely.
+
+
+## Configuration
+
+The settings described in this section are obligatory. By default use the values presented in snippets. The settings can be fine-tuned to fit a particular system's characteristic, e.g. in case the expired timeout messages need to be picked up more frequently to result in a more precise approximation of expiry time.
```
MarcinHoppe: If we recommend:

> By default use the values presented in snippets.

Then why they are not default in the code?

transports/sql/native-delayed-delivery.md
```diff
@@ -0,0 +1,52 @@
...
+
+The transport creates an additional queue that stores delayed messages. The table name has the format _`endpoint-name.suffix`_, using the suffix specified in the configuration:
```
MarcinHoppe: Technically speaking it is not a queue table. Maybe let's just state that it creates an additional table?

transports/sql/native-delayed-delivery.md
SzymonPobiega: I think they are code-defaults, @weralabaj ? I don't think we require any of these.

transports/sql/native-delayed-delivery.md
```diff
@@ -0,0 +1,52 @@
...
+
+snippet: DelayedDeliveryBatchSize
+
+When the delay time lapses, SQL Server transport moves a batch of messages to the destination queue, i.e. the endpoint's input queue. Note that this means the exact time of delivering delayed message is always approximate.
+
+
+## Backwards compatibility
+
+When upgrading to a version of the transport that supports native delayed delivery, it is safe to operate a combination of endpoints using native delayed delivery and endpoints using timeout manager at the same time:
+- Endpoints with native delayed delivery can send delayed messages to endpoints using persistence-based delayed delivery.
```
MarcinHoppe: persistence-based delayed delivery -> timeout manager.

transports/sql/native-delayed-delivery.md
```diff
@@ -0,0 +1,52 @@
...
+snippet: DelayedDeliveryBatchSize
+
+When the delay time lapses, SQL Server transport moves a batch of messages to the destination queue, i.e. the endpoint's input queue. Note that this means the exact time of delivering delayed message is always approximate.
+
+
+## Backwards compatibility
+
+When upgrading to a version of the transport that supports native delayed delivery, it is safe to operate a combination of endpoints using native delayed delivery and endpoints using timeout manager at the same time:
+- Endpoints with native delayed delivery can send delayed messages to endpoints using persistence-based delayed delivery. 
+- Endpoints with native delayed delivery can continue to receive delayed messages from endpoints using persistence-based delayed delivery.
```
MarcinHoppe: persistence-based delayed delivery -> timeout manager.

MarcinHoppe:
@weralabaj Thanks, looks much better now. Last batch of (minor) comments.

transports/sql/native-delayed-delivery.md
weralabaj: My mistake, MatureBatchSize and Suffix have defaults, I'll update.
@SzymonPobiega I may be wrong but I don't see where the default for interval is specified, although in the comment we say "Configures how often delayed messages are processed (every 5 seconds by default)."

weralabaj:
The only outstanding issue is to clarify if interval setting should have a default value (and does have it). Comment says one thing, but I can't find the relevant code ;)

SzymonPobiega:
Well, the default interval was `default(TimeSpan)` :joy:. That actually explains why my machine was hanging where there was a problem with DB. It kept running the query in a tight loop.

I've changed it [here](https://github.com/Particular/NServiceBus.SqlServer/commit/be9f5fc7eb3fbef4fb126b8a0365c4803f155c2c) to 1 second to be aligned with Timeout Manager. I've also updated the snippets.

weralabaj:
@MarcinHoppe should be ok now I think :)

PR #2818, opened 2017-06-06
[SPIKE] Parallel sample and snippet build

danielmarbach:


danielmarbach:
Not worth pursuing since it did not lead to speed improvements

Issue #2830, opened 2017-06-09
Feedback: 'Using NServiceBus with ASP.NET MVC'


ramonsmits:
Feedback for 'Using NServiceBus with ASP.NET MVC' https://docs.particular.net/samples/web/asp-mvc-application/

Location in GitHub: https://github.com/Particular/docs.particular.net/blob/master/samples/web/asp-mvc-application/sample.md

ping @Particular/nservicebus-maintainers @Particular/container-maintainers 

The V6 init code is broken. The V5 code is correct, the container is shared between both MVC and NServiceBus but the V6 doesn't achieve the same. In the V6 sample, both MVC and NServiceBus have their own container instance meaning singletons cannot be shared between for example controllers and handlers.

- DependencyResolver.SetResolver is invoked more than once
- RegisterControllers is invoked more than once

I have not been able to resolve these issues as it seems as a chicken/egg order issue as V6 does not register the IEndpointInstance in the passed container.

timbussmann:
@ramonsmits raised a proposal here: https://github.com/Particular/docs.particular.net/pull/2834

ramonsmits:
Closing this as the PR will fix this.

PR #2834, opened 2017-06-09
Change autofac configuration

timbussmann:
fixes https://github.com/Particular/docs.particular.net/issues/2830
@ramonsmits can you have a look at this approach? I think this should solve the issues you mentioned (e.g. allows to share a singleton between message handler and MVC controllers)

(if we want to apply this change, we should probably also adjust the v7 sample)

### DIFF
--- a/samples/web/asp-mvc-application/Core_6/AsyncPagesMVC/Global.asax.cs
+++ b/samples/web/asp-mvc-application/Core_6/AsyncPagesMVC/Global.asax.cs
@@ -39,10 +39,12 @@ public class MvcApplication :
 
         // Register the MVC controllers.
         builder.RegisterControllers(typeof(MvcApplication).Assembly);
-
-        // Set the dependency resolver to be Autofac.
+        // Register the endpoint as a factory as the instance isn't created yet.
+        builder.Register(ctx => endpoint);
+        
         var container = builder.Build();
 
+        // Set the dependency resolver to be Autofac.
         DependencyResolver.SetResolver(new AutofacDependencyResolver(container));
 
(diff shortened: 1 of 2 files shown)


samples/web/asp-mvc-application/Core_6/AsyncPagesMVC/Global.asax.cs
```diff
@@ -39,10 +39,12 @@ protected void Application_Start()
 
         // Register the MVC controllers.
         builder.RegisterControllers(typeof(MvcApplication).Assembly);
-
-        // Set the dependency resolver to be Autofac.
+        // Register the endpoint as a factory as the instance isn't created yet.
+        builder.Register(ctx => endpoint);
```
ramonsmits: Is this then resolved as a singleton/singleinstance? I'm not familiar enough with Autofac. @Particular/container-maintainers

ramonsmits: /cc @timbussmann

timbussmann: no this would be a transient scope afaik. Maybe this should be changed to `.SingleInstance()` in order to avoid potential issues when disposing child-scopes (or tag it as externally owned).

WilliamBZA: Correct, `SingleInstance` is recommended here

timbussmann:
changed registration to a singleton and fixed the v7 sample as well. I think this is good to go @Particular/docs-maintainers

weralabaj:
rebased

Issue #2723, opened 2017-05-12
Feedback: 'Heartbeat Plugin'


cremor:
Feedback for 'Heartbeat Plugin' https://docs.particular.net/servicecontrol/plugins/heartbeat

Location in GitHub: https://github.com/Particular/docs.particular.net/blob/master/servicecontrol/plugins/heartbeat.md

ping @Particular/servicecontrol-maintainers 

Is the section "[Expired heartbeat messages forwarded to Dead letter queue](https://docs.particular.net/servicecontrol/plugins/heartbeat#expired-heartbeat-messages-forwarded-to-dead-letter-queue)" still relevant? Because according to [Upgrade Version 6 to 6.2](https://docs.particular.net/nservicebus/upgrades/6to6.2) this was changed in 6.2

indualagarsamy:
Hi @cremor - you are right. The documentation needs to be updated to add the new behavior for TTBR messages.

SzymonPobiega:
@Particular/msmq-transport-maintainers could you please take a look at this? I don't have enough context to be able to update the doco.

indualagarsamy:
@SzymonPobiega - I'll have a go and create a PR for this.

indualagarsamy:
@cremor - once again, thanks for your feedback on this. Sorry, it took a little while longer than expected to get to this. Hope the new clarifications help. Cheers.

PR #2842, opened 2017-06-13
Added the changed TTBR behavior based on recent version.

indualagarsamy:
Fixes https://github.com/Particular/docs.particular.net/issues/2723

@Particular/msmq-transport-maintainers - can you please review and merge?

### DIFF
--- a/servicecontrol/plugins/heartbeat.md
+++ b/servicecontrol/plugins/heartbeat.md
@@ -82,7 +82,7 @@ Heartbeat messages have a time to be received (TTBR) set based on the TTL value.
 
 ### MSMQ
 
-Although NServiceBus configures the use of DLQ by default, messages that are defined with TTBR will not be automatically forwarded to the DLQ and will be discarded.
+Although NServiceBus configures the use of DLQ by default, messages that are defined with TTBR will not be automatically forwarded to the DLQ and will be discarded. Configuration can be specified to [override this behavior](/nservicebus/msmq/dead-letter-queues.md#enabling-dlq-for-messages-with-ttbr) so that these messages can be forwarded to the DLQ.
 
 WARN: When using NServiceBus Versions 6.1 or below, messages will be forwarded to the DLQ even if TTBR is set on the messages.  To avoid this behavior, DLQ can be disabled by configuring the [MSMQ connection strings](/nservicebus/msmq/connection-strings.md).  The heartbeat messages will be forwarded to the DLQ when ServiceControl is either stopped or very busy. In this case, the dead letter queue needs to be monitored and cleaned up.
 
--- a/servicecontrol/plugins/heartbeat.md
+++ b/servicecontrol/plugins/heartbeat.md
@@ -77,9 +77,9 @@ partial: disable
 
 ## Expired heartbeat messages forwarded to Dead letter queue
 
-Heartbeat messages have a time to be received (TTBR) set based on the TTL value. If ServiceControl does not consume the heartbeat messages before the TTBR expires then transports that support a dead letter queue (DLQ) will move these messages to the DLQ. This can happen when ServiceControl is stopped or very busy. The dead letter queue needs to be monitored and cleaned up.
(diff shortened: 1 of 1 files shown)


servicecontrol/plugins/heartbeat.md
```diff
@@ -77,9 +77,9 @@ partial: disable
 
 ## Expired heartbeat messages forwarded to Dead letter queue
 
-Heartbeat messages have a time to be received (TTBR) set based on the TTL value. If ServiceControl does not consume the heartbeat messages before the TTBR expires then transports that support a dead letter queue (DLQ) will move these messages to the DLQ. This can happen when ServiceControl is stopped or very busy. The dead letter queue needs to be monitored and cleaned up.
+Heartbeat messages have a time to be received (TTBR) set based on the TTL value.  If ServiceControl does not consume the heartbeat messages before the TTBR expires then transports that support a dead letter queue (DLQ) will discard these messages. However if the endpoint that is sending the heartbeat messages is using NServiceBus Versions 6.1 or below, these messages will be moved to the DLQ. This can happen when ServiceControl is stopped or very busy. In this case, the dead letter queue needs to be monitored and cleaned up.
```
andreasohlund: >If ServiceControl does not consume the heartbeat messages before the TTBR expires then transports that support a dead letter queue (DLQ) will discard these messages.

This doesn't sound right? 

All transports that support TTBR will discard. The ones that have a DLQ feature "might" move the message there?

indualagarsamy: But https://docs.particular.net/nservicebus/upgrades/6to6.2#timetobereceived-for-msmq ?

indualagarsamy: Isn't the new behavior now? I did clarify the old behavior as:

>However if the endpoint that is sending the heartbeat messages is using NServiceBus Versions 6.1 or below, these messages will be moved to the DLQ. This can happen when ServiceControl is stopped or very busy. In this case, the dead letter queue needs to be monitored and cleaned up.

indualagarsamy:
@andreasohlund - can you review this please. I added the ASB section after having a discussion with @SeanFeldman.

andreasohlund:
@Particular/servicecontrol-maintainers please give this a once over and merge

servicecontrol/plugins/heartbeat.md
```diff
@@ -75,11 +75,18 @@ partial: ttlCode
...
+## Expired heartbeat messages
 
-Heartbeat messages have a time to be received (TTBR) set based on the TTL value. If ServiceControl does not consume the heartbeat messages before the TTBR expires then transports that support a dead letter queue (DLQ) will move these messages to the DLQ. This can happen when ServiceControl is stopped or very busy. The dead letter queue needs to be monitored and cleaned up.
+Heartbeat messages have a time to be received (TTBR) set based on the TTL value.  If ServiceControl does not consume the heartbeat messages before the TTBR expires then those messages may be discarded. Transports like MSMQ and ASB support DLQ and if configured the messages will be forwarded to the DLQ instead. 
 
 
 ### MSMQ
 
-If this dead letter queue behavior is not required then it can disabled. This frequently happens with MSMQ as NServiceBus configures the usage of the DLQ by default. Read [MSMQ connection strings](/nservicebus/msmq/connection-strings.md) on how to disabled dead letter queue usage.
+Although NServiceBus configures the use of DLQ by default, messages that are defined with TTBR will not be automatically forwarded to the DLQ and will be discared.
```
WilliamBZA: This could very well be my ignorance showing, but isn't this in contradiction with the previous statement?

>  Transports like MSMQ and ASB support DLQ and if configured the messages will be forwarded to the DLQ instead.

I read that as (with MSMQ) the message will be forward to DLQ if TTBR is after processing time.

Then we say:
>  messages that are defined with TTBR will not be automatically forwarded to the DLQ and will be discared

Which seems to suggest the opposite.

indualagarsamy:
@WilliamBZA - Can you see if the latest changes address your concerns better?

PR #2844, opened 2017-06-13
Dispose connection when open throws

MarcinHoppe:


MarcinHoppe:
@tmasternak @SzymonPobiega We talked about this, can you take a look?

PR #2846, opened 2017-06-13
Added note regarding Stream incompatibility with ServiceControl

ramonsmits:
@Particular/azure-maintainers Please review

### DIFF
--- a/nservicebus/azure-service-bus/topologies/index.md
+++ b/nservicebus/azure-service-bus/topologies/index.md
@@ -25,6 +25,8 @@ The `EndpointOrientedTopology` is backward compatible with the Azure Service Bus
 
 The `ForwardingTopology` was introduced to take advantage of the broker nature of the Azure Service Bus and to leverage its native capabilities. It is the recommended option for new projects.
 
+NOTE: Although the ForwardingTopology is recommended it is currently not compatible with ServiceControl.
+
 Both topologies create a single input queue per endpoint and implement [Publish-Subscribe](/nservicebus/messaging/publish-subscribe/) mechanism. However, there is a significant difference in the way that mechanism is implemented.
 
 No default topology is set by the Azure Service Bus transport. Topology has to be explicitly configured using [configuration API](/nservicebus/azure-service-bus/configuration/full.md).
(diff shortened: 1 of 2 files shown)


ramonsmits:
@Particular/azure-maintainers How can we recommend something that isn't compatible with the rest of our platform?

yvesgoeleven:
@ramonsmits as discussed on slack, because we have to, MS is recommending it going that direction only in the future

nservicebus/azure-service-bus/topologies/index.md
```diff
@@ -25,6 +25,8 @@ The `EndpointOrientedTopology` is backward compatible with the Azure Service Bus
 
 The `ForwardingTopology` was introduced to take advantage of the broker nature of the Azure Service Bus and to leverage its native capabilities. It is the recommended option for new projects.
 
+NOTE: Although the ForwardingTopology is recommended it is currently not compatible with ServiceControl.
```
SeanFeldman: "Although" is not needed.

SeanFeldman: Back tick ForwardingTopology name.

ramonsmits:
@SeanFeldman I've updated the notes.

SeanFeldman:
I'm on a fence with this change.

> currently NOT compatible with ServiceControl.

As a new reader I'll understand it as "no SC for you". While it's not compatible as-is, there's an option of an adapter. And if we warn about incompatibility, we should mention adapter as an option.

@yvesgoeleven @ramonsmits

yvesgoeleven:
@SeanFeldman maybe add a link to the adapter everywhere as well?

PS: I really really feel we should fix the SC, been repeating this for 3 years now, I know, but still...
PPS: No one reads doco until it's too late

SeanFeldman:
> No one reads doco until it's too late

Even then I would not want to read the grim note w/o a ray of hope that **is** possible.

yvesgoeleven:
@SeanFeldman hence the link instead

ramonsmits:
To have a bit of both:

> The `SupportedBrokeredMessageBodyTypes.Stream` body format is recommended, but this mode is currently NOT compatible with ServiceControl Out of the box. As an alternative, it is possible to use the ServiceControl adapter approach to make ServiceControl compatible.

PR #2853, opened 2017-06-15
Queue creation warning should be visible for NSBv6 as well

dvdstelt:
There wasn't any documentation visible for NServiceBus v6 that queues can be created automatically by enabling installers. The file was for version 4 & 5 only.
The name-change should make it visible for v6 and further as well.

/ping @Particular/docs-maintainers


DavidBoike:
Not even changing any text, you're such a slacker @dvdstelt.

PR #2860, opened 2017-06-16
Update wormhole

SimonCropp:


SimonCropp:
@SzymonPobiega this throws an exception at startup. can u have a look

```
System.Exception occurred
  HResult=0x80131500
  Message=Faults forwarding requires an error queue to be specified using 'EndpointConfiguration.SendFailedMessagesTo()'
  Source=NServiceBus.Core
  StackTrace:
   at NServiceBus.MsmqTransport.Initialize(SettingsHolder settings, String connectionString) in C:\BuildAgent\work\3206e2123f54fce4\src\NServiceBus.Core\Transports\Msmq\MsmqTransport.cs:line 37
```

SzymonPobiega:
@SimonCropp  I've pushed a commit. It makse it work with 6.3 msmq. Unfortunately MSMQ added a required setting in 6.3 ("errorQueue") which is not part of the seam and NServiceBus.Raw does not yet manage it. In next release of NSB raw I will add it there so that the downstreams that use Raw will not have to worry.

SimonCropp:
@SzymonPobiega thanks

PR #2867, opened 2017-06-21
Support for non versioned subscription

mauroservienti:
Connects to https://github.com/Particular/NServiceBus.RavenDB/issues/319

* [x] add doco as a partial
* [x] create code snippet as we release on MyGet


DavidBoike:
Looks good so far, obviously missing the snippet, which I just changed the API for. :wink:

SimonCropp:
most of this should be in an upgrade guide. the rest should be split into a versioned partial so that the text reflects the current behavior or the selected version

mauroservienti:
@SimonCropp 

> most of this should be in an upgrade guide.

Even if it's not an upgrade? I mean, it's a new opt-in feature, that's it. We don't have an upgrade story in this case.

> the rest should be split into a versioned partial so that the text reflects the current behavior or the selected version

Yeah, we talked about that yesterday. My only concern was that it might not be visible enough. But I don't know how the feature works. What is the version shown by default? Latest? In which case if we add a partial bound to version 4.2 will be shown by default? Otherwise I believe it might be hard to discover.

nservicebus/ravendb/subscription-versioning.md
```diff
@@ -32,3 +32,13 @@ Because the version has been baked in to the RavenDB document id since the very
...
+ 
+## Non-versioned subscriptions support
+
+As of `NServiceBus.RavenDB` V4.2.0 support for non-versioned subscriptions has beed introduced. Non-versioned subscription are an opt-in feature due to their incompatibility with the default subscription storage behavior. To enable support for storing subscriptions without storing the message version change endpoint configuration as follows:
+
+```
+<insert code snippet here>
+```
+
+NOTE: due to the breaking nature of this feature it is important to take into account the impact of enabling it on pre-existing installations. Once enabled all the stored subscriptions, if any, will be immidiately ignored and publishers will stop publishing events until all subscribers are restarted and new subscription requests reach the pusblisher endpoint. Opting in this new feature has no effects for newly deployed publisher endpoints that never received any subscription requests.
```
SimonCropp: The `NOTE:` part will not be rendered. so u will want to start with an upper case. ie `Due to...`

SimonCropp:
@mauroservienti sorry i misread when i glanced over "breaking nature. u r correct this should not be an upgrade guide.

But you will still need to do it as a partial since the snippet will only target the current major. here is the doco https://github.com/Particular/docs.particular.net#markdown-partials

mauroservienti:
partial created and referenced, hopefully I did it correctly, please @SimonCropp review.
If I recall correctly to create the snippet I need to wait for the nuget package to be at least on Myget otherwise I cannot reference it, correct?

DavidBoike:
@mauroservienti You can start with a local NuGet, it just won't build on TC. Then update when the final package hits at least MyGet.

Though this is such a small and optional change may be easier to just wait until the package is on MyGet.

mauroservienti:
uhm...I did something wrong with the partial. @Particular/docs-maintainers suggestions?

DavidBoike:
You've got: `version-compatibility_non-versioned-subscriptions_raven_[4.2,].partial.md`

But it's supposed to be in `subscription-versioning.md`

So it should be (I believe):

`subscription-versioning_non-versioned-subscriptions_raven_[4.2,].partial.md`

mauroservienti:
too hot here to notice that stupid mistake, changed. Let's see if TC is happy 😃 thanks!

SimonCropp:
@mauroservienti i had to fix the file name. the range for partials has to match the version dirs in snippets. so since there is no 4.2 here https://github.com/Particular/docs.particular.net/tree/master/Snippets/Raven the partial name needs to be 4.0

i also fixed some spelling

SimonCropp:
i dont know if the @Particular/docs-maintainers want to add a throw for the above?

mauroservienti:
Thanks @SimonCropp , so since the new feature will be in 4.2 once we have the package and the snippets, in a 4.2 folder, I can rename the partial to 4.2 to match the snippet. Correct?

SimonCropp:
@mauroservienti in this case i recommended putting a version on the snippet suffix https://github.com/Particular/docs.particular.net#version-suffix-on-snippets. it will then render with version in the snippet header like https://docs.particular.net/nservicebus/pipeline/manipulate-with-behaviors?version=core_4

SimonCropp:
ie so there should be no need to create an extra snippets dir for 4.2

mauroservienti:
So does it mean that we can upgrade samples and snippets to 4.2, but via version on the snippet suffix we can render that snippet only when 4.2 is selected?

SimonCropp:
@mauroservienti no. it means when 4.0 is selected the snippet will show but it will have a header clearly indicating it is targeting 4.2

mauroservienti:
rebased on master to bring in the persistence file re-org. Added the snippet since we pushed to MyGet
Not anymore WIP, ready to review @Particular/ravendb-persistence-maintainers

mauroservienti:
@DavidBoike `4.2.0` is on MyGet, this has been updated to latest build from master. If you're happy with the result this is ready to be merged.

PR #2869, opened 2017-06-22
Switching to closure access in lambda to avoid deadlocks in lifetime scope

danielmarbach:
The current recomended approach that we have on our autofac sample can lead to a deadlock.  Example

```
builder.Register(x =>
            {
                return Endpoint.Start(endpointConfiguration).GetAwaiter().GetResult();
            }).As<IEndpointInstance>().SingleInstance();
```

The code above registers as a lambda the endpoint start. Endpoint start returns a task. You use `GetAwaiter().GetResult()` to execute the task in a blocking manner. The lambda itself is lazy evaluated when the first resolve on `IEndpointInstance` happens. 

```
 Timeout.Execute(() =>
            {                
               IEndpointInstance endpoint = container.Resolve<IEndpointInstance>();

               var msg = new StartTestSaga() { MsgId = 1 };
                endpoint.Publish(msg);
            }, 2000);
```

The timer execute enters on a background thread the lambda which then calls into resolve. Resolve internally uses the lifetime scope of autofac. Autofacs lifetime scope acquires a lock. https://github.com/autofac/Autofac/blob/ea21e5fd7076cb10383c339ab00dbd0fa5def4a2/src/Autofac/Core/Lifetime/LifetimeScope.cs#L268 While holding the lock we enter the lambda defined above when then starts the endpoint synchronously. When NServiceBus starts it also uses the container provided by you which is already holding a lock. It then wants to resolve other instances from the container, tries to acquire the lock which is already taken by the very same thread and you are deadlocked. See the picture below

image

## Workaround

```
            IMessageSession messageSession = null;
            // Register endpoint instance
            builder.Register(ctx => messageSession).As<IMessageSession>().SingleInstance();

            // Build container
            var container = builder.Build();

            endpointConfiguration.UseContainer<AutofacBuilder>(customizations =>
            {
                customizations.ExistingLifetimeScope(container);
            });

            var endpoint = await Endpoint.Start(endpointConfiguration).ConfigureAwait(false);
            messageSession = endpoint;

```

### DIFF
--- a/samples/containers/autofac/Autofac_6/Sample/Program.cs
+++ b/samples/containers/autofac/Autofac_6/Sample/Program.cs
@@ -20,10 +20,8 @@ static class Program
 
         var builder = new ContainerBuilder();
 
-        builder.Register(x =>
-            {
-                return Endpoint.Start(endpointConfiguration).GetAwaiter().GetResult();
-            })
+        IEndpointInstance endpoint = null;
+        builder.Register(x => endpoint)
             .As<IEndpointInstance>()
             .SingleInstance();
 
@@ -42,6 +40,11 @@ static class Program
         endpointConfiguration.UseSerialization<JsonSerializer>();
         endpointConfiguration.UsePersistence<LearningPersistence>();
         endpointConfiguration.UseTransport<LearningTransport>();
+
+        #region EndpointAssignment
+        endpoint = await Endpoint.Start(endpointConfiguration)
+            .ConfigureAwait(false);
+        #endregion
 
         var endpointInstance = container.Resolve<IEndpointInstance>();
         var myMessage = new MyMessage();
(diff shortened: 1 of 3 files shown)


danielmarbach:
I wasn't sure whether a partial is really required here so I decided to go with an empty snippet for the previous version. Would it work to just version the newly added snippet for the v6 sample?

DavidBoike:
@danielmarbach I think an "empty" snippet like that will render as an empty box and look broken. It also might be a bit odd to have a second snippet without narration. I think you'll probably need a partial.

danielmarbach:
Are you sure about the empty box? 

About the narration part: The current page has no narration at all. So another snippet without narration would not really hurt in that regard. 

I'll add the partial

danielmarbach:
Switched to partial

PR #2887, opened 2017-06-26
Multiple orms

SimonCropp:
shows multiple approaches to data persistence in the sql transport+persistence sample


samples/sqltransport-sqlpersistence/sample.md
```diff
@@ -39,48 +39,121 @@ This sample contains three projects:
...
+Receiver\bin\Debug\net462\NServiceBus.Persistence.Sql
+```
+
+
+### Accessing the ambient database details
+
+When persisting data to the same database it is recommended to use the same [DBConnection](https://msdn.microsoft.com/en-us/library/system.data.common.dbconnection.aspx) and  [DBTransaction](https://msdn.microsoft.com/en-us/library/system.data.common.dbtransaction.aspx) that is used by the underlying persistence and transport. The approach to this differs depending on the approach used to persist the data. Below are listed several approaches to data access including raw ADO.net and several popular ORM.
+
+NOTE: The connection and transaction are managed by NServiceBus, so there is no need to explicitly commit a transaction or dispose the connection.
+
```
SzymonPobiega: Should we explain why it is advisable to use the NSB-managed connection? e.g.

> Using NSB-managed connection ensures that both user handlers and NSB-managed sagas execute in the same transaction context.

SzymonPobiega:
All the ORMs use the same `Order` entity. How about doing a bit of composite-style app: submit the Order message that contains customer, order, shipping and billing information and have each handler use a different ORM to store a different piece of data.

SimonCropp:
> How about doing a bit of composite-style app

I think that would just confuse people as it adds an extra concept that would need to be explained

SimonCropp:
merged manually here https://github.com/Particular/docs.particular.net/commit/e6cce7d4bae79d50fac0ac46f4609559448d3f87

PR #2895, opened 2017-06-27
Everyone & Anonymous permissions explained

ramonsmits:


transports/msmq/index.md
```diff
@@ -71,12 +71,14 @@ Although MSMQ has the concept of both [Public and Private queues](https://techne
...
-|----------------|----------------------|
-| Owning account | Write, Receive, Peek |
-| Administrators | Full                 |
-| Anonymous      | Write                |
-| Everyone       | Write                |
+| Group          | Permissions         | Description                       |
+|----------------|---------------------|-----------------------------------|
+| Owning account | Send, Receive, Peek | Set by NServiceBus                |
+| Administrators | Full                | Set by NServiceBus                |
+| Anonymous      | Send                | Set by NServiceBus prior to 6.1.0 |
```
indualagarsamy: @ramonsmits - Shouldn't this be Versions 6.0.10 and below instead of 6.1.0? 6.1.0 was the version that introduced tighter permissions?

transports/msmq/index.md
```diff
@@ -71,12 +71,14 @@ Although MSMQ has the concept of both [Public and Private queues](https://techne
...
+| Everyone       | Send                | Set by NServiceBus prior to 6.1.0 |
```
indualagarsamy: Same as the above, i.e. 6.1.10 and below ?

transports/msmq/index.md
```diff
@@ -71,12 +71,14 @@ Although MSMQ has the concept of both [Public and Private queues](https://techne
...
+
+NOTE: If you are running in *Workgroup* mode then Windows will grant Send permissions to Anonymous. Everyone is added by default by Windows but not with Send or Receive permisssions. Prior to NServiceBus 6.1.0 Send permissions for Anonymous and Everyone were granted at queue creation but this behavior was altered to not accidentally open up permissions.
```
indualagarsamy: >Everyone is added by default by Windows but not with Send or Receive permisssions.

Not true. In workgroup mode, Everyone will have `Send` but not receive. In domain mode, Everyone will not have `Send`

How about this instead?


NOTE: When running in *Workgroup* mode, Windows will grant `Send` permissions to the `Anonymous` and `Everyone` user group.  However, the `Receive` permissions are not explicitly granted. When running in *Domain* mode, Windows will *not* grant `Send` permissions to the `Everyone` user group.

WARNING: When running Versions 6.0.10 and below,  `Send` permissions for `Anonymous` and `Everyone` group are granted at queue creation time.  While this might be ok in a development environment, the queue level permissions will need to be hardened in Production for better security.

indualagarsamy: Minor change to my suggestion based on @mikeminutillo's feedback. 

NOTE: When running in *Workgroup* mode, Windows will grant `Send` permissions to the `Anonymous` and `Everyone` user group.  However, the `Receive` permissions are not explicitly granted. When running in *Domain* mode, Windows will *not* grant `Send` permissions to the `Everyone` user group. When running in this mode, the receiving endpoint's queue will need to explicitly grant `Send` permissions to the domain user accounts that is attempting to send a message.  

WARNING: When running Versions 6.0.10 and below,  `Send` permissions for `Anonymous` and `Everyone` group are granted at queue creation time.  While this might be ok in a development environment, the queue level permissions will need to be hardened in Production for better security.

mikeminutillo: Let me have a go at it. I want to lead with the message *Domains won't work without you doing something* and then let the rest flow from there

mikeminutillo: How's this:

When an NServiceBus endpoint creates a queue on a machine, the default permissions depend on whether the machine is connected to a domain or a workgroup. 

If the machine is connected to a domain, then only the domain user running the endpoint when the queue is created will have `Send` permissions granted. If two endpoints need to communicate and are running under the same domain account, no further configuration is required. If two endpoints need to communicate using different domain accounts then the `Send` permission on the receiving endpoint input queue needs to be granted to the domain account of the sending endpoint.

For endpoints using the publish-subscribe pattern, both the publisher and the subscriber must have permission to send messages to each other. This is to allow the `Subscribe` message to be sent from the subscribing endpoint to the publishing endpoint.

If the machine is connected to a workgroup then the `Send` permission is granted to the `Everyone` and `Anonymous` user groups by Windows. Any endpoint will be able to send messages to any other endpoint without further configuration. 

----

After this you need the warning. I would maybe even just link off to the security advisory. It's not just new queues either. If you have run an older endpoint in prod, it may have granted additional permissions to an existing queue.

transports/msmq/index.md
ramonsmits: @indualagarsamy Its part of https://github.com/Particular/NServiceBus/releases/tag/6.1.0

> #4304 Anonymous permissions added to existing MSMQ queues

Those hotfixes we that we would set permissions always, not only at creation, overriding permissions set by the user. Maybe we should add that as a note too?

transports/msmq/index.md
```diff
@@ -78,13 +78,26 @@ Although MSMQ has the concept of both [Public and Private queues](https://techne
...
+
+### Domain mode
+
+If the machine is connected to a domain, then only the domain user running the endpoint when the queue is created will have Send permissions granted. If multiple endpoints need to communicate and are running under the same domain account, no further configuration is required. If multiple endpoints need to communicate using different domain accounts then the Send permission on the receiving endpoint input queue needs to be granted to the domain account of the sending endpoint.
+
+### Workgroup mode
+
+If the machine is connected to a workgroup then the Send permission is granted to the Everyone and Anonymous user groups by Windows. Any endpoint will be able to send messages to any other endpoint without further configuration.
+
+NOTE: In versions 5.2.20 and 6.0.3 a bug was fixes where [installers always changed permissions on a queue, also if it already existed](https://github.com/Particular/NServiceBus/issues/4266).
```
mikeminutillo: This needs to be a new section/heading. Otherwise it looks to be a part of the Workgroup mode section.

transports/msmq/index.md
```diff
@@ -78,13 +78,26 @@ Although MSMQ has the concept of both [Public and Private queues](https://techne
...
 
 To retrieve the group names the [WellKnownSidType](https://msdn.microsoft.com/en-us/library/system.security.principal.wellknownsidtype.aspx) enumeration is used.
 
 MSMQ permissions are defined in the [MessageQueueAccessRights](https://msdn.microsoft.com/en-us/library/system.messaging.messagequeueaccessrights.aspx) enumeration.
 
-NOTE: Write access is granted to both `Everyone` and `Anonymous`. Endpoints can receive messages from other endpoints running under different accounts. To increase security and further lock down MSMQ write permissions remove `Everyone` and `Anonymous` and grant specific permissions to a subset of accounts.
+NOTE: To increase security and further lock down MSMQ send/receive permissions remove `Everyone` and `Anonymous` and grant specific permissions to a subset of accounts.
```
mikeminutillo: grant specific permissions to the subset of accounts that need them.

transports/msmq/index.md
mikeminutillo: a bug was fixed

SimonCropp: not sure linking to a bug adds anything. i would prefer we just stated that the behavior changed

ramonsmits:
@mikeminutillo @SimonCropp Addressed your feedback.

transports/msmq/index.md
```diff
@@ -71,18 +71,34 @@ Although MSMQ has the concept of both [Public and Private queues](https://techne
...
-| Everyone       | Write                |
+| Group          | Permissions         | Description                       |
+|----------------|---------------------|-----------------------------------|
+| Owning account | Send, Receive, Peek | Set by NServiceBus                |
+| Administrators | Full                | Set by NServiceBus                |
+| Anonymous      | Send                | Set by NServiceBus prior to 6.1.0 |
+| Everyone       | Send                | Set by NServiceBus prior to 6.1.0 |
+
+
+For endpoints using the publish-subscribe pattern, both the publisher and the subscriber must have permission to send messages to each other. This is to allow the Subscribe message to be sent from the subscribing endpoint to the publishing endpoint.
```
indualagarsamy: This is also true for Request-Response pattern. Basically, any endpoint that wishes to send messages to another endpoint requires permission. I think we should add that.

mikeminutillo: I thought it was less obvious for Pub-Sub which is why I thought it needed an explicit mention. You are right though, for people who are coming from a synchronous background (like HTTP or web services) it may not be obvious. Good catch.

ramonsmits:
@indualagarsamy I hope I have addresses your request-response feedback.

transports/msmq/index.md
SimonCropp: @ramonsmits i dont think @indualagarsamy is questioning when this happened. just that "prior to 6.1.0 " is ambiguous and not in keeping with our version language through doc.

I think the test should not be 

>  Set by NServiceBus prior to 6.1.0

and instead should be

>  Set by NServiceBus Versions 6.0.x and below

indualagarsamy:
@SimonCropp - I am going to merge this PR when the build server goes green. I have addressed your comments in a PR that was merged into this one.

SimonCropp:
@indualagarsamy ok

PR #2907, opened 2017-06-29
manual unsubscribe sample

mauroservienti:
Sample showing how to manually unsubscribe when using MSMQ as a subscriptions persistence.

@SimonCropp it's the first time I create a sample, I'm not aware of any specific convention for the code part, so please take a look and review.
In this case I used VS2017 and referenced nugets as Project References, is it OK?

### PoA 

* [x] sample code
* [x] sample document describing the mechanics


samples/msmq/manual-unsubscribe/Core_6/Publisher/ManualUnsubscribeHandler.cs
```diff
@@ -0,0 +1,28 @@
...
+        ISubscriptionStorage subscriptionStorage;
+
+        public ManualUnsubscribeHandler(ISubscriptionStorage subscriptionStorage)
+        {
+            this.subscriptionStorage = subscriptionStorage;
+        }
+
+        public async Task Handle(ManualUnsubscribe message, IMessageHandlerContext context)
+        {
+            await subscriptionStorage.Unsubscribe(
```
timbussmann: you can return the task instead of awaiting it.

samples/msmq/manual-unsubscribe/Core_6/SubscriptionManager/Program.cs
```diff
@@ -0,0 +1,42 @@
...
+            endpointConfiguration.SendFailedMessagesTo("error");
+            endpointConfiguration.AuditProcessedMessagesTo("audit");
+
+            var endpointInstance = await Endpoint.Start(endpointConfiguration)
+                .ConfigureAwait(false);
+
+            var unsubscribeMessage = new ManualUnsubscribe()
+            {
+                MessageTypeName = "Messages.SomethingHappened",
+                MessageVersion = "1.0.0",
```
timbussmann: I think we no longer need the message version, and it seems to complicate things quite a bit?

timbussmann:
in case of already having a real endpoint which manages the unsubscribing and full messages describing the operation, I wonder why that message wouldn't go to the subscriber which then calls a regular unsubscribe (although that's tricky from a message handler). But otherwise seems ok to me.

samples/msmq/manual-unsubscribe/Core_6/Messages/Messages.csproj
```diff
@@ -0,0 +1,54 @@
...
+    <Reference Include="System.Net.Http" />
+    <Reference Include="System.Xml" />
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Include="ManualUnsubscribe.cs" />
+    <Compile Include="SomethingHappened.cs" />
+    <Compile Include="Properties\AssemblyInfo.cs" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="NServiceBus">
```
SimonCropp: any reason for mixing old and new csproj formats?

samples/msmq/manual-unsubscribe/Core_6/Messages/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Reflection;
```
SimonCropp: remove the redundant file

samples/msmq/manual-unsubscribe/Core_6/Publisher/App.config
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
```
SimonCropp: remove this file

samples/msmq/manual-unsubscribe/Core_6/Publisher/Program.cs
```diff
@@ -0,0 +1,40 @@
...
+namespace Publisher
+{
+    class Program
+    {
+        static void Main(string[] args)
+        {
+            MainAsync(args).GetAwaiter().GetResult();
+        }
+
+        static async Task MainAsync(string[] args)
```
SimonCropp: remove the redundant args

samples/msmq/manual-unsubscribe/Core_6/Publisher/Program.cs
```diff
@@ -0,0 +1,40 @@
+﻿using NServiceBus;
+using NServiceBus.Features;
+using NServiceBus.Persistence;
+using NServiceBus.Persistence.Legacy;
+using System;
+using System.Threading.Tasks;
+
+namespace Publisher
```
SimonCropp: remove the namespace

samples/msmq/manual-unsubscribe/Core_6/Publisher/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Reflection;
```
SimonCropp: remove this file

samples/msmq/manual-unsubscribe/Core_6/Subscriber/App.config
```diff
@@ -0,0 +1,6 @@
+<?xml version="1.0" encoding="utf-8"?>
```
SimonCropp: remove this file

samples/msmq/manual-unsubscribe/Core_6/Subscriber/Program.cs
```diff
@@ -0,0 +1,39 @@
+﻿using NServiceBus;
+using NServiceBus.Persistence;
+using NServiceBus.Persistence.Legacy;
+using System;
+using System.Threading.Tasks;
+
+namespace Subscriber
```
SimonCropp: remove the namespace

samples/msmq/manual-unsubscribe/Core_6/Subscriber/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Reflection;
```
SimonCropp: remove this file

samples/msmq/manual-unsubscribe/Core_6/manual-unsubscribe.sln
```diff
@@ -0,0 +1,40 @@
...
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{363EF626-8FF2-4762-AD14-1F22A75A4AC0}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{363EF626-8FF2-4762-AD14-1F22A75A4AC0}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{363EF626-8FF2-4762-AD14-1F22A75A4AC0}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{363EF626-8FF2-4762-AD14-1F22A75A4AC0}.Release|Any CPU.Build.0 = Release|Any CPU
+		{65DA7417-903D-4E7B-B47A-3489DE533DC1}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{65DA7417-903D-4E7B-B47A-3489DE533DC1}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{65DA7417-903D-4E7B-B47A-3489DE533DC1}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{65DA7417-903D-4E7B-B47A-3489DE533DC1}.Release|Any CPU.Build.0 = Release|Any CPU
```
SimonCropp: remove the release configs

samples/msmq/manual-unsubscribe/Core_6/SubscriptionManager/Program.cs
```diff
@@ -0,0 +1,42 @@
...
+                SubscriberEndpoint = "Subscriber",
+                SubscriberTransportAddress = "Subscriber@ANIMAL"
+            };
+
+            await endpointInstance.Send("Publisher", unsubscribeMessage)
+                .ConfigureAwait(false); ;
```
SimonCropp: extra `;`

samples/msmq/manual-unsubscribe/Core_6/Messages/Messages.csproj
mauroservienti: I have no idea 😃 VS2017 made all the decisions

mauroservienti: Should I revert everything to `Packages.config`

samples/msmq/manual-unsubscribe/Core_6/Publisher/ManualUnsubscribeHandler.cs
mauroservienti: It doesn't compile with the `ConfigureAwait`, or one or the other

samples/msmq/manual-unsubscribe/Core_6/SubscriptionManager/Program.cs
mauroservienti: `MessageType` is not happy if I don't put the version

samples/msmq/manual-unsubscribe/Core_6/Messages/Properties/AssemblyInfo.cs
mauroservienti: done

samples/msmq/manual-unsubscribe/Core_6/Publisher/App.config
mauroservienti: done

samples/msmq/manual-unsubscribe/Core_6/Publisher/Program.cs
mauroservienti: done

samples/msmq/manual-unsubscribe/Core_6/Publisher/Program.cs
mauroservienti: done

samples/msmq/manual-unsubscribe/Core_6/Publisher/Properties/AssemblyInfo.cs
mauroservienti: done

samples/msmq/manual-unsubscribe/Core_6/Subscriber/App.config
mauroservienti: done

samples/msmq/manual-unsubscribe/Core_6/Subscriber/Program.cs
mauroservienti: done

samples/msmq/manual-unsubscribe/Core_6/Subscriber/Properties/AssemblyInfo.cs
mauroservienti: done

samples/msmq/manual-unsubscribe/Core_6/manual-unsubscribe.sln
mauroservienti: done

samples/msmq/manual-unsubscribe/Core_6/SubscriptionManager/Program.cs
mauroservienti: done

mauroservienti:
> in case of already having a real endpoint which manages the unsubscribing and full messages describing the operation, I wonder why that message wouldn't go to the subscriber which then calls a regular unsubscribe (although that's tricky from a message handler). But otherwise seems ok to me.

the customer case this started from is related to subscribers that don't exist anymore. SO basically a subscriber is decommissioned but its subscriptions are never removed at the publisher. The end result is that publisher go on publishing but messages will end up in the dead letter queue on the publisher machine because the destination queue doesn't exist.

SeanFeldman:
Question: would this sample work for any message-driven pub/sub?

SimonCropp:
i pushed some minor cleanups

mauroservienti:
thanks @SimonCropp 

> Question: would this sample work for any message-driven pub/sub?

@SeanFeldman yes, by using `ISubscriptionStorage` it is agnostic to the transport. Sample is based n MSMQ but works for SQL Transport as for example.

SeanFeldman:
In that case, would it be better to re-brand this sample as a "Manual unsubscribe for transports with message driven pub/sub"?

mauroservienti:
Could be, solution name `manual-unsubscribe` is already agnostic. We could move it out of the `MSMQ` folder and when creating the `sample.md` talk about unsubscribe in general and say that the sample is using MSMQ transport.

mauroservienti:
reviewers! 😃 thoughts on this?

> In that case, would it be better to re-brand this sample as a "Manual unsubscribe for transports with message driven pub/sub"?

SimonCropp:
happy with a rename. but it would be better to hold off on asking for reviews till there is some markdown content

mauroservienti:
added a first draft of the sample doc, please take a look and suggest edits

samples/msmq/manual-unsubscribe/sample.md
```diff
@@ -0,0 +1,59 @@
+---
+title: Manual unsubscribe
+summary: Shows how to manually remove subscriptions when subscribers are decommissioned.
+reviewed: 2017-07-10
+component: Core
+related:
+- transports
+- transports/msmq
+- transports/sql
```
SeanFeldman: Any specific reason not to include ASQ as well?

mauroservienti:
No. I just copied headers from another sample, I'll add it @SeanFeldman. Thanks.

samples/msmq/manual-unsubscribe/sample.md
mauroservienti: added.

SimonCropp:
why is this related to ?
```
- transports/sql
- transports/azure-storage-queues
```

mauroservienti:
They are transports with message driven subscriptions

SimonCropp:
is this still WIP?

mauroservienti:
Removed the WIP, the sample document is still very raw.

SimonCropp:
@mauroservienti 

>  the  sample document is still very raw.

So what needs to be done to cook it? and can i help?

mauroservienti:
Please start by reviewing it and suggesting sections to add or existing section to rework if any. We can then polish it.

I'd like to make it as transport aware as possible, because it works for all transport that don't support native pub/sub. Or do you think it's better to provide one single sample document and a variant of the code sample already configured for each transport we want to demo?

samples/manual-unsubscribe/sample.md
```diff
@@ -0,0 +1,111 @@
...
+A publisher handle the `ManualUnsubscribe` message and perform the following task:
+
+snippet: unsubscribe-handling
+
+The message handler relies on the `ISubscriptionStorage` NServiceBus abstraction to perform the unsubscribe request regardless of the subscription storage configured for the publisher.
+
+
+## SubscriptionManager
+
+`SubscriptionManager` is a sample endpoint instance that can be used by operations personnel to send unsubscribe requests to publishers whenever a subscriber is decommissioned:
```
HEskandari: >operations personnel

it is a bit ambiguous...maybe DevOps?

samples/manual-unsubscribe/sample.md
```diff
@@ -0,0 +1,111 @@
...
+
+snippet: SubscriptionManager-config
+
+
+## Messages
+
+The shared messages and events used by this sample.
+
+
+## Running the sample
```
mikeminutillo: The above four sections have the same heading level as this one. Would it be better to have a high-level `## Solution overview` section and then make each project a level 3 heading under it?

mikeminutillo: It is usually my preference to lead with the Running the sample section and then go into details below. Get the user to run the sample, and then explain it's parts.

samples/manual-unsubscribe/Core_6/Publisher/ManualUnsubscribeHandler.cs
```diff
@@ -0,0 +1,48 @@
...
+                subscriber: address,
+                messageType: messageType,
+                context: emptyContext
+            ));
+        await Task.WhenAll(tasks)
+            .ConfigureAwait(false);
+    }
+}
+
+#endregion
```
mikeminutillo: This snippet feels very long. It might be OK on-screen but my initial read-through was "that looks complicated". Can we abstract some parts away?

Why do we need to filter the subscribers and then call unsubscribe for each one? Why isn't calling unsubscribe(address, messageType) enough?

mikeminutillo:
Should the sample mention calling `Unsubscribe()` from the original endpoint as an option? People might find this and think this is only way to unsubscribe.

Is it possible to hand-craft an `Unsubscribe` message that looks like it came from the original endpoint? Would that be simpler than hacking the storage directly?

samples/manual-unsubscribe/Core_6/Publisher/ManualUnsubscribeHandler.cs
SimonCropp: > Why do we need to filter the subscribers and then call unsubscribe for each one? Why isn't calling unsubscribe(address, messageType) enough?

msmq has the machine name in address. so to not need to query, the subscription manager would need to know the full queue names for all the machines subscribed

SimonCropp: > Can we abstract some parts away?

I would like to. perhaps only show the inner code of the handle, and say "subscriptionStorage injected by DI"?

samples/manual-unsubscribe/sample.md
```diff
@@ -0,0 +1,115 @@
...
+#### The publisher configuration
+
+snippet: publisher-config
+
+This specific sample is configured to use [MSMQ Subscription Persistence](/persistence/msmq/subscription.md).
+
+
+#### Subscriber decommissioning
+
+In such a configuration, when a subscriber endpoint is decommissioned it may happen that subscriptions remain stored at the publisher. Publishers have no way to detect that a subscriber is no longer available. They will continue to publish events even for subscribers that no longer exist. This behavior will eventually lead to storage and quota issues.
```
mauroservienti: @mikeminutillo would you add here more thoughts about:

> Should the sample mention calling Unsubscribe() from the original endpoint as an option? People might find this and think this is only way to unsubscribe.

mauroservienti: Or at the very beginning of this document?

mauroservienti:
> Is it possible to hand-craft an Unsubscribe message that looks like it came from the original endpoint? Would that be simpler than hacking the storage directly?

it can be done, and it's the first thing (thanks to @timbussmann ) suggested in the support case that originated this. The issue is that being a control message is a very low level one and constructing and sending it is other than simple. In NServiceBus V6 is to script the creation of an unsubscribe message and send it directly to the publisher queue one need to call the `ControlMessageFactory` static class to get an instance of an unsubscribe message. And then:

```
unsubscribeMessage.Headers[Headers.SubscriptionMessageType] = eventType.AssemblyQualifiedName; //event you want to unsubscribe from
unsubscribeMessage.Headers[Headers.ReplyToAddress] = replyToAddress; //the address of the subscriber
unsubscribeMessage.Headers[Headers.SubscriberTransportAddress] = replyToAddress; //the address of the subscriber
unsubscribeMessage.Headers[Headers.SubscriberEndpoint] = endpoint;
unsubscribeMessage.Headers[Headers.TimeSent] = DateTimeExtensions.ToWireFormattedString(DateTime.UtcNow);
unsubscribeMessage.Headers[Headers.NServiceBusVersion] = GitFlowVersion.MajorMinorPatch;
```

(maybe not all the above headers are required)

The unsubscribe message can then be sent by a `IDispatchMessages` instance, that can be injected via IoC, and calling the `Dispatch` method:

```
var transportOperation = new TransportOperation(unsubscribeMessage, new UnicastAddressTag(destination));
var transportTransaction = context.GetOrCreate<TransportTransaction>();
await dispatcher.Dispatch(new TransportOperations(transportOperation), transportTransaction, context).ConfigureAwait(false);
```

this approach seems to me way too complex and bound to the NServiceBus version.

SimonCropp:
@mauroservienti why " links in indexes"? samples indexes are auto gen'd

mauroservienti:
cool, I didn't know that. Removed from PoA.

mikeminutillo:
Thanks @mauroservienti for https://github.com/Particular/docs.particular.net/pull/2907#issuecomment-317947840

That makes sense. It looks like this is just not easy to do. :)

samples/manual-unsubscribe/sample.md
SimonCropp: @mikeminutillo how is this? https://github.com/Particular/docs.particular.net/pull/2907/commits/7f281dd2ecd2a16dc00bfb5baf9bbfe3c50b1032

SimonCropp:
@mauroservienti @mikeminutillo from my perspective this is good enough to merge. any objections?

mikeminutillo:
None from me. I like the snippet split. I think it makes things much clearer.

mauroservienti:
> from my perspective this is good enough to merge. any objections?

go 👍

PR #2914, opened 2017-06-30
Reviewed encryption key identifiers and externalize-encryption

danielmarbach:
Connects to https://github.com/Particular/PlatformDevelopment/issues/1355

### DIFF
--- a/nservicebus/upgrades/encryption-key-identifiers.md
+++ b/nservicebus/upgrades/encryption-key-identifiers.md
@@ -2,6 +2,7 @@
 title: Upgrade encryption configuration
 summary: Instructions on how to upgrade the Rijndael property encryption configuration to use key IDs.
 component: PropertyEncryption
+reviewed: 2017-06-30
 related:
  - nservicebus/security/generating-encryption-keys
 isUpgradeGuide: true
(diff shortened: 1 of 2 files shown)


nservicebus/upgrades/encryption-key-identifiers.md
```diff
@@ -59,7 +60,7 @@ The following configuration examples demonstrate how to update existing endpoint
 
 #### XML Configuration
 
-It is possible to choose to keep the current encryption key and will still be able to decrypt messages already in flight:
+It is possible to keep the current encryption key and still be able to decrypt messages in flight:
 
 ```xml
```
danielmarbach: @Particular/docs-maintainers This file has inline code. Was not sure if it is worth to move this into snippets since the information in there seems to be fairly stable. Happy to do it if you think it is necessary

DavidBoike: No strong feelings here. It wouldn't get validated by the build anyway.

SimonCropp: given this is an upgrade guide i dont mind

PR #2916, opened 2017-06-30
Reviewed msmq trouble shooting

danielmarbach:
Connects to https://github.com/Particular/PlatformDevelopment/issues/1355


danielmarbach:
I also double checked if all the external links are still valid

PR #2918, opened 2017-06-30
Switch to native ADO transactions

SzymonPobiega:
It allows to remove the warning override in EF while maintaining the same guarantees.


SimonCropp:
Need to remove the ref to AmbientTransactionWarning in the .md

PR #2919, opened 2017-06-30
Added V7 information

ramonsmits:


SimonCropp:
@ramonsmits i delete the part about "Version 7 will be released after...". there are many places in docs where we have v7 pre-release info, adding a "will be released" note to all of them will be difficult to maintian and a pain to fix after the release.

PR #2923, opened 2017-07-04
clean up targeting the MSMQ Permissions PR

indualagarsamy:
@ramonsmits @mikeminutillo @SimonCropp - I have made some changes, making it more clearer, fixing the version language, etc. Please review. I started this as a PR to the one that @ramonsmits started.

### DIFF
--- a/transports/msmq/index.md
+++ b/transports/msmq/index.md
@@ -73,10 +73,10 @@ Although MSMQ has the concept of both [Public and Private queues](https://techne
 
 ## Permissions
 
-| Group          | Permissions         | Description                       |
-|----------------|---------------------|-----------------------------------|
-| Owning account | Send, Receive, Peek | Set by NServiceBus                |
-| Administrators | Full                | Set by NServiceBus                |
+| Group          | Permissions         | Description                                    |
+|----------------|---------------------|------------------------------------------------|
+| Owning account | Send, Receive, Peek | Set by NServiceBus                             |
+| Administrators | Full                | Set by NServiceBus                             |
 | Anonymous      | Send                | Set by NServiceBus in Versions 6.0.x and below |
 | Everyone       | Send                | Set by NServiceBus in Versions 6.0.x and below |
 
--- a/transports/msmq/index.md
+++ b/transports/msmq/index.md
@@ -80,9 +80,9 @@ Although MSMQ has the concept of both [Public and Private queues](https://techne
 | Anonymous      | Send                | Set by NServiceBus in Versions 6.0.x and below |
 | Everyone       | Send                | Set by NServiceBus in Versions 6.0.x and below |
 
-NOTE: Starting with Versions 6.1.0 and above, the installers will not automatically grant permissions to the `Anonymous` and `Everyone` group. It will respect the existing queue permissions that have been set up for the endpoint. 
+NOTE: In Versions 6.1.0 and above, the installers will not automatically grant permissions to the `Anonymous` and `Everyone` group. It will respect the existing queue permissions that have been set up for the endpoint. The permissions granted to these user accounts is based on standard windows behavior and not via NServiceBus.
 
-Any endpoint that sends a message to a target endpoint requires the `Send` permission to be granted for the sending user account on the target queue. For example, if an `endpoint A` is running as `userA` and is sending a message to `endpoint B`, then `userA` requires the `Send` permission to be granted on `endpoint B`'s queue. When using messaging patterns like request-response or publish-subscribe, the queues for both the sending endpoint and the target endpoint will require `Send` permissions to be granted to each others user accounts. 
+Any endpoint that sends a message to a target endpoint requires the `Send` permission to be granted for the sending user account on the target queue. For example, if an `endpoint A` is running as `userA` and is sending a message to `endpoint B`, then `userA` requires the `Send` permission to be granted on `endpoint B`'s queue. When using messaging patterns like request-response or publish-subscribe, the queues for both the endpoints will require `Send` permissions to be granted to each others user accounts. 
 
 When an endpoint creates a queue on a machine, the default permissions depend on whether the server is joined to a [domain or a workgroup](https://support.microsoft.com/en-us/help/884974/information-about-workgroup-mode-and-about-domain-mode-in-microsoft-me).
 
@@ -97,13 +97,14 @@ If the machine is joined to a domain, then at the time of queue creation, only t
 If the machine is connected to a workgroup, then the `Send` permission is granted to the `Everyone` and `Anonymous` user groups by Windows. Any endpoint will be able to send messages to any other endpoint without further configuration.
 
 
+### Well-Known group names and Queue access rights
 To retrieve the group names the [WellKnownSidType](https://msdn.microsoft.com/en-us/library/system.security.principal.wellknownsidtype.aspx) enumeration is used.
 
 MSMQ permissions are defined in the [MessageQueueAccessRights](https://msdn.microsoft.com/en-us/library/system.messaging.messagequeueaccessrights.aspx) enumeration.
 
 NOTE: To increase security and further lock down MSMQ send/receive permissions remove `Everyone` and `Anonymous` and grant specific permissions to the subset of accounts that need them.
 
-NOTE: From Versions 6 and above, if the default queue permissions are set, a log message will be written during the endpoint startup, reminding that the queue has default permissions and might require stricter permissions for production. During development, if running with an attached debugger, this message will be logged as `INFO` level, otherwise `WARN`.
+NOTE: In Versions 6 and above, if the default queue permissions are set, a log message will be written during the endpoint startup, reminding that the queue has default permissions and might require stricter permissions for production. During development, if running with an attached debugger, this message will be logged as `INFO` level, otherwise `WARN`.
 
 An example of the warning that is logged:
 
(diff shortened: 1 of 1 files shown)


transports/msmq/index.md
```diff
@@ -77,25 +77,24 @@ Although MSMQ has the concept of both [Public and Private queues](https://techne
...
 
-When an endpoint sends a command or event message that source requires permission at the target queue. For endpoints using the either the publish-subscribe or request-response patterns, both parties actually send messages and must have permission to send messages to each other. This is to allow the Subscribe message to be sent from the subscribing endpoint to the publishing endpoint in case of publish-subscribe or the response to be send back to the client in case of request-response.
+Any endpoint that sends a message to a target endpoint requires the `Send` permission to be granted for the sending user account on the target queue. For example, if an `endpoint A` is running as `userA` and is sending a message to `endpoint B`, then `userA` requires the `Send` permission to be granted on `endpoint B`'s queue. When using messaging patterns like request-response or publish-subscribe, the queues for both the sending endpoint and the target endpoint will require `Send` permissions to be granted to each others user accounts.
```
mikeminutillo: > the queues for both the sending endpoint and the target endpoint will require

It's not clear which is the "sending" and "target" endpoints in a pub-sub/req-res interaction. I would just say "both"

transports/msmq/index.md
```diff
@@ -77,25 +77,24 @@ Although MSMQ has the concept of both [Public and Private queues](https://techne
...
 
 
 ### Domain mode
 
-If the machine is connected to a domain, then only the domain user running the endpoint when the queue is created will have Send permissions granted. If multiple endpoints need to communicate and are running under the same domain account, no further configuration is required. If multiple endpoints need to communicate using different domain accounts then the Send permission on the receiving endpoint input queue needs to be granted to the domain account of the sending endpoint.
+If the machine is joined to a domain, then at the time of queue creation, only the domain user will have `Send` permissions granted. The `Everyone` user group will not have `Send` permissions. If all the endpoints which need to communicate are running under the same domain account, no further configuration is required. However, if the endpoints are run using different domain accounts, then the `Send` permission on the receiving endpoint's input queue needs to be explicitly granted to the domain user account of the sending endpoint.
```
mikeminutillo: > The `Everyone` user group will not have `Send` permissions.

Why would I assume that it is? Just because of the table above? Should the table include some kind of indication?

transports/msmq/index.md
```diff
@@ -77,25 +77,24 @@ Although MSMQ has the concept of both [Public and Private queues](https://techne
...
 
 
 ### Workgroup mode
 
-If the machine is connected to a workgroup then the Send permission is granted to the Everyone and Anonymous user groups by Windows. Any endpoint will be able to send messages to any other endpoint without further configuration.
+If the machine is connected to a workgroup, then the `Send` permission is granted to the `Everyone` and `Anonymous` user groups by Windows. Any endpoint will be able to send messages to any other endpoint without further configuration.
 
 
 To retrieve the group names the [WellKnownSidType](https://msdn.microsoft.com/en-us/library/system.security.principal.wellknownsidtype.aspx) enumeration is used.
```
mikeminutillo: This needs a heading or it looks like it's part of the `Workgroup mode` section above

transports/msmq/index.md
mikeminutillo: What about Anonymous?

transports/msmq/index.md
```diff
@@ -104,9 +103,9 @@ MSMQ permissions are defined in the [MessageQueueAccessRights](https://msdn.micr
 
 NOTE: To increase security and further lock down MSMQ send/receive permissions remove `Everyone` and `Anonymous` and grant specific permissions to the subset of accounts that need them.
 
-NOTE: From Version 6, if the default queue permissions are set, a log message will be written during the transport startup, reminding that the queue has default permissions. During development, if running with an attached debugger, this message will be logged as `INFO` level, otherwise `WARN`.
+NOTE: From Versions 6 and above, if the default queue permissions are set, a log message will be written during the endpoint startup, reminding that the queue has default permissions and might require stricter permissions for production. During development, if running with an attached debugger, this message will be logged as `INFO` level, otherwise `WARN`.
```
mikeminutillo: "From" doesn't flow well with "Versions X and above". Maybe just "In Versions 6 and above"?

transports/msmq/index.md
```diff
@@ -77,25 +77,24 @@ Although MSMQ has the concept of both [Public and Private queues](https://techne
 |----------------|---------------------|-----------------------------------|
 | Owning account | Send, Receive, Peek | Set by NServiceBus                |
 | Administrators | Full                | Set by NServiceBus                |
-| Anonymous      | Send                | Set by NServiceBus prior to 6.1.0 |
-| Everyone       | Send                | Set by NServiceBus prior to 6.1.0 |
+| Anonymous      | Send                | Set by NServiceBus in Versions 6.0.x and below |
+| Everyone       | Send                | Set by NServiceBus in Versions 6.0.x and below |
 
+NOTE: Starting with Versions 6.1.0 and above, the installers will not automatically grant permissions to the `Anonymous` and `Everyone` group. It will respect the existing queue permissions that have been set up for the endpoint.
```
mikeminutillo: "Starting with" does not flow well with "Versions X and above". What about just "In Version 6.1.0 and above"?

There's two key differences in that version, isn't there? Firstly, we no longer change permissions on existing queues at all (in the past we did). Secondly, we no longer set permissions for Anonymous and Everyone explicitly. What happens with those accounts is a windows default for all newly created queues (created by NSB or any other mechanism). I wonder if we need to be more explicit about that.

transports/msmq/index.md
indualagarsamy: I don't think anonymous users have send permissions, but I'd have to check to be certain.

indualagarsamy: @mikeminutillo - It's important to clarify that the "Everyone" user group will not have send permissions, because that's a huge difference between domain and workgroup and have caused confusions for a few users.

transports/msmq/index.md
```diff
@@ -77,36 +77,36 @@ Although MSMQ has the concept of both [Public and Private queues](https://techne
...
 
 
 ### Domain mode
 
-If the machine is connected to a domain, then only the domain user running the endpoint when the queue is created will have Send permissions granted. If multiple endpoints need to communicate and are running under the same domain account, no further configuration is required. If multiple endpoints need to communicate using different domain accounts then the Send permission on the receiving endpoint input queue needs to be granted to the domain account of the sending endpoint.
+If the machine is joined to a domain, then at the time of queue creation, only the domain user that created the queue, domain administrators and local administrators will have `Send` permissions granted. The `Everyone` user group and `Anonymous` user group will not have `Send` permissions. If all the endpoints which need to communicate are running under the same domain account, no further configuration is required. However, if the endpoints are run using different domain accounts, then the `Send` permission on the receiving endpoint's input queue needs to be explicitly granted to the domain user account of the sending endpoint.
```
ramonsmits: > If the machine is joined to a domain, then at the time of queue creation, only the domain user that created the queue, domain administrators and local administrators will have `Send` permissions granted

Should we end that line with ..  by NServiceBus. ?

transports/msmq/index.md
```diff
@@ -77,36 +77,36 @@ Although MSMQ has the concept of both [Public and Private queues](https://techne
 |----------------|---------------------|-----------------------------------|
 | Owning account | Send, Receive, Peek | Set by NServiceBus                |
 | Administrators | Full                | Set by NServiceBus                |
-| Anonymous      | Send                | Set by NServiceBus prior to 6.1.0 |
-| Everyone       | Send                | Set by NServiceBus prior to 6.1.0 |
+| Anonymous      | Send                | Set by NServiceBus in Versions 6.0.x and below |
+| Everyone       | Send                | Set by NServiceBus in Versions 6.0.x and below |
 
+NOTE: In Versions 6.1.0 and above, the installers will not automatically grant permissions to the `Anonymous` and `Everyone` group. It will respect the existing queue permissions that have been set up for the endpoint. The permissions granted to these user accounts is based on standard windows behavior and not via NServiceBus.
```
ramonsmits: Instead of using the back ticks everywhere, wouldn't it be better to use cursive instead by using asterisk? We are not using code or commandline things here.

indualagarsamy: I don't have a preference for either way. I did this to differentiate it in a consistent manner.

transports/msmq/index.md
indualagarsamy: No, because that's also default windows behavior when joined to a domain.  However I do see your point that we explicitly add administrators permissions. I think its safe to not specify either. As we have clearly stated the correct behavior.

ramonsmits: No, local and domain admins are not added by Windows, that is something only NServiceBus does. Not sure if that you just wrote the same.

In this section we are talking about the behavior of Windows aren't we? If we mention anything NServiceBus related then we should mention that I think.

indualagarsamy: @ramonsmits - not according to this:
https://technet.microsoft.com/en-gb/library/cc731882(WS.10).aspx

See: *Default queue permissions for new queues do not grant everyone send access*
which explains what permissions are set by default for workgroup and domain mode.

ramonsmits: Well.. that doco is correct :-) the table mentions exactly that in the left column about *private* queues. I just tested it on my Windows 10 (1703) and Windows 2012 R2 (6.3) which are both part of a domain and with @dvdstelt who has both Win 10 and Win 2012 in workgroup mode.

All don't have any administrator group (not local & not from the domain) added when you create a queue via Computer Management.

We explicitly add that for the queues that we create:

- https://github.com/Particular/NServiceBus.Transport.Msmq/blob/develop/src/NServiceBus.Transport.Msmq/MsmqQueueCreator.cs#L63

ramonsmits:
@indualagarsamy @mikeminutillo Added some changes. I think we are near completion. The text below the grid is now 100% supporting the data in the grid.

dvdstelt:
w00t

Issue #2547, opened 2017-03-13
Feedback: 'Message Headers' - CorrelationId when sending from a saga


Niklas-Peter:
Feedback for 'Message Headers' https://docs.particular.net/nservicebus/messaging/headers

Location in GitHub: https://github.com/Particular/docs.particular.net/blob/master/nservicebus/messaging/headers.md

ping @Particular/nservicebus-maintainers

> Message correlation connects request messages with their corresponding response messages. The Correlation Id of the response message is a Message Id of its corresponding request message.

When sending a request from within a saga and trying to receive the response in a handler outside the saga (special case, I can not receive the response in the saga, because I have to complete the saga after sending the request), this is obviously not true. The `CorrelationId` of the response is not set to the `MessageId` of the request, but to the message ID of the message initiating the saga.

The `RelatedTo` header contains the request's message ID. However, I do not want my response handler have to know, whether the corresponding request was sent from a saga or not. 

Is there a reason to not always use `RelatedTo` in a response handler for request-response correlation?

mikeminutillo:
Hi @Niklas-Peter 

Once a `CorrelationId` has been set for a conversation, it is sticky and is carried forward to every new message in the conversation. 

Your intuition to rely on `RelatedTo` for this is correct. `RelatedTo` will always be the `MessageId` of the message that was being handled when this message was sent.

Niklas-Peter:
@mikeminutillo 
> Once a CorrelationId has been set for a conversation, it is sticky and is carried forward to every new message in the conversation.

Why do we need an additional `NServiceBus.ConversationId` then?

mikeminutillo:
@Niklas-Peter in previous versions of NServiceBus it was possible to switch to a different `CorrelationId` within the same `ConversationId`. In Version 6 and above this is still possible (by manually setting a `CorrelationId` on an outgoing message) but I'm not sure why you'd want to do it. 

The `CorrelationId` header was originally added to enable synchronous Callbacks. In NServiceBus Version 6 and above, this type of interaction can only be initiated by the very first message in a conversation and any given conversation is unlikely to have multiple `CorrelationId`s.

I believe the `RelatedTo` header was added afterwards (probably to enable the Message Flow Diagram in ServiceInsight). 

We had a brief discussion internally about renaming some of these headers but in order to allow backwards compatibility we need to keep them as they are.

We will review the documentation and see what we can do to make it clearer.

SimonCropp:
[...]

PR #2926, opened 2017-07-05
Explain relationship between `ConversationId` and `CorrelationId`

mikeminutillo:
Based on feedback provided in https://github.com/Particular/docs.particular.net/issues/2547

I tried to clarify the relationships between `Correlation Id`, `Conversation Id`, and `RelatedTo`

### DIFF
--- a/nservicebus/messaging/headers.md
+++ b/nservicebus/messaging/headers.md
@@ -73,9 +73,11 @@ A [unique ID for the current message](/nservicebus/messaging/message-identity.md
 
 NServiceBus implements the [Correlation Identifier](http://www.enterpriseintegrationpatterns.com/patterns/messaging/CorrelationIdentifier.html) pattern by using a `Correlation Id` header.
 
-Message correlation connects request messages with their corresponding response messages. The `Correlation Id` of the response message is a `Message Id` of its corresponding request message.
+Message correlation connects request messages with their corresponding response messages. The `Correlation Id` of the response message is the `Correlation Id` of its corresponding request message. Each outgoing message which is sent outside of a message handler will have it's `Correlation Id` set to it's `Message Id`.
 
 An example usage of Correlation Identifier within NServiceBus is [callbacks](/nservicebus/messaging/callbacks.md).
+
+Messages sent from a Saga using the `ReplyToOriginator` method will have their `Correlation Id` set based on the message which caused the saga to be created. See [Notifying callers of status](/nservicebus/sagas/#notifying-callers-of-status) for more information about the `ReplyToOriginator` method.
 
 
 ### NServiceBus.ConversationId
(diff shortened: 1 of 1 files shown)


nservicebus/messaging/headers.md
```diff
@@ -73,21 +73,27 @@ A [unique ID for the current message](/nservicebus/messaging/message-identity.md
 
 NServiceBus implements the [Correlation Identifier](http://www.enterpriseintegrationpatterns.com/patterns/messaging/CorrelationIdentifier.html) pattern by using a `Correlation Id` header.
 
-Message correlation connects request messages with their corresponding response messages. The `Correlation Id` of the response message is a `Message Id` of its corresponding request message.
+Message correlation connects request messages with their corresponding response messages. The `Correlation Id` of the response message is the `Correlation Id` of its corresponding request message. Each outgoing message which is sent outside of a message handler will have it's `Correlation Id` set to it's `Message Id`.
```
udidahan: I thought the `Correlation Id` of the response was the `Message Id` of the request.

mikeminutillo: That ends up being true, but it's not how it's written. The code that attaches correlation id is [here](https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Correlation/AttachCorrelationIdBehavior.cs). In pseudocode it is:

- if there is an explicit Correlation Id set, use that
- if there is an incoming message
  - if it has a correlation id header, use that
  - otherwise, if it has a message id header, use that
- if we still don't have a correlation id, use the message id

That means, for the outgoing message without a corresponding incoming message (or an explicitly set correlation id), it gets set to the `Message Id`.  When processing that message, all correlated outgoing messages will get that `Correlation Id` which is the original message's `Message Id`.

So if you say:

1. **Endpoint A** sends **Request** to **Endpoint B**
2. **Endpoint B** handles **Request** and sends **Request 2** to **Endpoint C**

Then the `Correlation Id` on **Request 2** will be the `Message Id` of **Request** even though this isn't a request/response situation. In fact, if you did:

1. **Endpoint A** sends **Request** to **Endpoint B**
2. **Endpoint B** handles **Request** and sends **Response** back to **Endpoint A**
3. **Endpoint A** handles **Response** and publishes **Event**

The the `Correlation Id` on **Request**, **Response**, _and_ **Event** will all be the `Message Id` of **Request**.

`Correlation Id` leaks out of single request/response interactions and works very similarly to `Conversation Id` (which is why this PR was raised in the first place). If that behavior is incorrect then we might need to treat it as a bug and fix it.

udidahan: I would say that that behavior is incorrect. In the scenario you described:

> Endpoint A sends Request to Endpoint B
> Endpoint B handles Request and sends Response back to Endpoint A
> Endpoint A handles Response and publishes Event

The Response should have the `Correlation Id` but the Event shouldn't. The Event should have the `Related To` set to the `Message Id` of the Response, and both the Response and the Event should share the same `Conversation Id` that was set on the original Request.

mikeminutillo: Raised https://github.com/Particular/NServiceBus/issues/4841

PR #2953, opened 2017-07-12
Msmq subscription queue override improvements

ramonsmits:


SimonCropp:
i made some minor cleanups. the rest LGTM

perhaps a review from a @Particular/msmq-transport-maintainers would be good

dvdstelt:
@ramonsmits Could you please provide a proper description and an issue?

dvdstelt:
lol, I thought this was in msmq repo. Now I see it's doco. Still, could use 'something' for a description :-)

persistence/msmq/subscription_config_core_[6,).partial.md
```diff
@@ -2,8 +2,24 @@ To configure MSMQ as the subscription persistence:
...
 snippet: ConfiguringMsmqPersistence
 
-By default NServiceBus uses a queue called `NServiceBus.Subscriptions`. If not specified otherwise, all endpoints will share that queue to store subscriptions. In order to specify a different queue, add the following `configSections` and subsequent config entry:
+By default NServiceBus uses a queue called `NServiceBus.Subscriptions`. If not specified otherwise, all endpoints will share that queue to store subscriptions. The following warning message will get logged:
 
-snippet: MsmqSubscriptionAppConfig
+> NServiceBus.Features.MsmqSubscriptionPersistence Could not find configuration section for Msmq Subscription Storage and no name was specified for this endpoint. Going to default the subscription queue
 
-WARNING: When using MSMQ Subscription Persistence on multiple endpoints running on the same machine, every endpoint requires a dedicated subscription storage queue.
+WARNING: When using MSMQ Subscription Persistence on multiple endpoints running on the same machine, every endpoint **must have** a dedicated subscription storage queue.
```
dvdstelt: How about

> When using MSMQ Subscription Persistence in multiple different endpoints, running on the same machine, every endpoint **should have** a dedicated subscription storage queue. Sharing a single queue will result in unexpected behavior.

dvdstelt: I suggested the above change, because I keep on reading the above sentence incorrectly. I hope this helps.

persistence/msmq/subscription_config_core_[,4].partial.md
```diff
@@ -1,4 +1,4 @@
-WARNING: Storing subscriptions in MSMQ is not suitable when endpoints are scaled-out, because the subscription queue cannot be shared among multiple endpoints.
+WARNING: Storing subscriptions in MSMQ **must not** be used when scaling out across multiple machines or running side-by-side on the same machine. The subscription queue cannot be shared among multiple endpoints instances for the same endpoint.
```
Comment (dvdstelt) [inline persistence/msmq/subscription_config_core_[,4].partial.md]: I found "is not suitable" better than "must not", but could be because I'm not native English. "must not" doesn't sound right to me.

Comment (dvdstelt) [inline persistence/msmq/subscription_config_core_[,4].partial.md]: I worked my way up from the bottom, so probably better to read the below comment first ;-)

persistence/msmq/subscription_config_core_[6,).partial.md
indualagarsamy: @dvdstelt - multiple different seems to be redundant. Also, when you say, "Sharing a single queue will result in unexpected behavior" doesn't explain much.  How about:

If multiple endpoints are hosted on the same server, and if the endpoints are configured to use MSMQ subscription persistence, then each endpoint **must have** a dedicated subscription storage queue.

persistence/msmq/subscription_config_core_[6,).partial.md
```diff
@@ -2,8 +2,24 @@ To configure MSMQ as the subscription persistence:
...
+
+In order to specify a different queue, use the code api or specify a configuration section.
```
indualagarsamy: >specify a different queue

implies there is a default queue. Can you add more details on the default queue?

timbussmann:
> Storing subscriptions in MSMQ **must not** be used when scaling out across multiple machines 

I think that's not true as long as you configure SSD correctly (which you need to configure anyways when scaling out MSMQ)?

> or running side-by-side on the same machine.

also not entirely true I think. It can be used as long as each endpoint instance has it's dedicated queue and SSD is configured correctly.

> The subscription queue cannot be shared among multiple endpoints instances for the same endpoint.

that seems to be correct.

> When using MSMQ Subscription Persistence on multiple endpoints running on the same machine, every endpoint **must have** a dedicated subscription storage queue.

if the endpoints are different logical endpoints, I think there should be no issue by sharing the queue?

persistence/msmq/subscription_config_core_[,4].partial.md
Comment (ramonsmits) [inline persistence/msmq/subscription_config_core_[,4].partial.md]: I based the phrasing on : https://www.ietf.org/rfc/rfc2119.txt

persistence/msmq/subscription_config_core_[6,).partial.md
ramonsmits: @indualagarsamy  The default queue is mentioned in the parent document.

ramonsmits:
@timbussmann Currently this is in docs:

> When using MSMQ Subscription persistence, the default queue to store the subscription messages is NServiceBus.Subscriptions. It is important to override this queue name for each endpoint to avoid sharing the same subscription queue for different endpoints running on the same machine.

You imply that this doco is incorrect and that the current following warning statement is already covering your concerns about scaling out:

> Storing subscriptions in MSMQ is not suitable when endpoints are scaled-out, because the subscription queue cannot be shared among multiple endpoints, especially if the endpoints are running on different machines.


I also think that a single subscription queue - which is the default since v5 - does not cause any issue but current doco states *it is important to override it* but it is not clear why hence this doco update. If that statement is incorrect than we can highlight that:

* MSMQ subscription storage does not work for a scaled out publisher
* That receiving a subscription on an publisher running side-by-side will only update 1 instance
* By using a subscription queue per publisher makes it clear who owns the subscriptions

SimonCropp:
@timbussmann @dvdstelt any more feedback on this on?

PR #2960, opened 2017-07-13
Updated to 3.1.1

SzymonPobiega:


SimonCropp:
does the creation script need to be changed? https://docs.particular.net/transports/sql/operations-scripting#create-queues-the-create-queue-helper-methods

SzymonPobiega:
@SimonCropp yeah, good catch.

SimonCropp:
r there any other varchar/nvarchar changes required on that page?

SzymonPobiega:
Yeah. Fixing all of them now as part of this branch.

SzymonPobiega:
I think I fixed them all. I didn't include sending native delayed messages because I think this would be an overkill.

PR #2965, opened 2017-07-14
[WIP] ASB transport adapter sample

SzymonPobiega:
- [x] Make the sample more real-life by adding some complexity (e.g. body encoding + secure connection strings)
 - [x] Updated `sample.md`
- [x] Review


SzymonPobiega:
So far I managed to implement a simple case where both endpoints use a single namespace while the SC uses a different one. The endpoints use the new forwarding topology while the SC uses the old one (but it should not matter since we don't really do pubsub there).

@SeanFeldman I've looked at the doco and asked on #azure and if I understand correctly the cross-namespace routing (https://docs.particular.net/transports/azure-service-bus/multiple-namespaces-support#cross-namespace-routing) does not work for pub/sub. So a scenario where each endpoint uses a different namespace seems weird to me (since the endpoints could not do pub/sub). 

Based on that I think it makes sense to make the sample more real-life either by using multi-namespace with round robin or some other feature like the body encoding.

SeanFeldman:
We'll do body encoding + secure connection strings.

SeanFeldman:
**Update**

- I've updated the sample to use Secure Connection Strings and Body Encoding `Stream` to demonstrate how the business endpoints can follow the recommendation we're giving for the transport while continue using SC.
- Updated sample.md with the changes for the transport and relevant snippets.

@SzymonPobiega please review

SzymonPobiega:
@SeanFeldman I reviewed the code once again and figured that we need to go multi-namespace. I did it in 0d071f1a8d1ddbe50c7b1bae6701f7d97795e66c. The reason is, without that the `UseNamespaceAliasesInsteadOfConnectionStrings` is not really used and messages flow correctly. When I switched `Shipping` to use a separate namespace it all fell apart.
 * The destination address for the retry did not include the namespace alias because `FailedQ` header does not include it
 * The `ReplyToAddress` header gets corrupted by ASB 6.x in ServiceControl and the namespace alias is replaced by SC's connection string.

In https://github.com/Particular/ServiceControl.TransportAdapter/pull/10 I added two features required to make it work: destination override based on a custom header included in the failed message and header value restoration.

For the context, here's a link to my slack conversation with Yves https://particularsoftware.slack.com/archives/C1607JM6G/p1500490738006461

samples/servicecontrol/adapter-asb-multi-namespace/SCTransportAdapter_1/Adapter/Adapter.csproj
```diff
@@ -0,0 +1,64 @@
...
+    <Reference Include="ServiceControl.TransportAdapter, Version=1.0.0.0, Culture=neutral, PublicKeyToken=9fc386479f8a226c, processorArchitecture=MSIL">
+      <HintPath>..\..\..\..\..\packages\ServiceControl.TransportAdapter.1.2.0\lib\net452\ServiceControl.TransportAdapter.dll</HintPath>
+    </Reference>
+    <Reference Include="System" />
+    <Reference Include="System.Core" />
+    <Reference Include="Microsoft.CSharp" />
+    <Reference Include="System.Data" />
+    <Reference Include="System.Runtime.Serialization" />
+    <Reference Include="System.ServiceModel" />
+    <Reference Include="System.Xml" />
```
SimonCropp: is the xml ref required?

samples/servicecontrol/adapter-asb-multi-namespace/diagram.svg
```diff
@@ -0,0 +1 @@
...71604938272,0,0,0.04938271604938272,58.22222222222223,0)" xlink:href="#m"/><use transform="matrix(0.04938271604938272,0,0,0.04938271604938272,69.03703703703704,0)" xlink:href="#x"/><use transform="matrix(0.04938271604938272,0,0,0.04938271604938272,78.91358024691358,0)" xlink:href="#s"/><use transform="matrix(0.04938271604938272,0,0,0.04938271604938272,88.79012345679013,0)" xlink:href="#p"/></g><path fill="#fff" d="M182-182c0 78-84 86-111 141h115V0H12c-6-101 99-100 120-180 1-22-12-31-33-32-23 0-32 14-35 34l-49-3c5-45 32-70 84-70 51 0 83 22 83 69" id="D"/><use transform="matrix(0.04938271604938272,0,0,0.04938271604938272,0,0)" xlink:href="#D" id="g"/><path fill="#fff" d="M23 0v-37h61v-169l-59 37v-38l62-41h46v211h57V0H23" id="E"/><use transform="matrix(0.04938271604938272,0,0,0.04938271604938272,0,0)" xlink:href="#E" id="h"/><path fill="#fff" d="M128-127c34 4 56 21 59 58 7 91-148 94-172 28-4-9-6-17-7-26l51-5c1 24 16 35 40 36 23 0 39-12 38-36-1-31-31-36-65-34v-40c32 2 59-3 59-33 0-20-13-33-34-33s-33 13-35 32l-50-3c6-44 37-68 86-68 50 0 83 20 83 66 0 35-22 52-53 58" id="F"/><use transform="matrix(0.04938271604938272,0,0,0.04938271604938272,0,0)" xlink:href="#F" id="i"/></defs></g></svg>
```
SimonCropp: can we use mermaid instead?

SeanFeldman:
I'm back and will review this tomorrow.

samples/servicecontrol/adapter-asb-multi-namespace/diagram.svg
SeanFeldman: I'll give it a try @SimonCropp

SeanFeldman:
@SzymonPobiega I've renamed adapter specific headers to `"_adapter.Original...."` to indicate better that those are *not* NServiceBus specific header.
image

Also, in order to visualize each endpoints` failure simulation status, added a bit of sugar icing
image

SeanFeldman:
This is what the diagram looks like now:

image

Since I've change things, @SimonCropp could you please do a last review? Thanks.

PR #2969, opened 2017-07-16
Fix incorrect SQS/S3 permissions

ahofman:
Fix incorrect "PutBucket" permission. 
Change case of S3 and SQS prefices to be consistent with what AWS uses (all lower case)


ahofman:

PR #2974, opened 2017-07-18
Expand licensing info

sergioc:
Cross referenced additional licensing approach to make it easier to understand that the registry is used for licensing the entire platform and there's different ways to import into the registry.

### DIFF
--- a/serviceinsight/license.md
+++ b/serviceinsight/license.md
@@ -14,7 +14,7 @@ related:
 
 There are a few options for adding a license to ServiceInsight.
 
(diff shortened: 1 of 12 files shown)


serviceinsight/license.md
```diff
@@ -12,6 +12,13 @@ related:
 - nservicebus/licensing
 ---
 
+There are a few options for adding a license to ServiceInsight.
+
+Note that the license file might allow running multiple products (NServiceBus Core, ServiceControl and ServiceInsight). Therefore the license file can also be loaded from other platform tools to achieve the same end result, as long as the licensing method imports the license into the registry.
```
SimonCropp: make this an include https://github.com/Particular/docs.particular.net#markdown-includes  since it is used twice

PR #2977, opened 2017-07-18
Timeout manager circuit breaker behavior

ramonsmits:
Based on https://github.com/Particular/NServiceBus/issues/4833

@timbussmann documentation of current behavior.


weralabaj:
@ramonsmits I'm not sure exactly what that means

> it will be actually about 30 seconds more

30 seconds more than what? I think we need to clarify that sentence :)

ramonsmits:
I agree with @timbussmann and shortened the note, that also resolved the remark by @weralabaj

PR #2979, opened 2017-07-19
Update FileBasedRouting sample

timbussmann:
* Updated to NServiceBus.FileBasedRouting 1.2
* Switched back transport to MSMQ: Learning transport doesn't make much sense in here as the sample demonstrates routing config of commands and events, whereas the "event routes" will be ignored as the learning transport provides native pub/sub.
* Switched back to InMemoryPersistence, as MSMQ requires a persistence which provides a timeout storage.


SimonCropp:
> Switched back transport to MSMQ

Good catch. and thanks for fixing my mistake

timbussmann:
> Good catch. and thanks for fixing my mistake

It was a non-obvious "mistake" anyway. I just noticed that I get repeated warnings in the log that the event routing is discarded because of native pub/sub. Seems like the warnings do their job quite well :)

PR #2983, opened 2017-07-21
peek interval delay and SAAS remark

ramonsmits:


transports/sql/queuepeek.md
```diff
@@ -8,13 +8,14 @@ redirects:
...
 
+The default peek interval delay is 1 second. The recommended range for this setting is between 100 milliseconds to 10 seconds. If a value higher than the maximum recommended settings is used, a warning message will be logged. While a value less than 100 milliseconds will put too much unnecessary stress on the database, a value of larger than 10 seconds should also be use with caution as it may result in messages backing up in the queue. 
 
 ## Delay setting configuration
 
 Use the following code:
 
 snippet: sqlserver-config-delay
+
+Increasing this value can be useful if you use a SQL SAAS solution and pay per transactions but messages send to this specific endpoint do not have a strict processing time window as each peek will result in a transaction.
```
MarcinHoppe: Are you referring specifically to Azure SQL Database or other offerings as well?

ramonsmits: @MarcinHoppe I was indeed referring to SQL Azure, but it seems that is not true. Seems I mixed up with Azure (Table) Storage.

PR #2985, opened 2017-07-21
Add an article on SQS transport topology.

ahofman:
I've used a diagram from draw.io simply out of convenience - no need to provide a CC unlike lucidchart. If this is problematic I'm happy to revisit.


SimonCropp:
re the diagram.

I think this would be something that is simple enough to be dont in mermaid https://github.com/Particular/docs.particular.net#mermaid

Just messing around i did this 
```
Participant Subscriber As Subscriber
Participant Publisher As Publisher
Participant Persistence As Persistence (S3 Bucket)
Subscriber ->> Publisher: Subscribe to Message1
Publisher ->> Persistence: Store "Subscriber wants Message1"
Note over Publisher: Publish(Message1)
Persistence ->> Publisher: Read "Subscriber wants Message1"
Publisher->> Subscriber : Send Message1
```
Which produced this 
image

ahofman:
I ended up re-doing the diagram with Visio - the simple "boxes and arrows" diagram conveys the topology a little clearer than a sequence diagram (IMO), plus the AWS icons look nice (although now I think of it, is there a branding issue here? I notice we don't use Azure icons in the Azure related docs - is this intentional?)

SimonCropp:
Shouldnt the download arrows go the other way?

SimonCropp:
And yeah consistency of diagrams is not something we are great at ATM

ahofman:
I had intended the arrows to convey the physical connection flow - i.e. the endpoint initiates all connections to SQS and S3 - rather than logical message flow. Maybe I'll add this statement to the docs to make it clearer.

PR #2987, opened 2017-07-24
WIP: build hacks

SimonCropp:


bording:
This shouldn't be needed now that I've changed the restore step of the build back to NuGet 4.1.0.

PR #2998, opened 2017-07-26
Metrics cleanup/review

udidahan:
Moved the "Messages received performance statistics" section underneath the "Processing time" section as it appears more related to it than to the "Critical Time" section it was under before.

I think some more clarity is in order, though.

### DIFF
--- a/nservicebus/operations/metrics.md
+++ b/nservicebus/operations/metrics.md
@@ -59,7 +59,7 @@ Critical time is the time between when a message is sent and when it is fully pr
 
 ### Queue length
 
-This metric tracks the **estimated** number of messages in the input queue of an endpoint.
+This metric tracks the **estimated** number of messages in the input queue of an endpoint. The reason why this value needs to be estimated is that when scaling out certain queuing environments, under high load, the number of messages actually in a given input queue would not reflect all the messages *in transit* to that queue, thus reporting much lower values. For this reason, a different approach is used - one based on *links* rather than queues.
 
 A _link_ is a communication channel between a sender of the message and its receiver. Each link is uniquely identified by some combination of destination address, message assembly, and the [host identifier](/nservicebus/hosting/override-hostid.md#host-identifier) of the sender. The exact composition of link identifiers depends on the transport properties and type of message being sent.
 
--- a/nservicebus/operations/metrics.md
+++ b/nservicebus/operations/metrics.md
@@ -25,7 +25,7 @@ partial: reporting
 
 ## Metrics captured
 
-NServiceBus and ServiceControl capture a number of different metrics about a running endpoint including the processing time, "critical time", and number of messages in each queue.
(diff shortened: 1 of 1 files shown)


SimonCropp:
@udidahan do u think "more clarity" should be in a new issue/PR. or would you prefer it included in this one?

udidahan:
I think that "more clarity" could result in changes to this PR, so might as well do it here. Feel free to change the name of the PR as well to better reflect the changes.

SimonCropp:
@Particular/nservicebus-maintainers thoughts on "more clarity for critical time"?

timbussmann:
hm I'm happy with the critical time section. But I feel that the "Messages received performance statistics" could be improved. Reading @udidahan issue description, I have a feeling this is also what he meant?

mikeminutillo:
Does it help to pull all of the receive stats up to the same level as the other metrics https://github.com/Particular/docs.particular.net/pull/3010

I wonder if we should divide these between local metrics that the endpoint calculates entirely on it's own and integration metrics that combine data from two endpoints. For instance, Processing Time and the Receive stats are all Local metrics. Queue Length and Critical Time are both Integration stats. 

Integration stats have the potential to be less accurate because we need to gather data from two sources and combine it. Latency and Clock Drift become problems.

SimonCropp:
@udidahan is this clarified to your satisfaction?

udidahan:
I've made some additional changes (improvements?). Would appreciate another review.

nservicebus/operations/metrics.md
```diff
@@ -10,61 +10,76 @@ related:
...
 
 snippet: Metrics-Enable
 
 partial: reporting
 
+
 ## Metrics captured
 
-NServiceBus and ServiceControl capture a number of different metrics about a running endpoint including the processing time, "critical time", and number of messages in each queue.
+NServiceBus and ServiceControl capture a number of different metrics about a running endpoint including the processing time, the number of messages in each queue (differentiating between those pulled from the queue, those processed successfully, and those which failed processing), as well as "critical time".
```
mikeminutillo: Should we be mentioning NServiceBus and ServiceControl or should we just focus on the Metrics plugin?

mikeminutillo: Should we itemize the metrics here? As we add more, this list will just get longer and longer and it's effectively repeated below anyway.

weralabaj: Since they're mentioned below, IMO not necessary.

weralabaj: IMO it's ok to give a bit bigger perspective, but if you think it may be confusing feel free to edit.

udidahan:
I've haven't had the time to see how to better resolve that middle section, but seeing as the change is already better than how the page currently is, I'm happy to have it merged as-is.

PR #3001, opened 2017-07-26
Metric reporting to Azure Application Insights

ramonsmits:
Includes a clone of the SLA violation logic from https://github.com/Particular/NServiceBus.Metrics.PerformanceCounters as https://github.com/Particular/NServiceBus.Metrics does not provide this metric. Also see https://github.com/Particular/NServiceBus.Metrics/issues/41


SimonCropp:
i cleaned up some stuff

 - added a R# settings file
 - made the code consistent with other samples
 - fixed some spelling errors
 - fixed the broken yaml
 - added a link to app insights

samples/application-insight/Metrics_1/Endpoint/Program.cs
```diff
@@ -0,0 +1,99 @@
...
+            probeCollector.Flush();
+        }
+    }
+
+    // Simulates busy (almost no delay) / quite time (10 seconds delay) in a sine wave.
+    static async Task LoadSimilator(IEndpointInstance endpointInstance, CancellationToken token)
+    {
+        try
+        {
+            for (var angle = 0;; angle++)
```
SimonCropp: i think i would prefer

```
var angle = 0;
while (true){
    angle++;
```

SeanFeldman:
@ramonsmits shouldn't this sample be under `/azure`?

SimonCropp:
@seanfarmar i am not sure. it doens really speak to any of our azure extensions. in this case it is more coincidental that we are targeting a logging service that is hosted by azure. eg if we also did a variant that targets https://stackify.com/ i would say this should be co located with that. if anything i think this sample should live in https://docs.particular.net/samples/logging/

SeanFeldman:
In that case I'd expect it to be with other sample(s) under `metrics` or `reporting` samples.

SimonCropp:
i am actually going to raise a PR to move https://docs.particular.net/samples/metrics/ under logging. it doesnt need its own dir

SeanFeldman:
👍 then this sample will join it as well.

SimonCropp:
we can rename "logging" in the menu to "logging and metrics"

ramonsmits:
@SimonCropp I don't know if I like the grouping underneath *logging*. I wouldn't like to see this listed at the same level as log4net, NLog, etc.

What about *monitoring* ? That would be a nice parent for everything related to metrics. We can move the current performance counter stuff underneath it too and also the [next sample current in next in line in the launchwave](https://github.com/Particular/LaunchWave.DevOpsAdvancedMonitoring/issues/88).

SeanFeldman:
@ramonsmits not sure why the snippets are not rendering 😞

SeanFeldman:
Snippets do not render - I suspect because `Metrics_1` is not registered as a component. @SimonCropp  is that it?

SimonCropp:
@ramonsmits @SeanFeldman https://github.com/Particular/docs.particular.net/pull/3001/commits/64a43567ea265c54eee9757e2b01ae4bf9db5b2b

samples/application-insights/sample.md
```diff
@@ -0,0 +1,87 @@
...
+
+The easiest way to setup graphs for metrics in Application Insights is to go to the `Metrics Explorer`, add a new graph selecting the custom metric of choice and then use the push pin symbol to add it to a dashboard.
+
+The following graphs show the following metrics:
+
+ * Maximum critical time & critical time grouped by instance id.
+ * Average processing time, not grouped.
+ * Minimum hours countdown until Service Level Agreement violation.
+
+<img alt="example graph metrics" src="example-graph-metrics.png" width="50%" />
```
ramonsmits: @Particular/docs-maintainers Is this allowed? The images have been made on high DPI devices. If this is not allowed, we must resize the image in half (or actually 1/4th) but that would make the image blurry on high dpi devices.

SeanFeldman:

samples/application-insights/sample.md
```diff
@@ -0,0 +1,82 @@
...
+
+snippet: configure-ai-instrumentation-key
+
+## Use NServiceBus.Metrics to capture metric values
+
+To capture the metric values generated by NServiceBus, a class called `ApplicationInsightsFeature` is added to the sample. It is responsible for enabling the NServiceBus metrics feature.
+
+snippet: enable-nsb-metrics
+
+It ensures that the telemetry probe is subscribed to the event stream. During this stage the `ApplicationInsightsProbeCollector` is configured and wired. Information such as *endpoint name* and *input queue* are passed to allow advanced data analysis in Application Insights.
```
DavidBoike: > It ensures

What is "it"?

samples/application-insights/sample.md
```diff
@@ -0,0 +1,82 @@
...
+The `ApplicationInsightsProbeCollector` is used to pass NServiceBus captured metric data to the Application Insights instance configured earlier. Information passed is added as telemetry context properties, which will be attached to every metric (NServiceBus signal probe) and event (NServiceBus duration probe) captured via the telemetry client. The telemetry client is used to communicate with Application Insights.
+
+snippet: telemetry-client
+
+When the `ApplicationInsightsProbeCollector` is registered with the NServiceBus metrics API, it registers observer callbacks to receive probe names, signals and durations. Signals are converted into Application Insights custom events. Durations are converted into Application Insights metrics.
+
+snippet: observers-registration
+
+
+## Setting up charts for metrics and events (signals and durations)
```
DavidBoike: Very long heading. "Creating charts" would probably do.

Also notice you use the word "graphs" in the next paragraph - be consistent. Based on my google search the terms overlap A LOT but "chart" is probably the more accurate term for what we're dealing with.

samples/application-insights/sample.md
```diff
@@ -0,0 +1,82 @@
...
+
+It ensures that the telemetry probe is subscribed to the event stream. During this stage the `ApplicationInsightsProbeCollector` is configured and wired. Information such as *endpoint name* and *input queue* are passed to allow advanced data analysis in Application Insights.  
+
+snippet: register-probe
+
+Events and metrics are buffered and the Application Insights telemetry client does this every minute. If the endpoint instance would gracefully stop it needs to ensure that any pending events are flushed to Application Insight.
+
+snippet: flush-probe
+
+## Send captured metric data to Application Insights
```
DavidBoike: Pretty long. Would "Sending data to Application Insights" be sufficient?

samples/application-insights/sample.md
DavidBoike: It would be preferable to use Markdown `Example Graph Metrics` and let the Docs Engine worry about presentation.

Why would you need to resize it in 1/4th? The DocsEngine should present them no wider than the column width.

You can also control width in Markdown, check out the example from [Publishing from Web Applications](https://docs.particular.net/nservicebus/hosting/publishing-from-web-applications#storage-driven-transport-topology):

    Storage-driven transport publishing topology

In that case it was done to make the "tall" image not take up too much vertical real-estate.

I would use an absolute width, not a percentage.

samples/application-insights/sample.md
```diff
@@ -0,0 +1,82 @@
...
+
+The following graphs show the following metrics:
+
+ * Maximum critical time & critical time grouped by instance id.
+ * Average processing time, not grouped.
+ * Minimum hours countdown until Service Level Agreement violation.
+
+<img alt="example graph metrics" src="example-graph-metrics.png" width="50%" />
+
+Similar for events, the following graphs shows all events grouped by event name. This graph can be furthered filtered down to e.g. show failures only.
```
DavidBoike: Can we avoid "e.g."?

This graph can be further filtered, for example, to show failures only.

samples/application-insights/sample.md
```diff
@@ -0,0 +1,82 @@
...
+
+<img alt="example graph events" src="example-graph-events.png" width="50%" />
```
DavidBoike: I'd do the same as the image above.

samples/application-insights/sample.md
```diff
@@ -0,0 +1,82 @@
...
+
+Refer to the Microsoft documentation for more information on [creating dashboards from the metrics explorer](https://docs.microsoft.com/en-us/azure/application-insights/app-insights-dashboards)
```
DavidBoike: Missing period at end

DavidBoike:
This sample seems like a great candidate for inclusion into our recurring Twitter campaign. Could you try to write 5-6 tweets that would promote this sample? If you were just a regular developer browsing Twitter, what would make you go "That sounds cool" and click to see what it was all about?

samples/application-insights/sample.md
SeanFeldman: Indeed, that works even better because on 4K these images now render great in preview. Before they were quite small when the window was not taking the whole screen.

samples/application-insights/sample.md
SeanFeldman: Love how it renders now.

samples/application-insights/sample.md
```diff
@@ -0,0 +1,82 @@
...
+ * Successful messages per second
+ * Average message processing time in milliseconds
+ * Maximum critical time in milliseconds
+
+For a detailed explanation of these metrics refer to the [metrics captured section in the metrics documentation](/nservicebus/operations/metrics.md#metrics-captured).
+
+
+## Prerequisites
+
+An Application Insights account and instrumentation key. The instrumentation key can be obtained under the Application Insights instance properties pane in the Azure portal.
```
SeanFeldman: > The instrumentation key can be obtained under the Application Insights instance properties pane in the Azure portal.

Agreed. Should be moved to the section it belongs to.

SeanFeldman: I'll add a note about learning transport.

samples/application-insights/sample.md
SeanFeldman: Wait. You don't know what "IT" is?
https://www.youtube.com/watch?v=PN8il_zxNWM

SeanFeldman: ok, ok, I'm being silly here 🙂

SeanFeldman:
@DavidBoike incorporated your comments

SeanFeldman:
Possible tweets:
1. See how your @NServiceBus solution is doing with @AppInsights
1. Seeing is believing - visualize @NServiceBus metrics with @AppInsights
1. Running on @Azure? Get @NServiceBus @AppInsights here (sample link)

samples/application-insights/sample.md
```diff
@@ -18,7 +18,9 @@ For a detailed explanation of these metrics refer to the [metrics captured secti
 
 ## Prerequisites
 
-An Application Insights account and instrumentation key. The instrumentation key can be obtained under the Application Insights instance properties pane in the Azure portal.
+To run this sample, an Application Insights account and instrumentation key are required.
+
+Note: for simplicity, the sample uses the Learning transport.
```
DavidBoike: Although the sample uses Azure Application Insights, the solution itself does not have to run on an Azure message transport. This example uses the [Learning Transport](/transports/learning/) but could be modified to run on any [transport](/transports/).

samples/application-insights/sample.md
```diff
@@ -36,21 +38,25 @@ By default the sample code loads this key from an environment variable called `A
 
 snippet: configure-ai-instrumentation-key
 
+The instrumentation key can be obtained under the Application Insights instance properties pane in the Azure portal.
```
DavidBoike: I looked in the portal and I have no idea how to find this for instances we have, and I didn't even have to create one. You should point to Azure documentation on how to create one and then also describe exactly the clicks to get to the value you need.

SeanFeldman: I'll have to push back on this. The sample is not focusing on teaching how to use AI. Documenting anything on the portal exactly to the clicks is insane. And it will break fast 😉

SeanFeldman:
@DavidBoike done

PR #3006, opened 2017-07-27
move metrics

SimonCropp:
Having a single dir under samples just for a single sample doesnt really make sense. and the line between logging and metrics is kind of blury. So this moves metrics sample under logging dir and changes "Logging" in the menu to be "Logging and Metrics"


SimonCropp:
ping @Particular/metrics-maintainers

PR #3017, opened 2017-07-31
fixed binding redirects

yvesgoeleven:
more storage binding redirect fixes


SimonCropp:
good catch. when the next `nuget.exe` is release this should not be a problem anymore

yvesgoeleven:
thanks @SimonCropp

PR #3028, opened 2017-08-03
Update the ASB SC adapter sample to use namespace hierarchy

SeanFeldman:
Connects to https://github.com/Particular/NServiceBus.AzureServiceBus/issues/593

Namespace hierarchy is the 4th SC unsupported transport feature.
Adding it to the sample to show that adapter can handle it as well.
Documentation for SC Adapter will be updated separately.

### DIFF
--- a/samples/servicecontrol/adapter-asb-multi-namespace/sample.md
+++ b/samples/servicecontrol/adapter-asb-multi-namespace/sample.md
@@ -112,7 +112,7 @@ The following code configures the adapter to match advanced transport features e
  * [Secure connection strings](/transports/azure-service-bus/securing-connection-strings.md).
  * [Customized brokered message creation](/transports/azure-service-bus/brokered-message-creation.md) using `Stream`.
  * [Multiple namespace](/transports/azure-service-bus/multiple-namespaces-support.md#round-robin-namespace-partitioning).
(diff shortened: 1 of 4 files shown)


SeanFeldman:
@SzymonPobiega is on vacation - @Particular/servicecontrol-maintainers could you review? Thanks.

WilliamBZA:
I think @boblangley would be best to review this given his proximity to ASB

SeanFeldman:
The intent was more to see if this is aligned with the idea of the adapter, which I'm sure it is. ASB wise it's working (I've tested it 🙂). Thanks.

samples/servicecontrol/adapter-asb-multi-namespace/sample.md
```diff
@@ -111,6 +112,7 @@ The following code configures the adapter to match advanced transport features e
  * [Secure connection strings](/transports/azure-service-bus/securing-connection-strings.md).
  * [Customized brokered message creation](/transports/azure-service-bus/brokered-message-creation.md) using `Stream`.
  * [Multiple namespace](/transports/azure-service-bus/multiple-namespaces-support.md#round-robin-namespace-partitioning).
+  * [Namespace hierary](/transports/azure-service-bus/namespace-hierarchy.md)
```
SimonCropp: there seems to be an extra indent here. is that on purpose?

PR #3037, opened 2017-08-06
Wording

udidahan:
Primary changes to the opening paragraph plus some other minor changes later.

### DIFF
--- a/samples/servicecontrol/adapter-rabbitmq-different-topologies/sample.md
+++ b/samples/servicecontrol/adapter-rabbitmq-different-topologies/sample.md
@@ -10,7 +10,7 @@ related:
 ---
 
 
-This sample shows how to configure ServiceControl to monitor endpoints and retry messages when using the RabbitMQ direct routing topology. The direct routing topology is used by endpoints that are using NServiceBus Version 6. ServiceControl is built on NServiceBus Version 5 and can only use the conventional routing topology. The ServiceControl transport adapter is required to bridge both routing topologies.
+This sample shows how to configure ServiceControl to monitor endpoints and retry messages when using the RabbitMQ direct routing topology, the default topology of endpoints using RabbitMQ on NServiceBus Version 6. However, since ServiceControl is built on NServiceBus Version 5 and can only use the conventional routing topology, a kind of "transport adapter" is required to bridge the routing topologies. 
 
 
 ## Prerequisistes
@@ -57,7 +57,7 @@ The code base consists of three projects.
 
 ### Sales Endpoint
 
-The Sales project contains endpoint that simulates execution of business process by exchanging messages with the Sales endpoint. It includes message processing failure simulation mode (toggled by pressing `f`) which can be used to generate failed messages that demonstrate message retry functionality. The Sales endpoint uses the RabbitMQ direct routing topology and requires an adapter in order to communicate with ServiceControl.
+The Sales project contains an endpoint that simulates the execution of a business process by exchanging messages with the Sales endpoint. It includes a message processing failure simulation mode (toggled by pressing `f`) which can be used to generate failed messages that demonstrate message retry functionality. The Sales endpoint uses the RabbitMQ direct routing topology and requires an adapter in order to communicate with ServiceControl.
 
 The following code configures the Sales endpoint to communicate with the adapter:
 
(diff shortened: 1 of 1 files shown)


samples/servicecontrol/adapter-rabbitmq-different-topologies/sample.md
```diff
@@ -33,12 +33,12 @@ NOTE: In order to connect to a different RabbitMQ broker, ensure all connection
 
 ## Running the project
 
- 1. Start the projects: Adapter, Sales and Shipping (right-click on the project, select the `Debug > Start new instance` option). Make sure adapter starts first because on start-up it creates a queue that is used for heartbeats.
+ 1. Start the projects: Adapter, Sales and Shipping (right-click on the project, select the `Debug > Start new instance` option). Make sure the adapter starts first because on start-up it creates a queue that is used for heartbeats.
```
SimonCropp: minor: but i prefer "Ensure" over "make sure"

SimonCropp: > Start the projects: Adapter, Sales and Shipping (right-click on the project, select the `Debug > Start new instance` option).

this doesnt really make sense. if we want them to runn all project why confuse them with "right click". if they have downloaded the sln from docs all those will be default startup projects and we should just be able to tell them to "run the solution"

danielmarbach:
Ping @Particular/rabbitmq-transport-maintainers

samples/servicecontrol/adapter-rabbitmq-different-topologies/sample.md
mauroservienti: and it's not required, it comes from other samples IIRC that say the same thing. Heartbeats are sent from endpoints to SC Q, or the adapter Q, if that Q doesn't exist a warning is raised at the sender endpoint, it doesn't fail. As soon as the adapter starts Qs are created and the warning goes away. I'd entirely remove that text, replacing it with "press F5".

samples/servicecontrol/adapter-rabbitmq-different-topologies/sample.md
```diff
@@ -10,7 +10,7 @@ related:
 ---
 
 
-This sample shows how to configure ServiceControl to monitor endpoints and retry messages when using the RabbitMQ direct routing topology. The direct routing topology is used by endpoints that are using NServiceBus Version 6. ServiceControl is built on NServiceBus Version 5 and can only use the conventional routing topology. The ServiceControl transport adapter is required to bridge both routing topologies.
+This sample shows how to configure ServiceControl to monitor endpoints and retry messages when using the RabbitMQ direct routing topology, the default topology of endpoints using RabbitMQ on NServiceBus Version 6. However, since ServiceControl is built on NServiceBus Version 5 and can only use the conventional routing topology, a kind of "transport adapter" is required to bridge the routing topologies.
```
mauroservienti: > This sample shows how to configure ServiceControl to monitor endpoints and retry messages when using the RabbitMQ direct routing topology, the default topology of endpoints using RabbitMQ on NServiceBus Version 6.

@Particular/rabbitmq-transport-maintainers can you confirm that `DirectRoutingTopology` is the default in V6?
Reading [this](https://docs.particular.net/transports/rabbitmq/routing-topology#conventional-routing-topology) it's the other way round.

bording: @mauroservienti The direct routing topology is **not** the default for any version of the RabbitMQ transport. 

The scenario being tested here is when someone has chosen to use the direct routing topology instead of the default because there is no way to configure ServiceControl to use something other than the default. It has nothing to do with ServiceControl being built on NSB 5, either. The direct routing topology is also available on NSB 5. The problem is that there's no way to configure transport settings inside of ServiceControl.

mauroservienti:
Thanks @bording I'll adjust the wording based on your comments.

samples/servicecontrol/adapter-rabbitmq-different-topologies/sample.md
weralabaj: I believe the issue is that with "just hit F5" you can't guarantee the order in which projects are started, which may look as the sample is not working correctly.

weralabaj: So I'd leave it, if you think that doesn't make sense then just edit on master.

samples/servicecontrol/adapter-rabbitmq-different-topologies/sample.md
```diff
@@ -33,12 +33,12 @@ NOTE: In order to connect to a different RabbitMQ broker, ensure all connection
 
 ## Running the project
 
- 1. Start the projects: Adapter, Sales and Shipping (right-click on the project, select the `Debug > Start new instance` option). Make sure adapter starts first because on start-up it creates a queue that is used for heartbeats.
+ 1. Start the projects: Adapter, Sales and Shipping (right-click on the project, select the `Debug > Start new instance` option). Ensure the adapter starts first because on start-up it creates a queue that is used for heartbeats.
```
mauroservienti: What do we do about this sentence? As @SimonCropp pointed out samples should be <kbd>F5</kbd> ready. And in my tests the above is not required: if the startup order is not correct endpoints will emit warnings because heartbeats cannot be delivered, all warnings will go away as soon as the adapter starts and Qs are created.

udidahan:
Although I was the one that opened this PR, now that people who actually know what they're talking about are involved, I'll let you all drive this completion.

MarcinHoppe:
I think this is actually good to go when the build is green.

/cc @Particular/docs-maintainers

PR #3040, opened 2017-08-07
Moved doco on how the transport adapter works to the doco page

SzymonPobiega:
Also added note that SC events are not supported


SzymonPobiega:
@MarcinHoppe I've added a paragraph stating that the adapter does not support the integration events.

PR #3041, opened 2017-08-07
RavenDB Persistence:  note reminding to do dispose the user created document store

### LINKED ISSUE #3035: Feedback: 'RavenDB Persistence'
Feedback for 'RavenDB Persistence' https://docs.particular.net/persistence/ravendb/ and https://docs.particular.net/samples/ravendb/simple/

ping @Particular/ravendb-persistence-maintainers

should we be disposing of `DocumentStore` on endpoint shutdown? or add a note reminding people to do so?

mauroservienti:
Connects to #3035 
Fix #3035

### DIFF
--- a/persistence/ravendb/index.md
+++ b/persistence/ravendb/index.md
@@ -31,7 +31,8 @@ This optionally allows customization of the document session that is created for
 
 The session that is created is then made available.
 
-NOTE: When the RavenDB `DocumentStore` is created by the user at endpoint configuration time it's important to dispose it, by calling the `Dispose()` method, before shutting down the endpoint process.
+include: raven-dispose-warning
+
 
 ### Using in a Handler
 
--- a/persistence/ravendb/index.md
+++ b/persistence/ravendb/index.md
@@ -31,6 +31,7 @@ This optionally allows customization of the document session that is created for
 
 The session that is created is then made available.
 
+NOTE: When the RavenDB `DocumentStore` is created by the user at endpoint configuration time it's important to dispose it, by calling the `Dispose()` method, before shutting down the endpoint process.
 
 ### Using in a Handler
 
--- a/persistence/ravendb/raven-dispose-warning.include.md
+++ b/persistence/ravendb/raven-dispose-warning.include.md
@@ -1,0 +1,1 @@
+NOTE: When the RavenDB `DocumentStore` is created by the user at endpoint configuration time it's important to dispose it, by calling the `Dispose()` method, before shutting down the endpoint process.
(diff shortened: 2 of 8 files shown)


DavidBoike:
Looks good, although it could be an include. Should it be referenced from the [Custom Saga Finder sample](https://docs.particular.net/samples/saga/ravendb-custom-sagafinder/) sample too?

@SimonCropp wondering what your motivation in #3035 was though. Doesn't this fall into the bucket of reminding developers to dispose things that are disposable? Should we really be taking responsibility for that?

SimonCropp:
> Doesn't this fall into the bucket of reminding developers to dispose things that are disposable

yeah i dont really like that either. but AFAIK everywhere else we have disposable stuff we put it in a using or clean it up. in the raven sample we should definitely clean it up. i think it is fine not to explain the obvious that is outside the scope of the NSB domain, but this is a grey area. eg here https://docs.particular.net/persistence/ravendb/#shared-session can i scope the store in a using?  that would seem to be the correct assumption to make. when in reality they need to clean it up on endpoint shutdown

mauroservienti:
Can an include be referenced by a document not in the same folder? The sample.md is in a different folder and needs the same include

SimonCropp:
includes are global and can be located anywhere and included in any file

SimonCropp:
and u can ignore the build error. it is fixed on master

SimonCropp:
and yes i think it should be an include. but i think the samples should correctly dispose of the document store

DavidBoike:
I think we should dispose the DocStore too. Just toss it in a using after Initialize that ends after the endpoint is stopped on the relevant Program.cs files.

DavidBoike:
In fact, if we do dispose the DocStore, I'm not even sure we really need to have the info boxes to tell developers to dispose disposable things. Putting it in a using would get the message across effectively enough.

SimonCropp:
@DavidBoike i think for the samples that makes sense. but, for example, i could not work out how to have a using that conveys the same thing here https://docs.particular.net/persistence/ravendb/connection#external-store-at-initialization-for-a-specific-persister

DavidBoike:
Hmmm, yeah that's a problem. Guess a NOTE is better than the gymnastics to include all the disposes in that snippet.

SimonCropp:
but either way i think the snippet here https://docs.particular.net/persistence/ravendb/connection#external-store-at-initialization-for-a-specific-persister needs tweeking. since from the usage i see there i would assume that the underlying persister will handle the dispose for me. at least that is what having the construction of the store scoped inside the delegate implies

PR #3042, opened 2017-08-07
SQS Ops Scripts

danielmarbach:
Connects to https://github.com/Particular/PlatformDevelopment/issues/1108


danielmarbach:
@Particular/aws-maintainers here is the current WIP. Early feedback is already welcome.

Snippets/Sqs/Sqs_All/QueueDeletion/Tests.cs
```diff
@@ -0,0 +1,47 @@
...
+
+    [TestFixture]
+    [Explicit]
+    public class Tests
+    {
+
+        [Test]
+        public void DeleteAllQueues()
+        {
+            QueueDeletionUtils.DeleteAllQueues().GetAwaiter().GetResult();
```
SimonCropp: i dont think `.GetAwaiter().GetResult();` is required here?

Snippets/Sqs/Sqs_All/Sqs_All.csproj
```diff
@@ -0,0 +1,17 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net462</TargetFramework>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="ApprovalTests" Version="3.*" />
```
SimonCropp: do u need ApprovalTests?

Snippets/Sqs/Sqs_All/Sqs_All.csproj
```diff
@@ -0,0 +1,17 @@
...
+    <PackageReference Include="Mono.Cecil" Version="0.*" />
```
SimonCropp: same goes for cecil

Snippets/Sqs/Sqs_All/Sqs_All.csproj
```diff
@@ -0,0 +1,17 @@
...
+    <PackageReference Include="NServiceBus" Version="6.*" />
+    <PackageReference Include="NServiceBus.AmazonSQS" Version="3.0.0-alpharelease0057" />
+    <PackageReference Include="NUnit" Version="3.*" />
+    <PackageReference Include="System.Management.Automation" Version="6.*" />
+    <ProjectReference Include="..\..\Common\Common.csproj" />
+    <Reference Include="System.Messaging" />
```
SimonCropp: suspect this got pulled over from msmq?

danielmarbach: yes, I planned to remove stuff I don't need. Not done yet

Snippets/Sqs/Sqs_All/QueueDeletion/Tests.cs
danielmarbach: No, need to fix this

SimonCropp:
looks good so far

Snippets/Sqs/Sqs_All/ErrorQueue/ErrorQueue.cs
```diff
@@ -0,0 +1,102 @@
...
+                {
+                    QueueUrl = queueUrl,
+                    WaitTimeSeconds = 1,
+                    MaxNumberOfMessages = 10,
+                })
+                .ConfigureAwait(false);
+
+            if (receivedMessages.Messages.Count == 0)
+            {
+                return;
```
ahofman: Is this safe - are we sure that if we long-poll for 1 second that we will definitely not miss any messages that may be still in the queue? Would it be safer to long-poll for 20 seconds?

Snippets/Sqs/Sqs_All/ErrorQueue/ErrorQueue.cs
```diff
@@ -0,0 +1,102 @@
...
+                    continue;
+                }
+
+                var failedQueueName = ReadFailedQueueHeader(message);
+                var failedQueueUrlResponse = await client.GetQueueUrlAsync(failedQueueName)
+                    .ConfigureAwait(false);
+                var failedQueueUrl = failedQueueUrlResponse.QueueUrl;
+                // Large message don't need to be handled separately since the S3BodyKey is preserved
+                await client.SendMessageAsync(new SendMessageRequest(failedQueueUrl, message.Body)) // what to do with the attributes?
+                    .ConfigureAwait(false);
```
ahofman: When the pump moves poison messages to the error queue, it doesn't move SQS message attributes with it, I wouldn't worry about the attributes for now

Snippets/Sqs/Sqs_All/QueueCreation/QueueCreationUtils.cs
```diff
@@ -0,0 +1,38 @@
...
+                    {
+                        QueueName = QueueNameHelper.GetSqsQueueName(queueName)
+                    };
+                    var value = Convert.ToInt32((maxTimeToLive ?? DefaultTimeToLive).TotalSeconds).ToString();
+                    sqsRequest.Attributes.Add(QueueAttributeName.MessageRetentionPeriod, value);
+                    await client.CreateQueueAsync(sqsRequest)
+                        .ConfigureAwait(false);
+                }
+            }
+            catch (QueueNameExistsException)
```
ahofman: When a queue is created with the same name as another queue that was deleted less than 60sec ago, `CreateQueueAsync` throws `QueueDeletedRecentlyException` - the user would just need to wait 60 sec and try again. Do we want to add this handling here?

Snippets/Sqs/Sqs_All/QueueCreation/QueueCreation.ps1
```diff
@@ -0,0 +1,127 @@
...
+)
+
+# main queue
+CreateQueue -QueueName $EndpointName -MaxTimeToLive $MaxTimeToLive
+
+# timeout queue
+CreateQueue -QueueName "$EndpointName.timeouts" -MaxTimeToLive $MaxTimeToLive
+
+# timeout dispatcher queue
+CreateQueue -QueueName "$EndpointName.timeoutsdispatcher" -MaxTimeToLive $MaxTimeToLive
```
ahofman: Important to be aware that SQS queue names are case-sensitive (unlike most other transports) - is this the same case that endpoints expect?

Snippets/Sqs/Sqs_All/QueueDeletion/QueueDeletionUtils.cs
```diff
@@ -0,0 +1,99 @@
...
+            using (var client = ClientFactory.CreateSqsClient())
+            {
+                await DeleteQueues(client)
+                    .ConfigureAwait(false);
+            }
+        }
+
+        static async Task DeleteQueues(IAmazonSQS client)
+        {
+            var queuesToDelete = await client.ListQueuesAsync("")
```
ahofman: This deletes every queue in the region, may be a bit heavy-handed :) The queue name prefix is typically used to isolate application environments in the same region (e.g., "DEV", "STAGING", "PROD", etc). 

Perhaps allow the queue name prefix to be passed in as a parameter?

(Unfortunately this suggestion cross-cuts pretty much everything in this PR - sorry :P)

Snippets/Sqs/Sqs_All/QueueCreation/QueueCreation.ps1
danielmarbach: Good point. No the core uses

https://github.com/Particular/NServiceBus/blob/926db772c885d729591352713809d40267d18bc3/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutManager.cs#L66 and

https://github.com/Particular/NServiceBus/blob/926db772c885d729591352713809d40267d18bc3/src/NServiceBus.Core/DelayedDelivery/TimeoutManager/TimeoutManager.cs#L86

Will change. Good catch

danielmarbach: I added a test which spins up endpoints

Snippets/Sqs/Sqs_All/QueueDeletion/QueueDeletionUtils.cs
danielmarbach: I don't think it cross cuts everything. Most other scripts just accept the queue name or the endpoint name. Do you agree that it is enough to just add a prefix param to the deletion scripts?

Snippets/Sqs/Sqs_All/QueueCreation/QueueCreationUtils.cs
danielmarbach: Not sure if it is required. I think for the queue creation it is OK to bomb in that case since we cannot operate with the queues not present but we can leave with queues still there in the worst case right?

danielmarbach: Never mind I get what you are saying. Do we also need to deal with the truncation?

ahofman: I'm not sure about truncation - I added that to the transport as an internal-only configuration option solely to get the ATs working, I'm not sure if it's a good way to handle real systems. To date I have just been relying on users choosing endpoint names that are short enough - this gives a very straightforward experience when a user goes looking for their queue in the SQS portal. I'm a little concerned that the pretruncation will be more confusing for new users and will give a worse experience compared to just getting an exception saying "queue name is too long", but I'm open to suggestions here.

In any case I don't think we should add the pretruncation support here because it's not possible for users to configure it in their endpoints at this time.

Snippets/Sqs/Sqs_All/QueueCreation/QueueCreationUtils.cs
```diff
@@ -0,0 +1,37 @@
...
+
+    #region sqs-create-queues
+
+    public static class QueueCreationUtils
+    {
+        static TimeSpan DefaultTimeToLive = TimeSpan.FromDays(4);
+
+        public static async Task CreateQueue(string queueName, TimeSpan? maxTimeToLive = null, string queueNamePrefix = null, bool preTruncateQueueNames = false)
+        {
+            try
```
SimonCropp: should this try be narrowed to only surround the CreateQueueAsync?

Snippets/Sqs/Sqs_All/ErrorQueue/ErrorQueue.cs
```diff
@@ -0,0 +1,100 @@
...
+                    MaxNumberOfMessages = 10,
+                })
+                .ConfigureAwait(false);
+
+            if (receivedMessages.Messages.Count == 0)
+            {
+                return;
+            }
+
+            foreach (var message in receivedMessages.Messages)
```
SimonCropp: minor suggestion. i find this more readable 

            var foundMessage = receivedMessages.Messages
                .SingleOrDefault(message => message.MessageId != messageId);

            if (foundMessage != null)
            {
                var failedQueueName = ReadFailedQueueHeader(foundMessage);
                var failedQueueUrlResponse = await client.GetQueueUrlAsync(failedQueueName)
                    .ConfigureAwait(false);
                var failedQueueUrl = failedQueueUrlResponse.QueueUrl;
                // Large message don't need to be handled separately since the S3BodyKey is preserved
                await client.SendMessageAsync(new SendMessageRequest(failedQueueUrl, foundMessage.Body)) // what to do with the attributes?
                    .ConfigureAwait(false);
                await client.DeleteMessageAsync(queueUrl, foundMessage.ReceiptHandle)
                    .ConfigureAwait(false);
                return;
            }

Snippets/Sqs/Sqs_All/ErrorQueue/ErrorQueue.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            await InspectMessagesUntilFound(client, messageId, queueUrl)
+                .ConfigureAwait(false);
+        }
+
+        static string ReadFailedQueueHeader(Message message)
+        {
+            var headers = ExtractHeaders(message);
+            var queueName = headers.Single(x => x.Key == "NServiceBus.FailedQ").Value;
+            return QueueNameHelper.GetSqsQueueName($"{queueName}");
```
SimonCropp: shouldnt `$"{queueName}"` be `queueName`?

SimonCropp:
overall very nice work :)

Snippets/Sqs/Sqs_All/QueueCreation/QueueCreationUtils.cs
SimonCropp: i generally dont like adding truncation. they are hard for the user to reason about. IMO it is not something we should be adding abstraction logic to

danielmarbach:
Ok so I incorporate all the feedback. From a code perspective, this is done. Will now start covering all the snippets in the markdown.

danielmarbach:
@Particular/aws-maintainers this is done from my perspective.

Snippets/Sqs/Sqs_All/ErrorQueue/ErrorQueue.cs
danielmarbach: I changed the long poll to 20 secs

Snippets/Sqs/Sqs_All/ErrorQueue/ErrorQueue.cs
```diff
@@ -0,0 +1,107 @@
...
+                // Large message don't need to be handled separately since the S3BodyKey is preserved
+                await client.SendMessageAsync(new SendMessageRequest(failedQueueUrl, foundMessage.Body)) // what to do with the attributes?
+                    .ConfigureAwait(false);
+                await client.DeleteMessageAsync(queueUrl, foundMessage.ReceiptHandle)
+                    .ConfigureAwait(false);
+
+                return;
+            }
+
+            await InspectMessagesUntilFound(client, messageId, queueUrl)
```
ahofman: This could be difficult to use if the user wants to return many messages to the source queue because this receives other messages in the error queue and marks them as invisible. The user may get in a loop of run script for 1 message -> wait for the visibility timeout to expire (20sec?) -> run script for next message -> wait for timeout -> etc, which would really hurt for 00's of messages. Is this something to worry about?

Snippets/Sqs/Sqs_All/QueueNameHelper.cs
```diff
@@ -29,8 +29,8 @@ public static string GetSqsQueueName(string destination, string queueNamePrefix
 
             if (s.Length > 80)
             {
-                throw new Exception(
-                    $"Address {destination} with configured prefix {queueNamePrefix} is longer than 80 characters and therefore cannot be used to create an SQS queue. Use a shorter queue name.");
+                throw new ArgumentException(
```
danielmarbach: These changes should be ported to the PowerShell version as well (currently copy pasted into each PS file)

Snippets/Sqs/Sqs_All/QueueCreation/CreateEndpointQueues.cs
```diff
@@ -24,18 +24,21 @@ static class CreateEndpointQueues
 
         public static async Task CreateQueuesForEndpoint(string endpointName, TimeSpan? maxTimeToLive = null, string queueNamePrefix = null, bool includeRetries = false)
         {
+            // timeout dispatcher queue
+            // This queue is created first because it has the longest name. 
+            // If the endpointName and queueNamePrefix are too long this call will throw and no queues will be created. 
+            // In this event, a shorter value for endpointName or queueNamePrefix should be used.
+            await QueueCreationUtils.CreateQueue($"{endpointName}.TimeoutsDispatcher", maxTimeToLive, queueNamePrefix)
```
danielmarbach: 👍

Snippets/Sqs/Sqs_All/ErrorQueue/ErrorQueue.cs
danielmarbach: How would you mitigate this? I originally had the wait timeout pretty low but that has other implications

SimonCropp: I think a warning in the .md will suffice

ahofman: I'd suggest changing the method to accept a collection of message ids instead of just one.

At a glance it looks like the [RabbitMQ scripts](https://docs.particular.net/transports/rabbitmq/operations-scripting#return-message-to-source-queue-the-retry-helper-methods) have a similar issue.

Are these scripts intended to be a starting point / reference / example only or are they meant to be bullet-proof for production usage as-is? If the former I'd say leave it and add a warning per @SimonCropp suggestion.

Snippets/Sqs/Sqs_All/QueueNameHelper.cs
ahofman: Done

Snippets/Sqs/Sqs_All/ErrorQueue/ErrorQueue.cs
SimonCropp: @ahofman they are intended as a starting point. how "bullet-proof" we make them depends on how much effort it takes and how much more difficult it makes the snippets to understand/explain. in the long run i would like to see this stuff packaged up into helper libs/nugets. but that is outside the scope of the current work

ahofman: Cool good to know :) I've added a warning to the .md.

Snippets/Sqs/Sqs_All/QueueCreation/QueueCreation.ps1
```diff
@@ -125,8 +128,8 @@ Add-Type @'
 
             if (s.Length > 80)
             {
-                throw new Exception(
-                    string.Format("Address {0} with configured prefix {1} is longer than 80 characters and therefore cannot be used to create an SQS queue. Use a shorter queue name.", destination, queueNamePrefix));
+                throw new ArgumentException(
```
danielmarbach: This has to be changed in all ps files :(

danielmarbach:
@SimonCropp  https://github.com/Particular/docs.particular.net/pull/3042#discussion_r132893704 has not been addressed. Will correct it on master tomorrow

PR #3053, opened 2017-08-10
Treat powershell as text

danielmarbach:
@bording @Scooletz I need your gitfoo skills to verify this


danielmarbach:
Repostandards seems to be fine https://github.com/Particular/RepoStandards/blob/master/.gitattributes

PR #3060, opened 2017-08-11
Add warning about minimum version of RabbitMQ required

bording:
Closes https://github.com/Particular/NServiceBus.RabbitMQ/issues/365

### DIFF
--- a/transports/rabbitmq/index.md
+++ b/transports/rabbitmq/index.md
@@ -14,7 +14,7 @@ tags:
 
 Provides support for sending messages over [RabbitMQ](http://www.rabbitmq.com/) using the [RabbitMQ .NET Client](https://www.nuget.org/packages/RabbitMQ.Client/).
 
-WARNING: The transport is not compatible with versions of the RabbitMQ broker earlier than 3.4.0.
+WARNING: The transport is not compatible with RabbitMQ broker Versions 3.3.X and below.
 
 
 ## Configuring the endpoint
--- a/transports/rabbitmq/index.md
+++ b/transports/rabbitmq/index.md
@@ -14,6 +14,8 @@ tags:
 
 Provides support for sending messages over [RabbitMQ](http://www.rabbitmq.com/) using the [RabbitMQ .NET Client](https://www.nuget.org/packages/RabbitMQ.Client/).
 
+WARNING: The transport is not compatible with versions of the RabbitMQ broker earlier than 3.4.0.
+
 
 ## Configuring the endpoint
 


transports/rabbitmq/index.md
```diff
@@ -14,6 +14,8 @@ tags:
 
 Provides support for sending messages over [RabbitMQ](http://www.rabbitmq.com/) using the [RabbitMQ .NET Client](https://www.nuget.org/packages/RabbitMQ.Client/).
 
+WARNING: The transport is not compatible with versions of the RabbitMQ broker earlier than 3.4.0.
```
SimonCropp: > earlier than 3.4.0.

can u make this consistent with https://github.com/Particular/docs.particular.net#version-language

adamralph: @SimonCropp I wanted to edit this, but I'm struggling to come up with the right language based on those guidelines. Would it be necessary to flip the direction of the constraint? E.g.

> The transport is only compatible with Versions 3.4.0 and above of the RabbitMQ broker.

SimonCropp: or

>  The transport is not compatible with RabbitMQ broker Versions 3.3 and below.

adamralph: The problem with that is that it's not explicit that 3.3 includes 3.3.1, 3.3.2, etc. It is implied, but I think some readers will miss the implication.

SimonCropp: > The transport is not compatible with RabbitMQ broker Versions 3.3.X and below.

adamralph: Yeah, I guess that could work.

So, putting two options forward:

> The transport is not compatible with RabbitMQ broker Versions 3.3.X and below.

or

> The transport is only compatible with RabbitMQ broker Versions 3.4.0 and above.

@SimonCropp @Particular/rabbitmq-transport-maintainers @Particular/docs-maintainers any preference?

WojcikMike: I prefer: 
> The transport is not compatible with RabbitMQ broker Versions 3.3.X and below.

Mostly because this is what we tested. It's the fact that we will not work with versions lower than 3.4.0.

bording:
I still prefer my original sentence over what's here now. I don't care for the random Capitalization of "Versions", and in general find the recommended language to be grammatically awkward.

However, given that I'm not going to try and challenge those conventions at this point, 👍 for merging as it currently is.

SimonCropp:
@bording 

> I don't care for the random Capitalization of "Versions", and in general find the recommended language to be grammatically awkward.

re "Capitalization of Versions". yeah that has never felt right to me either. but years ago we threw it out to a vote and that won.

re "recommended language to be grammatically awkward"  i get that, and i agree, but it is also unambiguous. the problem with " earlier than 3.4.0." is some people ended up not understanding if it is inclusive or not. if you can think of a way of phrasing it that is unambiguous and less awkward that would be great.

Given we are reasonably consistent regarding version, and it would be easy to move them all, i am more than happy to have the conventions challenged.

PR #3071, opened 2017-08-14
update metrics

SimonCropp:
@Particular/metrics-maintainers cant work out what the fix is for this. is there an upgrade guide?


SimonCropp:
ignore the rabbit error. that is fixed on master

SimonCropp:
given this could effect users, is there an upgrade guide?

dvdstelt:
@SimonCropp I don't see any upgrade guide for any minor version? And the tooltip provides all the information needed, as far as I'm concerned? It's just that doco wasn't properly updated. I'm also not too sure about the warning, but at least now the build (should) work.

dvdstelt:
Yes, build just succeeded, only error left is Rabbit error.

SimonCropp:
@dvdstelt thanks

PR #3084, opened 2017-08-16
netcore sample proposal

SimonCropp:
So we cant use `<TargetFrameworks>` to muti-target both 461 and netcoreapp2.0. since that means people would need the netcore2.0 sdk installed to compile, and hence run, 461. i think it is too early to force that. So this is a temporary solution until a time we can say "you have to have netcoreapp installed"

 * create a new `ProjectName.NetCore.csproj` that includes the main project but overwrites the `<targetframeworks>` to `netcoreapp2.0`
 * duplicate the sample sln and include the new project instead of the old

note the above steps could be initially scripted. will also need to verify it works on VS 2017 update 2

### DIFF
--- a/samples/serializers/bond/Bond_1.3/BondSample.NetCore.sln
+++ b/samples/serializers/bond/Bond_1.3/BondSample.NetCore.sln
@@ -1,0 +1,19 @@
+
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 15
+MinimumVisualStudioVersion = 15.0.26430.04
+MinimumVisualStudioVersion = 15.0.26430.04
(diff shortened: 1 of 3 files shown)


SimonCropp:
@DavidBoike yeah as part of this we can go with `TargetFramework` instead of `TargetFrameworks`

and my plan was to once of script (and hack+fix what fails) this for all samples that require netcore. i do not think it is possible it possible to solve in a way that doesnt have many edge cases. and given it is a temp thing i would prefer to avoid engine hacks for it

MarcinHoppe:
@SimonCropp We talked about it with @andreasohlund and we think this is important enough to work on it right away.

Are you going to do it across the board (where possible)?

/cc @Particular/platform-dev-squad

PR #3090, opened 2017-08-16
Add support for inherited events in custom routing sample

boblangley:
This PR adds support for automatic registration when handling a base type of an event that is published by another endpoint.


SimonCropp:
is this not required for core7?

boblangley:
Nope, I will update the PR. If you see this @SzymonPobiega I would still appreciate the review of the change. I did test it, but the sample doesn't utilize it.

SimonCropp:
should this actually be in doco and linked as a note from the sample?

boblangley:
>should this actually be in doco and linked as a note from the sample?

@SimonCropp Sorry, I don't understand.

I did update the core 7 sample, and changed the note about registration to indicate inherited event types will be added automatically.

PR #3091, opened 2017-08-17
add nuget doco

SimonCropp:


README.md
```diff
@@ -777,13 +777,89 @@ Use `var` everywhere.
...
+NServiceBus (>= 6.0.0 && < 7.0.0)
+RabbitMQ.Client (>= 4.1.0 && < 5.1.0)
+```
+
+NuGet will then resolve the lowest within those ranges. This make it more difficult to smoke test new versions of those dependencies using samples. As such, for all dependencies that are important to use the latest, the extra dependencies are explicitly included with wildcards.
+
+```
+<PackageReference Include="NServiceBus" Version="6.*" />
+<PackageReference Include="NServiceBus.RabbitMQ" Version="4.*" />
+<PackageReference Include="RabbitMQ.Client" Version="5.*" />
```
SeanFeldman: I would really like to see a `Label` or `Description` attribute on `PackageReference` with a comment that this is needed to force a specific version...

SimonCropp: i would prefer that was captured through code reviews. adding a comment to that stuff is prob several thousand lines to touch

SeanFeldman: Code review is point in time.
Much later it becomes much harder to understand.

SeanFeldman: Also, this would be done _moving forward_. Not retroactively.

SimonCropp: in that case can you raise that as plat dev since this approach to managing nuget dependencies spans all repos. eg it is used in many test projects

PR #3094, opened 2017-08-17
Messages without saga disappear when not handled explicitly 

dvdstelt:
Messages for a saga that arrive without a saga instance _and_ no implementation of a IHandleSagaNotFound just disappear. This wasn't documented.

### DIFF
--- a/nservicebus/sagas/saga-not-found.md
+++ b/nservicebus/sagas/saga-not-found.md
@@ -9,7 +9,7 @@ related:
 - samples/saga
 ---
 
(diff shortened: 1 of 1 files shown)


danielmarbach:
Isn't this a message loss scenario?

DavidBoike:
@danielmarbach Not really, it's a "working as designed."

Think about a timeout to ship via UPS after we've already successfully shipped via Fedex and completed the saga.

Or a "verify your email address" saga where the user clicks the verification link 3 months later.

dvdstelt:
100% agree with David, I expected them to disappear. I just wanted to point something to that little piece of doco and couldn't find it. So I first verified if my assumption was correct and then updated the doco accordingly. Especially with timeouts, some customers would end up with thousands of error msgs a day for not found saga instances.

PR #3114, opened 2017-08-23
No support for transaction scope in netcore2

SzymonPobiega:


transports/sql/transactions_ambient-core-warning_sqltransport_[4,).partial.md
```diff
@@ -0,0 +1 @@
+WARNING: The Transaction scope mode is not available in .NET Core 2.0 because the implementation of `SqlConnection` does not support enlisting in an ambient transaction.
```
SimonCropp: does "is not available" equate to throwing an exception?

SzymonPobiega: `SqlConnection` throws at runtime. This is because MS claims it will probably work (to some degree) in 2.1.

SimonCropp: any MS doco we can link to here?

SzymonPobiega: This https://github.com/dotnet/corefx/issues/22484#issuecomment-316829947 is the only "documentation" I found.

SimonCropp: yep. that is not worth it

PR #3131, opened 2017-08-24
store and forward to nsb v7

SimonCropp:


weralabaj:
@SimonCropp is that wip?

PR #3136, opened 2017-08-24
Only use a single saga property for correlation

ramonsmits:
Some changes that should make clear that:

- All messages mappings MUST map to the same saga property
- A saga can only have a single mapping property

If multiple saga properties need to be mapped or messages needed to be mapped to different saga data properties that a customer saga finder must be used.

/cc: @DavidBoike

### DIFF
--- a/nservicebus/sagas/message-correlation.md
+++ b/nservicebus/sagas/message-correlation.md
@@ -14,8 +14,7 @@ Correlation is needed in order to find existing saga instances based on data on 
 
 To declare this use the `ConfigureHowToFindSaga` method and use the `Mapper` to specify to which saga property each message maps to.
 
-NOTE: NServiceBus will only allows message mappings that all correlate to a single saga property and that all property types must match. Use a [custom saga finder](saga-finding.md). to correlate different properties and/or types.
-
+NOTE: NServiceBus  only allows message mappings where the message properties correlate to a single saga property and have the same type. 
 
 partial: note
 
--- a/nservicebus/sagas/message-correlation.md
+++ b/nservicebus/sagas/message-correlation.md
@@ -12,7 +12,9 @@ related:
 
 Correlation is needed in order to find existing saga instances based on data on the incoming message. In the example the `OrderId ` property of the `CompleteOrder` message is used to find the existing saga instance for that order.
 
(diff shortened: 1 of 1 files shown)


nservicebus/sagas/message-correlation.md
```diff
@@ -12,7 +12,9 @@ related:
 
 Correlation is needed in order to find existing saga instances based on data on the incoming message. In the example the `OrderId ` property of the `CompleteOrder` message is used to find the existing saga instance for that order.
 
-To declare this use the `ConfigureHowToFindSaga` method and use the `Mapper` to specify to which saga property each message maps to. Note that NServiceBus will only allow correlation on a single saga property and that the property types must match.
+To declare this use the `ConfigureHowToFindSaga` method and use the `Mapper` to specify to which saga property each message maps to.
+
+NOTE: NServiceBus will only allows message mappings that all correlate to a single saga property and that all property types must match. Use a [custom saga finder](saga-finding.md). to correlate different properties and/or types.
```
DavidBoike: This is going to result in two back-to-back NOTE boxes with the partial below.

I'm also not sure it's wise to recommend custom saga finders this far up the page and highlighted so much. We should be going out of our way to downplay custom saga finders and instead educating on what to do instead.

weralabaj: I'm with @DavidBoike on this one, to me personally it seems the previous version was clearer. Also we mention saga finders later, so why spell it out here so explicitly?
Was that change driven by a customer @ramonsmits ?

weralabaj:

SimonCropp:
@ramonsmits can u respond to the questions here or find someone else to take over the PR?

SimonCropp:
@Particular/nservicebus-maintainers anyone want to take this over? or should we close?

ramonsmits:
@weralabaj Yes, this was driven by a customer case. The problem with the customer was that they are not using V6 thus not enforced.

The current version doesn't highlight that by default you MUST only have one property for correlation. It is very easy to not read that at all.

What about we make that a NOTE and drop the second line about custom saga finders?

weralabaj:
> What about we make that a NOTE and drop the second line about custom saga finders?

Sounds good

SimonCropp:
this will need a rebase

weralabaj:
@ramonsmits do you need help finishing it?

ramonsmits:
@weralabaj I've rebased the PR and shortened the note.

PR #3144, opened 2017-08-25
Added note about default saga not found behavior

ramonsmits:
In the main saga guidance it is not really clear that the default behavior is that message are discarded if the message does not start a saga, but a saga instance does not exist. A lot of customers might not expect this behavior and can cause message/data loss.

### DIFF
--- a/nservicebus/sagas/index.md
+++ b/nservicebus/sagas/index.md
@@ -21,7 +21,18 @@ Design processes with more than one remote call to use sagas.
 
 While it may seem excessive at first, the business implications of the system getting out of sync with the other systems it interacts with can be substantial. It's not just about exceptions that end up in the logs.
 
-NOTE: If a message is received while there is no saga instance, then by default this message is discarded and succesfully processed. If the received message must not be discarded and throw an error so that it gets moved to the error queue then it is required to override saga not found behavior by implementing `IHandleSagaNotFound` and throw an exception. 
+## No saga instance available
+
+If a message is received while there is no saga instance, then by default this message gets discarded and succesfully processed. When you have a saga instance that is already completed then it makes sense to discard other messages.
+
+This can also happen for saga instance creation. A message is received that expects a saga instance to be there but that message is not received yet. In this case the message should not be discarded. There are two options to resolve this:
+
+- Let the saga instance be created for message B
+- Override the saga not found behavior and throw an exception
+
+By getting the saga instance created for message B that is received before message A we now have to make sure that the behavior of the saga will remain correct. For example, when a message is send when B is received then maybe state from message A needs to be included that is added to the saga instance when A is received. This state is not yet stored in the saga thus the message cannot be send yet. We now need to check that when message A is received if message B is already processed and then have the message send. This can make the saga much more complex!
+
+A simpler solution is to throw an error when the saga instance does not exist so that it gets moved to the error queue then it is required to override saga not found behavior by implementing `IHandleSagaNotFound` and throw an exception. 
 
 ## A simple Saga
 
--- a/nservicebus/sagas/index.md
+++ b/nservicebus/sagas/index.md
@@ -23,16 +23,20 @@ While it may seem excessive at first, the business implications of the system ge
 
 ## No saga instance available
 
-If a message is received while there is no saga instance, then by default this message gets discarded and succesfully processed. When you have a saga instance that is already completed then it makes sense to discard other messages.
+If a message is received while there is no saga instance, then by default this message gets discarded. When you have a saga instance that is already completed then it makes sense to discard other messages.
 
-This can also happen for saga instance creation. A message is received that expects a saga instance to be there but that message is not received yet. In this case the message should not be discarded. There are two options to resolve this:
+NOTE: Assume out of order delivery as this can happen due to error recovery and/or concurrent message processing.
 
-- Let the saga instance be created for message B
-- Override the saga not found behavior and throw an exception
+This message discarding behavior can also happen for saga instance creation. When a message is received that expects a saga instance to be there but that message that would create the saga instance has not yet been received or completed processing.
 
-By getting the saga instance created for message B that is received before message A we now have to make sure that the behavior of the saga will remain correct. For example, when a message is send when B is received then maybe state from message A needs to be included that is added to the saga instance when A is received. This state is not yet stored in the saga thus the message cannot be send yet. We now need to check that when message A is received if message B is already processed and then have the message send. This can make the saga much more complex!
+There are two options to avoid discarding incoming messages in this scenario:
 
-A simpler solution is to throw an error when the saga instance does not exist so that it gets moved to the error queue then it is required to override saga not found behavior by implementing `IHandleSagaNotFound` and throw an exception. 
+- Let the saga instance be created for the message processing that previously assumed the saga instance would already exist using `IAmStartedBy<T>`
+- Override the saga not found behavior and throw an exception using `IHandleSagaNotFound`
+
+By letting the saga instance also be created for a message that previously relied on another message that would have create the saga instance we now have to make sure that the behavior of the saga will remain correct. Lets assume that we relied on message A to be processed before B but we now also have the saga instance created when B is received then maybe state from message A needs to be included that is added to the saga instance when A is received. This state is not yet stored in the saga if message B is processed before message A thus the message cannot be send yet. We now need to check that when message A is received if message B is already processed and then have the message send. This can make the saga much more complex!
+
+A simpler solution is to throw an error when the saga instance does not exist so that it eventually gets moved to the error queue after all immediate and delayed retries attempts are done. Override saga not found behavior by [implementing `IHandleSagaNotFound` and throw an exception](saga-not-found.md). 
 
 ## A simple Saga
 
--- a/nservicebus/sagas/index.md
+++ b/nservicebus/sagas/index.md
@@ -21,7 +21,7 @@ Design processes with more than one remote call to use sagas.
 
 While it may seem excessive at first, the business implications of the system getting out of sync with the other systems it interacts with can be substantial. It's not just about exceptions that end up in the logs.
 
-NOTE: If a message is received while there is not saga instance then by default this message is discarded and succesfully processed. If the received message must not be discarded and throw an error so that it gets moved to the error queue then it is required to override saga not found behavior by implementing `IHandleSagaNotFound` and throw an exception. 
+NOTE: If a message is received while there is no saga instance, then by default this message is discarded and succesfully processed. If the received message must not be discarded and throw an error so that it gets moved to the error queue then it is required to override saga not found behavior by implementing `IHandleSagaNotFound` and throw an exception. 
 
 ## A simple Saga
 
--- a/nservicebus/sagas/index.md
+++ b/nservicebus/sagas/index.md
@@ -24,21 +24,22 @@ While it may seem excessive at first, the business implications of the system ge
 
 ## No saga instance available
 
-If a message that doesn't start a saga is received but no related saga instance is found, then by default that message is discarded, i.e. it is possible that a saga instance had been already completed when further messages arrive and such messages should be discarded.
+If a message that is handled by a saga is received, but no related saga instance is found, then by default that message is discarded, i.e. it is possible that a saga instance had been already completed when further messages arrive and such messages should be discarded.
 
-NOTE: Always assume that messages can be delivered out of order, i.e. due to invoking the error recovery mechanism, network latency, and/or concurrent message processing.
+NOTE: Always assume that messages can be delivered out of order, e.g. due to invoking the error recovery mechanism, network latency, and/or concurrent message processing.
 
 A similar message discarding behavior can also be invoked in case of saga instance creation. When a message is received that expects a saga instance to be there but that message that would create the saga instance has not yet been received or completed processing.
 
-There are two options to avoid discarding incoming messages in this scenario:
+To ensure messages are not discarded when they arrive out of order:
 
-- Let the saga instance be created for the message processing that previously assumed the saga instance would already exist using `IAmStartedBy<T>`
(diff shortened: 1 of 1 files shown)


SimonCropp:
@Particular/nservicebus-maintainers can you review this

nservicebus/sagas/index.md
```diff
@@ -21,6 +21,7 @@ Design processes with more than one remote call to use sagas.
 
 While it may seem excessive at first, the business implications of the system getting out of sync with the other systems it interacts with can be substantial. It's not just about exceptions that end up in the logs.
 
+NOTE: If a message is received while there is not saga instance then by default this message is discarded and succesfully processed. If the received message must not be discarded and throw an error so that it gets moved to the error queue then it is required to override saga not found behavior by implementing `IHandleSagaNotFound` and throw an exception.
```
andreasohlund: We need to mention that this only happens if the message isn't configured to start a new saga.

Perhaps:

>Messages not configured to start new saga instances will be discarded and considered successful if no existing instance is found. If the received message must not be discarded..

Thoughts?

timbussmann: > If a message is received while there is ~~not~~ no saga instance then by default this message is discarded and succesfully processed.

> If the received message must not be discarded and throw an error so that it gets moved to the error queue then it is required to override saga not found behavior by implementing `IHandleSagaNotFound` and throw an exception. 

that sentence is too complex. I think this would require a larger section with a corresponding snippet.

timbussmann:
should we consider to provide a feature/config option to make messages to fail if no saga was found?

weralabaj:

SimonCropp:
@ramonsmits can u respond to the questions here or find someone else to take over the PR?

SimonCropp:
@Particular/nservicebus-maintainers anyone want to take this over? or should we close?

ramonsmits:
@timbussmann I think having such a feature enabled by default would make the user aware of a decision to make. Maybe this can be added to https://github.com/Particular/PlatformDevelopment/issues/166 ?

ramonsmits:
@SimonCropp I restored the branch and made a change but that isn't listed here. Do I need to create a new PR?

SimonCropp:
@ramonsmits i see some new changes. what ones r u missing

ramonsmits:
@SimonCropp Now I see them too, maybe it was some caching or eventual consistency thingy.

weralabaj:
@ramonsmits not sure if tim's comments were addressed? what needs to be done here to complete?

nservicebus/sagas/index.md
```diff
@@ -21,6 +21,18 @@ Design processes with more than one remote call to use sagas.
 
 While it may seem excessive at first, the business implications of the system getting out of sync with the other systems it interacts with can be substantial. It's not just about exceptions that end up in the logs.
 
+## No saga instance available
+
+If a message is received while there is no saga instance, then by default this message gets discarded and succesfully processed. When you have a saga instance that is already completed then it makes sense to discard other messages.
```
timbussmann: > gets discarded and succesfully processed

that sounds a bit contradicting. Wouldn't it state the same when just going with `this message gets discarded`?

nservicebus/sagas/index.md
```diff
@@ -21,6 +21,18 @@ Design processes with more than one remote call to use sagas.
...
+
+This can also happen for saga instance creation. A message is received that expects a saga instance to be there but that message is not received yet. In this case the message should not be discarded. There are two options to resolve this:
```
timbussmann: > This can also happen for saga instance creation.

I think this sentence isn't understandable by itself. Start with something like "This behavior" instead of just "This", but without the next sentence, it's not clear how discarding messages can happen for messages that start a saga?

timbussmann: > A message is received that expects a saga instance to be there but that message is not received yet

two times referring `a message`, it's not super clear that the second mentioned message is a different message.

timbussmann: > In this case the message should not be discarded.

but this is the current behavior. Maybe restate the behavior to something like "There are two options to avoid discarding incoming messages in this scenario"?

nservicebus/sagas/index.md
```diff
@@ -21,6 +21,18 @@ Design processes with more than one remote call to use sagas.
...
+
+- Let the saga instance be created for message B
```
timbussmann: it's not clear what message B is, message B hasn't been introduced before.

nservicebus/sagas/index.md
```diff
@@ -21,6 +21,18 @@ Design processes with more than one remote call to use sagas.
...
+- Override the saga not found behavior and throw an exception
```
timbussmann: does this need some links?

nservicebus/sagas/index.md
```diff
@@ -21,6 +21,18 @@ Design processes with more than one remote call to use sagas.
...
+
+By getting the saga instance created for message B that is received before message A we now have to make sure that the behavior of the saga will remain correct. For example, when a message is send when B is received then maybe state from message A needs to be included that is added to the saga instance when A is received. This state is not yet stored in the saga thus the message cannot be send yet. We now need to check that when message A is received if message B is already processed and then have the message send. This can make the saga much more complex!
```
timbussmann: > when a message is send when B is received then maybe state from message A needs to be included that is added to the saga instance when A is received

this sentence contains 3 `when`, it's quite complex.

nservicebus/sagas/index.md
timbussmann: maybe state that the saga should use the `IAmStartedBy...`  interface?

ramonsmits:
Thanks for the review @timbussmann I'll address this next week.

SimonCropp:
this will need a rebase

weralabaj:
@ramonsmits do you need help finishing it?

ramonsmits:
@timbussmann @weralabaj I've made some changes based on Tim his feedback. Is this now understandable or does it need some sequence diagram to explain the scenarios?

nservicebus/sagas/index.md
```diff
@@ -22,6 +22,24 @@ Design processes with more than one remote call to use sagas.
 While it may seem excessive at first, the business implications of the system getting out of sync with the other systems it interacts with can be substantial. It's not just about exceptions that end up in the logs.
 
 
+## No saga instance available
```
weralabaj: We need a better title, it's not very descriptive, e.g. "Saga design considerations"?

nservicebus/sagas/index.md
```diff
@@ -22,6 +22,24 @@ Design processes with more than one remote call to use sagas.
...
+
+If a message that doesn't start a saga is received but no related saga instance is found, then by default that message is discarded, i.e. it is possible that a saga instance had been already completed when further messages arrive and such messages should be discarded.
+
+NOTE: Always assume that messages can be delivered out of order, i.e. due to invoking the error recovery mechanism, network latency, and/or concurrent message processing.
+
+A similar message discarding behavior can also be invoked in case of saga instance creation. When a message is received that expects a saga instance to be there but that message that would create the saga instance has not yet been received or completed processing.
```
weralabaj: I've reviewed content above, I hope I kept the essence of what you're saying, but starting from this place I'm not sure if I understand the intention.

E.g. I'm a bit confused about the fact that you introduce the discarding of a message here, separately. Isn't it the same behavior as above, only a different scenario/reason? If so, that should be moved to the paragraph before the "NOTE:" part.

nservicebus/sagas/index.md
```diff
@@ -22,6 +22,24 @@ Design processes with more than one remote call to use sagas.
...
+
+There are two options to avoid discarding incoming messages in this scenario:
```
weralabaj: How about:

> To ensure messages are not discarded when they arrive out of order:

nservicebus/sagas/index.md
```diff
@@ -22,6 +22,24 @@ Design processes with more than one remote call to use sagas.
...
+
+- Let the saga instance be created for the message processing that previously assumed the saga instance would already exist using `IAmStartedBy<T>`
```
weralabaj: I feel it'd be much simpler to explain using an example, e.g. let's take saga from some sample we have, that is started by a message X, and then updated when a message Y arrives... Otherwise it'll be hard to follow.

nservicebus/sagas/index.md
```diff
@@ -22,6 +22,24 @@ Design processes with more than one remote call to use sagas.
...
+- Override the saga not found behavior and throw an exception using `IHandleSagaNotFound`
+
+By letting the saga instance also be created for a message that previously relied on another message that would have create the saga instance we now have to make sure that the behavior of the saga will remain correct. Lets assume that we relied on message A to be processed before B but we now also have the saga instance created when B is received then maybe state from message A needs to be included that is added to the saga instance when A is received. This state is not yet stored in the saga if message B is processed before message A thus the message cannot be send yet. We now need to check that when message A is received if message B is already processed and then have the message send. This can make the saga much more complex!
```
weralabaj: I feel it'd be much simpler to explain using a specific example, e.g. let's take saga from some sample we have, that is started by a message X, and then updated when a message Y arrives...  But rather than talking about A/B or X/Y have concrete, meaningful message names. Otherwise, it'll be hard to follow. 

Was that the intention here? If so, I can take a stab at describing it.

nservicebus/sagas/index.md
```diff
@@ -22,6 +22,24 @@ Design processes with more than one remote call to use sagas.
...
+
+A simpler solution is to throw an error when the saga instance does not exist so that it eventually gets moved to the error queue after all immediate and delayed retries attempts are done. Override saga not found behavior by [implementing `IHandleSagaNotFound` and throw an exception](saga-not-found.md).
```
weralabaj: I'm not clear - is that an alternative or complementary step? ie. we use it together with having two IAmStartedBy... messages, or instead of it?

nservicebus/sagas/index.md
ramonsmits: Above the note, I explain the default discard behavior. Below the note a scenario where you don't want to have that behavior due to out of order processing which is often not thought of.

nservicebus/sagas/index.md
ramonsmits: Instead of. So either you deal with our of order delivery by

- implementing `IAmStartedBy` for all messages that could be received out of order, manage the ordering in the saga and have a state machine that decides all messages have arrived and continue.
- rely on error recovery by implementing IHandleSagaNotFound and check the messages types that must not be discarded as a saga instance is expected to exist.

nservicebus/sagas/index.md
ramonsmits: Maybe call this "What happens when a saga does not exist?"

ramonsmits: @timbussmann @weralabaj Below the note, it primarily is about out of order delivery/processing. The note states always assume out of order delivery. How to introduce the user to this?

ramonsmits:
@timbussmann @weralabaj 

> to list some scenarios where messages can be discarded.

I think I did, I even made it a note :-)

> NOTE: Always assume that messages can be delivered out of order, e.g. due to invoking the error recovery mechanism, network latency, and/or concurrent message processing.

Maybe I don't understand the comment?

nservicebus/sagas/index.md
weralabaj: I'd opt for the generic title, like "Saga design considerations" because if it's too narrow then the reader is more likely to assume it's not relevant to them (I think). From the content it's very clear what the problem is, you repeat a few times "not found" and "out of order", so search should be fine too.

> How to introduce the user to this?

I think the introduction is fine, just it looks like you talk about something, part 1, then is the note, and then you continue talking about the same thing, part 2, but the connection between part 1 and 2 is not very clear. At least for me it sounds like part 2 is again starting from scratch, which is a bit confusing and that's why I'm not sure if I got the intention right.

If you say, yes, I'm talking about the same thing, and there's a note in the middle of it, then I'm fine, it's all clear ;)

weralabaj:
@ramonsmits I hope my comment here https://github.com/Particular/docs.particular.net/pull/3144#discussion_r142911211 helps to clarify, if not, let's have a quick call?

nservicebus/sagas/index.md
ramonsmits: @weralabaj Well, they are two different sections. The first mentioning the saga not found discard behavior, the second that this can have implications if you assume ordered delivery but in reality it is not and how to deal with out or order delivery. Maybe adding headings makes it all more clear.

weralabaj:
@ramonsmits I made some edits, to make the content fit better what is already on that page, please have a look and let me know if that's ok.

nservicebus/sagas/index.md
```diff
@@ -83,11 +56,44 @@ This interface tells NServiceBus that the saga not only handles `StartOrder`, bu
 partial: at-least-one
 
 
+### Dealing with out of order delivery
```
ramonsmits: @weralabaj I don't understand why this is nested under *A simple saga*

weralabaj: it's not, it'd be under "Starting a saga" header, https://docs.particular.net/nservicebus/sagas/#starting-a-saga

weralabaj: I thought it'd make sense, since that part concentrates on caveats of staring a saga.

ramonsmits: @weralabaj Sorry, that was due github not showing all text during the compare.

In my opinion the discarding behavior and out or order are mentioned too late. I think they are key elements to mention very early on. Maybe we need an introduction at the top? Stating the things like stateful, correlation, out of order and discarding?

weralabaj: Hm... looking at the whole article I don't have a good idea of where to put it earlier, first we explain what saga is, and then move on to starting... So quite early in fact. Although I agree the article is getting a bit long and maybe that's the problem?

Not sure, open for edits/specific suggestions.

weralabaj:
rebased

PR #3156, opened 2017-08-30
clarifications for endpoint tuning defaults

SeanFeldman:
The default is not spelled out, which leave questions.
After looking at the override, it might feel that max throughput needs to be overridden as well when changing settings.

With this change it states clearly that throughput is **not** throttled.

@Particular/nservicebus-maintainers please review

### DIFF
--- a/nservicebus/operations/tuning_throttling_core_[,5].partial.md
+++ b/nservicebus/operations/tuning_throttling_core_[,5].partial.md
@@ -1,4 +1,4 @@
-NOTE: by default endpoing is not throttled.
+NOTE: By default endpoints are not throttled.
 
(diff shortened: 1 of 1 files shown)


nservicebus/operations/tuning_throttling_core_[,5].partial.md
```diff
@@ -1,3 +1,5 @@
+NOTE: By default endpoints are not throttled.
```
Comment (timbussmann) [inline nservicebus/operations/tuning_throttling_core_[,5].partial.md]: 👍 on the statement, but does it need to be a note?

Comment (weralabaj) [reply nservicebus/operations/tuning_throttling_core_[,5].partial.md]: Given that they used to be, at least for some license types, I'd vote to leave it as a note, so it's more visible.

PR #3160, opened 2017-08-31
Doco that ilmerge is not supported

andreasohlund:


andreasohlund:
Do we need to spell out the reason?

weralabaj:
@andreasohlund I think even one sentence explaining the decision would be good to add.

andreasohlund:
@weralabaj added a short reasoning

PR #3164, opened 2017-09-01
MSMQ DLQ Custom Check

mikeminutillo:
Sample that watches the MSMQ DLQ using a ServiceControl Custom Check


mikeminutillo:
Would it make sense for this sample to include a way to populate the DLQ?

andreasohlund:
Sounds resonable, one option is to create a bogus queue and send a message with a short TTBR to it. Would need to be a native msmq client send simce we don't set the DLQ flag on messages with a TTBR

SimonCropp:
> Would it make sense for this sample to include a way to populate the DLQ?

yep. but that can be a phase two

PR #3171, opened 2017-09-04
Multi instance migration sample

SzymonPobiega:
This is the first part of multi-instance migration doco. It is based on the previous multi-instance sample. I think we need a new sample that is more complex and would show:
 * How endpoints using different catalogs on same instance can take advantage of multi-catalog
 * How topologies with 3+ SQL instances can use switch
 * How pub/sub works

Not sure if we should tackle these three aspects in one sample or create three separate samples. Thoughts?


samples/sqltransport/multi-instance-migration/SqlTransport_4/Receiver.V3/Program.cs
```diff
@@ -0,0 +1,45 @@
...
+        transport.EnableLegacyMultiInstanceMode(ConnectionProvider.GetConnection);
+        endpointConfiguration.UsePersistence<InMemoryPersistence>();
+        endpointConfiguration.SendFailedMessagesTo("error");
+        endpointConfiguration.EnableInstallers();
+
+        #endregion
+
+        SqlHelper.EnsureDatabaseExists(ConnectionProvider.DefaultConnectionString);
+
+        endpointConfiguration.Conventions().DefiningMessagesAs(t => t.Assembly == typeof(ClientOrder).Assembly && t.Namespace == "Messages");
```
SimonCropp: why do we need conventions?

SimonCropp:
and u can ignore the azure build error. it is fixed on master

samples/sqltransport/multi-instance-migration/sample.md
```diff
@@ -0,0 +1,86 @@
+---
+title: Multi-Instance Mode to Bridge migration
+summary: Migration of SQL Server transport Multi-Instance Mode topology to Bridge
+reviewed: 2017-09-04
+component: SqlTransport
+related:
+ - transports/sql/deployment-options
+---
+
```
SimonCropp: perhaps a short intro?

samples/sqltransport/multi-instance-migration/SqlTransport_4/Receiver.V3/Program.cs
SzymonPobiega: Yes because the messages assembly is shared between V3 and V4 versions.

samples/sqltransport/multi-instance-migration/sample.md
SzymonPobiega: I'll add it today.

weralabaj:
I think this is a lot to tackle in one sample ;) Definitely pub sub in multi instance setup would be nice as a standalone, so if somebody is considering this configuration they know how it works and what it involves.

I don't understand this part, however

> How topologies with 3+ SQL instances can use switch

SzymonPobiega:
> How topologies with 3+ SQL instances can use switch

@weralabaj `NServiceBus.Bridge` includes a thing called `Switch` which is basically a bridge with more than one ends (as many as you want) that are called *ports*. It allows to bridge more than two brokers but requires configuring destination ports. This part is not yet documented though.

weralabaj:
@SzymonPobiega hm... more than two brokers, meaning potentially two different transports? or have to be the same transport?
either way sounds a bit complicated, so also a good candidate for a separate sample.

samples/sqltransport/multi-instance-migration/sample.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Prerequisites
+
+include: sql-prereq
+
+The databases created by this sample are `NsbSamplesSqlMultiInstanceReceiver3`, `NsbSamplesSqlMultiInstanceSender3`, `NsbSamplesSqlMultiInstanceReceiver4`, `NsbSamplesSqlMultiInstanceSender3` and `NsbSamplesSqlMultiInstanceBridge`.
```
weralabaj: automatically created? if yes, IMO better to be explicit :)

SzymonPobiega:
@weralabaj more than two brokers might mean e.g. three different transports but in this context I was thinking about 3-4 SQL instances in the multi-instance topology (this does not sound super unrealistic).

samples/sqltransport/multi-instance-migration/sample.md
```diff
@@ -0,0 +1,86 @@
...
+ 1. The Receiver displays information that an order was submitted.
+ 1. The Sender displays information that the order was accepted.
+
+
+## Code walk-through
+
+This sample contains the following projects:
+
+ * Sender.V3 - A console application responsible for sending the initial `ClientOrder` message and processing the follow-up `ClientOrderAccepted` message.
+ * Receiver.V3 - A console application responsible for processing the order message.
```
weralabaj: A console application running/using NServiceBus Version 6 and SQL Server transport Version 3.

samples/sqltransport/multi-instance-migration/sample.md
```diff
@@ -0,0 +1,86 @@
...
+
+The Receiver mimics a back-end system. In Version 3 it is configured to use SQL Server transport in the [*multi-instance*](/transports/sql/deployment-options.md#modes-overview-multi-instance) mode.
+
+snippet: ReceiverConfigurationV3
+
+It receives `ClientOrder` message sent by Sender and replies to them with `ClientOrderAccepted`.
+
+snippet: Reply
+
+In Version 4 the receiver does not requires any special configuration. The reply is sent back to the bridge:
```
weralabaj: does not require~s~

transports/upgrades/sqlserver-31to4.md
```diff
@@ -18,3 +18,15 @@ The Transaction scope [transaction mode](/transports/sql/transactions.md) is not
 To run the upgraded project on .NET Core the transport needs to be switched to one of the native transactions modes. Consider using the [Outbox](/nservicebus/outbox) to maintain the same *exactly-once processing* guarantees.
 
 NOTE: Transaction scope is supposed to be supported by `SqlConnection` in future versions of .NET Core. 
+
+## Multi-instance mode
+
+The multi-instance mode has been deprecated in Version 4. NServiceBus topologies with queues distributed between multiple catalogs and SQL Server instances can be migrated using a combination of [Transport Bridge](/nservicebus/bridge/) and multi-catalog [addressing](/transports/sql/addressing.md):
```
weralabaj: Some of this part would be good as an intro to the sample :)

transports/upgrades/sqlserver-31to4.md
```diff
@@ -18,3 +18,15 @@ The Transaction scope [transaction mode](/transports/sql/transactions.md) is not
...
+ * When communicating with an endpoint that connects to a different SQL Server instance, send messages through a bridge. The [multi-instance migration sample](/samples/sqltransport/multi-instance-migration) demonstrates this approach.
+ * When communicating with an endpoint that connects to a different catalog on the same SQL Server instance, send messages directly and configure multi-catalog addressing.
+
+Both of these feature are available for NServiceBus 6 (and SQL Server transport 3.1) so the topology migration can take place before switching to NServiceBus 7.
```
weralabaj: Both of these feature**s**

transports/upgrades/sqlserver-31to4.md
SzymonPobiega: Done.

SzymonPobiega:
@weralabaj I rebased it so it should now build fine. Can you approve?

PR #3177, opened 2017-09-06
Saga completion - addition

SeanFeldman:
Connects to https://github.com/Particular/NServiceBus/issues/4943

How to handle scenario when transaction doesn't span transport and persistence.

@andreasohlund here's an initial stub at it.

### DIFF
--- a/nservicebus/sagas/index.md
+++ b/nservicebus/sagas/index.md
@@ -74,13 +74,13 @@ When a message is received that could possibly be handled by a saga, and no exis
 
 ### Transactional behavior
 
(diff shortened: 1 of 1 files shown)


SeanFeldman:
Build is not breaking because of the change in this PR, can be ignored.

weralabaj:
as i understand this https://github.com/Particular/docs.particular.net/pull/3180 is the replacement ? cc @andreasohlund

andreasohlund:
#3180 targets this branch so its a suggestion

SeanFeldman:
I like what @andreasohlund did in #3180. Once it's merged into this PR, we can do a sanity check and merge this one.

andreasohlund:
@SimonCropp thanks!!

@SeanFeldman are you ok with the final result?

SeanFeldman:
https://www.youtube.com/watch?v=q3svW8PM_jc

PR #3180, opened 2017-09-07
Clarify saga completion caveats

andreasohlund:
@SeanFeldman rearranged the text slightly, what do you think?

### DIFF
--- a/nservicebus/sagas/index.md
+++ b/nservicebus/sagas/index.md
@@ -74,9 +74,9 @@ When a message is received that could possibly be handled by a saga, and no exis
 
 ### Transactional behavior
 
-Completing a saga is a destructive operation so transation support of the selected transport and persitence must be considered to ensure correctness. If the persistence is able to participate in the same transaction as the incoming receive operation, either using DTC or by sharing the transports storage transaction (SQLServer transport), no futher actions is needed.
+Completing a saga is a destructive operation so transaction support of the selected transport and persitence must be considered to ensure correctness. If the persistence is able to participate in the same transaction as the incoming receive operation, either using DTC or by sharing the transports storage transaction (SQLServer transport), no futher action is needed.
 
-For combinations where the above is not true further actions is needed to avoid saga completion causing incorrect business behavior. The problematic scenario is sagas completing together with sending/publishing messages. In this case failure after the saga is completed might cause outgoing messages to not be dispatched and when the incoming message is retried the saga is no longer found resulting in outgoing messages being lost.
+For combinations where the above is not true action is needed to avoid saga completion causing incorrect behavior. The problematic scenario is sagas completing together with sending/publishing outgoing messages. In this case failure after the saga is completed might cause outgoing messages to not be dispatched and when the incoming message is retried the saga is no longer found resulting in outgoing messages being lost.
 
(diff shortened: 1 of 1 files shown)


nservicebus/sagas/index.md
```diff
@@ -72,12 +72,17 @@ For more information about setting (requesting) the timeouts and handling them,
...
 
-1. Enabling the [Outbox feature](/nservicebus/outbox) if supported by the chosen persistence.
-1. Setting a timeout or sending a message to self to invoke saga completion as a single operation that will be safely retried in case of a failure.
+If the persistence can't participate in the same transaction as the incoming receive operation, then an additional action is needed to avoid saga completion causing incorrect behavior. The problematic scenario is when sagas are being completed together with sending/publishing outgoing messages. In case of failure after the saga is completed, the outgoing messages may not be dispatched. However, when the incoming message is retried the completed saga is not found, which results in outgoing messages being lost.
+
+This issue can be avoided by:
+
+1. Enabling the [Outbox feature](/nservicebus/outbox), if supported by the chosen persistence.
+1. Making sure that no outgoing messages will be dispatched by completing the saga from a timeout.
+1. Never complete the saga but instead set a flag/timestamp and use some native mechanism of the selected storage to cleanup old saga instances.
```
weralabaj: > Never complete the saga but instead set a flag/timestamp and use some native mechanism of the selected storage to cleanup old saga instances.

@andreasohlund do we actually want to suggest that? or is that only recommended in situations when the outgoing message has to be dispatched at the same time that saga is completed?

andreasohlund: I've seen it use for clearing out "never ending sagas" in the past so I think its legit? See http://udidahan.com/2009/09/01/dont-delete-just-dont/

nservicebus/sagas/index.md
```diff
@@ -72,12 +72,17 @@ For more information about setting (requesting) the timeouts and handling them,
...
 
-1. Enabling the [Outbox feature](/nservicebus/outbox) if supported by the chosen persistence.
-1. Setting a timeout or sending a message to self to invoke saga completion as a single operation that will be safely retried in case of a failure.
+If the persistence can't participate in the same transaction as the incoming receive operation, then an additional action is needed to avoid saga completion causing incorrect behavior. The problematic scenario is when sagas are being completed together with sending/publishing outgoing messages. In case of failure after the saga is completed, the outgoing messages may not be dispatched. However, when the incoming message is retried the completed saga is not found, which results in outgoing messages being lost.
+
+This issue can be avoided by:
+
+1. Enabling the [Outbox feature](/nservicebus/outbox), if supported by the chosen persistence.
+1. Making sure that no outgoing messages will be dispatched by completing the saga from a timeout.
```
SeanFeldman: It's either completing from a timeout _or_ sending an explicit command to self. I'd like to keep the option of a command to self. Even though, in the end, both are messages 🙂

nservicebus/sagas/index.md
SeanFeldman: How about instead of starting off with a such a strong statement as "Never complete..." start with 

> Replace saga completion with soft delete by setting a flag.....

PR #3181, opened 2017-09-07
Delete logging-configuration_profile_host_[,7].partial.md

seanfarmar:
The `IConfigureLoggingForProfile<T>` is obsolete in host v7


weralabaj:
@seanfarmar if I got your intention right, then the correct approach would be to rename partial to nservicebus/hosting/nservicebus-host/logging-configuration_profile_host_[,6].partial.md, then it'd be valid for versions up to 6. Would that be correct?

weralabaj:
@seanfarmar see now, should be what you were after ;)

weralabaj:
if correct, feel free to merge yourself

PR #3184, opened 2017-09-07
Rabbit topology selection is mandatory

andreasohlund:
Connects to https://github.com/Particular/NServiceBus.RabbitMQ/issues/427

### DIFF
--- a/transports/rabbitmq/index.md
+++ b/transports/rabbitmq/index.md
@@ -27,7 +27,7 @@ The RabbitMQ transport requires a connection string to connect to the broker. Se
 
 ### Topology selection
 
-Topologies is used to control how native Rabbit constructs like Queues, Topologies and bindings are created. As of Version 5 selecting topology is mandatory. For backwards compatiblility use the `ConventionalRoutingTopology` which was the previous default:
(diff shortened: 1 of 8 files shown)


andreasohlund:
@Particular/rabbitmq-transport-maintainers just to get it going, please flesh it out as needed

transports/rabbitmq/index.md
```diff
@@ -26,6 +26,15 @@ snippet: rabbitmq-config-basic
 The RabbitMQ transport requires a connection string to connect to the broker. See [Connection settings](/transports/rabbitmq/connection-settings.md) for options on how to provide the connection string.
 
 
+### Topology selection
+
+Topologies are used to control how native Rabbit constructs like queues and exchanges are created. In Versions 5 and above, selecting a topology is mandatory. For backwards compatiblility use the `ConventionalRoutingTopology`, which was the previous default:
+
+snippet: rabbitmq-config-topology
```
adamralph: It looks like the build is failing because it can't find this snippet. Does it have to be in the `Rabbit_All` project instead of the `Rabbit_5` one?

andreasohlund: Think we need to pull it out to a separate partial since it only applies to v5?

andreasohlund: I've pushed a fix

PR #3196, opened 2017-09-08
All SQL Transport 4 samples to beta*

DavidBoike:
`4.0.0-*` does not work because a NuGet bug causes unlisted unstable builds from the NServiceBus 4.0 era to get picked up.


DavidBoike:
@Particular/sqlserver-transport-maintainers Build is failing because the Store & Forward sample is using obsolete `EnableLegacyMultiInstanceMode()`.

What should be done?

PR #3206, opened 2017-09-11
Fix snippet

MarcinHoppe:


SimonCropp:
that is actually intentional since the endpoint config would come after that

MarcinHoppe:
The missing bracket looks like something is off. Maybe signal the missing code with an ellipsis or something?

SimonCropp:
Yeah u r correct. It is dodgy. I will have a better look at it tomorrow

MarcinHoppe:
Feel free to use this PR as a start 😄.

PR #3229, opened 2017-09-18
Fix ugly bootstrapping INFO box

DavidBoike:


DavidBoike:
Engine doesn't like an article with no index.md in the same folder. Taking a different approach.

PR #3237, opened 2017-09-19
Adding a v1 MSMQ Persistence sample

### LINKED ISSUE #3110: Add sample for MSMQ persistence
ping @Particular/msmq-transport-maintainers

andreasohlund:
Do we need a v6 sample as well?

Fixes #3110

### DIFF
--- a/samples/msmq/persistence/sample.md
+++ b/samples/msmq/persistence/sample.md
@@ -14,9 +14,10 @@ Ensure that MSMQ has been installed.
 
 ## Code walk-through
 
-This sample shows basic usage of MSMQ as storage for subscriptions. The application publish an empty message to itself, via the MSMQ transport, and writes to the console when the message is received. 
+This sample shows basic usage of MSMQ as storage for subscriptions. The application publishes an empty message to itself, via the MSMQ transport, and writes to the console when the message is received. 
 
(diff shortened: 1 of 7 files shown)


samples/msmq/persistence/sample.md
```diff
@@ -0,0 +1,24 @@
...
+
+
+## Code walk-through
+
+This sample shows basic usage of MSMQ as storage for subscriptions. The application publishes an empty message to itself, via the MSMQ transport, and writes to the console when the message is received. 
+
+Note that the default queue name, `NServiceBus.Subscriptions`, is overriden to avoid sharing the subscription storage queue with other endpoints on the same machine. The queue name is `{EndpointName}.Subscriptions`.
+
+
+### Configuration
```
indualagarsamy: @andreasohlund - I think it might be important to add a note that, you have to use other persistence options for Timeouts as timeouts is not supported via MSMQ like for subscription storage. InMemory while ok for the sample is not a good production safe option and a note will clarify that much better. Thoughts?

andreasohlund: Done, just added a generic note about storage needs. The docs for InMemory already have the relevant "don't use in production" warnings (https://docs.particular.net/persistence/in-memory)

PR #3239, opened 2017-09-19
WIP: Conditional net core

SimonCropp:


SimonCropp:
@Particular/docs-maintainers thoughts on this. it allows us to avoid `*.core.csproj` and `*.core.sln` files. can anyone see any flaws in my logic?

SimonCropp:
@bording any thoughts/flaws in this?

bording:
@SimonCropp I'd want to spend some time looking into the `BundledNETCoreAppPackageVersion` to make sure that's a safe/reliable way detect a 2.0 SDK being installed. Where did you find information on it?

Have you tried these projects with a VS install that doesn't have the 2.0 SDK installed (but must have the 1.0 SDK installed since we're using SDK-style projects)? I recall seeing some stuff about conditional `TargetFramework` properties either not being supported, or only being supported in a limited way, which would then make this not work correctly in all scenarios.

I don't have link handy for more information about that, so I could be misremembering...

SimonCropp:
re `BundledNETCoreAppPackageVersion` i reverse engineered it. 

i also posed it on stack overflow https://stackoverflow.com/questions/46311964/how-to-control-targetframework-based-on-if-netcore-2-x-sdk-is-installed

SimonCropp:
sooo @Particular/docs-maintainers any thoughts?

DavidBoike:
@SimonCropp still need someone to try running it with 15.3 without netcore2 installed? I think my "clean" windows VM might still be in that configuration.

SimonCropp:
replaced by https://github.com/Particular/docs.particular.net/pull/3272

PR #3248, opened 2017-09-20
add quartz

SimonCropp:


SimonCropp:
@andreasohlund done

samples/scheduling/quartz/sample.md
```diff
@@ -0,0 +1,81 @@
...
+related:
+- nservicebus/messaging/timeout-manager
+---
+
+This sample illustrates the use of [Quartz.NET](https://www.quartz-scheduler.net/) to send messages from within an NServiceBus endpoint.
+
+> Quartz.NET is a full-featured, open source job scheduling system that can be used from smallest apps to large scale enterprise systems.
+
+
+Note: The approach used in this sample can mitigate some of the architectural drawback of the [Scheduler](/nservicebus/scheduling/). The scheduler is build on top of the [Timeout Manager](/nservicebus/messaging/timeout-manager.md) which leverages the queuing system to trigger scheduled actions. Under heavy load, and associated delay in processing of messages, there may be some disparity between the expected time of a scheduled action and actual execution time.
```
mikeminutillo: drawbacks (should be pluralized)

built on top (rather than build on top)

Given that Quartz is also a "scheduler" would it be better to refer to the NServiceBus Scheduler in this sample?

I found this confusing: 

> Under heavy load, and associated delay in processing of messages, there may be some disparity between the expected time of a scheduled action and actual execution time. 

Perhaps

> Under heavy load there may be some disparity between the expected time of a scheduled action and actual execution time due to the delay between timeout messages being generated and processed.

samples/scheduling/quartz/sample.md
```diff
@@ -0,0 +1,81 @@
...
+ 1. At startup Scheduler will schedule a message send to Receiver every 3 seconds.
+ 1. Receiver will handle the message.
+
+
+## Code Walk-through
+
+INFO: This sample uses a pre-release package of Quartz version 3 to reduce the friction with the NServiceBus async messaging api.
+
+
+### Logging
```
mikeminutillo: I would drop this section to the bottom. It's useful but not core to the sample

samples/scheduling/quartz/sample.md
```diff
@@ -0,0 +1,81 @@
...
+A helper to inject and extract the `IEndpointInstance` from the Quartz scheduler context.
+
+snippet: QuartzContextExtensions
+
+Quartz also support Dependency Injection (DI) via the [JobFactory API](https://www.quartz-scheduler.net/documentation/quartz-2.x/tutorial/miscellaneous-features.html).
+
+
+### Configure and start the scheduler
+
+The endpoint is started and the `IEndpointInstance` is injected into the Quartz scheduler context
```
mikeminutillo: Missing a period.

It might be worth adding one more sentence to explain why you want the endpoint instance in the scheduler context.

> Quartz.NET Jobs have access to items inside the scheduler context when they run.

SimonCropp:
> Do we need to provide guidance about scaling out here

@mikeminutillo good point i will add a note and point to the relevant quartz doco

PR #3250, opened 2017-09-20
Hangfire sample

janovesk:


janovesk:
Have a look @Particular/docs-maintainers, cc @SimonCropp.

samples/scheduling/hangfire/sample.md
```diff
@@ -0,0 +1,92 @@
...
+related:
+- nservicebus/messaging/timeout-manager
+---
+
+This sample illustrates the use of [Hangfire](https://www.hangfire.io/) to send messages from within an NServiceBus endpoint.
+
+> Hangfire - An easy way to perform background processing in .NET and .NET Core applications. Hangfire is an open-source framework that helps you to create, process and manage your background jobs.
+
+
+Note: The approach used in this sample can mitigate some of the architectural drawbacks of the [NServiceBus Scheduler](/nservicebus/scheduling/). The NServiceBus scheduler is built on top of the [Timeout Manager](/nservicebus/messaging/timeout-manager.md) which leverages the queuing system to trigger scheduled actions. Under heavy load, there may be some disparity between the expected time of a scheduled action and actual execution time due to the delay between timeout messages being generated and processed.
```
SimonCropp: figure we can make this an include shared with the other sample after this PR is merged

samples/scheduling/hangfire/Core_6/Scheduler/Program.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        // create and start scheduler instance
+        var scheduler = new BackgroundJobServer();
+
+        #endregion
+
+        #region scheduleJob
+        
+        // Tell Hangfire to schedule the job and trigger every minute 
+        RecurringJob.AddOrUpdate("myJob", () => SendMessageJob.Run(), Cron.Minutely);
```
SimonCropp: out of curiosity how does the static `RecurringJob` use the instance `scheduler`?

janovesk:
They don't have an async API for now. It's coming in the next major. They will, however, make sure to wait for any scheduled jobs returning Tasks, so should work fine.

Also, no good way to pass the IEndpointInstance to the jobs except using a static or using their container support. I went with static as the container support would just make the sample bigger and would be explicitly for a particular container.

samples/scheduling/hangfire/Core_6/Scheduler/Program.cs
janovesk: Through some kind of magic. You have the same thing with configuration. You have a static `GlobalConfiguration` class that you use to configure and that is used behind the cover when you do `var scheduler = new BackgroundJobServer();`

SimonCropp: hmm. weird. but it is actually what i saw in some snippets, so dont doubt u. just seems to be asking for some side effects. not that this is our problem given the context of this sample. 

thanks for the clarification

samples/scheduling/hangfire/sample.md
```diff
@@ -0,0 +1,93 @@
...
+ 1. Start both the Scheduler and Receiver projects.
+ 1. At startup Scheduler will schedule sending a message to Receiver every minute.
+ 1. Receiver will handle the message.
+
+
+## Code Walk-through
+
+### Endpoint Helper
+
+A helper to make the `IEndpointInstance` created by NServiceBus available inside Hangfire jobs. In this sample it is implemented as a static property.
```
weralabaj: > A helper to make the `IEndpointInstance` created by NServiceBus available inside Hangfire jobs. 

I don't understand that sentence, can you rephrase?

samples/scheduling/hangfire/sample.md
```diff
@@ -0,0 +1,93 @@
...
+snippet: scheduleJob
+
+
+### Cleanup
+
+The Hangfire server should be disposed when the endpoint is shut down.
+
+snippet: shutdown
+
+For cleanup purpose either a static variable may need to be kept or the shutdown done as part of the container cleanup.
```
weralabaj: > For cleanup purpose either a static variable may need to be kept or the shutdown done as part of the container cleanup.

I don't understand that sentence, can you rephrase?

samples/scheduling/hangfire/Core_6/Scheduler/Program.cs
danielmarbach: Why not change to `RecurringJob.AddOrUpdate("myJob", () => SendMessageJob.Run(endpointInstance), Cron.Minutely);` and then 

```
public class SendMessageJob
    {
        public static Task Run(IMessageSession session)
        {
            return session.Send("Samples.HangfireScheduler.Receiver", new MyMessage());
        }
    }
```

janovesk: That will not work, unfortunately. Hangfire serialized the expression and the arguments into a JSON blob, which stored until the trigger and then deserialized and run.

janovesk: Which can run at a different machine if you run multiple instances etc

PR #3254, opened 2017-09-20
Typo

### LINKED ISSUE #3253: Feedback: 'Why Use SQS?'
Feedback for 'Why Use SQS?' https://docs.particular.net/transports/sqs/why-use-sqs

Location in GitHub: https://github.com/Particular/docs.particular.net/blob/master/transports/sqs/why-use-sqs.md

This sentence at the very bottom:

> To solve these problems, the guidance from Amazon is to use SQL as the transport.

should probably mention **SQS** and not **SQL**?

MarcinHoppe:
Resolve #3253


MarcinHoppe:
@Particular/aws-maintainers Please confirm that it's really SQS that is recommended here.

PR #3258, opened 2017-09-22
WIP Showing context decorator for dependencies that access session or handler context

danielmarbach:


SimonCropp:
this will need a rebase

danielmarbach:
I need feedback from @Particular/nservicebus-maintainers whether we want to proceed with this or not

timbussmann:
I think such a snippet/sample would be helpful 👍 

My thoughts on the current state: This approach relies heavily on the container and correct registration of the factory, endpoint instances and so on. Instead of leaving this generic, I think a concrete example using one of the popular containers would be easier to understand. As an alternative I still wonder whether an AsyncLocal approach wouldn't be easier to understand (maybe as a 2nd snippet/sample)?

danielmarbach:
I don't see the value of showing it for one container only. Then we'd have to jump the whoops and show it for all containers that we support. Closing this then

danielmarbach:
Raised https://github.com/Particular/PlatformDevelopment/issues/1482

PR #3262, opened 2017-09-22
Enable installer

timbussmann:
the sample needs installers to be explicitly enabled so the saga table is created when running the sample.


DavidBoike:
@timbussmann Just merge one-line fixes like this when they're green.

PR #3267, opened 2017-09-25
Add reasoning for deprecating host

weralabaj:


nservicebus/upgrades/acs-host-7to8.md
```diff
@@ -12,7 +12,19 @@ upgradeGuideCoreVersions:
  - 7
 ---
 
-The NServiceBus Azure Host will be deprecated as of Version 9 and users are recommended to switch to self hosting for new endpoints. Upgrading existing endpoints is still supported for Version 8
+The NServiceBus Azure Host will be deprecated as of Version 9 and users are recommended to switch to self hosting for new endpoints. Upgrading existing endpoints is still supported for Version 8.
```
danielmarbach: Should we make this an include and also include into the host upgrade guide?

nservicebus/upgrades/acs-host-7to8.md
```diff
@@ -12,7 +12,19 @@ upgradeGuideCoreVersions:
...
+
+The following are the main reasons why we are deprecating the NServiceBus Host:
+
+- Ease of configuration: Endpoint configuration and related APIs have been significantly improved over the last few years, therefore self-hosting is much simpler to set up than it used to be when Host was first introduced.
```
danielmarbach: > few years, therefore

to

> few years. Therefore

?

nservicebus/upgrades/acs-host-7to8.md
```diff
@@ -12,7 +12,19 @@ upgradeGuideCoreVersions:
...
+
+- Ease and transparency of deployment: The Host under the hoods used the `sc.exe` tool, but required performing additional manual steps to ensure reliability. Those steps were described in the documentation but were sometimes missed, which could cause problems at runtime. With the new recommended [installation procedure](/nservicebus/hosting/windows-service.md#installation) users are guided through the process and less likely to omit crucial configuration steps.
```
andreasohlund: >The Host under the hoods used the `sc.exe` tool

Not correct, we use TopShelf that in turn calls into https://msdn.microsoft.com/en-us/library/system.serviceprocess.serviceinstaller.install(v=vs.110).aspx

see https://github.com/Particular/Topshelf/blob/master/src/Topshelf/Hosts/AbstractInstallerHost.cs

Perhaps we can just say that instead of abstracting Service installation we now "force" them to be more explicit about how to install their service?

nservicebus/upgrades/acs-host-7to8.md
andreasohlund: Agreed, we need this in the windows host upgrade guide as well. 

That said some parts will differ since eg. usage of `SC.exe` isn't relevant when running in azure?

weralabaj: Hm... I could omit the link and make it more generic.

nservicebus/upgrades/acs-host-7to8.md
weralabaj: eh... ok :)

weralabaj:
@andreasohlund @danielmarbach incorporated your feedback

nservicebus/upgrades/host-deprecate.include.md
```diff
@@ -0,0 +1,11 @@
+The following are the main reasons why we are deprecating the NServiceBus Hosts:
```
andreasohlund: Do we need the plural "hosts" , since its an include perhaps we can just say

>The main reasons for deprecating the host is:

Thoughts?

nservicebus/upgrades/host-7to8.md
```diff
@@ -11,7 +11,9 @@ upgradeGuideCoreVersions:
  - 7
 ---
 
-The NServiceBus will be deprecated as of Version 9 and users are recommended to switch to self hosting for new endpoints. Upgrading existing endpoints is still supported for Version 8
+The NServiceBus will be deprecated as of Version 9 and users are recommended to switch to self hosting for new endpoints. Upgrading existing endpoints is still supported for Version 8.
```
andreasohlund: The NServiceBus will => The NServiceBus Host will

nservicebus/upgrades/host-deprecate.include.md
```diff
@@ -0,0 +1,11 @@
...
+
+- **Configuration transparency**: The hosts used to have various profiles and roles which automatically set certain options like disabling transactions. The effect was not obvious without reading the documentation. It could result in misunderstandings and undesired, surprising effects in the system at runtime. With self-hosting users have full, explicit, fine-grained control over all aspects of the endpoint configuration and runtime behavior.
+
+- **Easier troubleshooting**: The hosts introduce another layer of abstraction which can make troubleshooting more difficult. Some problems require deep dive into internal implementation and were hard to solve without support.
+
+- **Better performance**: Using the hosts with default, generic settings introduces certain overhead. Users using self-hosting typically observe shorter startup times and less memory consumption. They can also fine-tune configuration to their specific scenarios.
```
danielmarbach: Should we try to use passive here and not mention Users and they? Something like

Self-hosting typically leads to shorter startup times and less memory consumption. With self-hosting, more control over the configuration is enabled.

weralabaj: @danielmarbach done https://github.com/Particular/docs.particular.net/pull/3267/files#diff-e766edf7c7e1bc214e324d65daa82a0c

andreasohlund:
How about we announce this internally in engineering ?

PR #3272, opened 2017-09-25
BLOCKED: Core split v2

SimonCropp:


SimonCropp:
also verified it works in VS without netcore installed

bording:
I saw the the SO answer pointing to using `UsingMicrosoftNETSdk` instead, so 👍 to that.

samples/azure/storage-persistence/ASP_2/Client/Client.csproj
```diff
@@ -1,7 +1,12 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
-  <PropertyGroup>
+  <PropertyGroup Condition=" '$(UsingMicrosoftNETSdk)' == 'true' ">
+    <TargetFrameworks>net461;netcoreapp2.0</TargetFrameworks>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(UsingMicrosoftNETSdk)' != 'true' ">
     <TargetFrameworks>net461</TargetFrameworks>
```
bording: Since this is a single framework, shouldn't these be `TargetFramework` instead of `TargetFrameworks`?

SimonCropp:
ok it seems u cant mix multiple TargetFrameworks with "multiple startup projects". ie it just uses the first TargetFrameworks and ignores the others. which means people cannot run our samples that target

marking this a blocked. if no one has any other ideas i will abandon this

SimonCropp:
spoke to @davkean. he said this is not intentional and to raise a bug. so: https://github.com/dotnet/project-system/issues/2855

SimonCropp:
update: https://github.com/dotnet/project-system/issues/2855 has been scheduled for VS 15.6 https://github.com/dotnet/project-system/milestone/19

weralabaj:
@bording Do you have enough context to continue this PR? I don't know what's that related to :/

bording:
@weralabaj This is related to moving samples away from needing two different solutions & projects for a single sample just to have samplea that runs on .NET Core.

The idea is that you could provide a sample that has multiple target frameworks, which allows you to pick which one to use. This PR uses the following condition to do that:
```
<PropertyGroup Condition=" '$(UsingMicrosoftNETSdk)' == 'true' ">
    <TargetFrameworks>net461;netcoreapp2.0</TargetFrameworks>
</PropertyGroup>
<PropertyGroup Condition=" '$(UsingMicrosoftNETSdk)' != 'true' ">
    <TargetFramework>net461</TargetFramework>
</PropertyGroup>
```
Because at the time this PR was opened, there was a chance that someone could have VS 2017, but not the .NET Core 2.0 SDK installed. Now that VS 2017 15.4 does actually install the 2.0 SDK, I think we probably don't need the condition.

However, because of the bug in Visual Studio, this can't be done yet. If you're multitargeting the project there's no way to have multiple projects start up and also pick which framework to use, so these samples become much harder to use.

I'd say that we should close this PR and once the bug is VS is resolved, we can consider making this change then.

weralabaj:
> I'd say that we should close this PR and once the bug is VS is resolved, we can consider making this change then.

Fine, but who will keep track of it?

bording:
> Fine, but who will keep track of it?

I'm watching the project system issue that Simon opened, so I guess I'll track it? Should we file an issue here as well to say that this should be revisited once it's fixed?

weralabaj:
I think we don't need to duplicate issues, if it's already tracked in other place then that's fine.

PR #3273, opened 2017-09-26
ServiceInsight SSL validation override document.

HEskandari:
Connects to https://github.com/Particular/ServiceInsight/issues/717


serviceinsight/self-signed-certificate-connection.md
```diff
@@ -0,0 +1,24 @@
...
+reviewed: 2016-09-28
+summary: ServiceInsight refuses to connect to a ServiceControl instance running on SSL with self-signed certificate.
+component: ServiceInsight
+tags:
+- SSL
+- ServiceControl
+- Certificate
+---
+
+ServiceControl can be configured to run on a SSL, but if a self-signed certificate is chosen, ServiceInsight would not allow connections to that instance. While running a self-sign certificate is not recommended, ServiceInsight provides an override configuration entry to allow you to connect to such instances.
```
weralabaj: run over SSL

weralabaj: > can be configured to run on a SSL

run **over** SSL

weralabaj: >  While running a self-sign certificate is not recommended

Should we explain why?

SimonCropp:
this will need a rebase

HEskandari:
I have marked this as WIP. It would require some changes as the configuration code has changed after the review by the SI maintainers.

serviceinsight/self-signed-certificate-connection.md
adamralph: Discussed with @Particular/serviceinsight-maintainers. We will hyperlink "running a self-sign certificate is not recommended".

HEskandari:
@adamralph a very good point. I'll add it to SI menu under the 'Usage' section.

weralabaj:
@HEskandari I see the linked issue has been closed https://github.com/Particular/ServiceInsight/issues/717, can we now complete this PR?

MarcinHoppe:
Technically speaking it is HTTPS, which can run on top of SSL (no longer recommended and pretty much obsolete) or TLS (the current version).

HEskandari:
@weralabaj Thanks for the reminder, I'll finalize it today.

HEskandari:
@weralabaj fast forwarding a few broken builds, cause reasons....I think this can be merged now.

adamralph:
@Particular/serviceinsight-maintainers should this document not mention that this feature only exists in 1.9.0 onwards? Also, I think we probably should not have closed https://github.com/Particular/ServiceInsight/issues/717 (and released 1.9.0) before this was merged.

HEskandari:
I don't think we track feature by version on SI documents. As an example, some of the 'views' like Sequence Diagramn are not available on older version of SI. Not saying this is a bad idea though. It is just that it is more than just this page if we want to keep things consistent.

👍 on your comment about the release.

PR #3280, opened 2017-09-27
Forwarding Address sample

mikeminutillo:
Enables an endpoint to leave a forwarding address for specific message types. This is useful when migrating a handler to another endpoint to capture any in-flight messages.


SzymonPobiega:
Looks good @mikeminutillo but I think we can make it much simpler if we could assume that forwarding is configured with transport addresses, not endpoints names. Given that forwarding is a niche scenario, I guess that's a fair assumption.

With this assumption we would only need a map of `Type` to `string` (representing the transport address of where to forward) and we could use a single behavior in the physical context or even in transport receive context (before the outbox). We would use types in the *enclosed message types* to lookup the destination. What's more, we could actually specify the forwarding with type name instead of actual type. This would give us freedom to not even reference the message assembly in the forwarding endpoint.

SimonCropp:
this will need a rebase

mikeminutillo:
@SzymonPobiega I disagree. I think this is the type of thing that might be in your environment for weeks as you reconfigure your routing and wait for in-flight messages to drain. For most transports it won't matter because Logical->Physical is a no-op. For MSMQ it will be more important for this to be logical mapping. I think we want to keep physical addresses out of code where possible.

SzymonPobiega:
@mikeminutillo OK, you convinced me. But still, I think a single behavior in the incoming pipeline should be enough. That behavior would take the endpoint name based on the incoming message type and run that through `ResolveRoute` and then fork to the routing context.

In general the sooner we execute the forwarding, the less invasive it is (i.e. we don't want to run the audit behaviors and so on).

weralabaj:
@SzymonPobiega @mikeminutillo can you maybe discuss it over a call and agree what to do with this sample, i.e. if change is needed (then mark as WIP) or should stay as is?

mikeminutillo:
@weralabaj we did have a call and agreed to leave this as two behaviors for now.

Route Distribution requires a logical message. That's why we do the routing in the Incoming Logical Message context.

Forking to the Routing Context requires a physical message. That's why we wait until we fall back to the Incoming Physical Message context to fork out. 

We did briefly consider adding a behavior to the Auditing Context to drop a header in saying we forwarded the message but we decided it wasn't needed for this sample.

PR #3286, opened 2017-09-28
Use new mutator api

andreasohlund:
We are moving away from providing DI for mutators in the future, see Particular/NServiceBus#4978 (comment). This PR prepares for this by guiding users to the `RegisterMutator` api that was introduced in v6


Snippets/Core/Core_6/Headers/Writers/HeaderWriterReply.cs
```diff
@@ -55,7 +52,7 @@ class MessageToReply :
         {
         }
 
-        class Mutator :
+        class Mutator :
```
bording: Trailing whitespace! @andreasohlund Time to install https://marketplace.visualstudio.com/items?itemName=MadsKristensen.TrailingWhitespaceVisualizer!

andreasohlund:
Installed, thanks for the heads up

PR #3290, opened 2017-09-28
cross namespace subscriptions

yvesgoeleven:
Describes cross namespace subscriptions for endpoint oriented topology


SeanFeldman:
Looks good, but failing on TC with `namespace_subscription_registered_endpoint` missing.

PR #3295, opened 2017-09-29
Make Pub/Sub sample a better landing page

DavidBoike:
Sounds like this sample is going to be used as a landing page for @indualagarsamy's monoliths blog post.

@dbelcham please take a look.

### DIFF
--- a/samples/pubsub/sample.md
+++ b/samples/pubsub/sample.md
@@ -10,9 +10,16 @@ redirects:
  - nservicebus/publish-subscribe-sample
 related:
  - nservicebus/messaging/publish-subscribe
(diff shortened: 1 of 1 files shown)


samples/pubsub/sample.md
```diff
@@ -10,9 +10,16 @@ redirects:
  - nservicebus/publish-subscribe-sample
 related:
  - nservicebus/messaging/publish-subscribe
+ - tutorials/intro-to-nservicebus
```
SimonCropp: intro-to-nservicebus make no more sense on this sample than it does on any other sample

PR #3297, opened 2017-09-30
Fixes #3287 - Include required subscription setting in SQL usage

### LINKED ISSUE #3287: Feedback: The SQL Usage section does not mention the required subscriptions settings
Was just following along and got an exception on startup. Might be nice to add a link to the subscription settings or to all three SQL options add a 'default'.

https://docs.particular.net/persistence/sql/?version=sqlpersistence_3#usage

Let me know what you'd like (if anything). More than happy to make the edit.

janpieterz:
fixes https://github.com/Particular/docs.particular.net/issues/3287


SimonCropp:
@janpieterz good catch thanks

PR #3300, opened 2017-10-02
Further tweaks to Outbox opening paragraphs

adamralph:


SzymonPobiega:
I merged this immediately because it sounds much better. On the other hand, I think that part of this should be rewritten. Especially that part about simulating. People might have a feeling that we are faking something. I'll think about how to rephrase it. Thanks @adamralph ❤️

PR #3303, opened 2017-10-02
[WIP] Dependency Injection Asp.Net Core sample

mauroservienti:
Solves this problem https://gitter.im/Particular/NServiceBus?at=59d1e36fb59d55b82360cb5e
I'm not sure if we need to add a Controller showing that the same `MyService` can be injected there as well.


mauroservienti:
build is failing:

```
System.Exception : Could not find nuget for alias 'aspnetcore'.
```

Give that the sample uses `Autofac` should I move it in the `Autofac` directory?

samples/dependency-injection/aspnetcore/core_7/Sample/Program.cs
```diff
@@ -0,0 +1,47 @@
...
+        await host.StopAsync()
+            .ConfigureAwait(false);
+    }
+
+    static void AttemptToLaunchBrowser(string address)
+    {
+        Console.WriteLine($"Attempting to open browser to: {address}");
+        try
+        {
+            using (Process.Start("explorer.exe", address))
```
mauroservienti: Being a net core only sample does this line makes this Windows only?

SimonCropp: Yes hence the try catch

SimonCropp:
The divergence is to allow supporting poping up the webpage

mauroservienti:
Being a web project a web page is automatically started even with the previous configuration when the project is started e.g. from VS

SimonCropp:
i have hit many bugs with iis express+vs2017+netcore. The current approach works around them. Happy to change it back if u want. I would just not prefer to shave the yaks

mauroservienti:
I haven't hit any, also web projects can be run as simple console apps not hosted in IIS at all. Anyway let's leave this as is.

SimonCropp:
Afaik if u run it in console mode it does not open the browser

mauroservienti:
VS, change startup mode from IIS to Console, F5 and browser pops up.

SimonCropp:
Is that without my process start change?

mauroservienti:
yes, just change this:

image

PR #3306, opened 2017-10-03
API usage for DisableInstaller configuration for MSMQ.

indualagarsamy:
Relates to https://github.com/Particular/NServiceBus.Transport.Msmq/pull/45


Snippets/MsmqTransport/MsmqTransport_1/DoNotCreateQueuesOnInstall.cs
```diff
@@ -0,0 +1,14 @@
+﻿using NServiceBus;
+
+class DoNotCreateQueuesOnInstall
+{
+    DoNotCreateQueuesOnInstall(EndpointConfiguration endpointConfiguration)
+    {
+        #region do-not-create-queues-on-install
+
+        var transport = endpointConfiguration.UseTransport<MsmqTransport>();
+        transport.DoNotCreateQueuesOnInstall();
```
andreasohlund: Have we considered switching to not create by default ? (and make creation opt-in)

indualagarsamy: By default the installers are not run. Therefore the default behavior is that these queues will not created. Queues are only created if the `EnableInstallers` API is called. This API was named explicitly so its clear to the user that queues will not be created even if EnableInstallers() is being called.

andreasohlund:
Got it

PR #3312, opened 2017-10-04
Doco and upgrade guide for new v7 critical error default

andreasohlund:
Related to https://github.com/Particular/NServiceBus/pull/4994

### DIFF
--- a/Snippets/Core/Core_6/CriticalErrors.cs
+++ b/Snippets/Core/Core_6/CriticalErrors.cs
@@ -50,17 +50,6 @@ namespace Core6
             #endregion
         }
 
-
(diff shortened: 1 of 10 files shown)


timbussmann:
I think these sentences need more explanation what that means for the application/process:

> The default behavior is to stop the endpoint should a critical error occur. 

> The default behavior is to shutdown the bus using

> The default behavior is to dispose the bus using

andreasohlund:
@timbussmann Added a WARN to clarify. See what you think?

andreasohlund:
Looks good in general, made a final tweak to the upgrade guide.Please merge
if it makes sense to you

PR #3314, opened 2017-10-05
WIP: Added other EIPs

ramonsmits:
I would want to add more products on that page besides BizTalk. We often get the question how is NServiceBus better than X where X is a product is not a service bus and I usually link to this page.

Maybe we need to restructure this page too and not only just sprinkle it with some more product names?

Thoughts @dvdstelt @udidahan @Particular/sales ?

### DIFF
--- a/nservicebus/architecture/index.md
+++ b/nservicebus/architecture/index.md
@@ -9,7 +9,7 @@ redirects:
 - nservicebus/architecture/nservicebus-and-biztalk
 ---
 
-NServiceBus is designed for communication between business-oriented services in SOA and DDD projects. It is not a replacement for RPC technologies such as WCF or broker integration tools like BizTalk. Successful systems use a mix of approaches and technologies for communication, not just NServiceBus.
+NServiceBus is designed for communication between business-oriented services in SOA and DDD projects. It is not a replacement for RPC technologies such as WCF or broker integration tools like BizTalk or other EIPs like TIBCO, Mule ESB or Boomi and many others. Successful systems use a mix of approaches and technologies for communication, not just NServiceBus.
 
 This article discusses the similarities and differences between NServiceBus and its Microsoft counterparts.
 


ramonsmits:
Might we want to include the following video on that page where Udi is explaining differences?

https://drive.google.com/file/d/0B_pXCFG6roi6UnNpcC04VjNBNkk/view

weralabaj:
> EIPs

You mean ESBs? We should not use the abbreviations if we don't explain them first, so make sure we have that in the article somewhere ;)

As for TIBCO - they have a few tools, so I think we need to be more specific as to which one are we talking about. If I remember correctly TIBCO randezvous is not a broker.

nservicebus/architecture/index.md
```diff
@@ -9,7 +9,7 @@ redirects:
 - nservicebus/architecture/nservicebus-and-biztalk
 ---
 
-NServiceBus is designed for communication between business-oriented services in SOA and DDD projects. It is not a replacement for RPC technologies such as WCF or broker integration tools like BizTalk. Successful systems use a mix of approaches and technologies for communication, not just NServiceBus.
+NServiceBus is designed for communication between business-oriented services in SOA and DDD projects. It is not a replacement for RPC technologies such as WCF or broker integration tools like BizTalk or other EIPs like TIBCO, Mule ESB or Boomi and many others. Successful systems use a mix of approaches and technologies for communication, not just NServiceBus.
```
udidahan: EIP = Enterprise Integration Patterns. That doesn't accurately describe the things you mentioned.

SimonCropp:
@ramonsmits thoughts on the feedback?

SimonCropp:
ok closign this one. the change doesnt actually make sense

"It is not a replacement for... EIPs like TIBCO, Mule ESB or Boomi and many others"

Even replacing EIP with ESB, we are a direct competitor to Mule ESB. so the change as is, is not valid

PR #3315, opened 2017-10-05
Add HandleCurrentMessageLater changes to v6 upgrade guide

timbussmann:
as this was supported in v5 and we're patching v6 to throw an exception when HandleCurrentMessageLater is used in conjunction with the Outbox (as this would cause a message loss in v6), we concluded that we should add this to the v6 upgrade guide as well.

related issue: https://github.com/Particular/NServiceBus/issues/4993

### DIFF
--- a/nservicebus/upgrades/5to6/handlers-and-sagas.md
+++ b/nservicebus/upgrades/5to6/handlers-and-sagas.md
@@ -52,6 +52,11 @@ The signature for the mutators now passes context arguments that give access to 
 See [header manipulation](/nservicebus/messaging/header-manipulation.md) for one example on how this might look.
 
 
+### HandleCurrentMessageLater and the Outbox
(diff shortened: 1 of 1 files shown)


weralabaj:
@timbussmann can you link a related issue? sounds like there's one. will be then easier to follow if somebody needs the history/context. thx!

SimonCropp:
so we now throw if someone uses HandleCurrentMessageLater  when Outbox is enabled?

weralabaj:
Seems that's what Tim said in the description, @SimonCropp are you suggesting we should be more specific in the doco and say explicitly it'll throw an exception? IMO seems it's enough to say it's not available, but no objections to adding more.

SimonCropp:
include the text of the exception so it is discoverable when people google it

PR #3320, opened 2017-10-05
splitting version specific settings into partials

SeanFeldman:
Connects to https://github.com/Particular/NServiceBus.AzureServiceBus/pull/646

@Particular/azure-maintainers looks like DescriptionFactory was replaced when we needed to split it into the old doco.


SeanFeldman:
@Particular/azure-maintainers please review and merge if no comments.

PR #3324, opened 2017-10-06
Update Application Insights sample discoverability

boblangley:
This PR:

- Adds tags / related to the meta data for the sample.
- Moves the sample under `samples/logging` from the root to make it more discoverable.


samples/logging/application-insights/sample.md
```diff
@@ -1,7 +1,15 @@
...
 title: Report metrics to Azure Application Insights
 component: Metrics
-reviewed: 2017-07-29
+reviewed: 2017-10-06
+tags:
+- Azure
+- Metrics
+related:
+ - nservicebus/operations/metrics
+ - samples/azure/
```
boblangley: Not sure if `samples/azure` is appropriate here. I would be nice if this sample could be listed under both `Logging and Metrics` and `Azure`

samples/logging/application-insights/sample.md
```diff
@@ -1,7 +1,16 @@
...
+summary: Shows how to report NServiceBus metrics data to Azure Application Insights
 component: Metrics
-reviewed: 2017-07-29
+reviewed: 2017-10-06
+tags:
+- Azure
+- Metrics
+related:
+ - nservicebus/operations/metrics
+ - samples/azure
```
boblangley: Not sure if `samples/azure` is appropriate here. I would be nice if this sample could be listed under both `Logging and Metrics` and `Azure` - commented again after fixing the path

samples/logging/application-insights/sample.md
```diff
@@ -1,7 +1,16 @@
 ---
 title: Report metrics to Azure Application Insights
+summary: Shows how to report NServiceBus metrics data to Azure Application Insights
 component: Metrics
-reviewed: 2017-07-29
+reviewed: 2017-10-06
+tags:
+- Azure
+- Metrics
```
DavidBoike: Content verification says "Each tag must have more than one page" ... guessing Metrics is the one that has only one. Add it to other related pages or remove the tag.

samples/logging/application-insights/sample.md
DavidBoike: Not sure how that will render. I think best course is to get this merged and see how it shows up and then adjust.

Issue #2931, opened 2017-07-06
Feedback: 'Scheduling'


SeanFeldman:
Feedback for 'Scheduling' https://docs.particular.net/nservicebus/scheduling/

Location in GitHub: https://github.com/Particular/docs.particular.net/blob/master/nservicebus/scheduling/index.md

What are the benefits/drawbacks of using `IPipelineContext` over `IEndpointInstance`/`IMessageSession`?

ping @Particular/nservicebus-maintainers

danielmarbach:
ping @Particular/nservicebus-maintainers

timbussmann:
@SeanFeldman do you mean using something like this instead of the sample code:

    await endpointInstance.ScheduleEvery(
        timeSpan: TimeSpan.FromMinutes(5),
        task: pipelineContext =>
        {
            var message = new CallLegacySystem();
            // use endpointInstance instead of piplineContext
            return endpointInstance.Send(message);
        })
    .ConfigureAwait(false);

as the scheduled task is basically executed similar to a handler, using the pipeline context does ensure outgoing messages are correctly batched and enlisted in the incoming transaction if possible (so no ghost messages when the scheduled tasks fails later on). It comes with the same behavior as in a regular message handler, so it would be incorrect to use the message session instead. Do you think this needs additional clarification on this page?

SeanFeldman:
Yes @timbussmann, I think that note/cla would be helpful. Have seen several occasions where `IEndpointInstance` was used instead of `IPipelineContext` causing exactly the issues you've described 🙂

timbussmann:
I'll see whether I can add some useful note to the existing docs.

timbussmann:
@SeanFeldman do you think a code comment in the snippet would be good enough?

SeanFeldman:
Better than nothing 😃

PR #3335, opened 2017-10-10
Add note about context parameter

timbussmann:
fixes https://github.com/Particular/docs.particular.net/issues/2931 by adding a comment to underline that you should use the context parameter to send messages instead of using the endpointInstance.

### DIFF
--- a/Snippets/Core/Core_6/Scheduling/Scheduling.cs
+++ b/Snippets/Core/Core_6/Scheduling/Scheduling.cs
@@ -18,6 +18,7 @@
                     timeSpan: TimeSpan.FromMinutes(5),
                     task: pipelineContext =>
                     {
+                        // use the pipelineContext parameter to send messages
                         var message = new CallLegacySystem();
                         return pipelineContext.Send(message);
                     })
--- a/Snippets/Core/Core_7/Scheduling/Scheduling.cs
+++ b/Snippets/Core/Core_7/Scheduling/Scheduling.cs
@@ -18,6 +18,7 @@
                     timeSpan: TimeSpan.FromMinutes(5),
                     task: pipelineContext =>
                     {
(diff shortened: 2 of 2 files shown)


weralabaj:
Looking at the original issue @timbussmann seems like we still didn't answer the question:

> What are the benefits/drawbacks of using IPipelineContext over IEndpointInstance/IMessageSession?

I'd additionally add a note covering what you said in the issue, i.e. "as the scheduled task is basically executed similar to a handler, using the pipeline context does ensure outgoing messages are correctly batched and enlisted in the incoming transaction if possible (so no ghost messages when the scheduled tasks fails later on). It comes with the same behavior as in a regular message handler, so it would be incorrect to use the message session instead."

right below the snippet.

timbussmann:
@weralabaj I wanted to avoid such a statement on this page, as this is a general applicable rule when it comes to using `IMessageSession` vs. `PipelineContext`. I think there is indeed a higher risk that somebody ignores the passed parameter for some reason and uses the instance, but that could could also be anywhere else essentially. My statement is valid for all pipeline related message operations in NSB6+, I'm not sure I want to have such a statement on scheduled task explicitly.

weralabaj:
I don't see a problem with that, the reader that lands on this page doesn't necessarily need a general guidance, and that's relevant in the context of that snippet. Or do I misunderstand?
It's quite easy to ignore the code comment, as that doesn't seem to be important and doesn't explain why you should use this thing.
But if you think it's too much, then just merge as is.

timbussmann:
> It's quite easy to ignore the code comment, as that doesn't seem to be important and doesn't explain why you should use this thing.

not sure whether readers ignoring comments would not also ignore other parts of doco.
I think this is a simple and small improvement which was based on @SeanFeldman feedback. I don't see a huge need for improving this section tbh.

Issue #3243, opened 2017-09-19
Feedback: 'ServiceControl'


LiamFeldman:
Feedback for 'ServiceControl' https://docs.particular.net/servicecontrol/

Location in GitHub: https://github.com/Particular/docs.particular.net/blob/master/servicecontrol/index.md

ping @Particular/servicecontrol-maintainers

The connection between SC and SC Instances is not explained on the SC landing page. Confusing for new users.

weralabaj:
@mikeminutillo do you still want to work on it or should I look for somebody else?

mikeminutillo:
There is some work happening in the background to try and make the difference between ServiceControl and ServiceControl Instances happening in the background. Once that has been sorted out, it will make updating the docs much clearer.

weralabaj:
@mikeminutillo what do you think about simply adding such a paragraph for now https://github.com/Particular/docs.particular.net/pull/3338? not sure what background work is happening, but still I think that info would be useful to have there.

weralabaj:
@LiamFeldman Mike clarified that relationship in this PR https://github.com/Particular/docs.particular.net/issues/3243. If you think there's still something to improve, please, let us know!

And thanks again for suggesting the improvement :)

PR #3338, opened 2017-10-10
clarify sc vs sc instances relationship

weralabaj:
Fixes https://github.com/Particular/docs.particular.net/issues/3243


mikeminutillo:
I'm not sure this quite right. You wouldn't normally have more than one ServiceControl instance per system being monitored. Part of the problem is that until Monitoring instances are introduced, we don't have anything to compare Monitoring instances to. I'll re-arrange some things and push my take.

mikeminutillo:
https://github.com/Particular/docs.particular.net/pull/3346

weralabaj:
I like the other version, makes sense, so merging. Do you think mentioning that one scenario to use multiple instances is when you use different transports would be useful, though?

Issue #3340, opened 2017-10-10
Feedback: 'Assembly scanning changes in Version 6'


SeanFeldman:
Feedback for 'Assembly scanning changes in Version 6' https://docs.particular.net/nservicebus/upgrades/5to6/assembly-scanning

Location in GitHub: https://github.com/Particular/docs.particular.net/blob/master/nservicebus/upgrades/5to6/assembly-scanning.md

The warning is incorrect for this upgrade guide. Version 6 is not a pre-release as the message indicates 
![image](https://user-images.githubusercontent.com/1309622/31399225-cc913582-ada8-11e7-92e4-348aae44b8e5.png)

ping @Particular/nservicebus-maintainers

PR #3341, opened 2017-10-11
Remove incorrect snippets

andreasohlund:
Only used in the 5 to 6 upgrade guide so its safe to remove

Closes https://github.com/Particular/docs.particular.net/issues/3340

### DIFF
--- a/Snippets/Core/Core_6/Scanning/ScanningPublicApi.cs
+++ b/Snippets/Core/Core_6/Scanning/ScanningPublicApi.cs
@@ -69,18 +69,6 @@ namespace Core6.Scanning
             #endregion
         }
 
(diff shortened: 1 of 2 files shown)


Snippets/Core/Core_6/Scanning/ScanningPublicApi.cs
```diff
@@ -69,18 +69,6 @@ void ScanningExcludeTypes(EndpointConfiguration endpointConfiguration, Type type
             #endregion
         }
 
-        void ScanningUpgrade(EndpointConfiguration endpointConfiguration)
```
weralabaj: IMO that one should stay as is.

andreasohlund: fixed

PR #3342, opened 2017-10-11
Guidance on MessageQueueException (0x80004005): Message Queue service is not available.

ramonsmits:
Based on https://eu1.salesforce.com/apex/Case?num=00031575

### DIFF
--- a/transports/msmq/troubleshooting.md
+++ b/transports/msmq/troubleshooting.md
@@ -50,8 +50,8 @@ The cause of this exception is that the MSMQ has run out of space for holding on
 This exception may occur if the MSMQ service is stopped or crashed.
 
 - If self-hosting, [make sure a critical error callback is configured](/nservicebus/hosting/critical-errors.md#custom-handling).
-- Make sure that [Windows Service Restart Recovery](/nservicebus/hosting/windows-service.md#installation-restart-recovery) is enabled to restart Windows Services automatically when they stop or crash.
-- [Create Windows Service depencies for your endpoint to the MSMQ service](/nservicebus/hosting/windows-service.md#installation-service-dependencies) for endpoints to automatically be stopped/started in case the MSMQ service is restarted (only then, they will not restart on a stop/start, only restart).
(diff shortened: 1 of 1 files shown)


transports/msmq/troubleshooting.md
```diff
@@ -45,6 +45,14 @@ at System.Messaging.MessageQueue.SendInternal(Object obj, MessageQueueTransactio
 
 The cause of this exception is that the MSMQ has run out of space for holding on to messages. This could be due to messages sent that could not be delivered, or messages received that have not been processed.
 
+## MessageQueueException (0x80004005): Message Queue service is not available.
```
weralabaj: @ramonsmits seems that is in the wrong place? should be under the next paragraph (### Resolution) I think, see here how it'll look like https://github.com/Particular/docs.particular.net/blob/32233ecd66b8777d1e500f59936b232659b31306/transports/msmq/troubleshooting.md - the resolution part corresponds to the previous problem only I believe.


As for the content how about:

This exception may occur if the MSMQ service is stopped or crashed.

### Resolution

- If self-hosting, ensure that the [critical error handling is configured](/nservicebus/hosting/critical-errors.md#custom-handling) correctly and custom callback method has been provided.
- Ensure that the [Windows Service Restart Recovery is enabled to restart Windows Services automatically when they stop or crash](/nservicebus/hosting/windows-service.md#installation-restart-recovery).
- For every endpoint [configure dependencies on the MSMQ service](/nservicebus/hosting/windows-service.md#installation-service-dependencies). The endpoints will then be automatically stopped/restarted in case the MSMQ service is restarted. Note the endpoints will be only restarted if the MSMQ service is restarted, but not if it's only stopped or only started.

weralabaj: One thing to clarify - don't all those resolution steps assume you're self-hosting? If so, shouldn't we skip or move or repeat "if self-hosting" from the first point?

ramonsmits:
@weralabaj 
>One thing to clarify - don't all those resolution steps assume you're self-hosting? If so, shouldn't we skip or move or repeat "if self-hosting" from the first point?

Steps 2 and 3 are applicable on any window service so self hosted and nservicebus.host ones.

PR #3343, opened 2017-10-11
Updated reference to updated API

dvdstelt:
## Issue 

SQL Transport [3,) concurrency documentation was referring to old API. Result was that customer on NSBv6 and SQL Transport v3 was unable to find how to change concurrency with current documentation.

## Solution

Change API reference and added link to documentation on how to change concurrency.

### DIFF
--- a/transports/sql/concurrency_content_sqltransport_2.partial.md
+++ b/transports/sql/concurrency_content_sqltransport_2.partial.md
@@ -1,0 +1,12 @@
+## Versions 2.0.x
+
+Each endpoint running SQL Server transport spins up a fixed number of threads (controlled by `MaximumConcurrencyLevel` property of `TransportConfig` section) for each input queue. Each thread runs in a loop, polling the database for messages awaiting processing.
+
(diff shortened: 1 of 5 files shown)


dvdstelt:
Question, should the link be added to older documentation as well?
If so, should every version point to concurrency doco reflecting its version, or the general concurrency url?

weralabaj:
Since the tuning page has multiple versions I'd keep it as is and add the similar link to other partials too. Unless it turns out that the link is different depending on the version (e.g. because the linked version doesn't exist, etc.).

dvdstelt:
Updated all 3 documents with the link to a specific version, where applicable.

weralabaj:
@dvdstelt personally, I'd just link to generic version in all partials but I see what you're doing there, makes sense IMO. we should get some form of remembering the version implemented in the engine :/

dvdstelt:
>personally, I'd just link to generic version in all partials

That's how I read your comment. I asked what the default is that we use :-)
I vote for changing it to that though, because your next remark made me remember that doing that is the default. 

> IMO. we should get some form of remembering the version implemented in the engine :/

That would be THE solution. It's tough currently navigating the website for older versions. :-)

weralabaj:
> I vote for changing it to that though, because your next remark made me remember that doing that is the default.

that may be hard to manage over time, especially once we have the "remember the version" feature (hope we will one day :)).

dvdstelt:
That's why I vote for changing it to 'not reference a specific version', although that wasn't really clear. So to restate, let's change it back to not include version number.

PR #3344, opened 2017-10-11
WIP - Cleanup MSMQ Deployment Scripts

indualagarsamy:
WIP - Separating powershell scripts from c# code for better readability.

TO DO:
Add more usage scripts for the powershell.


SimonCropp:
@Particular/docs-maintainers  not to comment yet on "if i think this is a good idea". but if this approach is taken it has some non trivial repercussions to keep content consistent. some other examples of pages that would need to change https://docs.particular.net/transports/sql/operations-scripting https://docs.particular.net/transports/sqs/operations-scripting  

also related https://github.com/Particular/DocsEngine/issues/333

indualagarsamy:
Thanks for the review @SimonCropp. I think it makes sense to address this at the docs engine level. I guess when we do that, it might make it easier if we keep the original format as is. I'll close this PR.

PR #3364, opened 2017-10-19
Azure configuration source replacement

yvesgoeleven:
Closes https://github.com/Particular/NServiceBus.Host.AzureCloudService/issues/88

Documents how users can achieve behavior similar to the original azure configuration source. How to retrieve config settings from either roleenvironment if available, or from xml config otherwise.

### DIFF
--- a/Snippets/CloudServicesHost/CloudServicesHost_8/AzureConfigurationSourceReplacement.cs
+++ b/Snippets/CloudServicesHost/CloudServicesHost_8/AzureConfigurationSourceReplacement.cs
@@ -1,5 +1,4 @@
-﻿using NServiceBus;
-using System;
+﻿using System;
 using System.Diagnostics;
 using System.IO;
 using System.Reflection;
@@ -9,7 +8,7 @@ using System.Web.Hosting;
 
 class AzureConfigurationSourceReplacement
 {
-    AzureConfigurationSourceReplacement(EndpointConfiguration endpointConfiguration)
+    string AzureConfigurationSourceReplacementExample()
     {
         #region azure-configuration-source-replacement
 
@@ -29,6 +28,8 @@ class AzureConfigurationSourceReplacement
         }
 
         #endregion
+
+        return value;
     }
 }
 
@@ -76,13 +77,7 @@ static class SafeRoleEnvironment
 
     }
 
-    public static bool IsAvailable
-    {
-        get
-        {
-            return isAvailable;
-        }
-    }
+    public static bool IsAvailable => isAvailable;
 
     public static string CurrentRoleInstanceId
     {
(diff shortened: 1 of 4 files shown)


yvesgoeleven:
@Particular/azure-maintainers ready for review

Snippets/CloudServicesHost/CloudServicesHost_8/AzureConfigurationSourceReplacement.cs
```diff
@@ -0,0 +1,242 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.Reflection;
+using System.Configuration;
+using System.Web.Configuration;
+using System.Web.Hosting;
+
+class AzureConfigurationSourceReplacement
+{
+    AzureConfigurationSourceReplacement(EndpointConfiguration endpointConfiguration)
```
SeanFeldman: `endpointConfiguration` is not used, remove?

Snippets/CloudServicesHost/CloudServicesHost_8/AzureConfigurationSourceReplacement.cs
```diff
@@ -0,0 +1,242 @@
...
+    }
+}
+
+#region configuration-resolver
+
+static class ConfigurationResolver
+{
+    public static Configuration GetConfigurationHandler()
+    {
+        if (IsWebsite()) return WebConfigurationManager.OpenWebConfiguration(HostingEnvironment.ApplicationVirtualPath);
```
SeanFeldman: Condition on the same line with execution? Let's do it right, on separate lines. 🙂

Snippets/CloudServicesHost/CloudServicesHost_8/AzureConfigurationSourceReplacement.cs
SeanFeldman: Hmmm, I think a snippet could be something like 

```c#
string GetConfigurationSection(string sectionName, string attributeName)
{
  // awesome code you provided
  return value;
}
```

I can guarantee people just going to copy/paste and use it as-is. So why to force them to create their own methods and go an extra step just to re-use the code? 🙂

Snippets/CloudServicesHost/CloudServicesHost_8/AzureConfigurationSourceReplacement.cs
```diff
@@ -0,0 +1,242 @@
...
+
+        return ConfigurationManager.OpenExeConfiguration(ConfigurationUserLevel.None);
+    }
+
+    static bool IsWebsite()
+    {
+        return HostingEnvironment.IsHosted;
```
SeanFeldman: good candidate for expression body

PR #3379, opened 2017-10-25
Documentation for the new transport configuration API for MSMQ

indualagarsamy:
Relates to transport API changes here: https://github.com/Particular/NServiceBus.Transport.Msmq/pull/54


indualagarsamy:
@Particular/msmq-transport-maintainers - please review.

transports/msmq/transportconfig_usenontransactionalqueues_msmqtransport_[1,).partial.md
```diff
@@ -0,0 +1,13 @@
+
+### UseNonTransactionalQueues
+
+This setting should be used with caution. As the queues are not transactional, any message that has an exception during processing will not be rolled back to the queue. Therefore this setting must only be used where loss of messages is an acceptable scenario.
```
andreasohlund: Perhaps a note that this is not the same as transactionmode.none and that all endpoints needs to agree on this setting to be able to comminicate?

transports/msmq/transportconfig_usenontransactionalqueues_msmqtransport_[1,).partial.md
```diff
@@ -0,0 +1,13 @@
+
+### UseNonTransactionalQueues
+
+This setting should be used with caution. As the queues are not transactional, any message that has an exception during processing will not be rolled back to the queue. Therefore this setting must only be used where loss of messages is an acceptable scenario. 
+
+While there may be a performance gain when using non-transactional queues, it should be carefully weighed against message loss. 
+
+snippet: use-nontransactional-queues
+
+WARN: Endpoints that use non-transactional queues will not be able to send messages to endpoints that use transactional queues.
```
indualagarsamy: @andreasohlund - I have a warning towards the end. How about combining both like this:

How about something like below:

WARN: This setting is not the same as using TransactionMode.None. This setting implies that the physical queue where the messages will be stored will not be a transactional queue. Endpoints that use non-transactional queues will not be able to send messages to endpoints that use transactional queues. Therefore it is important for all the endpoints that intend to communicate with each other to use the same setting.

PR #3390, opened 2017-10-27
Update installer workflow

weralabaj:
Based on Simon's support case reply. 

@SimonCropp please, review


persistence/sql/installer-workflow.md
```diff
@@ -16,34 +16,50 @@ WARNING: Read about [SQL Persistence NuGet Packages](/persistence/sql/#nuget-pac
...
 
 
 ### Development Workflow
 
-A standard development work operates as follows:
+The main purpose in this type of environment is having the `Just hit F5` experience, by automatically creating required tables in a local database (or remote database used only for development). Keep in mind that the risk and potential impact of accidentally corrupting the database schema or existing data is minimal in this environment.
+
+A typical workflow on developers machine consists of the following steps:
+
+ 1. In the project add references to both `NServiceBus.Persistence.Sql` and `NServiceBus.Persistence.Sql.MsBuild` packages.
```
weralabaj: Or maybe that should be below/above as a prerequisite, rather than one of the steps?

weralabaj:
Also thinking this section https://docs.particular.net/persistence/sql/#installation can be moved to this page and only linked to from index.md, thoughts on that?

weralabaj:
Also thinking... It may make sense to move the whole installation section https://docs.particular.net/persistence/sql/?version=sqlpersistence_3.1#installation to this page and have a page dedicated to the installation process?

I think script creation is fine in the index, as this is setting/configuration aspect, but installation is a whole process and that linking back and forth is becoming a bit awkward. Thoughts?

PR #3396, opened 2017-10-31
MSMQ Distributor Troubleshooting section

ramonsmits:
I created this troubleshooting section based on a recent support call.


transports/msmq/distributor/configuration.md
```diff
@@ -76,3 +76,87 @@ If self-hosting the endpoint here is the code required to enlist the endpoint wi
 snippet: ConfiguringWorker
 
 Similar to self-hosting, if running NServiceBus prior to Version 6, ensure the `app.config` of the worker contains the `MasterNodeConfig` section to point to the hostname where the distributor process is running.
+
+## Troubleshooting
```
indualagarsamy: Should the troubleshooting section be under configuration? I think it makes sense to be on its own markdown file?

transports/msmq/distributor/configuration.md
```diff
@@ -76,3 +76,87 @@ If self-hosting the endpoint here is the code required to enlist the endpoint wi
...
+
+Make sure that the maximum concurrency level is increased on the distributor. By default, the maximum concurrency level is 1. Also, make sure that the workers have the maximum concurrency level set higher.
+
+#### Prefetching
+
+When your workers are using NServiceBus 6, then there is the ability to configure the worker capacity independently from the maximum concurrency. By assigning a capacity value higher then the maximum concurrency you are able to reduce the latency to start working on a next message as by default the worker first reports it is available for work and then gets a new message forwarded by the distributor.
+
+#### Sender side distribution
+
+Using [sender side distribution](h/transports/msmq/sender-side-distribution) instead of using the distributor might be a better alternative if you are able to use NServiceBus 6 and is much more suitable for scenarios where IO is the bottleneck.
```
indualagarsamy: > might be a better alternative

that sounds prescriptive. Doesn't it depend on the pros and cons of each situation? i.e. sending lots of messages to a worker that's unavailable vs only sending ones that check in with the distributor for example?

transports/msmq/distributor/configuration.md
weralabaj: Given that this page is quite long already, I'm +1 for a separate troubleshooting page

transports/msmq/distributor/configuration.md
```diff
@@ -76,3 +76,87 @@ If self-hosting the endpoint here is the code required to enlist the endpoint wi
...
+```  
+Name: Distributor
+Version: 5.0.5
+Enabled by Default: No
+Status: Enabled
+Dependencies: None
+Startup Tasks: None
+```
+
+### The distributor queue is building up while it is running
```
weralabaj: > The distributor queue is building up while it is running

You mean a situation like "growing distributor queue"? or something like that?

transports/msmq/distributor/configuration.md
```diff
@@ -76,3 +76,87 @@ If self-hosting the endpoint here is the code required to enlist the endpoint wi
...
+
+The distributor is probably unaware of any workers. When a worker starts it enlists its capacity at the distributor and this event is written to the log.
```
weralabaj: > it enlists its capacity at the distributor

simpler would be "it signals that it's available to the distributor" or similar

transports/msmq/distributor/configuration.md
```diff
@@ -76,3 +76,87 @@ If self-hosting the endpoint here is the code required to enlist the endpoint wi
...
+
+Make sure that:
+
+- In the **worker** `App.Config`:
+  - The master node is correctly configured in the `MasterNodeConfig` configuration section.
+  - The `DistributorDataAddress` and `DistributorControlAddress` attributed have the correct values in the `UnicastBusConfig` configuration section.
+- The workers are running
+- The workers might require a maximum concurrency level higher then 1 if your monitoring your resources are not indicating any bottlenecks.
+ 
+### The distributor is doing all the work
```
weralabaj: > The distributor is doing all the work

how will you observe that?

transports/msmq/distributor/configuration.md
```diff
@@ -76,3 +76,87 @@ If self-hosting the endpoint here is the code required to enlist the endpoint wi
...
+
+When your workers are using NServiceBus 6, then there is the ability to configure the worker capacity independently from the maximum concurrency. By assigning a capacity value higher then the maximum concurrency you are able to reduce the latency to start working on a next message as by default the worker first reports it is available for work and then gets a new message forwarded by the distributor.
+
+#### Sender side distribution
+
+Using [sender side distribution](h/transports/msmq/sender-side-distribution) instead of using the distributor might be a better alternative if you are able to use NServiceBus 6 and is much more suitable for scenarios where IO is the bottleneck. 
+
+#### Do not run all distributors on a single box
+
+Distributor endpoints are often deployed on *the distributor machine*. This is one machine that hosts all distributor endpoints. This becomes the bottleneck under load as all messages flow throught this machine. An alternative is to have multiple distributor machines on physically different networks interface cards.
```
weralabaj: > have multiple distributor machines on physically different networks interface cards.

machines on cards or using cards, etc?

transports/msmq/distributor/configuration.md
```diff
@@ -76,3 +76,87 @@ If self-hosting the endpoint here is the code required to enlist the endpoint wi
...
+
+Alternative, when first hosting all endpoints on a single machine do not use the distributor to now run *all* endpoints on multiple workers. First balance your current endpoints by moving endpoints. Eventually you will have only a single endpoint running on a machine. If this machine now becomes your bottleneck then you might first want to upscale this first and if upscaling becomes too costly then make use of the distributor.
```
weralabaj: This section is really not clear to me:

> do not use the distributor to now run *all* endpoints on multiple workers

you mean that when you have the distributor and multiple workers on a single machine then each of the workers should not "run" all endpoints? not sure what's the scenario here...

>First balance your current endpoints by moving endpoints.

Balance... the load? or what? move endpoints where? between workers, i.e. that each worker runs some of the endpoints? can you clarify?

>  Eventually you will have only a single endpoint running on a machine.

How's that related to having many workers? How many workers will you have on each machine and what's preventing you from having one worker run multiple endpoints on a single machine?

> If this machine now becomes your bottleneck then you might first want to upscale this first and if upscaling becomes too costly then make use of the distributor.

I think we should start from this instead of having this at the end, but since this is troubleshooting guide I'm not clear why we consider here the scenario of not having a distributor at all? Maybe would make sense more in the "when to use distributor section" (another page) or maybe after reprhasing it'll be more visible why it's mentioned here.

transports/msmq/distributor/configuration.md
```diff
@@ -76,3 +76,87 @@ If self-hosting the endpoint here is the code required to enlist the endpoint wi
...
+
+#### Do not run all distributors on a single box
+
+Distributor endpoints are often deployed on *the distributor machine*. This is one machine that hosts all distributor endpoints. This becomes the bottleneck under load as all messages flow throught this machine. An alternative is to have multiple distributor machines on physically different networks interface cards.
+
+Alternative, when first hosting all endpoints on a single machine do not use the distributor to now run *all* endpoints on multiple workers. First balance your current endpoints by moving endpoints. Eventually you will have only a single endpoint running on a machine. If this machine now becomes your bottleneck then you might first want to upscale this first and if upscaling becomes too costly then make use of the distributor. 
+
+#### Upscaling
+
+The distributor is very chatty. Network latency and bandwidth will lower the overal throughput when the CPU isn't your current bottleneck, using the distributor might not be the ideal solution as using the distributor will result in roughly 4 times the number of send messages thus has a severy impact on your network resources.
```
weralabaj: > The distributor is very chatty.

That is very informal :P Also "very chatty" is relative, i'd be more precise, e.g. "For every message distributor generates X additional messages to coordinate the work" or something along those lines.

weralabaj: > Network latency and bandwidth will lower the overal throughput when the CPU isn't your current bottleneck, using the distributor might not be the ideal solution as using the distributor will result in roughly 4 times the number of send messages thus has a severy impact on your network resources

That needs to be reprhased

transports/msmq/distributor/configuration.md
```diff
@@ -76,3 +76,87 @@ If self-hosting the endpoint here is the code required to enlist the endpoint wi
...
+
+
+### Each worker is getting events directly, bypassing the distributor 
+
+The worker is very likely to not be running as a worker, but as a regular endpoint. In this configuration the worker will subscribe using its own address and not using the distributor.
```
weralabaj: How will you A) observe the problem B) verify that the endpoint is not running as a worker?

transports/msmq/distributor/configuration.md
```diff
@@ -76,3 +76,87 @@ If self-hosting the endpoint here is the code required to enlist the endpoint wi
 snippet: ConfiguringWorker
 
 Similar to self-hosting, if running NServiceBus prior to Version 6, ensure the `app.config` of the worker contains the `MasterNodeConfig` section to point to the hostname where the distributor process is running.
+
```
weralabaj: After reviewing the whole section I think it'd make sense to add a section to each point explaining how this behavior may be observed/verified. Something brief, e.g. "In the QueueExplorer (or another tool) you see... " or "message header X contains A instead of B".

transports/msmq/distributor/configuration.md
weralabaj: CPU and other metrics - idle, low on worker nodes, but distributor node is very busy

transports/msmq/distributor/configuration.md
ramonsmits: Add log entry text

transports/msmq/distributor/configuration.md
ramonsmits: Server monitoring indicates worker machines are IDLE while the machine hosting the distributor is busy.

transports/msmq/distributor/configuration.md
ramonsmits: Same event, with same message ID is sent to each worker instead of only a single copy being send to the distributor. resulting in more than once semantics.

transports/msmq/distributor/configuration.md
ramonsmits: severy => severe

ramonsmits:
@weralabaj I've added some big changes based on our conversation last week. I've also moved some content to different pages and reordered sections on the configuration page.

weralabaj:
I'll review today

weralabaj:
@indualagarsamy do you want to give it another look?

weralabaj:
@ramonsmits The build complains about non-existent link, and indeed this section https://github.com/Particular/docs.particular.net/pull/3396/files#diff-001422e1760a2a548bb529afcf3d9076L60 (Routing with the distributor, or here https://docs.particular.net/transports/msmq/distributor/#routing-with-the-distributor) was removed, was that intentional?
If yes, can you fix the build errors and either remove the links or point to a different location?

ramonsmits:
@weralabaj I moved that section to the configuration page to the bottom under *Advanced*. I'll check where is got linked from.

ramonsmits:
@weralabaj thanks for those grammar corrections!

weralabaj:
@ramonsmits here's the list of broken links https://builds.particular.net/viewLog.html?buildId=256730&buildTypeId=DocsWeb_ContentVerification

SeanFeldman:
Looks like very helpful changes for Distributor documentation.
Is there an ETA on this PR?

weralabaj:
No ETA, from my side there's only one outstanding issue - fixing links. Once this is done, it can be merged by anyone really.

ramonsmits:
@weralabaj I'm getting the following error:

```
Error:'Anchor link doesn't exist #override-distributor-queues'. Url:'#override-distributor-queues'.
```

Do not understand why as the following header exists:
```
### Override distributor queues
```

indualagarsamy:
@ramonsmits - can you rebase and check if it resolves your issue? Sometimes if the sections are added (after you've created your branch) that could happen.

weralabaj:
@ramonsmits I think the issue is that in order to avoid anchor links clashes on some pages, when generating anchor link we take into account all headers starting from h2 and nest them, i.e. in that particular case the anchor should be #advanced-override-distributor-queues (because the "Override distributor queues" is h3 or ### nested under "advanced" h2 or ## header). I'll modify and check if that resolves the issue.

ramonsmits:
> i.e. in that particular case the anchor should be #advanced-override-distributor-queues 

@weralabaj That makes sense, thanks for fixing!

PR #3400, opened 2017-11-02
Use AWS_REGION instead of AWS_DEFAULT_REGION

danielmarbach:
It seems that using AWS_REGION is the more commonly used approach over AWS_DEFAULT_REGION which is mostly used by CLI.


danielmarbach:
Please review

PR #3407, opened 2017-11-03
Complete metric sample with NewRelic

danielmarbach:


danielmarbach:
I'm still processing metric data behind the scenes so I might update the graphics with a bit more ups and downs to make it a better showcase

samples/logging/new-relic/Metrics_2/Endpoint/Endpoint.csproj
```diff
@@ -0,0 +1,16 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFrameworks>net461</TargetFrameworks>
+    <OutputType>Exe</OutputType>
+    <LangVersion>7.1</LangVersion>
+    <AutoGenerateBindingRedirects>false</AutoGenerateBindingRedirects>
```
timbussmann, line 7: not sure why this is here, but I had to remove this from the v8 version because this was causing assembly loading issues due core's reference to System.Memory

timbussmann: ugh, commented on the wrong PR 😆 ignore this.

PR #3408, opened 2017-11-04
small changes proposal

mikedevbo:
Hi,

Small changes proposal.

Regards
Michał


weralabaj:
@mikedevbo Thanks for the PR! I assume the reason for changes is to be consistent about transport usage across all tutorial lessons, is that correct?

@DavidBoike It looks good to me, I didn't smoke-test though. Not sure if using MSMQ in those two lessons was intentional or was it a simple omission when switching to learning transport a while ago? I assume the latter, but please, confirm.

DavidBoike:
Thanks @mikedevbo! Not sure how I missed this.

PR #3406, opened 2017-11-03
[WIP]Clarify how the uniform session behaves when inside or outside of the pipeline

andreasohlund:
WIP since we need to extract some doco pages on how IMessageHandler context behaves compared to IMessageSession

PR #3410, opened 2017-11-06
Clarify IUniformSession behavior

timbussmann:
An attempt to better state the behavior of IUniformSession and what it represents.
Removed the list about the different behavior, as https://docs.particular.net/nservicebus/messaging/send-a-message gives a better and completer explanation.

### DIFF
--- a/nservicebus/messaging/uniformsession.md
+++ b/nservicebus/messaging/uniformsession.md
@@ -26,7 +26,7 @@ Install the `NServiceBus.UniformSession` NuGet package available for endpoints u
 
 snippet: uniformsession-usage
 
-`IUniformSession` does represent either a `IMessageSession` or `IMessageHandlerContext`. When injecting `IUniformSession`, the instance will automatically resolve to the correct session type based on the call hierarchy. To learn more about the session types, read the ['sending messages' documentation page](/nservicebus/messaging/send-a-message.md).
+`IUniformSession` represents either a `IMessageSession` or `IMessageHandlerContext` depending on where it's used. Injected `IUniformSession` instances will automatically resolve to the correct session type based on the call hierarchy. To learn more about the session types, read the ['sending messages' documentation page](/nservicebus/messaging/send-a-message.md).
 
 
 ## Safeguards
--- a/nservicebus/messaging/uniformsession.md
+++ b/nservicebus/messaging/uniformsession.md
@@ -26,7 +26,7 @@ Install the `NServiceBus.UniformSession` NuGet package available for endpoints u
 
 snippet: uniformsession-usage
 
-`IUniformSession` represents either a `IMessageSession` or `IMessageHandlerContext` depending on where it's used. Injected `IUniformSession` instances will automatically resolve to the correct session type based on the call hierarchy. To learn more about the session types, read the ['sending messages' documentation page](/nservicebus/messaging/send-a-message.md).
+`IUniformSession` represents either a `IMessageSession` or `IMessageHandlerContext` depending on where it's used. Injected `IUniformSession` instances will automatically resolve to the correct session type based on the call hierarchy. To learn more about the session types, read the [Sending messages](/nservicebus/messaging/send-a-message.md) article.
 
 
 ## Safeguards
--- a/nservicebus/messaging/uniformsession.md
+++ b/nservicebus/messaging/uniformsession.md
@@ -26,10 +26,7 @@ Install the `NServiceBus.UniformSession` NuGet package available for endpoints u
 
 snippet: uniformsession-usage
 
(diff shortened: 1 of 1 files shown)


nservicebus/messaging/uniformsession.md
```diff
@@ -26,10 +26,7 @@ Install the `NServiceBus.UniformSession` NuGet package available for endpoints u
 
 snippet: uniformsession-usage
 
-Due to the duality of `IUniformSession` (it can represent either `IMessageSession` or `IMessageHandlerContext`), `ReusedComponent` will behave different depending on where it is used.
-
- * If used in the MVC controller as shown above all messages will be immediately dispatched
- * If used in the handler as shown above messages are [only dispatched when the handler completed](/nservicebus/messaging/batched-dispatch.md)
+`IUniformSession` does represent either a `IMessageSession` or `IMessageHandlerContext`. When injecting `IUniformSession`, the instance will automatically resolve to the correct session type based on the call hierarchy. To learn more about the session types, read the ['sending messages' documentation page](/nservicebus/messaging/send-a-message.md).
```
andreasohlund: >`IUniformSession` does represent either a `IMessageSession` or `IMessageHandlerContext`

How about: 

`IUniformSession` does represent either a `IMessageSession` or `IMessageHandlerContext` depending on where its used.

 Injected `IUniformSession` instances will automatically resolve to the correct session type based on the call hierarchy

andreasohlund: Or perhaps even shorter:

`IUniformSession` will either represent a `IMessageSession` or a `IMessageHandlerContext` by resolving to the correct session type based on the call hierarchy when injected.

danielmarbach: The problem is that now this text is detached from the sample code. The bullet list was in direct reference to the sample code written for that purpose which contains an MVC controller. Some people learn and memoize by seeing concrete examples. By removing the mentioning of the MVC controller we would also need to remove that part of the sample code which might make this harder to grasp for those people.

timbussmann: @andreasohlund brought up that the current statement on the MVC-bullet point is confusing as it's possible that users wrap the send operation in a custom `TransactionScope` which might delay the actual "dispatch" of the message. We agreed that it also doesn't make sense to go into too much details about the differences between session and context, as this is already documented.

danielmarbach: If you take into consideration the TransactionScope I agree. On the other hand wrapping sends in a TransactionScope is something that is not very common even in legacy applications. From the point on where NSB was opened for multiple transports the transaction scope approach was not supported on all transports. But here we are describing the actual behavior in comparison to message session and wrapping with a scope is a niche scenario. I'm not against changing it but if we do then we should probably also properly link the text to the code snippets shown. By removing the mentioning of the controller it is no longer the case.

andreasohlund:
@timbussmann does this replace https://github.com/Particular/docs.particular.net/pull/3406 ?

nservicebus/messaging/uniformsession.md
danielmarbach: I reflected a bit more about this, reread the text with the proposed changes and I'm Ok with changing it as @andreasohlund proposed in the shorter version sentence that I upvoted. I still think it would not have hurt to repeat the behaviour in reference to the above snippet since it prevents me from needing to go to another page but I guess if you and andreas feel different then I'm probably too biased towards letting it there and thus I have to let it go ;)

timbussmann:
@andreasohlund yes, that was the intention.

nservicebus/messaging/uniformsession.md
```diff
@@ -26,10 +26,7 @@ Install the `NServiceBus.UniformSession` NuGet package available for endpoints u
 
 snippet: uniformsession-usage
 
-Due to the duality of `IUniformSession` (it can represent either `IMessageSession` or `IMessageHandlerContext`), `ReusedComponent` will behave different depending on where it is used.
-
- * If used in the MVC controller as shown above all messages will be immediately dispatched
- * If used in the handler as shown above messages are [only dispatched when the handler completed](/nservicebus/messaging/batched-dispatch.md)
+`IUniformSession` represents either a `IMessageSession` or `IMessageHandlerContext` depending on where it's used. Injected `IUniformSession` instances will automatically resolve to the correct session type based on the call hierarchy. To learn more about the session types, read the ['sending messages' documentation page](/nservicebus/messaging/send-a-message.md).
```
danielmarbach: @Particular/docs-maintainers we weren't sure if using the ticks to separate the title in the link is aligned with other docs pages. Any thoughts?

weralabaj: FYI it's not ;) Usually we just put the title with capital letters (or however it's spelled on the page itself) as the link. see my change. no big deal though, that's typically fixed as part of the review process, so don't stress about it ;)

PR #3424, opened 2017-11-09
Doco for the new SC plugins

SzymonPobiega:
Relates to https://github.com/Particular/DotNetCoreLaunch/issues/27


SzymonPobiega:
@boblangley I moved the plugins doco for the new packages one level up, to separate folders but I do believe that should not be their final place. I think that SA doco should go somewhere under Sagas and that HB and CC doco should be moved to NServiceBus/Operations where the Metrics doco lives.

weralabaj:
> I think that SA doco should go somewhere under Sagas and that HB and CC doco should be moved to NServiceBus/Operations where the Metrics doco lives.

👍 But I think it should be linked in the menu in both places (i.e. where it is now + what you suggested)

SzymonPobiega:
@boblangley I think given it builds now, it is ready for review.

SzymonPobiega:
I've added missing upgrade guide and moved the content to correct place

PR #3433, opened 2017-11-14
Random isn't thread-safe

ramonsmits:


weralabaj:
@ramonsmits any context for this change? support case, question?

ramonsmits:
@weralabaj Was looking into sender side distribution for UMG and noticed this issue. Under concurrent load this isn't random at all due to it not being thread-safe.

PR #3438, opened 2017-11-16
Throttling troubleshooting with AmazonSQS

ramonsmits:


weralabaj:
@danielmarbach Do you still plan on reviewing it?

transports/sqs/index.md
```diff
@@ -79,3 +79,28 @@ The SQS transport uses the default [retries and timeouts](http://docs.aws.amazon
...
+```
+2017-11-14 23:10:24,314|ERROR|18|NServiceBus.Transports.SQS.MessageDispatcher|Exception from Send.
+Amazon.SQS.AmazonSQSException: Request is throttled. ---> Amazon.Runtime.Internal.HttpErrorResponseException: The remote server returned an error: (403) Forbidden. ---> System.Net.WebException: The remote server returned an error: (403) Forbidden.
+```
+
+When this happens while receiving a message then this can safely be ignored as (recoverability features](/nservicebus/recoverability/) should resolve this. If you are sending messages outside of a receiving message context you will have to be be aware that your send can fail. Especially when sending messages in bulk, specifically in parallel, you can get throttled if the sender has enough resources to suddenly flood the AmazonSQS infrastructure.
+
+A storage query that reads 100,000 items/records and for each a message gets produces and then sending these in parallel is vulneralble to get throttled.
+
+- Limit the amount of concurrent send tasks
```
danielmarbach: I think this is way to specific with the numbers and mixes vocabulary like parallel and concurrent. When I ran the handler code analysis I could show that mostly the semaphore approach is slower than useful batching.

Also not sure if it is feasible to link to a gist which is not maintained as part of this repo. I'll try to rework this a bit as soon as I have more time

danielmarbach: I have absolutely no capacity right now. Withdrawing myself from this PR. Maybe someone else from @Particular/aws-maintainers can chime in or we'll get back when we finally have time. Thanks @ramonsmits for raising it

transports/sqs/index.md
```diff
@@ -79,3 +79,28 @@ The SQS transport uses the default [retries and timeouts](http://docs.aws.amazon
...
+
+When sending messages it can happen that your sending is throttled. When that happens the following exception is logged.
+```
+2017-11-14 23:10:24,314|ERROR|18|NServiceBus.Transports.SQS.MessageDispatcher|Exception from Send.
+Amazon.SQS.AmazonSQSException: Request is throttled. ---> Amazon.Runtime.Internal.HttpErrorResponseException: The remote server returned an error: (403) Forbidden. ---> System.Net.WebException: The remote server returned an error: (403) Forbidden.
+```
+
+When this happens while receiving a message then this can safely be ignored as [recoverability features](/nservicebus/recoverability/) should resolve this. If you are sending messages outside of a receiving message context you will have to be be aware that your send can fail. Especially when sending messages in bulk, specifically in parallel, you can get throttled if the sender has enough resources to suddenly flood the Amazon SQS infrastructure.
+
+A storage query that reads 100,000 items/records and for each a message gets produces and then sending these in parallel is vulneralble to get throttled.
```
mauroservienti: I'd entirely remove this section and the rest of the changes. What mentioned before is enough IMO for everyone to a get a clear understanding of what could happen.

ramonsmits:
@mauroservienti @danielmarbach Updated it based on your feedback

mauroservienti:
What am I missing with this url `/nservicebus/handlers/async-handlers#large-amount-of-concurrent-message-operations`? It's the cause of the build failure

mauroservienti:
@Particular/docs-maintainers can you please take a look at the build error? I’m struggling to understand ho the link should be formatted. Thanks 😄

SzymonPobiega:
@mauroservienti I fixed a broken link

mauroservienti:
Thanks @SzymonPobiega

ramonsmits:
Rebased to hopefully get a succesfull build yet again....

ramonsmits:
@mauroservienti You requested changes, but was that only the link? That has been fixes.

transports/sqs/index.md
```diff
@@ -74,3 +74,20 @@ The SQS transport uses the default [retries and timeouts](http://docs.aws.amazon
...
+```
+2017-11-14 23:10:24,314|ERROR|18|NServiceBus.Transports.SQS.MessageDispatcher|Exception from Send.
+Amazon.SQS.AmazonSQSException: Request is throttled. ---> Amazon.Runtime.Internal.HttpErrorResponseException: The remote server returned an error: (403) Forbidden. ---> System.Net.WebException: The remote server returned an error: (403) Forbidden.
+```
+
+When this happens while receiving a message then this can safely be ignored as [recoverability features](/nservicebus/recoverability/) should resolve it. When sending messages outside a message handler those message operations can fail. Sending messages in bulk might increase the chance of being throttled on send to avoid flodding the Amazon SQS infrastructure.
+
+By batching multiple concurrent message operations together in a smart way it is possible to adhere to the throttling constraints. For example send a maximum of 100 messages concurrently as outlined in the [sending large amount of messages guidelines](/nservicebus/handlers/async-handlers.md#concurrency-large-amount-of-concurrent-message-operations).
```
SeanFeldman: > By batching multiple concurrent message operations together in a smart way

What do you mean by "in a smart way"?

SeanFeldman: > it is possible to adhere to the throttling constraints.

Do you mean "avoid throttling"?

SeanFeldman: > For example send a maximum of 100 messages concurrently as outlined in the [sending large amount of messages guidelines]...

Perhaps say "use batching" pointingn to the doco w/o stating a specific number as we don't do that in the referenced number. A batch could be 120 or 200 msgs as well.

ramonsmits:
@SeanFeldman I think I've addressed your feedback.

SeanFeldman:
I took the liberty to update is a little @ramonsmits.
Approved.
Once sentence that still doesn't sit right with me is
> Sending messages in bulk increases the chance of being throttled on send.

There's a subtle difference between sending messages in bulk in a single operation vs sending messages one per sending operation and grouping those operations together. Reading the sentence I couldn't determine which one you refer to. Suspect it's the latter.

ramonsmits:
@SeanFeldman Good change, I forgot the fact that the messages pump could also get throttled!

> For incoming messages this can be safely ignored as [recoverability feature](/nservicebus/recoverability/) should resolve it. For outgoing messages sent outside of a handler context, send operation can fail.

I don't agree, recovery would eventually move the message to the error queue. When sending large message sets within an incoming message we probably have a severe problem as we are using batched dispatch! Is batched dispatch with SQS doing any concurrent sends @Particular/aws-maintainers  ?

SeanFeldman:
I slightly paraphrased what the had originally @ramonsmits 🙂
Let's get on a call to discuss.

SeanFeldman:
In my discussion with @ramonsmits we've concluded the following.

Three scenarios where request could get throttled by SQS:
1. Incoming message (receiving)
1. Sending outside of a handler
1. Sending from within a handler

**Incoming message (receiving)**
Message will be retried until it's removed. If connection is throttled, delivery counter is not incremented - @Particular/aws-maintainers could you confirm this?

**Sending outside of a handler**

Large number of concurrent sends using `async/await`, using a list of tasks to be executed with `Task.WaitAll()` could cause throttling and partial message delivery as a result of throttling exception thrown for some of send operations. The failure would need to be handled manually.

**Sending from within a handler**
_Assuming_ no immediate dispatching is invoked, messages would be **NOT** be batched by the transport using [`SendMessagesBatch` SQS API](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_SendMessageBatch.html). Rather, they'll be sent out similar to how messages are sent out outside of a handler context, which could [result in throttling and ghost messages](https://github.com/Particular/NServiceBus.AmazonSQS/blob/deff356da78937cdea00dca846b7f5a4f0a6ea46/src/NServiceBus.AmazonSQS/MessageDispatcher.cs#L40-L46).

transports/sqs/index.md
```diff
@@ -74,3 +74,20 @@ The SQS transport uses the default [retries and timeouts](http://docs.aws.amazon
...
+
+### AmazonSQSException: Request is throttled
+
+SQS can throttle some send operations. When that happens the following exception is logged.
+```
+2017-11-14 23:10:24,314|ERROR|18|NServiceBus.Transports.SQS.MessageDispatcher|Exception from Send.
+Amazon.SQS.AmazonSQSException: Request is throttled. ---> Amazon.Runtime.Internal.HttpErrorResponseException: The remote server returned an error: (403) Forbidden. ---> System.Net.WebException: The remote server returned an error: (403) Forbidden.
+```
+
+For incoming messages this can be safely ignored as [recoverability feature](/nservicebus/recoverability/) should resolve it. For outgoing messages sent outside of a handler context, send operation can fail. Sending messages in bulk increases the chance of being throttled on send.
```
SeanFeldman: Based on our discussion with @ramonsmits, the outgoing part is not quite right.

SzymonPobiega:
@ramonsmits @SeanFeldman is anything blocking this one?

SeanFeldman:
@SzymonPobiega there is. 
1. Pinged @Particular/aws-maintainers with a question here https://github.com/Particular/docs.particular.net/pull/3438#issuecomment-359007389, but there was no response.
1.See https://github.com/Particular/docs.particular.net/pull/3438#issuecomment-361171297
@ramonsmits do you have time to update that section?

danielmarbach:
@SeanFeldman 

> Message will be retried until it's removed. If connection is throttled, delivery counter is not incremented - @Particular/aws-maintainers could you confirm this?

Can you elaborate what drove you to that conclusion? SQS transport uses a while loop for the immediate retries and retries until core tells it that the error was handled. So normal recoverability seems to apply here.

SeanFeldman:
@danielmarbach exactly the reason I've pinged 🙂 Thanks for pointing that out.

As for the other scenarios where throttling would take place, are those accurate?

danielmarbach:
@SeanFeldman The other part is accurate

SeanFeldman:
Thank you @danielmarbach 

@ramonsmits the info in https://github.com/Particular/docs.particular.net/pull/3438#issuecomment-359007389 updated and can be incorporated.

ramonsmits:
@SeanFeldman @danielmarbach please review my new changes.

transports/sqs/index.md
```diff
@@ -80,14 +80,40 @@ NOTE: NServiceBus will perform [immediate](/nservicebus/recoverability/#immediat
...
+
+When throttling happens the following exception is logged:
+
 ```
 2017-11-14 23:10:24,314|ERROR|18|NServiceBus.Transports.SQS.MessageDispatcher|Exception from Send.
 Amazon.SQS.AmazonSQSException: Request is throttled. ---> Amazon.Runtime.Internal.HttpErrorResponseException: The remote server returned an error: (403) Forbidden. ---> System.Net.WebException: The remote server returned an error: (403) Forbidden.
 ```
 
-For incoming messages this can be safely ignored as [recoverability feature](/nservicebus/recoverability/) should resolve it. For outgoing messages sent outside of a handler context, send operation can fail. Sending messages in bulk increases the chance of being throttled on send.
+Throttling is more likely to happen when sending a large number messages concurrently (in parallel). For example when using async/await, using a list of tasks to be executed with `Task.WaitAll()` could cause throttling and partial message delivery if any of the sends result in an thottling error.
```
danielmarbach: I'd remove the `(in parallel)` part. 

I'm not sure we need to be so specific and mention `a list of tasks with Task.WaitAll`. Does it really help to add this. I think sending messages concurrently is enough but that is just my 2 cents.

If you leave it as is at least replace `Task.WhenAll`

transports/sqs/index.md
```diff
@@ -80,14 +80,40 @@ NOTE: NServiceBus will perform [immediate](/nservicebus/recoverability/#immediat
...
+
+To avoid possible Amazon throttling errors, limit the maximum number of concurrent sends. For example, allow only a small amount of messages to be sent concurrently as outlined in the [sending large amount of messages](/nservicebus/handlers/async-handlers.md#concurrency-large-amount-of-concurrent-message-operations) guidelines or don't use concurrent sends at all and sequentially send your messages.
```
danielmarbach: > don't use concurrent sends at all and sequentially send your messages

how about just

> or send messages sequentially

transports/sqs/index.md
```diff
@@ -80,14 +80,40 @@ NOTE: NServiceBus will perform [immediate](/nservicebus/recoverability/#immediat
...
+Throttling can happen during any send or receive operation and can happen during the following scenarios:
+
+- Incoming message (receiving)
+- Sending from within a handler
+- Sending outside of a handler
+
+
+#### Incoming message (receiving)
+
+For incoming messages throttling errors can be safely ignored as the message pump will just try to fetch the next available message again.
```
danielmarbach: Remove `just`

transports/sqs/index.md
```diff
@@ -80,14 +80,40 @@ NOTE: NServiceBus will perform [immediate](/nservicebus/recoverability/#immediat
...
+
+#### Sending from within a handler
+
+The [recoverability feature](/nservicebus/recoverability/) should resolve it as any message sends and fails results in an exception which will be processed by the recoverability logic. If an incoming message would continiously result in a throttling error then the message will eventually be moved to the error queue.
```
danielmarbach: Failing message sends raise an exception when throttled. The exception will be handled by the recoverability feature...

An incoming message that continuously fails dude to throttling errors will be moved to the error queue.

danielmarbach: continuously

transports/sqs/index.md
```diff
@@ -80,14 +80,40 @@ NOTE: NServiceBus will perform [immediate](/nservicebus/recoverability/#immediat
...
+
+A throttling error could result in ghost messages. Ghost messages are messages send even though the handler failed due to not being able to transmit all messages to the transport succesfully. This makes a throttling error similar to any other error could and can occur both with the default [batched message dispatch](/nservicebus/messaging/batched-dispatch) or when using [immediate dispatch](nservicebus/messaging/send-a-message#dispatching-a-message-immediately).
```
danielmarbach: Ghost messages are only relevant if a handler produces multiple messages

danielmarbach: The second sentence is quite complex and hard to follow.

You want to say that ghost messages can occur regardless of the dispatch mode?

transports/sqs/index.md
```diff
@@ -80,14 +80,40 @@ NOTE: NServiceBus will perform [immediate](/nservicebus/recoverability/#immediat
...
+
+#### Sending outside of a handler
+
+As message sending does not happen within a handler any failures during sending will not rely on the [recoverability feature](/nservicebus/recoverability/). Any retry logic must be manually implemented. If a throttling exception occurs 1 or more message are likely not transmitted to the transport if these is no custom error logic.
```
danielmarbach: The first and second sentence applies to NSB in general. Do we have that documented somewhere else which we could link to?

The third sentence looks like a sentence fragment. Are you suggesting that users should always have this retry logic in place for sends outside the pipeline?

transports/sqs/index.md
ramonsmits: > Ghost messages are only relevant if a handler produces multiple messages

@danielmarbach I think that isn't true, also if it only sends a single message and the incoming message isn't ACKd or committed succesfully.

ramonsmits: > You want to say that ghost messages can occur regardless of the dispatch mode?

Yes, we do not have a definition of what a ghost message is and googling on the term also does not give you a definition.

ramonsmits:
@danielmarbach I've addressed most feedback. Removed some overly detailed info and rewritten a few lines for readability.

transports/sqs/index.md
danielmarbach: @ramonsmits yeah you are right. But that is by definition the case with transports that only support TX level receive only right? So should we just link to that doco page instead?

transports/sqs/index.md
```diff
@@ -102,17 +102,22 @@ Throttling can happen during any send or receive operation and can happen during
...
 
 #### Sending outside of a handler
 
-As message sending does not happen within a handler any failures during sending will not rely on the [recoverability feature](/nservicebus/recoverability/). Any retry logic must be manually implemented. If a throttling exception occurs 1 or more message are likely not transmitted to the transport if these is no custom error logic.
+As message sending does not happen within a handler any failures during sending will not rely on the [recoverability feature](/nservicebus/recoverability/). Any retry logic must be manually implemented.
+
+If no custom error logic is implemented and a throttling exception occurs then or more messages are likely not transmitted to the transport. Custom retry logic could either retry all regenerated messages again or have each individual message wrapped and retried.
```
danielmarbach: > If no custom error logic is implemented and a throttling exception occurs then or more messages are likely not transmitted to the transport.

How about

When throttling occurs with no custom error logic implemented one or more messages might not have been transmitted to the transport. The custom retry logic could either retry all messages to be sent again or have each individual message retried.

transports/sqs/index.md
```diff
@@ -74,3 +74,51 @@ The SQS transport uses the default [retries and timeouts](http://docs.aws.amazon
...
+Failing message sends raise an exception when throttled. The exception will be handled by the [recoverability feature](/nservicebus/recoverability/). An incoming message that continuously fails due to throttling errors will be moved to the error queue.
+
+A throttling error could result in partial message delivery while the incoming message is not processed succesfully and can occur regardless of using the default [batched message dispatch](/nservicebus/messaging/batched-dispatch) or when using [immediate dispatch](nservicebus/messaging/send-a-message#dispatching-a-message-immediately).
+
+Throttling errors are similar to any other technical error that can occur.
+
+
+#### Sending outside of a handler
+
+As message sending does not happen within a handler any failures during sending will not rely on the [recoverability feature](/nservicebus/recoverability/). Any retry logic must be manually implemented.
```
SeanFeldman: > As message sending does not happen within a handler any failures during sending will not rely on the recoverability feature.

How about

As message sending does not happen within a handler **context** any failures during sending will not ~~rely on~~ **be covered by** the recoverability feature.

transports/sqs/index.md
```diff
@@ -74,3 +74,51 @@ The SQS transport uses the default [retries and timeouts](http://docs.aws.amazon
...
+
+If no custom error logic is implemented and a throttling exception occurs then or more messages are likely not transmitted to the transport. Custom retry logic could either retry all regenerated messages again or have each individual message wrapped and retried.
```
SeanFeldman: I'm struggling with this sentence.

> If no custom error logic is implemented and a throttling exception occurs then or more messages are likely not transmitted to the transport.

was **"then or more"** meant to be something else that got removed?

> messages are likely not transmitted to the transport

technically messages were "transmitted to the transport", they were "not delivered to the SQS service".

ramonsmits:
@SeanFeldman @danielmarbach addressed your feedback.

danielmarbach:
🎆 thanks @ramonsmits for being the champion on this one!

PR #3444, opened 2017-11-17
Note on in-flight messages with previous serialization format

ramonsmits:


nservicebus/serialization/index.md
```diff
@@ -35,11 +35,13 @@ The [Newtonsoft JSON Serializer](newtonsoft.md) provides an effective general-pu
 
 The default serializer used in NServiceBus projects is the custom [XmlSerializer](xml.md).
 
+NOTE: If you are transitioning to a new serializer then messages that are currently 'in flight' are formatted using the previous format. These messages cannot be deserialized prior to NServiceBus 6. Make sure *all* queues of all endpoints are drained/empty. **This includes [Saga Timeout](/nservicebus/sagas/timeouts.md) and [Deferred/Delayed](/nservicebus/messaging/delayed-delivery.md) messages via timeout persistence.** Live migration is only support from NServiceBus 6 and higher.
```
bording: I'd prefer to see this note be in a partial so that we can have different content for NSB 6+. Right now it's confusing that the "These messages cannot be deserialized prior to NServiceBus 6" sentence is right in the middle of the warning, but it doesn't actually explain why this is true, and assumes you'll understand that it's referring to the additional serilzliers section below... which is in a 6+ partial.

Creating two versions of this next will clarify things. In the < 6 version, you can be clearer that this isn't supported, so you need to ensure there are no in-flight messages.

For the 6+ version, you can just say something about needing to add an additional serializer to handle this.

weralabaj:
@ramonsmits do you need some help with this?

ramonsmits:
@bording Applied changes based on your feedback.

bording:
I'd like to take one more look at this, but I haven't had a chance to do so yet.

weralabaj:
@bording Feel free to merge the PR once you review ;)

bording:
Instead of blocking this while I find time to do a review pass, I'll go ahead and merge it, and then do a followup PR later.

Issue #3311, opened 2017-10-04
Doco for the new conversation id override strategy


andreasohlund:
For the sample we'll likely suggest a custom attribute to flag the ID fields.

```
    class PlaceOrder   

        [BusinessId] //user defined attribute
        string OrderId;
```

Me and @weralabaj could come up with the following options that we can mention as well:

### Hardcode aka "Wall of code"

```
   if(context.Message is PlaceOrder)
       id = context.Message.OrderId;
```

### Marker interface aka "categorized wall of code"
```
interface IOrderRelatedMessage
    string OrderId;

interface ICustomerRelatedMessage
    string CustomerId;
```
### Convention on names + reflection aka "Magic"
```
    if(context.Message.IsCommand)
    {
        string entity = ParseEntityName(context.Message.Type);

        id = CachedReflectionInvoked(context.Message,$"{entity}Id")

        return true;
    }
  ```

weralabaj:
@andreasohlund not sure why we need to

> Record demo

where do you want to post it?

andreasohlund:
I was going to post in our demo slack channel?

weralabaj:
Ah ok, since it's a public repo I wasn't sure if you want to include it in the doco page or sth ;)

andreasohlund:
PR merged, will start with this shortly

weralabaj:
ping @andreasohlund

andreasohlund:
I was planning to do this as soon as https://github.com/Particular/docs.particular.net/pull/3447 is in

andreasohlund:
Doco PR raised https://github.com/Particular/docs.particular.net/pull/3615

andreasohlund:
Doco is merged, no longer convinced that we need a sample, closing this as done

PR #3447, opened 2017-11-19
Added snippet to show how to override conversation ID header value.

ramonsmits:


ramonsmits:
@andreasohlund I'm trying to override the value, but it doesn't work.

> Cannot set the NServiceBus.ConversationId header to 'XXXXX' as it cannot override the incoming header value ('2f4076a0-d8de-4297-9d18-a830015dd42a').

Why not??

andreasohlund:
We concluded that overriding should only be happening when sending outside
messagehandlers. Do you have a use case for overriding on
handlers?(essentially ”breaking” the message flow chain)

andreasohlund:
@timbussmann I remember that you had some concerns that we're being to strict?

Snippets/Core/Core_7/Headers/OverrideConversationId.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System.Threading.Tasks;
+using Core7.Handlers;
+using NServiceBus;
+
+public class ConversationIdHeaderOverride
```
andreasohlund: @ramonsmits I'm about to add doco for https://github.com/Particular/NServiceBus/blob/74a1c4bddfc5167deeb8fcd010fe0644f25823fa/src/NServiceBus.AcceptanceTests/Core/Causation/When_overriding_conversation_id_generation.cs#L53 

I wonder if we should make that the only documented option for v7? (or is there still a usecase for "poluting" the business code with setting the header)

nservicebus/messaging/headers.md
```diff
@@ -89,6 +89,11 @@ Identifier of the conversation that this message is part of. It enables the trac
 
 The first message that is sent in a new flow is automatically assigned a unique `Conversation Id` that is then propagated to all the messages that are subsequently sent, thus forming a _conversation_. Each message that is sent within a conversation also has a `RelatedTo` value that identifies the originating message that caused it to be sent. 
 
+There are scenarios where you want to override the Conversation Id because having a single huge conversation doesn't make sense. This can be done by overriding the header with a new value:
```
andreasohlund: I'd say its not about a single huge one but rather having a business related id instead to make things like `PlaceOrder` and `CancelOrder` show up in the same conversation (and display nicely in SI) by using the OrderId as the conversation id?

I was planning to pull this more philosophical part out to a dedicated page as part of https://github.com/Particular/docs.particular.net/issues/3311

So perhaps we can shorten this to "Overriding the ConversationId is done setting the conversation id header value when sending a message"

Then I can deal with the full explanation as part of #3311?

timbussmann:
@andreasohlund yeah I was wondering about the risk of allows allowing overwriting the conversationid. But so far I think we haven't heard from users having issues with the current behavior.

Snippets/Core/Core_7/Headers/OverrideConversationId.cs
ramonsmits: To be honest, I think the business code is the perfect place to indicate what the conversation ID is. I personally would want this magic per message to be hidden in a mutator or pipeline unless it can be really applied generically.

ramonsmits: For example, you could use a convention based on message ID, take the first part of a message ID to identify the conversation:

nservicebus/messaging/headers.md
DavidBoike: It can be about a single huge one. Picture a long-running process like a scheduler (or implemented as a "We finished one, send a message to do another") that periodically spawns "child conversations" where currently the entire conversation would span multiple months and thousands of messages.

ramonsmits: THat is exactly what I meant with huge, its never ending saga with timeouts.

ramonsmits:
@timbussmann We did, in the past I processed a case where this was an issue because in ServiceInsight it was impossible to look at the conversation.

ramonsmits:
@andreasohlund updated the current document but still would like to see the ability to override the behavior. The only way to do that would I think manipulate the incoming headers someway or to use IMessageSession within a receive operation which both really shouldn't be needed.

timbussmann:
> We did, in the past I processed a case where this was an issue because in ServiceInsight it was impossible to look at the conversation.

I think we had a v6 issue where it wasn't possible to set a custom id as it was always overriden by the pipeline at a later point.

Snippets/Core/Core_3/Headers/OverrideConversationId.cs
```diff
@@ -0,0 +1,12 @@
+﻿using NServiceBus;
```
timbussmann: why adding v3 snippets?

Snippets/Core/Core_6/Headers/OverrideConversationId.cs
```diff
@@ -0,0 +1,17 @@
...
+using NServiceBus;
+
+public class ConversationIdHeaderOverride
+{
+    public async Task Handle()
+    {
+        IMessageHandlerContext context = null;
+        #region override-conversation-id
+        var sendOptions = new SendOptions();
+        sendOptions.SetHeader(Headers.ConversationId, "MyCustomConversationId/" + System.Guid.NewGuid());
```
timbussmann: I think there is no scenario (except 3rd party integration) where an incoming message doesn't have a conversation id present. Given that this will fail, it doesn't seem to be a valid snippet to set a conversation id in a handler.

Snippets/Core/Core_3/Headers/OverrideConversationId.cs
ramonsmits: Because it wasn't any effort to add.

Snippets/Core/Core_6/Headers/OverrideConversationId.cs
ramonsmits: @timbussmann You are right, this was still under the assumption that is would work. I'll update the interface.

andreasohlund:
>but still would like to see the ability to override the behavior. The only way to do that would I think manipulate the incoming headers someway or to use IMessageSession within a receive operation which both really shouldn't be needed.

Agreed, can you raise a core issue to track this?

andreasohlund:
>pipeline unless it can be really applied generically.

That was the intent with guiding the users towards a custom attribute/interface/etc , see https://github.com/Particular/docs.particular.net/issues/3311

Ie to mark their business message in some way, but have the actual extraction be generic to not clutter the code that emits the message. Does that make sense?

nservicebus/messaging/headers.md
```diff
@@ -89,6 +89,16 @@ Identifier of the conversation that this message is part of. It enables the trac
 
 The first message that is sent in a new flow is automatically assigned a unique `Conversation Id` that is then propagated to all the messages that are subsequently sent, thus forming a _conversation_. Each message that is sent within a conversation also has a `RelatedTo` value that identifies the originating message that caused it to be sent. 
 
+There are scenarios where you want to manually set the Conversation Id, for example, a CancelOrder needs to become part of an existing order conversation. This can be done by overriding the header with a custom value:
+
+snippet: override-conversation-id
+
+It is not possible to override an existing Conversation Id, trying to do this results in the following error:
```
andreasohlund: Should this be a `NOTE:` ?

Perhaps also add that messages emitted from handlers will get the conversation id of the incoming message so its not really possible to override conversation ids from message handlers?

ramonsmits: @andreasohlund The exception that follows it is part of it. I don't think that can be combined.

andreasohlund: ok

ramonsmits:
@andreasohlund raised https://github.com/Particular/NServiceBus/issues/5088

nservicebus/messaging/headers.md
ramonsmits: @andreasohlund the section already has the following:

weralabaj:
@ramonsmits it's a bit hard to follow what's going on in this issue, please, once you resolve all the "content" issues and it's ready for docs maintainers review, request it explicitly to give us a signal, ok?

andreasohlund:
Build fails with 

>System.Exception : Some redundant snippets were detected. override-conversation-id (>= 4.0.0 && < 5.0.0) override-conversation-id (>= 3.0.0 && < 4.0.0)
   at RedundantSnippets.ThrowForRedundantSnippets(IReadOnlyList`1 availableSnippets, IReadOnlyList`1 usedSnippetKeys) in C:\BuildAgent\work\3ba50a7be36aeb6b\src\ParticularDocsCore\Sample\RedundantSnippets.cs:line 15
   at ArticleSnippetsReader.Finish() in C:\BuildAgent\work\3ba50a7be36aeb6b\src\ParticularDocsCore\ArticleSnippetsReader.cs:line 33
   at SiteRunner.RenderSite(String sourceDirectory, String targetDirectory, String lastShaPrefix) in C:\BuildAgent\work\3ba50a7be36aeb6b\src\ParticularDocsCore\SiteRunner.cs:line 37
   at ContentVerificationTest.RunFullSiteGeneration() in C:\BuildAgent\work\3ba50a7be36aeb6b\src\DocsWeb.Tests\ContentVerificationTest.cs:line 22

weralabaj:
the page has versions header saying it only covers version 5 and above so either we need to change that or remove the snippets

weralabaj:
rebased

PR #3452, opened 2017-11-21
Made the .net core 2.0 incompatibility more discoverable

ramonsmits:
Based on https://github.com/Particular/NServiceBus.SqlServer/issues/417

### DIFF
--- a/transports/sql/transactions_ambient-core-warning_sqltransport_[4,).partial.md
+++ b/transports/sql/transactions_ambient-core-warning_sqltransport_[4,).partial.md
@@ -1,1 +1,11 @@
-WARNING: The Transaction scope mode is not available in .NET Core 2.0 because the implementation of `SqlConnection` does not support enlisting in an ambient transaction.
+
+#### .NET Core 2.0 unsupported
+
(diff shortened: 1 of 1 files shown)


transports/sql/transactions_ambient-core-warning_sqltransport_[4,).partial.md
```diff
@@ -1 +1,12 @@
...
+
+The default transaction mode, i.e. transaction scope, is currently not supported in .NET Core 2.0. The attempt to use it will result in the following exception:
+
+```
+NServiceBus.Transport.SQLServer.MessagePump|Sql receive operation failed
+System.NotSupportedException: Enlisting in Ambient transactions is not supported.
+```
+Choose one of the other transaction modes to resolve this.
+
+WARNING: The Transaction scope mode is NOT available in .NET Core 2.0 because the implementation of `SqlConnection` does not support enlisting in an ambient transaction.
```
weralabaj: I think we can skip the warning, it's just repeating the same thing, I think a separate section is explicit enough.

PR #3450, opened 2017-11-21
Updated SQLP samples to target new releases (V3 and V4)

SzymonPobiega:
- [x] https://docs.particular.net/samples/saga/migration/
- [x] https://docs.particular.net/samples/sql-persistence/transitioning-correlation-ids/
- [x] https://docs.particular.net/samples/sql-persistence/saga-rename/
- [x] https://docs.particular.net/samples/sql-persistence/simple/
- [x] https://docs.particular.net/samples/saga/sql-sagafinder/
- [x] https://docs.particular.net/samples/outbox/sql/
- [x] https://docs.particular.net/samples/sqltransport-sqlpersistence/
- [x] update snippets
- [x] updated docs pages

PR #3454, opened 2017-11-21
Fix broken samples

dbelcham:
The TC build script for the samples and snippets on TeamCity has been broken for a while allowing broken solutions to get deployed. This PR fixes most of those broken samples/snippets, with the exception of the ones in the SQL Persistence PR #3450 

The build script fix will be put in place once this PR is ready and merged to `master` (after #3450 has been committed and ff'd into this PR. That will break all of the other open PRs at that time. They will need to pull in the changes from this (rebase from `master`). An announcement will need to be made to let those PR owners know what they have to do.

### DIFF
--- a/Snippets/NHibernate/NHibernate_8/Usage.cs
+++ b/Snippets/NHibernate/NHibernate_8/Usage.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using global::NHibernate.Cfg;
 using NServiceBus;
+using NServiceBus.Persistence;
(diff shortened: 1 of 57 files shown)


SeanFeldman:
Smoke tested v8 versions of `azure-service-bus`, `polymorphic-events-asb`, and `native-integration-asb`. Everything works.

weralabaj:
FYI https://github.com/Particular/docs.particular.net/pull/3450 has been merged

PR #3458, opened 2017-11-24
Merging of two tutorial pages into one

dvdstelt:
When searching for monitoring tutorials, I got several results in search.
I immediately expected the tutorials for monitoring to be on the page called "Tutorials", before I looked further in search results for "Monitoring tutorials".

Both pages are currently two rather small & empty tutorial pages. One for regular NServiceBus, the other for monitoring.
Why not merge them and have a single "Tutorials" page with all tutorials?


dvdstelt:
Question : Is it possible to keep the monitoring `index.md` but remove it from search results?

weralabaj:
> Question : Is it possible to keep the monitoring index.md but remove it from search results?

I don't think so, but why would you do that?

> Why not merge them and have a single "Tutorials" page with all tutorials?

@DavidBoike @mikeminutillo Do you have any objections? I don't have a context for that, to me it seems like a logical change.

dvdstelt:
> I don't think so, but why would you do that?

For people accessing https://docs.particular.net/tutorials/monitoring/ instead of https://docs.particular.net/tutorials/
I now removed the index, it will probably throw a 404 now? But a simple redirect would probably be even better ;-)

weralabaj:
Ah... I see :D That should solve the problem! https://github.com/Particular/docs.particular.net/pull/3458/commits/fda419351b5f210509d20de81c1a1b5a70d54f72

mikeminutillo:
:+1: I'm OK with that. The only reason I created the monitoring tutorials page in the first place is because the engine wouldn't let me have a folder with no index. We had no links or anything to it.

PR #3459, opened 2017-11-24
SC.Monitoring `Monitoring/EndpointName` setting

tmasternak:


servicecontrol/monitoring-instances/installation/creating-config-file.md
```diff
@@ -14,6 +14,14 @@ The configuration of a Monitoring instance can be adjusted via ServiceControl Ma
 
 Prior to modifying these configuration settings review [Setting a Custom Hostname](setting-custom-hostname.md):
 
+### Monitoring/EndpointName
+The endpoint name to be used by Monitoring instance and as a result it's [input queue](/nservicebus/endpoints/specify-endpoint-name.md#input-queue).
```
SeanFeldman: > The endpoint name to be used by Monitoring instance and as a result it's input queue.

The endpoint name to be used by Monitoring instance and it**s** ~~as a result it's~~ input queue.

mikeminutillo: In the setup guide we referred to this queue as the Monitoring queue. Should we re-use that term here?

> The name of the Monitoring queue. Will be used as the input queue for the endpoint.

tmasternak: @mikeminutillo I like the suggestion but I think that it's about misleading that the setting name is called `EndpointName` and the description will tell it defines Monitoring queue (btw I think that the setting should stay as it is). How about

>The endpoint name to be used by Monitoring instance and the name of the Monitoring queue.

PR #3466, opened 2017-11-27
A note added about different sets of metrics.

Scooletz:
This PR adds a note, clarifying that 

> depending on the version of the reporting package and the way metrics are gathered, the set of available metrics may vary.


Scooletz:

PR #3475, opened 2017-11-29
Upgrade Autofac version

mikedevbo:
Hi,

Quick fixes.

When press F5 there is an exception:
Could not load file or assembly 'Autofac, Version=4.0.1.0, Culture=neutral, PublicKeyToken=17863af14b0044da' or one of its dependencies. The located assembly's manifest definition does not match the assembly reference. (Exception from HRESULT: 0x80131040)

After upgrade Autofac to version 4.6.2.0 everything works as expected.


weralabaj:
Thanks @mikedevbo, I didn't get exceptions, but indeed there were warnings in the builds output. Thank you for addressing this.

PR #3490, opened 2017-12-05
ServiceControl Monitoring configuration - Host id info and queue rewrite.

ramonsmits:
Changes based on https://github.com/Particular/LaunchWave.DevOpsAdvancedMonitoring/issues/438

/cc @danielmarbach


ramonsmits:
@tmasternak This probably can be closed or maybe altered if we drop the instance API on the Metrics.ServiceControl plugin?

nservicebus/operations/metrics/service-control.md
```diff
@@ -24,16 +24,26 @@ snippet: SendMetricDataToServiceControl
...
+The default instance name is `particular.monitoring` which is also used as the input queue for ServiceControl Monitoring.
 
 partial: interval
 
 ### Instance Id
 
 An override for `$.diagnostics.hostid` and `$.diagnostics.hostdisplayname`.
 
-By default the monitoring plug-in will use [host identifiers](/nservicebus/hosting/override-hostid.md) to identify the monitored endpoint instance.
+This value is shown in the list of instances on the endpoint details page in ServicePulse Monitoring. By default the monitoring plug-in will use the [NServiceBushost host id hash identifier](/nservicebus/hosting/override-hostid.md) to identify the monitored endpoint instance.
```
tmasternak: This needs to be changed after https://github.com/Particular/LaunchWave.DevOpsAdvancedMonitoring/issues/438 has been fixed.

So it should be by: By default it will show HostId which is guid. One can override it with HostDisplayName. InsanceId should not be used unless using Azure CloudHost.

ramonsmits:
> I think that samples should use HostDisplayName.

@tmasternak What do you mean with that?

tmasternak:
@ramonsmits for example [this](https://github.com/Particular/docs.particular.net/pull/3490/files#diff-5da0120272964b47b6e103f79b3616c6R32).
InstanceId should be treated as a last resort option when user can do what they want with HostId and HostDisplayName. That's why the samples should show the best practice approach which is "Use HostId and HostDipslayName".

ramonsmits:
@tmasternak but we have that guidance already, and that is where I'm pointing to.

> It is recommended to [override the *host id* and *host display name* via NServiceBus core](/nservicebus/hosting/override-hostid.md) and to use the API without the `InstanceId` argument.

tmasternak:
@ramonsmits but the samples are not following our own guidance. The code should not show any usage of `instanceId` but `hostId` and `hostDisplayName` instead.

ramonsmits:
@tmasternak That is correct, the guidance on that already exists and linked to including a , this guidance is on API on the SC metrics. So the code corresponds to that API. The section of the snippet makes the intent (I think) pretty clear.

If we don't document this, then we must mark this API as obsolete.

ramonsmits:
Updated due to conflict.

tmasternak:
> That is correct, the guidance on that already exists and linked to including a , this guidance is on API on the SC metrics. So the code corresponds to that API. The section of the snippet makes the intent (I think) pretty clear.

@ramonsmits that's a good point!

PR #3491, opened 2017-12-05
Change the name of the file to match the content

WojcikMike:
Rename the file name to match the content


mikeminutillo:
Happy with the rename. Do we need to adjust the page title or put in any redirects?

PR #3493, opened 2017-12-06
Editorial review

MarcinHoppe:


udidahan:
I re-read the [section on multiple message types starting a saga](https://github.com/Particular/docs.particular.net/blob/master/nservicebus/sagas/index.md#multiple-message-types-starting-a-saga), specifically this part:

> Assume that message A should be processed before message B. Often in such a situation the saga will persist some data from message A to be read when the saga processes message B. However, if message B is processed first and starts the saga, then the data from message A has not been persisted yet.

That doesn't make sense to me.

MarcinHoppe:
@udidahan I took at stab at improving the description by referring to the example: #3508. Can you take a look?

udidahan:
It's quite a bit better. Thanks.

I've also made a tiny tweak to one sentence in PR #3513.

Let me know what you think.

MarcinHoppe:
I like it. Let's give @kbaley a chance to take a look at both PRs and merge.

PR #3498, opened 2017-12-07
Reordering logger config

ramonsmits:
I copy this snippet almost daily and never want to set the directory. Probably lots of our customers too.


SzymonPobiega:
❤️

PR #3503, opened 2017-12-11
Note on reverting SC hostname to localhost

ramonsmits:
Based on support call.


WilliamBZA:
Tweaked some of the grammar, but what is the context here?

ramonsmits:
@WilliamBZA Customer had configure a remote host, found out that it wasn't secure even though it uses IIS with ARR. Requested support, and I saw that it was not using localhost. Ran SCMU, set it to localhost so that the SC was not remotely accessible anymore.

ramonsmits:
Rebased to trigger new build

ramonsmits:
@WilliamBZA Can this be merged?

PR #3504, opened 2017-12-12
Switch Monitoring demo to direct download

mikeminutillo:
Building as a PR to check whether the build server will like it


mikeminutillo:
Seems to be OK

ramonsmits:
Can we test this somewhere?

mikeminutillo:
@ramonsmits I think you'd need to run the docs engine locally to test this. Or we push to prod, test it, and revert if it's broken.

PR #3522, opened 2017-12-20
How to import failed audits

SzymonPobiega:
Connects to https://github.com/Particular/PlatformDevelopment/issues/1593

@WilliamBZA not sure if you already started on this. In case you did we can compare/merge;-)

I think we should like to that doco from the SC's custom checks log message.


WilliamBZA:
I had started putting it in the main troubleshooting article, but it's probably better suited as a distinct page.

I wouldn't link directly to this page, but would like to a search that would find this page (we already do this type of thing in SC and SP)

WilliamBZA:
I suppose we should wait until we're ready to release 1.45.0 before merging this?

SzymonPobiega:
Yeah, let's wait. In case of SC I'd rather release a feature first and then doco because there is no way to mark doco with release numbers.

WilliamBZA:
Marking as WIP until we're ready to release

WilliamBZA:
WIP removed

PR #3540, opened 2018-01-04
[WIP] Wormhole ping pong sample Core V7

mauroservienti:
being based on MSMQ it's full framework only.

### DIFF
--- a/samples/wormhole/ping-pong/Wormhole_2/Client/Client.csproj
+++ b/samples/wormhole/ping-pong/Wormhole_2/Client/Client.csproj
@@ -1,0 +1,15 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFrameworks>net461</TargetFrameworks>
+    <OutputType>Exe</OutputType>
+    <LangVersion>7.1</LangVersion>
+  </PropertyGroup>
+  <ItemGroup>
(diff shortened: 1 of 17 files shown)


mauroservienti:
@SzymonPobiega can you please take a look? The sample compiles and seems to function however sent messages never arrive to the expected destination. They are picked up by GatewayA but never delivered to GatewayB.

SzymonPobiega:
@mauroservienti yeah, I'll check it today.

SzymonPobiega:
@mauroservienti it seems like the gateway itself it broken for the current beta of V7 :-( I need to fix it before this sample could be merged.

mauroservienti:
Thanks @SzymonPobiega I’ll leave this marked as WIP till the Gateway is fixed.

andreasohlund:
>it seems like the gateway itself it broken for the current beta of V7 :-( I need to fix it before this sample could be merged.

Is this still true?

SzymonPobiega:
Yeah. I didn't have time to work on this yet.

andreasohlund:
What is actually broken with the GW?

SzymonPobiega:
@andreasohlund The main issue was the difference in the distribution strategy APIs. It now compiles so I should be able to prepare an RC soon.

SzymonPobiega:
@andreasohlund after giving it some thought I decided to not work on the wormhole for now and see if the bridge can be easily used instead. I propose to close this for now.

PR #3543, opened 2018-01-05
MSMQ outgoing queues messages count

ramonsmits:
Based on a support incident. I could not find relevant info on the web so added it to our MSMQ troubleshooting.


transports/msmq/troubleshooting.md
```diff
@@ -51,6 +51,8 @@ at System.Messaging.MessageQueue.SendInternal(Object obj, MessageQueueTransactio
 
 The cause of this exception is that the MSMQ has run out of space for holding on to messages. This could be due to messages sent that could not be delivered, or messages received that have not been processed.
 
+Also check the Outgoing queues if messages send to remote servers are received and processed. By default a message remains in the Outgoing queue of the sender until the message is not only delivered but also processed at the receiver. For more information please read [MSMQ dead-letter queues](dead-letter-queues.md).
+
```
andreasohlund: Perhaps add that the reason the messages stay in the outgoing queues until processed is because NSB turns DLQ on by default?

ramonsmits:
@andreasohlund I've explicitly now mentioned that default is NServiceBus enabling this.

PR #3544, opened 2018-01-05
Monitoring 3rd party sample for Core V7

mauroservienti:


danielmarbach:
@mauroservienti I took the liberty to update this sample to use the `NServiceBus.CustomChecks` component. There was a build issue claiming there is no `CustomChecks7` component which made me realize that we switched to the uniform plugin approach.

danielmarbach:
@boblangley since you have been involved in creating that plugin could you please review?

danielmarbach:
Gave it a second look and LGTM. Merging

PR #3547, opened 2018-01-08
SQS ConnectionString not supported

danielmarbach:


transports/sqs/configuration-options.md
```diff
@@ -9,6 +9,8 @@ redirects:
 - nservicebus/sqs/configuration-options
 ---
 
+NOTE: The transport does not support `transport.ConnectionString(...)` to specify the connection string via code.
```
yvesgoeleven: maybe add the reason as well?

danielmarbach: I thought about it and typed the following:

> The transport reads relevant connection information from environment variables by using the standard AWS SDK credential source unless specified otherwise. 

But the I realized this statement would not be entirely true since versions before v4 do not support reading `AWS_REGION` variable. I felt a partial is overkill for the explanation. Should we still include it in the general note? Or do you have other suggestions?

danielmarbach:
Merging this we can always add more context if really required

PR #3548, opened 2018-01-08
SQS Read region from ENV variable

danielmarbach:
Closes https://github.com/Particular/docs.particular.net/pull/3548


timbussmann:
is that config necessary? You stated in another issue:

> versions before v4 do not support reading AWS_REGION variable.

so I'd assume that v4 does automatically respect the AWS_REGION variable?

danielmarbach:
@timbussmann v4 support the `AWS_REGION` variable because it uses the standard constructor of the SDK types. Versions before v4 specify the credential source automatically. Since they allow to overload the credential source we cannot assume the environment variable will always be read. With v4 it is different. We always use the default ctor, for fancy stuff you have to overload the factory and then it is the responsibility of the implementor to provide the necessary settings. 

By introducing the environment variable reading in this sample also for older versions we guaranteed a better transition to v4 and beyond and it simplifies the sample doc management since we don't need unnecessary partials.

PR #3551, opened 2018-01-08
SQS Partial for getting started and mentioning bucket only required for large messages

### LINKED ISSUE #3533: Feedback: 'Amazon SQS Transport' Getting Started section is confusing
Feedback for 'Amazon SQS Transport' https://docs.particular.net/transports/sqs/

Location in GitHub: https://github.com/Particular/docs.particular.net/blob/master/transports/sqs/index.md

The snippet in `Getting Started` is lacking a bit of context. The section states what environment variables are used for the access key and the secret. Directly afterwards is a snippet containing `var s3Configuration = transport.S3("myBucketName", "my/key/prefix");` without stating any information on why that configuration is required.

Also, a link to https://docs.particular.net/transports/sqs/configuration-options would probably be helpful. According to the configuration options page, the `Region` setting is mandatory although the `Getting Started` snippet does not contain any region configuration?

danielmarbach:
Closes #3533

Splits index into partial mentioning `AWS_REGION` and indicating that the S3 bucket is only required for sending large messages


Snippets/Sqs/Sqs_1/Usage.cs
```diff
@@ -7,6 +7,7 @@ class Usage
         #region SqsTransport
 
         var transport = busConfiguration.UseTransport<SqsTransport>();
+        // bucket only required for messages larger than 256KB
```
timbussmann: maybe clarify which parameter you referring to, e.g. "S3 bucket only required [...]"

Snippets/Sqs/Sqs_2/Usage.cs
```diff
@@ -8,6 +8,7 @@ class Usage
 
         var transport = endpointConfiguration.UseTransport<SqsTransport>();
         transport.Region("ap-southeast-2");
+        // bucket only required for messages larger than 256KB
```
timbussmann: should the s3 bucket configuration be a dedicated section instead?

danielmarbach: Not sure if really required for a getting started

timbussmann: > Not sure if really required for a getting started

that's what I meant. I'm not sure this needs to be in the getting started section at all?

danielmarbach: Left it as is. We can always remove it at a later stage if we think it is too complex for a getting started section.

PR #3552, opened 2018-01-08
Outbox - Section about concurrency and deduplication

ramonsmits:
Based on support call

### DIFF
--- a/nservicebus/outbox/index.md
+++ b/nservicebus/outbox/index.md
@@ -108,7 +108,7 @@ Try to avoid mixing transactional and non transactional tasks. If performing non
 
 ## Outbox expiration duration
 
-Part of the outbox is guarantee that the data remains consistent if the same message is processed more-than-once. A message with the exact same identification can be detected and ignored. This data is stored in the outbox record and kept for a configurable duration. The default expiration duration depends per persistence implementation. All implemetations can be configured on how long an outbox record needs to be stored and some can be configured on how often outbox cleanup needs to occur.
+Part of the outbox is to guarantee that the data remains consistent if the same message is processed more-than-once. A message with the exact same identification can be detected and ignored. This data is stored in the outbox record and kept for a configurable duration. The default expiration duration depends per persistence implementation. All implemetations can be configured on how long an outbox record needs to be stored and some can be configured on how often outbox cleanup needs to occur.
 
 Ensure that the retention period of the outbox is longer than the maximum time the message can be retried, including delayed retries and manual retries via ServiceControl. Additional care must be taken by operators of ServicePulse to not retry messages older than the outbox retention period.
 
@@ -124,7 +124,7 @@ NOTE: If the system is processing a high volume of messages per second having a 
 
 ## Outbox cleanup interval
 
-How outbox cleanup is performed is different per persistence implementation. Most persisters run a cleanup task every minute. Depending on the throughput this cleanup interval can be run more often to make a single cleanup operation purge less records per interval.
+How outbox cleanup is performed depends per persistence implementation. Most persisters run a cleanup task every minute. Depending on the throughput this cleanup interval can be run more often to make a single cleanup operation purge less records per interval.
 
 
 ## Persistence
--- a/nservicebus/outbox/index.md
+++ b/nservicebus/outbox/index.md
@@ -81,10 +81,7 @@ WARNING: When verifying outbox functionality, it can sometimes be helpful to tem
 
 ## Message identity and idempotent processing
 
-If handlers are idempotent then outbox might not be required at all. Outbox relies on message identity to guarantee consistency. When using outbox on both the sender and receiver message identifiers are consistent if a handler is invoked more-than-once for a message. 
-If a sender does not use the outbox or when sending a message outside of a handler the sender *must* generate and apply the same message identifier to guarantee consistency at the receiver.
-The same message identity is also important to have deduplication working on the infrastructure level on some transport that support this. [Read more on how to set a message identifier.](/nservicebus/messaging/message-identity.md)
-
+If handlers are idempotent then outbox might not be required at all. Outbox relies on message identity to guarantee consistency. When using outbox on both the sender and receiver message identifiers are consistent if a handler is invoked more-than-once for a message. If a sender does not use the outbox or when sending a message outside of a handler the sender *must* generate and apply the same message identifier to guarantee consistency at the receiver. [Read more on how to set a message identifier.](/nservicebus/messaging/message-identity.md)
 
 ## Concurrency
 
--- a/nservicebus/outbox/index.md
+++ b/nservicebus/outbox/index.md
@@ -98,6 +98,7 @@ The following scenarios might cause handlers to be invoked multiple times:
 
 In the above scenarios, multiple physical copies of the same logical message exist in the queue. If an endpoint allows concurrent message processing, or it is scaled out, its possible that the multiple copies of the message can be processed concurrently.
 
+
 ## Non transactional resources
 
 If non-transactional resources are being accessed then make sure processing is idempotent. Non-transactional resources are not guaranteed to be handled in an idempotent way by the Outbox. If more-than-once invocation is not wanted, then this can be prevented almost completely by making sure that only a single endpoint instance is running and having [configured that endpoint with a maximum concurrency of 1](/nservicebus/operations/tuning.md#tuning-concurrency). Only data modified via a shared storage session is made idempotent via the outbox and has exact-once semantics. See [persistence specific documention](/nservicebus/outbox/#Persistence) on how to obtain the shared storage session.
@@ -113,6 +114,14 @@ Part of the purpose of the outbox is to guarantee that the data remains consiste
 
 Ensure that the retention period of the outbox is longer than the maximum time the message can be retried, including delayed retries and manual retries via ServiceControl. Additional care must be taken by operators of ServicePulse to not retry messages older than the outbox retention period.
 
+
+## Outbox cleanup interval
+
+Outbox cleanup varies depending on the persistence that is chosen. Most persisters run a cleanup task every minute.
+
+Depending on the throughput of the system's endpoints, the Outbox cleanup interval can be run more frequently. This will allow each cleanup operation to purge the fewest records possible each time it runs, which will allow the cleanup to execute faster.
+
+
 ## Required storage
 
 Outbox requires storage. The amount of storage space required for Outbox can be calculated as follows:
@@ -122,13 +131,6 @@ Outbox requires storage. The amount of storage space required for Outbox can be 
 A single outbox record, after all transport operations have been dispatched, usually requires less than 50 bytes of which most are taken for storing the original message ID as this is a string value.
 
 NOTE: If the system is processing a high volume of messages, having a long deduplication timeframe will increase the amount of storage space that is required by Outbox.
-
-## Outbox cleanup interval
-
-Outbox cleanup varies depending on the persistence that is chosen. Most persisters run a cleanup task every minute.
-
-Depending on the throughput of the system's endpoints, the Outbox cleanup interval can be run more frequently. This will allow each cleanup operation to purge the fewest records possible each time it runs, which will allow the cleanup to execute faster.
-
 
 ## Persistence
 
--- a/nservicebus/outbox/index.md
+++ b/nservicebus/outbox/index.md
@@ -78,9 +78,23 @@ In order to gradually convert an entire system from the DTC to the Outbox:
 
 WARNING: When verifying outbox functionality, it can sometimes be helpful to temporarily [stop the MSDTC Service](https://technet.microsoft.com/en-us/library/cc770732.aspx). This ensures that the Outbox is working as expected, and no other resources are enlisting in distributed transactions.
 
+## Concurrency
+
+Outbox does not prevent invoking handlers or sagas more then once, it only prevents that any changes configured in persistence is only committed once. If an endpoint allows concurrent message processing or is scaled out it can happen that the same message is processed more then once. If you are using non-transactional resources then make sure processing is idempotent. If this is unwanted then make sure you only run a single endpoint and have [configured the endpoint to process messages sequentially by setting the maximum concurrency to 1](/nservicebus/operations/tuning.md#tuning-concurrency).
+
+## Deduplication
+
+Part of the outbox is deduplication which makes sure that a message with the exact same identification can be detected and ignored. After all messages have been send this data is stored for while. The default duration depends per persistence implementation and all can be configured on how long an outbox record needs to be stored and some can be configured on how often outbox cleanup needs to occur.
+
+Keep in mind that the maximum lifetime is not set too low! If the time window is set to 30 minutes, a message is forwarded to the error queue and manually retried by an operator potentially hours or days later while the original message was already correctly processed then this manual retry could apply the same data modifications again because the outbox record that prevents this is already purged as it was expired.
 
 ## Persistence
 
 The Outbox feature requires persistence in order to perform deduplication and store outgoing downstream messages.
 
-Refer to the dedicated pages for [NHibernate](/persistence/nhibernate/outbox.md), [RavenDB](/persistence/ravendb/outbox.md), [SQL](/persistence/sql/outbox.md) or [ServiceFabric](/persistence/service-fabric/outbox.md) persistence.
+Refer to the dedicated persistence pages:
+
+- [NHibernate](/persistence/nhibernate/outbox.md)
+- [RavenDB](/persistence/ravendb/outbox.md)
+- [SQL](/persistence/sql/outbox.md)
+- [ServiceFabric](/persistence/service-fabric/outbox.md)
--- a/nservicebus/outbox/index.md
+++ b/nservicebus/outbox/index.md
@@ -99,7 +99,7 @@ In those scenarios, multiple physical copies of the same logical message exist i
 
 ## Non transactional resources
 
-If non-transactional resources are being accessed then make sure processing is idempotent. Non-transactional resources are not guaranteed to be handled in an idempotent way by the Outbox. If more-than-once invocation is unwanted then this can be almost prevented completely by making sure only a single endpoint instance is running and have [configured the endpoint to process messages sequentially by setting the maximum concurrency to 1](/nservicebus/operations/tuning.md#tuning-concurrency). Only data modified via SynchronizedStorageSession is made idempotent via the outbox and has exact-once semantics.
+If non-transactional resources are being accessed then make sure processing is idempotent. Non-transactional resources are not guaranteed to be handled in an idempotent way by the Outbox. If more-than-once invocation is unwanted then this can be almost prevented completely by making sure only a single endpoint instance is running and have [configured the endpoint to process messages sequentially by setting the maximum concurrency to 1](/nservicebus/operations/tuning.md#tuning-concurrency). Only data modified via a shared storage session is made idempotent via the outbox and has exact-once semantics. See [persistence specific documention](#Persistence) on how to obtain the shared storage session.
 
 NOTE: This problem is not specific to Outbox but mentioned here as sometimes it is assumed Outbox does deduplication that prevents more-than-once invocation.
 
--- a/nservicebus/outbox/index.md
+++ b/nservicebus/outbox/index.md
@@ -78,15 +78,40 @@ In order to gradually convert an entire system from the DTC to the Outbox:
 
 WARNING: When verifying outbox functionality, it can sometimes be helpful to temporarily [stop the MSDTC Service](https://technet.microsoft.com/en-us/library/cc770732.aspx). This ensures that the Outbox is working as expected, and no other resources are enlisting in distributed transactions.
 
+
+## Message identity and idempotent processing
+
+If handlers are idempotent then outbox might not be required at all. When implementating idempotent processing keep in mind that any generated message *must* apply the same message identifier if invoked more then once to have deduplication working on the infrastructure level or to solve this at the logical level by validating message data.
+
+- [Message Identity](/nservicebus/messaging/message-identity.md)
+
+
 ## Concurrency
 
-Outbox does not prevent invoking handlers or sagas more then once, it only prevents that any changes configured in persistence is only committed once. If an endpoint allows concurrent message processing or is scaled out it can happen that the same message is processed more then once. If you are using non-transactional resources then make sure processing is idempotent. If this is unwanted then make sure you only run a single endpoint and have [configured the endpoint to process messages sequentially by setting the maximum concurrency to 1](/nservicebus/operations/tuning.md#tuning-concurrency).
+Outbox cannot prevent that handlers or sagas are invoked at most once. Under certain scenarios this can happen:
 
-## Deduplication
+- Sender sends multiple physical message with the same message identifier (submit order form)
+- Transport can send multiple messages when having a failover, they operate in a at-least-once delivery model
+- The outbox dispatch stage can be invoked more then once if the previous dispatch (partially) failed and is now retried.
 
-Part of the outbox is deduplication which makes sure that a message with the exact same identification can be detected and ignored. After all messages have been send this data is stored for while. The default duration depends per persistence implementation and all can be configured on how long an outbox record needs to be stored and some can be configured on how often outbox cleanup needs to occur.
+Multiple physical copied of the same logical message now exist. If an endpoint allows concurrent message processing or is scaled out it can happen that these multiple copies are processed concurrently.
+
+
+## Non transactional resources
+
+Outbox does not perform deduplication! If you are using non-transactional resources then make sure processing is idempotent. If this is unwanted then this can be almost prevented completely by making sure only a single endpoint instance and have [configured the endpoint to process messages sequentially by setting the maximum concurrency to 1](/nservicebus/operations/tuning.md#tuning-concurrency).
+
+NOTE: This problem is not specific to Outbox but mentioned here as sometimes it is assumed Outbox does deduplication.
+
+Try to avoid mixing transaction and non transactional tasks. If performing non transactional tasks send a message to perform this tasks in isolation.
+
+
+## Outbox expiration duration
+
+Part of the outbox is guarantee that the data remains consistent if the same message is processed more-than-once. A message with the exact same identification can be detected and ignored. This data is stored in the outbox record and kept for a configurable duration. The default expiration duration depends per persistence implementation. All implemetations can be configured on how long an outbox record needs to be stored and some can be configured on how often outbox cleanup needs to occur.
 
 Keep in mind that the maximum lifetime is not set too low! If the time window is set to 30 minutes, a message is forwarded to the error queue and manually retried by an operator potentially hours or days later while the original message was already correctly processed then this manual retry could apply the same data modifications again because the outbox record that prevents this is already purged as it was expired.
+
 
 ## Required storage
 
@@ -98,16 +123,9 @@ A single outbox record - after all transport operations have been dispatched - u
 
 NOTE: If the system is processing a high volume of messages per second having a large deduplication timeframe might not be ideal as outbox records will occupy storage.
 
-## Outbox cleanup
+## Outbox cleanup interval
 
-How outbox cleanup is performed is different per persistence implementation. Most persisters run a cleanup task every minute. Depending on the throughput this cleanup interval can be run more often to make a single cleanup operation purge less records
-
-
-## Message identity and idempotent processing
-
-If handlers are idempotent then outbox might not be required at all. When implementating idempotent processing keep in mind that any generated message *must* apply the same message identifier if invoked more then once to have deduplication working on the infrastructure level or to solve this at the logical level by validating message data.
-
-- [Message Identity](/nservicebus/messaging/message-identity.md)
+How outbox cleanup is performed is different per persistence implementation. Most persisters run a cleanup task every minute. Depending on the throughput this cleanup interval can be run more often to make a single cleanup operation purge less records per interval.
 
 
 ## Persistence
--- a/nservicebus/outbox/index.md
+++ b/nservicebus/outbox/index.md
@@ -25,36 +25,36 @@ The Outbox feature can be used instead of the DTC, to mimic the same level of co
 
 ## How it works
 
-The Outbox feature is implemented using the [Outbox](http://gistlabs.com/2014/05/the-outbox/) and the [Deduplication](https://en.wikipedia.org/wiki/Data_deduplication#In-line_deduplication) patterns.
+The Outbox feature is implemented using the [outbox](http://gistlabs.com/2014/05/the-outbox/) and the [deduplication](https://en.wikipedia.org/wiki/Data_deduplication#In-line_deduplication) patterns.
 
-Every time an incoming message is processed, a copy of that message is stored in the persistent _Outbox storage_. Whenever a new message is received, the framework determines whether that message has been processed already by checking for its presence in Outbox storage.
+Every time an incoming message is processed, a copy of that message is stored in the persistent _outbox storage_. Whenever a new message is received, the framework determines whether that message has been processed already by checking for its presence in outbox storage.
 
-If the message is not found in Outbox storage then it is processed in the regular way, shown in the following diagram:
+If the message is not found in outbox storage then it is processed in the regular way, shown in the following diagram:
 
 ![No DTC Diagram](outbox.svg)
 
 Processing a new incoming message consists of the following steps:
 
  * The handlers responsible for processing the message are invoked.
- * The _downstream messages_ (messages sent while processing the message, e.g. from inside handlers) are stored in Outbox storage and business data is saved. Both operations are executed within a single (non-distributed) transaction.
- * The downstream messages are sent and then marked as _dispatched_ in Outbox storage.
+ * The _downstream messages_ (messages sent while processing the message, e.g. from inside handlers) are stored in outbox storage and business data is saved. Both operations are executed within a single (non-distributed) transaction.
+ * The downstream messages are sent and then marked as _dispatched_ in outbox storage.
 
-If an incoming message is found in Outbox storage, then it is treated as a duplicate and is not processed again.
+If an incoming message is found in outbox storage, then it is treated as a duplicate and is not processed again.
 
-Even though an incoming message has been processed and business data has been saved, the framework might fail to immediately send the downstream messages. For example, the message queueing system may be unavailable, or the NServiceBus endpoint may be stopped before the messages can be dispatched. However, whenever an NServiceBus endpoint is started or is already running, and the framework detects downstream messages in Outbox storage which are not marked as dispatched, it will attempt to dispatch them at that time.
+Even though an incoming message has been processed and business data has been saved, the framework might fail to immediately send the downstream messages. For example, the message queueing system may be unavailable, or the NServiceBus endpoint may be stopped before the messages can be dispatched. However, whenever an NServiceBus endpoint is started or is already running, and the framework detects downstream messages in outbox storage which are not marked as dispatched, it will attempt to dispatch them at that time.
 
 Note: On the wire level, the Outbox guarantees `at-least-once` message delivery, meaning downstream messages may be sent and processed multiple times. At the handler level, however, the Outbox guarantees only transactionally exact-once or non-transactional at-least-once message processing, identical to distributed transactions. This higher guarantee level is due to the deduplication that happens on the receiving side.
 
 
 ## Important design considerations
 
-WARNING: Because the Outbox uses a single (non-distributed) database transaction to store all data, the business data and Outbox storage *must exist in the same database*.
+WARNING: Because the Outbox uses a single (non-distributed) database transaction to store all data, the business data and outbox storage *must exist in the same database*.
 
  * The Outbox feature works only for messages sent from NServiceBus message handlers.
  * Endpoints using DTC can only communicate with endpoints using the Outbox if either of the following conditions are satisfied:
    * Endpoints using the Outbox do not send messages to endpoints using DTC. However, endpoints using DTC can send messages to endpoints using the Outbox.
    * If endpoints using the Outbox send messages to endpoints using DTC, then the handlers processing those messages are [idempotent](https://en.wikipedia.org/wiki/Idempotence).
- * The Outbox may generate duplicate messages if outgoing messages are successfully dispatched but the _Mark as dispatched_ phase fails. This may happen for a variety of reasons, including _Outbox storage_ connectivity issues and deadlocks.
+ * The Outbox may generate duplicate messages if outgoing messages are successfully dispatched but the _Mark as dispatched_ phase fails. This may happen for a variety of reasons, including _outbox storage_ connectivity issues and deadlocks.
 
 
 ## Enabling the Outbox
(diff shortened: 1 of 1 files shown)


nservicebus/outbox/index.md
```diff
@@ -78,9 +78,45 @@ In order to gradually convert an entire system from the DTC to the Outbox:
 
 WARNING: When verifying outbox functionality, it can sometimes be helpful to temporarily [stop the MSDTC Service](https://technet.microsoft.com/en-us/library/cc770732.aspx). This ensures that the Outbox is working as expected, and no other resources are enlisting in distributed transactions.
 
+## Concurrency
+
+Outbox does not prevent invoking handlers or sagas more then once, it only prevents that any changes configured in persistence is only committed once. If an endpoint allows concurrent message processing or is scaled out it can happen that the same message is processed more then once. If you are using non-transactional resources then make sure processing is idempotent. If this is unwanted then make sure you only run a single endpoint and have [configured the endpoint to process messages sequentially by setting the maximum concurrency to 1](/nservicebus/operations/tuning.md#tuning-concurrency).
```
timbussmann: I'm not sure what the first two sentences really add to this section, is that related to concurrency?

> If an endpoint allows concurrent message processing or is scaled out it can happen that the same message is processed more then once.

this seems to mix up retries and concurrency a bit? the same message can be processed multiple times due to retries, but how can the same message be processed multiple times concurrently?

more **than** once

nservicebus/outbox/index.md
```diff
@@ -78,9 +78,45 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
+
+## Deduplication
+
+Part of the outbox is deduplication which makes sure that a message with the exact same identification can be detected and ignored. After all messages have been send this data is stored for while. The default duration depends per persistence implementation and all can be configured on how long an outbox record needs to be stored and some can be configured on how often outbox cleanup needs to occur.
```
timbussmann: How about:
The outbox ensures that messages with the same identifier are ignored. Identifiers of processed messages have to be stored for the deduplication to work correctly. The default storage duration depends on the choosen persistence and can be configured.

nservicebus/outbox/index.md
```diff
@@ -78,9 +78,45 @@ In order to gradually convert an entire system from the DTC to the Outbox:
 
 WARNING: When verifying outbox functionality, it can sometimes be helpful to temporarily [stop the MSDTC Service](https://technet.microsoft.com/en-us/library/cc770732.aspx). This ensures that the Outbox is working as expected, and no other resources are enlisting in distributed transactions.
 
+## Concurrency
+
+Outbox does not prevent invoking handlers or sagas more then once, it only prevents that any changes configured in persistence is only committed once. If an endpoint allows concurrent message processing or is scaled out it can happen that the same message is processed more then once. If you are using non-transactional resources then make sure processing is idempotent. If this is unwanted then make sure you only run a single endpoint and have [configured the endpoint to process messages sequentially by setting the maximum concurrency to 1](/nservicebus/operations/tuning.md#tuning-concurrency).
+
+## Deduplication
```
timbussmann: should that be `Deduplication storage`?

nservicebus/outbox/index.md
```diff
@@ -78,9 +78,45 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
+
+Part of the outbox is deduplication which makes sure that a message with the exact same identification can be detected and ignored. After all messages have been send this data is stored for while. The default duration depends per persistence implementation and all can be configured on how long an outbox record needs to be stored and some can be configured on how often outbox cleanup needs to occur.
+
+Keep in mind that the maximum lifetime is not set too low! If the time window is set to 30 minutes, a message is forwarded to the error queue and manually retried by an operator potentially hours or days later while the original message was already correctly processed then this manual retry could apply the same data modifications again because the outbox record that prevents this is already purged as it was expired.
```
timbussmann: how about:
Note: Setting the deduplication storage duration value too short can result in duplicate messages in case the message is retried after the configured storage duration.

nservicebus/outbox/index.md
```diff
@@ -78,9 +78,45 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
+
+## Required storage
+
+Outbox requires storage. The required space can be calculated as follows:
+
+    Total outbox records = Message througput per second * Deduplication period in seconds
+
+A single outbox record - after all transport operations have been dispatched - usually requires less then 50 bytes of which most are taken for storing the original message ID as this is a string value.
```
timbussmann: the "after all transport operations have been dispatched" adds no valuable information to the sentence.

timbussmann: less **than**

timbussmann: the "as this is a string value." part adds no valuable information to the sentence.

ramonsmits:
@andreasohlund Applied your feedback

nservicebus/outbox/index.md
```diff
@@ -78,9 +78,45 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
+
+## Deduplication
+
+Part of the outbox is deduplication which makes sure that a message with the exact same identification can be detected and ignored. After all messages have been send this data is stored for while. The default duration depends per persistence implementation and all can be configured on how long an outbox record needs to be stored and some can be configured on how often outbox cleanup needs to occur.
+
+Keep in mind that the maximum lifetime is not set too low! If the time window is set to 30 minutes, a message is forwarded to the error queue and manually retried by an operator potentially hours or days later while the original message was already correctly processed then this manual retry could apply the same data modifications again because the outbox record that prevents this is already purged as it was expired.
+
+## Required storage
+
+Outbox requires storage. The required space can be calculated as follows:
```
timbussmann: how about
"The required storage space required by the Outbox can be calculated as follows:"

nservicebus/outbox/index.md
```diff
@@ -78,9 +78,45 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
+
+    Total outbox records = Message througput per second * Deduplication period in seconds
+
+A single outbox record - after all transport operations have been dispatched - usually requires less then 50 bytes of which most are taken for storing the original message ID as this is a string value.
+
+NOTE: If the system is processing a high volume of messages per second having a large deduplication timeframe might not be ideal as outbox records will occupy storage.
+
+## Outbox cleanup
```
timbussmann: skip this section

nservicebus/outbox/index.md
```diff
@@ -78,9 +78,45 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
+
+How outbox cleanup is performed is different per persistence implementation. Most persisters run a cleanup task every minute. Depending on the throughput this cleanup interval can be run more often to make a single cleanup operation purge less records
+
+
+## Message identity and idempotent processing
+
+If handlers are idempotent then outbox might not be required at all. When implementating idempotent processing keep in mind that any generated message *must* apply the same message identifier if invoked more then once to have deduplication working on the infrastructure level or to solve this at the logical level by validating message data.
+
+- [Message Identity](/nservicebus/messaging/message-identity.md)
```
timbussmann: use a sentence instead of a single bullet point

nservicebus/outbox/index.md
```diff
@@ -78,9 +78,45 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
+
 
 ## Persistence
 
 The Outbox feature requires persistence in order to perform deduplication and store outgoing downstream messages.
```
timbussmann: this information seems to be quire redundant at this point in the document?

nservicebus/outbox/index.md
```diff
@@ -78,9 +78,45 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
 
-Refer to the dedicated pages for [NHibernate](/persistence/nhibernate/outbox.md), [RavenDB](/persistence/ravendb/outbox.md), [SQL](/persistence/sql/outbox.md) or [ServiceFabric](/persistence/service-fabric/outbox.md) persistence.
+Refer to the dedicated persistence pages:
```
timbussmann: For more information refer to the [...]

nservicebus/outbox/index.md
ramonsmits: @timbussmann It does, as when the messages have not yet been dispatches that contains all serialized messages which can take up megabytes of data.

ramonsmits: Same with the string data, those two fields are non fixed length values which cannot be calculated upfront but can have a huge impact on the required storage. So maybe I need to word it differently, but not sure how.

ramonsmits:
@timbussmann Partially applied your feedback

nservicebus/outbox/index.md
timbussmann: > but can have a huge impact on the required storage

but unlikely given that by default the message id is defined by the core and that is a very well known size?

also, this seems to be very dependent on the actual storage mechanism, so I still feel that "as this is a string value" adds no real value to this section.

nservicebus/outbox/index.md
```diff
@@ -79,8 +79,60 @@ In order to gradually convert an entire system from the DTC to the Outbox:
 WARNING: When verifying outbox functionality, it can sometimes be helpful to temporarily [stop the MSDTC Service](https://technet.microsoft.com/en-us/library/cc770732.aspx). This ensures that the Outbox is working as expected, and no other resources are enlisting in distributed transactions.
 
 
+## Message identity and idempotent processing
+
+If handlers are idempotent then outbox might not be required at all. When implementating idempotent processing keep in mind that any generated message *must* apply the same message identifier if invoked more-then-once to have deduplication working on the infrastructure level or to solve this at the logical level by validating message data. [Read more on how to set a message identifier.](/nservicebus/messaging/message-identity.md)
```
timbussmann: > When implementating idempotent processing keep in mind that any generated message *must* apply the same message identifier if invoked more-then-once to have deduplication working on the infrastructure level or to solve this at the logical level by validating message data.

this part is a bit confusing. Are you referring to the scenario where you have an endpoint without outbox sending messages to an outbox enabled endpoint?

nservicebus/outbox/index.md
```diff
@@ -79,8 +79,60 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
+
+
+## Concurrency
+
+Outbox cannot prevent that handlers or sagas are invoked at most once. Under certain scenarios this can happen:
```
timbussmann: how about:
Outbox cannot guarantee that handlers or sagas are invoked exactly once.

nservicebus/outbox/index.md
```diff
@@ -79,8 +79,60 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
+- Sender sends multiple physical message with the same message identifier (submit order form)
+- Transport can send multiple messages when having a failover, they operate in a at-least-once delivery model
+- The outbox dispatch stage can be invoked more then once if the previous dispatch (partially) failed and is now retried.
+
+Multiple physical copied of the same logical message now exist. If an endpoint allows concurrent message processing or is scaled out it can happen that these multiple copies are processed concurrently.
+
+
+## Non transactional resources
+
+Outbox does not perform deduplication! If you are using non-transactional resources then make sure processing is idempotent. If this is unwanted then this can be almost prevented completely by making sure only a single endpoint instance and have [configured the endpoint to process messages sequentially by setting the maximum concurrency to 1](/nservicebus/operations/tuning.md#tuning-concurrency).
```
SzymonPobiega: Do not use exclamation marks. And in fact outbox does involve deduplication when concurrency is set to 1. I would suggest just mentioning that non-transactional resources are not guaranteed to be handled in an idempotent way. Only the data updated via `SynchronizedStorageSession` is.

nservicebus/outbox/index.md
```diff
@@ -79,8 +79,60 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
+If handlers are idempotent then outbox might not be required at all. When implementating idempotent processing keep in mind that any generated message *must* apply the same message identifier if invoked more-then-once to have deduplication working on the infrastructure level or to solve this at the logical level by validating message data. [Read more on how to set a message identifier.](/nservicebus/messaging/message-identity.md)
+
+
+## Concurrency
+
+Outbox cannot prevent that handlers or sagas are invoked at most once. Under certain scenarios this can happen:
+
+- Sender sends multiple physical message with the same message identifier (submit order form)
+- Transport can send multiple messages when having a failover, they operate in a at-least-once delivery model
+- The outbox dispatch stage can be invoked more then once if the previous dispatch (partially) failed and is now retried.
```
timbussmann: more than once

timbussmann: previous dispatch (partially) failed and is retried.

nservicebus/outbox/index.md
timbussmann: > Under certain scenarios this can happen:

this is abit confusing as it gives the impression the following bullet points are a sequence of actions. how about:

> The following scenarios might cause handlers to be invoked multiple times:

nservicebus/outbox/index.md
```diff
@@ -79,8 +79,60 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
+
+Try to avoid mixing transaction and non transactional tasks. If performing non transactional tasks send a message to perform this tasks in isolation.
+
+
+## Outbox expiration duration
+
+Part of the outbox is guarantee that the data remains consistent if the same message is processed more-than-once. A message with the exact same identification can be detected and ignored. This data is stored in the outbox record and kept for a configurable duration. The default expiration duration depends per persistence implementation. All implemetations can be configured on how long an outbox record needs to be stored and some can be configured on how often outbox cleanup needs to occur.
+
+Keep in mind that the maximum lifetime is not set too low! If the time window is set to 30 minutes, a message is forwarded to the error queue and manually retried by an operator potentially hours or days later while the original message was already correctly processed then this manual retry could apply the same data modifications again because the outbox record that prevents this is already purged as it was expired.
```
SzymonPobiega: Do not use `!`. Maybe 'Ensure that the retention period of the outbox is longer than the maximum time the message can be retried, including Second Level Retries and manual retries via SC. Additional care must be taken by operators of ServicePulse to not retry messages older than the outbox retention period`.

nservicebus/outbox/index.md
```diff
@@ -79,8 +79,60 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
+
+## Concurrency
+
+Outbox cannot prevent that handlers or sagas are invoked at most once. Under certain scenarios this can happen:
+
+- Sender sends multiple physical message with the same message identifier (submit order form)
+- Transport can send multiple messages when having a failover, they operate in a at-least-once delivery model
+- The outbox dispatch stage can be invoked more then once if the previous dispatch (partially) failed and is now retried.
+
+Multiple physical copied of the same logical message now exist. If an endpoint allows concurrent message processing or is scaled out it can happen that these multiple copies are processed concurrently.
```
timbussmann: how about:
In those scenarios, multiple physical copies of the same logical exist in the queue.

nservicebus/outbox/index.md
SzymonPobiega: If handlers are idempotent than outbox is definitely not required. Also, if a handler is idempotent then message IDs don't really matter e.g. is a handler is adding items to a set then it does not care about message IDs.

Deterministic assignment of message IDs is important if users are doing custom ID-based deduplication.

nservicebus/outbox/index.md
timbussmann: > Outbox does not perform deduplication!

this sentence is a contradiction to several statements in this doco page which mention the usage of deduplication.

nservicebus/outbox/index.md
SzymonPobiega: I think we should just mention that outbox does not prevent concurrent attempts to process the same message (e.g. when message is duplicated on the wire) but ensures that the outcome of the processing is persisted only once.

nservicebus/outbox/index.md
ramonsmits: @SzymonPobiega I mention this as often customer are unaware about the message sending/publishing side of things. Very often the data operation is safe to multiple invocations but not the messages sending part. The fact that in additional invocations the handler generates the same logical messages but with a different message identifier is I think not what you want and you always want that to be atleast 'static' as to what the outbox does or be deterministic. You are saying basically only do idempotent processing based on the logical message and ignore the message identifier if I'm not mistaken.

ramonsmits: In this case I meant when you consider not using the outbox. I'll try and rephrase this.

nservicebus/outbox/index.md
ramonsmits: Rephrased based on @SzymonPobiega and your comments in the note:

ramonsmits:
@timbussmann @SzymonPobiega I've made a new edit and addressed most feedback. Please review.

nservicebus/outbox/index.md
```diff
@@ -79,8 +79,61 @@ In order to gradually convert an entire system from the DTC to the Outbox:
 WARNING: When verifying outbox functionality, it can sometimes be helpful to temporarily [stop the MSDTC Service](https://technet.microsoft.com/en-us/library/cc770732.aspx). This ensures that the Outbox is working as expected, and no other resources are enlisting in distributed transactions.
 
 
+## Message identity and idempotent processing
+
+If handlers are idempotent then outbox might not be required at all. When considering not using the outbox and implementing idempotent processing keep in mind that any generated message *must* apply the same message identifier if invoked more-then-once to have deduplication working on the infrastructure level or to solve this at the logical level by validating message data. [Read more on how to set a message identifier.](/nservicebus/messaging/message-identity.md)
```
timbussmann: > When considering not using the outbox and implementing idempotent processing keep in mind that any generated message *must* apply the same message identifier if invoked more-then-once to have deduplication working on the infrastructure level or to solve this at the logical level by validating message data.

I'm still not fully understanding this sentence. Why do you have to set the same message identifier when not using the outbox? What infrastructure is doing deduplication?

nservicebus/outbox/index.md
```diff
@@ -79,8 +79,61 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
+
+
+## Concurrency
+
+Outbox cannot guarantee that handlers or sagas are invoked exactly once. Outbox does not prevent concurrent attempts to process the same message but ensures that the outcome of the processing is persisted only once for resources that share the outbox transaction.
```
timbussmann: > Outbox cannot guarantee that handlers or sagas are invoked exactly once.

this still seems to be a contradiction to what this document states at the beginning:

> [...] At the handler level, however, the Outbox guarantees `exactly-once` message processing [...]

timbussmann: if I understand this correctly, this is more of an issue where outbox cannot deduplicate two different physical but isomorphic messages ?

nservicebus/outbox/index.md
```diff
@@ -79,8 +79,61 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
+- Transport can send multiple messages when having a failover, they operate in a at-least-once delivery model
+- The outbox dispatch stage can be invoked more-than-once if the previous dispatch (partially) failed and is retried.
+
+In those scenarios, multiple physical copies of the same logical message exist in the queue. If an endpoint allows concurrent message processing or is scaled out it can happen that these multiple copies are processed concurrently.
+
+
+## Non transactional resources
+
+If non-transactional resources are being accessed then make sure processing is idempotent. Non-transactional resources are not guaranteed to be handled in an idempotent way by the Outbox. If more-than-once invocation is unwanted then this can be almost prevented completely by making sure only a single endpoint instance is running and have [configured the endpoint to process messages sequentially by setting the maximum concurrency to 1](/nservicebus/operations/tuning.md#tuning-concurrency). Only data modified via SynchronizedStorageSession is made idempotent via the outbox and has exact-once semantics.
```
timbussmann: do we really want to name `SynchronizedStorageSession`? I'd say 95% of user never really work with this class directly and don't know what this is?

nservicebus/outbox/index.md
```diff
@@ -79,8 +79,61 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
+
+    Total outbox records = Message througput per second * Deduplication period in seconds
+
+A single outbox record - after all transport operations have been dispatched - usually requires less than 50 bytes of which most are taken for storing the original message ID as this is a string value.
+
+NOTE: If the system is processing a high volume of messages per second having a large deduplication timeframe might not be ideal as outbox records will occupy storage.
+
+## Outbox cleanup interval
+
+How outbox cleanup is performed is different per persistence implementation. Most persisters run a cleanup task every minute. Depending on the throughput this cleanup interval can be run more often to make a single cleanup operation purge less records per interval.
```
timbussmann: to many `is` in this sentence.

nservicebus/outbox/index.md
```diff
@@ -79,8 +79,61 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
+
+NOTE: This problem is not specific to Outbox but mentioned here as sometimes it is assumed Outbox does deduplication that prevents more-than-once invocation.
+
+Try to avoid mixing transactional and non transactional tasks. If performing non transactional tasks send a message to perform this task in isolation.
+
+
+## Outbox expiration duration
+
+Part of the outbox is guarantee that the data remains consistent if the same message is processed more-than-once. A message with the exact same identification can be detected and ignored. This data is stored in the outbox record and kept for a configurable duration. The default expiration duration depends per persistence implementation. All implemetations can be configured on how long an outbox record needs to be stored and some can be configured on how often outbox cleanup needs to occur.
```
timbussmann: to guarantee

ramonsmits:
Rebased to trigger a new build.

nservicebus/outbox/index.md
ramonsmits: > Why do you have to set the same message identifier when not using the outbox?

If you want to dedupe on the infrastructure level or on the endpoint level based on message identity. All outgoing messages must be the same with or without outbox.

> What infrastructure is doing deduplication?

nservicebus/outbox/index.md
ramonsmits: @timbussmann To what do you refer?

ramonsmits: > this is more of an issue where outbox cannot deduplicate two different physical but isomorphic messages ?

No, also the same physical copy in case the handler is invoked but the outbox record not yet committed and the message to be reprocessed. Better said, outbox does not guarantee at-most-once execution, only transactionally exact-once or non-transactional at-least-once.

nservicebus/outbox/index.md
ramonsmits: > I'd say 95% of user never really work with this class directly and don't know what this is?@timbussmann Assumption alert, I think a lot of customers do not use sagas and do their own storage IO which requires them to use the session.

ramonsmits: I've rewritten that sentence, not using the type, but mentioning the concept and pointing to the section for an implementation overview.

nservicebus/outbox/index.md
ramonsmits: This section is not a contradiction. The start of the document is about comparing transactional semantics of MSDTC with Outbox and for both they are transactionally exact-once. It even explicitly mentions that this is exact-once at 'a higher level'. On a low level it is still at-least-once processing hence this section explaining that it still is at-least-once.

nservicebus/outbox/index.md
SzymonPobiega: @ramonsmits you only need stable/deterministic message IDs if you are doing ID-based deduplication (e.g. Outbox or some infra). Otherwise you don't care (e.g. when you deduplicate based on hash of a body)

nservicebus/outbox/index.md
```diff
@@ -79,8 +79,61 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
+- The outbox dispatch stage can be invoked more-than-once if the previous dispatch (partially) failed and is retried.
+
+In those scenarios, multiple physical copies of the same logical message exist in the queue. If an endpoint allows concurrent message processing or is scaled out it can happen that these multiple copies are processed concurrently.
+
+
+## Non transactional resources
+
+If non-transactional resources are being accessed then make sure processing is idempotent. Non-transactional resources are not guaranteed to be handled in an idempotent way by the Outbox. If more-than-once invocation is unwanted then this can be almost prevented completely by making sure only a single endpoint instance is running and have [configured the endpoint to process messages sequentially by setting the maximum concurrency to 1](/nservicebus/operations/tuning.md#tuning-concurrency). Only data modified via a shared storage session is made idempotent via the outbox and has exact-once semantics. See [persistence specific documention](#Persistence) on how to obtain the shared storage session.
```
SzymonPobiega: I think we should replace it with a guidance to not mix transactional and non-transactional resources. Since outbox by definition requires transactional store, when using outbox you should never ever touch any non-transactional resource in a handler.

ramonsmits: @SzymonPobiega That is just mentioned here below:

ramonsmits:
@SzymonPobiega I get this error
>  Error:'Anchor link doesn't exist #Persistence'. Url:'#Persistence'.

While I do have a h2 with persistence.

SzymonPobiega:
@ramonsmits maybe try a link with a page name at least? It seems like the doco engine does not recognize the links to a second in the current document.

ramonsmits:
@Particular/editors could anyone review this PR?

nservicebus/outbox/index.md
dbelcham: @ramonsmits the 2nd sentence here is unclear. The first part of it seems to make sense, but after "to have deduplication working..." it gets confusing, to the point where I'm not sure if the first part of the sentence even relates to it. 

Try breaking that sentence down into smaller pieces. Go as far as making bullet points of what you want to say. Then turn them into short sentences that each, in isolation, are completely clear. Read each sentence without the others around it. If the sentence still makes sense without the surrounding context, then it's in good shape. Eventually, if warranted, those can be joined together to make compound sentences.

nservicebus/outbox/index.md
dbelcham: > Better said, outbox does not guarantee at-most-once execution, only transactionally exact-once or non-transactional at-least-once

If it's better said that way, then I'd suggest re-writing the section to say it that way.

I can see how this section could be construed as contradictory. If these are two different situations then you need to be clear, in both locations, which situation you're talking about. i.e. if the first section is talking about MSDTC with Outbox, then make it clear by using that phrase or those proper nouns.

nservicebus/outbox/index.md
```diff
@@ -79,8 +79,61 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
+If handlers are idempotent then outbox might not be required at all. When considering not using the outbox and implementing idempotent processing keep in mind that any generated message *must* apply the same message identifier if invoked more-then-once to have deduplication working on the infrastructure level or to solve this at the logical level by validating message data. [Read more on how to set a message identifier.](/nservicebus/messaging/message-identity.md)
+
+
+## Concurrency
+
+Outbox cannot guarantee that handlers or sagas are invoked exactly once. Outbox does not prevent concurrent attempts to process the same message but ensures that the outcome of the processing is persisted only once for resources that share the outbox transaction.
+
+The following scenarios might cause handlers to be invoked multiple times:
+
+- Sender sends multiple physical message with the same message identifier (submit order form)
```
dbelcham: These bullet points read awkwardly. I *think* its because they are poorly split sentence fragments. Try re-writing them as full sentences on each bullet point. Use more than one sentence on each bullet if that makes it clearer. Once you have them reading clearly then you can explore removing wording and turning them into sentence fragments.

nservicebus/outbox/index.md
```diff
@@ -79,8 +79,61 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
+- Transport can send multiple messages when having a failover, they operate in a at-least-once delivery model
+- The outbox dispatch stage can be invoked more-than-once if the previous dispatch (partially) failed and is retried.
+
+In those scenarios, multiple physical copies of the same logical message exist in the queue. If an endpoint allows concurrent message processing or is scaled out it can happen that these multiple copies are processed concurrently.
```
dbelcham: > In those scenarios....

might be better as
> In the above scenarios...

dbelcham: > If an endpoint allows concurrent message processing or is scaled out it can happen that these multiple copies are processed concurrently

try
> If an endpoint allows concurrent message processing, or it is scaled out, its possible that the multiple copies of the message can be processed concurrently.

nservicebus/outbox/index.md
```diff
@@ -79,8 +79,61 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
+
+
+## Non transactional resources
+
+If non-transactional resources are being accessed then make sure processing is idempotent. Non-transactional resources are not guaranteed to be handled in an idempotent way by the Outbox. If more-than-once invocation is unwanted then this can be almost prevented completely by making sure only a single endpoint instance is running and have [configured the endpoint to process messages sequentially by setting the maximum concurrency to 1](/nservicebus/operations/tuning.md#tuning-concurrency). Only data modified via a shared storage session is made idempotent via the outbox and has exact-once semantics. See [persistence specific documention](/nservicebus/outbox/#Persistence) on how to obtain the shared storage session.
```
dbelcham: > If more-than-once invocation is unwanted then this can be almost prevented completely by making sure only a single endpoint instance is running and have....

might be better as
> If more-than-once invocation is not wanted, then this can be prevented almost completely by making sure that only a single endpoint instance is running and having configured that endpoint with a maximum concurrency of 1.

nservicebus/outbox/index.md
```diff
@@ -79,8 +79,61 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
+
+NOTE: This problem is not specific to Outbox but mentioned here as sometimes it is assumed Outbox does deduplication that prevents more-than-once invocation.
+
+Try to avoid mixing transactional and non transactional tasks. If performing non transactional tasks send a message to perform this task in isolation.
+
+
+## Outbox expiration duration
+
+Part of the outbox is to guarantee that the data remains consistent if the same message is processed more-than-once. A message with the exact same identification can be detected and ignored. This data is stored in the outbox record and kept for a configurable duration. The default expiration duration depends per persistence implementation. All implemetations can be configured on how long an outbox record needs to be stored and some can be configured on how often outbox cleanup needs to occur.
```
dbelcham: > Part of the outbox is to guarantee....

should this be?
> Part of the purpose of the outbox is to guarantee....

dbelcham: No need to "dash" `more-than-once` in the first sentence.

dbelcham: > This data is stored in the outbox record

Which data? Use something like "The message indentification data is stored..." instead of the word "this" to add clarity

dbelcham: > This data is stored in the outbox record and kept for a configurable duration. The default expiration duration depends per persistence implementation. All implemetations can be configured on how long an outbox record needs to be stored and some can be configured on how often outbox cleanup needs to occur.

would be better as
> The identification data for each outbox record is retained. The default duration that the data is retained for will vary depending on the persistence chosen for the outbox. That default duration, as well as the frequency of data removal, can be overrode for all outbox persistences.

nservicebus/outbox/index.md
```diff
@@ -79,8 +79,61 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
+
+Ensure that the retention period of the outbox is longer than the maximum time the message can be retried, including delayed retries and manual retries via ServiceControl. Additional care must be taken by operators of ServicePulse to not retry messages older than the outbox retention period.
+
+## Required storage
+
+Outbox requires storage. The required space can be calculated as follows:
```
dbelcham: > The required space can be calculated as follows

better as
> The amount of storage space required for Outbox can be calculated as follows:

nservicebus/outbox/index.md
```diff
@@ -79,8 +79,61 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
+
+    Total outbox records = Message througput per second * Deduplication period in seconds
+
+A single outbox record - after all transport operations have been dispatched - usually requires less than 50 bytes of which most are taken for storing the original message ID as this is a string value.
```
dbelcham: change the two dashes (`-`) in the first sentence to be commas

nservicebus/outbox/index.md
```diff
@@ -79,8 +79,61 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
+
+NOTE: If the system is processing a high volume of messages per second having a large deduplication timeframe might not be ideal as outbox records will occupy storage.
```
dbelcham: >If the system is processing a high volume of messages per second having a large deduplication timeframe might not be ideal as outbox records will occupy storage.

better as
> If the system is processing a high volume of messages, having a long deduplication timeframe will increase the amount of storage space that is required by Outbox.

nservicebus/outbox/index.md
```diff
@@ -79,8 +79,61 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
+
+## Outbox cleanup interval
+
+How outbox cleanup is performed depends per persistence implementation. Most persisters run a cleanup task every minute. Depending on the throughput this cleanup interval can be run more often to make a single cleanup operation purge less records per interval.
```
dbelcham: > How outbox cleanup is performed depends per persistence implementation

better as
> Outbox cleanup varies depending on the persistence that is chosen.

dbelcham: > Depending on the throughput this cleanup interval can be run more often to make a single cleanup operation purge less records per interval

Depending on the throughput of what? "Depending on the throughput of the endpoints...."? If that's the case, then I'd change it to this:
> Depending on the throughput of the system's endpoints, the Outbox cleanup interval can be run more frequently. This will allow each cleanup operation to purge the fewest records possible each time it is runs, which will allow the cleanup to execute faster.

SzymonPobiega:
@ramonsmits do you have capacity to process @dbelcham 's comments?

ramonsmits:
@dvdstelt Just made me aware of this case https://c.eu1.visual.force.com/apex/Case?num=00034468 and this PR aligns with this issue.

DavidBoike:
@ramonsmits then you should address the comments. You’re currently the blocker here.

nservicebus/outbox/index.md
ramonsmits: Changed the beginning to reflect this.

ramonsmits:
@dbelcham @timbussmann Applied changes based on your feedback, please review.

nservicebus/outbox/index.md
```diff
@@ -43,7 +43,7 @@ If an incoming message is found in Outbox storage, then it is treated as a dupli
...
 
-Note: On the wire level, the Outbox guarantees `at-least-once` message delivery, meaning downstream messages may be sent and processed multiple times. At the handler level, however, the Outbox guarantees `exactly-once` message processing, similar to distributed transactions. This higher guarantee level is due to the deduplication that happens on the receiving side.
+Note: On the wire level, the Outbox guarantees `at-least-once` message delivery, meaning downstream messages may be sent and processed multiple times. At the handler level, however, the Outbox guarantees only transactionally exact-once or non-transactional at-least-once message processing, identital to distributed transactions. This higher guarantee level is due to the deduplication that happens on the receiving side.
```
timbussmann: > the Outbox guarantees only transactionally exact-once or non-transactional at-least-once message processing

I don't understand that sentence.

> identital 

identical

nservicebus/outbox/index.md
```diff
@@ -79,8 +79,64 @@ In order to gradually convert an entire system from the DTC to the Outbox:
 WARNING: When verifying outbox functionality, it can sometimes be helpful to temporarily [stop the MSDTC Service](https://technet.microsoft.com/en-us/library/cc770732.aspx). This ensures that the Outbox is working as expected, and no other resources are enlisting in distributed transactions.
 
 
+## Message identity and idempotent processing
+
+If handlers are idempotent then outbox might not be required at all. Outbox relies on message identity to guarantee consistency. When using outbox on both the sender and receiver message identifiers are consistent if a handler is invoked more-then-once for a message.
```
timbussmann: > When using outbox on both the sender and receiver message identifiers are consistent if a handler is invoked more-then-once for a message. 

I don't understand this sentence or what it's trying to tell me. Also, more-than-once.

nservicebus/outbox/index.md
```diff
@@ -79,8 +79,64 @@ In order to gradually convert an entire system from the DTC to the Outbox:
 WARNING: When verifying outbox functionality, it can sometimes be helpful to temporarily [stop the MSDTC Service](https://technet.microsoft.com/en-us/library/cc770732.aspx). This ensures that the Outbox is working as expected, and no other resources are enlisting in distributed transactions.
 
 
+## Message identity and idempotent processing
+
+If handlers are idempotent then outbox might not be required at all. Outbox relies on message identity to guarantee consistency. When using outbox on both the sender and receiver message identifiers are consistent if a handler is invoked more-then-once for a message. 
+If a sender does not use the outbox or when sending a message outside of a handler the sender *must* generate and apply the same message identifier to guarantee consistency via the outbox when that logic is invoked more-then-once.
```
timbussmann: I think the last part of the sentence (when that logic is invoked more-then-once.) can be removed?

nservicebus/outbox/index.md
ramonsmits: @timbussmann how about:

> At the handler level, however, the Outbox guarantees only exact-once processing when using transactional storage and behaves identically to distributed transactions. If handlers use nontransactional resources processing behaves at-least-once message processing.

ramonsmits:
@timbussmann applied changes on your feedback, also added a suggestion https://github.com/Particular/docs.particular.net/pull/3552#discussion_r169123782 let me know if that rewrite is understandable.

nservicebus/outbox/index.md
```diff
@@ -43,7 +43,7 @@ If an incoming message is found in Outbox storage, then it is treated as a dupli
...
 
-Note: On the wire level, the Outbox guarantees `at-least-once` message delivery, meaning downstream messages may be sent and processed multiple times. At the handler level, however, the Outbox guarantees `exactly-once` message processing, similar to distributed transactions. This higher guarantee level is due to the deduplication that happens on the receiving side.
+Note: On the wire level, the Outbox guarantees `at-least-once` message delivery, meaning downstream messages may be sent and processed multiple times. At the handler level, however, the Outbox guarantees only transactionally exact-once or non-transactional at-least-once message processing, identical to distributed transactions. This higher guarantee level is due to the deduplication that happens on the receiving side.
```
timbussmann: > the Outbox guarantees only transactionally exact-once or non-transactional at-least-once message processing,

I still don't understand that sentence. It does seem to be quite different from your proposal on my last comment about this sentence @ramonsmits ?

nservicebus/outbox/index.md
```diff
@@ -79,8 +79,64 @@ In order to gradually convert an entire system from the DTC to the Outbox:
 WARNING: When verifying outbox functionality, it can sometimes be helpful to temporarily [stop the MSDTC Service](https://technet.microsoft.com/en-us/library/cc770732.aspx). This ensures that the Outbox is working as expected, and no other resources are enlisting in distributed transactions.
 
 
+## Message identity and idempotent processing
+
+If handlers are idempotent then outbox might not be required at all. Outbox relies on message identity to guarantee consistency. When using outbox on both the sender and receiver message identifiers are consistent if a handler is invoked more-than-once for a message.
```
timbussmann: I don't like this whole section:

> Outbox relies on message identity to guarantee consistency.

guarantee what consistency? It relies of the message identity to deduplicate.

> When using outbox on both the sender and receiver message identifiers are consistent if a handler is invoked more-than-once for a message. 

I'm wondering about the correctness of this sentence. The outbox should prevent more-than-once invocation of a handler. If a handler is invoked more than once, then only because the first attempt wasn't successful and the potentially sent messages were rolled back. I'm not aware of any mechanism to guarantee deterministic message identifiers.

> If a sender does not use the outbox or when sending a message outside of a handler the sender *must* generate and apply the same message identifier to guarantee consistency at the receiver.

this sentence is also weird. The only reason you'd have to generate deterministic message identifiers is in case there is a risk of sending the same message multiple times. This can only be caused by a rippling effect and and also can only be solved by using determinisitic message identifiers across the whole set of endpoints. I don't think this one sentence here is in any way sufficient documentation on how to address and solve this problem.

> The same message identity is also important to have deduplication working on the infrastructure level on some transport that support this.

too much of a technical detail.

nservicebus/outbox/index.md
```diff
@@ -79,8 +79,64 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
+## Message identity and idempotent processing
+
+If handlers are idempotent then outbox might not be required at all. Outbox relies on message identity to guarantee consistency. When using outbox on both the sender and receiver message identifiers are consistent if a handler is invoked more-than-once for a message. 
+If a sender does not use the outbox or when sending a message outside of a handler the sender *must* generate and apply the same message identifier to guarantee consistency at the receiver.
+The same message identity is also important to have deduplication working on the infrastructure level on some transport that support this. [Read more on how to set a message identifier.](/nservicebus/messaging/message-identity.md)
+
+
+## Concurrency
+
+Outbox cannot guarantee that handlers or sagas are invoked exactly once. Outbox does not prevent concurrent attempts to process the same message but ensures that the outcome of the processing is persisted only once for resources that share the outbox transaction.
```
timbussmann: In the end, this sentence just confirms the outbox limitations of only providing "exactly-once" semantics for the outbox storage + transport. All other resources continue to be exposed to duplicate messages. I'd rather ensure to clarify this concept and avoid this paragraph all together.

nservicebus/outbox/index.md
```diff
@@ -79,8 +79,64 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
+
+The following scenarios might cause handlers to be invoked multiple times:
+
+- When a sender sends multiple physical message with the same message identifier. An example is a order form which is submitted multiple times.
+- When a transport performs a failover and operates in a at-least-once delivery model and has not marked a message as processed while this is the case but the acknowledgement was not processed by the tranport yet.
+- When a failure occurs during the outbox dispatch stage. The dispatch stage will to be invoked more-than-once and each time message can message can partially be delivered to the transport infrastructure.
+
+In the above scenarios, multiple physical copies of the same logical message exist in the queue. If an endpoint allows concurrent message processing, or it is scaled out, its possible that the multiple copies of the message can be processed concurrently.
```
timbussmann: I think the point about the concurrency is the relevant point. All of the above listed scenarios should be handled correctly in case there is no race condition, so I'm not sure about the value of listing those scenarios and I suggest to focus on the concurrency aspect instead.

Thinking more about these concurrency race conditions I fail to see the larger problem justifying this documentation section. In case of concurrent processing of a duplicate message, one of the messages should fail due to a concurrency exception on the outbox storage. This will rollback the message and keep the "exactly once semantic" of the outbox. Again, the only cases which cause side effects are non-outbox-storage related side-effects, which we cannot prevent in general with the outbox anyway (= this is a know limitation of the outbox).

nservicebus/outbox/index.md
```diff
@@ -79,8 +79,64 @@ In order to gradually convert an entire system from the DTC to the Outbox:
...
+
+Outbox requires storage. The amount of storage space required for Outbox can be calculated as follows:
+
+    Total outbox records = Message througput per second * Deduplication period in seconds
+
+A single outbox record, after all transport operations have been dispatched, usually requires less than 50 bytes of which most are taken for storing the original message ID as this is a string value.
+
+NOTE: If the system is processing a high volume of messages, having a long deduplication timeframe will increase the amount of storage space that is required by Outbox.
+
+## Outbox cleanup interval
```
timbussmann: isn't this section part of the "outbox experiation duration" section?

nservicebus/outbox/index.md
ramonsmits: > guarantee what consistency? It relies of the message identity to deduplicate.

It relies on the message identity to guard consistency in the datastore, no other values are used for that. Again, outbox doesn't dedupe, more than one copy of the message can exist on the wire, it does nothing to prevent this.


> The outbox should prevent more-than-once invocation of a handler.If a handler is invoked more than once, then only because the first attempt wasn't successful and the potentially sent messages were rolled back. I'm not aware of any mechanism to guarantee deterministic message identifiers.

nservicebus/outbox/index.md
ramonsmits: > Again, the only cases which cause side effects are non-outbox-storage related side-effects, which we cannot prevent in general with the outbox anyway (= this is a know limitation of the outbox).

dbelcham:
@ramonsmits I re-worked the wording for grammar and organization of the content. Can you verify that the content is technically accurate now?

@timbussmann can you verify that your concerns have been addressed?

timbussmann:
happy with the changes. @dbelcham changes introduced an lot of casing differences which a.) should have been done in a separate PR and b.) are not consistent. I can spot quite a few more "Outbox"es, is that intentional or should that be fixed as well (outside this PR)?

PR #3556, opened 2018-01-09
Added note on RavenDB shared session

ramonsmits:
Based on:

- https://discuss.particular.net/t/nsb-5-raven-msmq-when-moving-to-outbox-shouldnt-we-be-using-a-shared-document-session/293
- https://github.com/Particular/NServiceBus.RavenDB/issues/71

PR #3558, opened 2018-01-09
Raven sharedsession api

DavidBoike:
Replaces #3556

@ramonsmits please review. @mauroservienti if you're *totally bored* only.

Accepts that the SharedSession API is not well named, so divides the sections into Shared Session which only talks about **using** the session within handler/saga (as this happens automatically, much better in V6 than in V5, but that's beside the point) and a separate section on customizing the document session, explaining that the name of that API being SharedSession is a bit unfortunate but clarifying what it REALLY does.

Also includes an error box for V2-3 (both targeting CoreV5) that the SharedSession API (for customizing) and the Outbox feature are incompatible, due to a bug that at this point (due to V5's remaining support and non-critical nature) we have no plans to fix unless a customer desperately needs it, as the whole thing is completely fixed in V6 by design.


DavidBoike:
@ramonsmits If you're happy with this, go ahead and squash-merge it.

ramonsmits:
Applied 2 small changes, have to await for the build server to complete the merge.

DavidBoike:
An ASB8 sample broke due to a change in `NServiceBus.Bridge.InterceptMessageForwarding, NServiceBus.Bridge` just pushed to NuGet today. Using my admin ability to squash merge anyway.

PR #3563, opened 2018-01-10
Idempotency required for addressing logic to work properly

SeanFeldman:
Related to #695

Addresses https://github.com/Particular/NServiceBus.AzureServiceBus/issues/695#issuecomment-356504603
> I believe we also have an issue with `AddressingLogic` encapsulating `SanitizationStrategy`.
>
> For example, let's say `HierarchyComposition` adds a `scadapter` prefix to the the path `veryLong...QueuePath` of a queue that was already 260 characters long. This will result in a queue path `scadapter/veryLong...QueuePath` and will exceed the maximum queue length, causing sanitization to be applied. Let's say sanitization is turning path into a GUID. The result will be a path `<guid-value>` rather than expected **`scadapter/<guid-value>`**.


SeanFeldman:
@Particular/azure-maintainers please review.

PR #3565, opened 2018-01-11
Version information regarding NServiceBus.Host

ramonsmits:


SzymonPobiega:
@ramonsmits this information is already available in the drop down in the host doco pages where you select the version. Why do you need it to be included explicitly?

If you think it is useful I think we should add it to the https://docs.particular.net/components/ page or generate a page-per-component which shows the mapping. This information is already maintained in the `components\core-dependencies` folder.

ramonsmits:
@SzymonPobiega I saw a similar overview on RavenDB page and I really liked it!

SzymonPobiega:
Merging this for now. It will be eventually replaced by a smart tag in the docs engine that includes such a table based on the data from components\core-dependencies.

PR #3567, opened 2018-01-11
NativeDeferral clarifications

SeanFeldman:
A warning about delays longer than 15 mins should stand out and not blend with the rest of the description.


SeanFeldman:
@danielmarbach please review

PR #3569, opened 2018-01-12
WIP: License info log file

ramonsmits:


ramonsmits:
@Particular/docs-maintainers I created this based on log events I ran for a customer. Not sure if I should proceed and if it even is useful to have as documentation. There can be all kinds of license states that probably have different log output per major core version. ANy thoughts?

DavidBoike:
I don't like having dates in there, you could probably start each message at the WARN/INFO or whatever.

This page is already versioned, should use partials not headings.

HEskandari:
Also, need this for v6 and v7. The partials targetting those versions (in case there have been no changes) would solve that. Also, +1 on removing unimportant bits from the log. To me, the important parts are the log level and the message.

ramonsmits:
@DavidBoike @HEskandari So you do think it has value? Then I will slowly finish this.

HEskandari:
@ramonsmits I do believe it has value. I do get occasional questions about this. Haven't seen the related partials but consider adding few more details like maybe:

- When the performance is or is not degraded when the license is expired.
- Documenting the debugging mode experience (where a Window is shown)

ramonsmits:
@HEskandari really good additions, I had the exact same questions Today about an expired license.

SzymonPobiega:
Still WIP?

SzymonPobiega:
Ping

PR #3581, opened 2018-01-17
RavenDB sample(s) on Linux

DavidBoike:


samples/ravendb/simple/sample.md
```diff
@@ -29,20 +29,13 @@ This sample shows a simple Client + Server scenario.
 
 ### Raven Config
 
-Configure the endpoint to use RavenDB persistence.
+Configure the endpoint to use RavenDB persistence. The Url/Port may need to be changed to match the database in use.
```
bording: Why are "Url" and "Port" capitalized?

bording:
Oh, one other thing. The new v5 Core solution needs a R# settings file.

samples/ravendb/simple/sample.md
DavidBoike: Url will be capitalized in the snippet, though that technically contains the port too. Perhaps just "The Url may be..." would be better.

bording: 👍 For "The Url" since that includes the port.

PR #3583, opened 2018-01-18
Document supported platforms

timbussmann:


DavidBoike:
So, as prepared right now this will appear as an unversioned document, which may be somewhat misleading. Do we want to add the `component: NServiceBus`, and then add partials for other versions?

timbussmann:
the goal was to create a documentation of supported platforms as independent as possible from the actual versions. Not sure I see value in tying this to the core version unless we explicitly want to state details about supported versions.

timbussmann:
I'm going ahead and merge this document as is for now as I feel we talked about the topic of naming versions quite a bit. Let's raise another issue if you feel we need to adjust the information stated about supported platforms @DavidBoike

PR #3588, opened 2018-01-19
Mention endpoint hosting for webapi for Service Fabric, closes #3561

### LINKED ISSUE #3561: Feedback: 'Service Fabric Hosting'
Feedback for 'Service Fabric Hosting' https://docs.particular.net/nservicebus/hosting/service-fabric-hosting/

Location in GitHub: https://github.com/Particular/docs.particular.net/blob/master/nservicebus/hosting/service-fabric-hosting/index.md

I think it is probably worth pointing out that Send Only endpoints in service fabric would not start the endpoint in a communication listener.  If you are sending a command from ASPNET, for example, then you’d start the endpoint in the Startup class just as you would outside of Service Fabric.

danielmarbach:


danielmarbach:
@Particular/azure-maintainers please review and merge

nservicebus/hosting/service-fabric-hosting/index.md
```diff
@@ -24,6 +24,8 @@ With stateless services, the number of instances of a service can range from one
 
 snippet: StatelessEndpointCommunicationListener
 
+NOTE: If Stateless services are used to host web application frameworks such as ASP.NET, ASP.NET Core or WebAPI it is recommended to start the endpoint inside the service collection extensions instead of the communication listener as shown in [using NServiceBus in an ASP.NET Core WebAPI Application](/samples/web/send-from-aspnetcore-webapi/). This approach is simpler than having to coordinate multiple communication listener since no coordination is needed between the listeners.
```
SeanFeldman: From a MSFT source it was suggested to use Azure WebApps service as it's geared to web apps much better than SF.

SeanFeldman: Aside from the comment above, also
> it is recommended to start the endpoint inside the service collection extensions 

perhaps should be a link to the recommendation?

danielmarbach: Isn't linking to the webapi sample enough?

About the webapps recommendation. Should we add another note saying for classical webapps the WebApps services is recommended in azure?

Issue #3596, opened 2018-01-23
Feedback: 'Heartbeat Plugin' project hosting link broken


timbussmann:
Feedback for 'Heartbeat Plugin' https://docs.particular.net/servicecontrol/plugins/heartbeat

Location in GitHub: https://github.com/Particular/docs.particular.net/blob/master/servicecontrol/plugins/heartbeat.md

The "project hosting" link directs to `https://github.com/Particular/?query=Heartbeat` but it seems this doesn't yield any useful result. Other packages seem to link directly to their specific repository (e.g. file based routing).

ping @Particular/servicecontrol-maintainers

SzymonPobiega:
@timbussmann I think this is now fixed after merging the monitoring menu PR. Can you check?

WilliamBZA:
The query should be `https://github.com/Particular/?q=heartbeat` instead of `https://github.com/Particular/?query=Heartbeat`

We'll have to update the components.yaml file.

SzymonPobiega:
Hmmm I can't find that link :(

WilliamBZA:
https://github.com/Particular/docs.particular.net/blob/master/components/components.yaml#L277

WilliamBZA:
I'll do a PR

mikeminutillo:
I assume this fixed after the #3598 was closed?

If not, please re-open.

PR #3598, opened 2018-01-23
Update components.yaml

WilliamBZA:
Connects to https://github.com/Particular/docs.particular.net/issues/3596

### DIFF
--- a/components/components.yaml
+++ b/components/components.yaml
@@ -169,7 +169,7 @@
   DocoUrl: monitoring/custom-checks
   GitHubOwner: Particular/servicecontrol-maintainers
   Category: Other
-  ProjectUrl: https://github.com/Particular/?query=CustomChecks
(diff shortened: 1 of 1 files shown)


timbussmann:
LGTM. But is there any reason we're not linking to the specific repo now that those are unified (e.g. https://github.com/Particular/NServiceBus.Heartbeat)?

WilliamBZA:
I would guess there might be people using the older packages? Similar to [this](https://github.com/Particular/PlatformDevelopment/issues/1334) discussion

timbussmann:
@WilliamBZA according to this issue, the old repos are going to be deleted as the new ones contain tags for the old versions? Even more a reason to not link to a query I'd say?

WilliamBZA:
@timbussmann Only V3 and V4 are being killed off. We still have V5 and V6 plugins.

PR #3613, opened 2018-01-31
Use wildcard dependencies in perf counters sample

timbussmann:
Most samples are using wildcard dependencies for the nsb packages. Is there any reason this sample doesn't align with the other samples regarding specifying dependencies? @Particular/metrics-maintainers


timbussmann:
already fixed by https://github.com/Particular/docs.particular.net/commit/74b9642b1853a4b39a026e6e5d7ae00ec288c8e6

PR #3616, opened 2018-02-08
Update index.md

SimonCropp:


dbelcham:
Good catch @SimonCropp 
I opened a PR (#3619) with the proper API for that section and will close this in favour of that one.

PR #3619, opened 2018-02-09
Update auditing upgrade details for v6 to v7

dbelcham:
Replace exception name with proper configuration option for v7 auditing

### DIFF
--- a/nservicebus/upgrades/6to7/index.md
+++ b/nservicebus/upgrades/6to7/index.md
@@ -28,8 +28,8 @@ Configuring Audit via the following APIs have been deprecated:
  * `HKEY_LOCAL_MACHINE\SOFTWARE\ParticularSoftware\ServiceBus\AuditQueue` registry key
  
 
-Instead use one of the following:
-BadImageFormatException
+Instead use the following:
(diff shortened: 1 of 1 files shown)


nservicebus/upgrades/6to7/index.md
```diff
@@ -28,8 +28,8 @@ Configuring Audit via the following APIs have been deprecated:
  * `HKEY_LOCAL_MACHINE\SOFTWARE\ParticularSoftware\ServiceBus\AuditQueue` registry key
  
 
-Instead use one of the following:
-BadImageFormatException
+Instead use the following:
+`endpointConfiguration.AuditProcessedMessagesTo("targetAuditQueue");`
```
timbussmann: do we add a "dummy parameter" to the APIs in general or should we just use `endpointConfiguration.AuditProcessedMessagesTo(...);` or `endpointConfiguration.AuditProcessedMessagesTo` ?

timbussmann: looking at some of the previous upgrade guides, using an actual parameter value seems to be quite common.

PR #3623, opened 2018-02-15
NServiceBus Core - Added note about running least privilege and running installers

ramonsmits:
Based on discussion on installers in https://discuss.particular.net/t/is-there-a-connection-retry-for-the-transport/388

### DIFF
--- a/nservicebus/operations/installers.md
+++ b/nservicebus/operations/installers.md
@@ -15,7 +15,7 @@ Installers ensure that endpoint specific artifacts (e.g. queues, directories, da
 
 ## Running installers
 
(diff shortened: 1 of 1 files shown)


ramonsmits:
@timbussmann I opted for a note as it is application to the snippets below it. I've rewritten it, I hope it reads much better now, it is shorter too.

nservicebus/operations/installers.md
```diff
@@ -15,6 +15,8 @@ Installers ensure that endpoint specific artifacts (e.g. queues, directories, da
 
 ## Running installers
 
+NOTE: Running installers requires the process to run with permissions to create/delete tables, queues, folders and other resources. We recommend to apply the [principle of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege) and prevent running an endpoint process with too many permissions. This means only running the installers during setup with elevated permissions and not while processing messages.
```
timbussmann: Like this version much better 👍 I think the last sentence could use some tweaking, how about something like this:
"Consider running the installers only during the initial setup using elevated permissions"?

PR #3628, opened 2018-02-20
RavenDB Outbox cleanup: Added string value for -1 milliseconds and restored milliseconds

ramonsmits:


SzymonPobiega:
@ramonsmits do you agree with Daniel's opinion? Should we bring a third person?

PR #3629, opened 2018-02-21
clarified unique instance ID's when using callbacks with brokers

adamralph:
/cc @Particular/callback-maintainers


danielmarbach:
LGTM

PR #3638, opened 2018-02-27
Update distributed-instances.md

WilliamBZA:


servicecontrol/servicecontrol-instances/distributed-instances.md
```diff
@@ -52,6 +52,8 @@ As an example two ServiceControl instances are used where one is the master and
 </configuration>
 ```
 
+WARN: All instances of ServiceControl MUST have a unique name
```
danielmarbach: Should this be `WARNING:`?

danielmarbach: How about incorporating this into the step by step guide instead of a warn as well as into the known limitations?

WilliamBZA: It should be warning.

I can get on board with incorporating into the step-by-step guide, but I still think it's worth calling this out as a dedicated warning somewhere on the page as this is the kind of thing that is difficult to debug in production.

PR #3639, opened 2018-02-27
Update routing.md

janv8000:
Deleted text is already present in the "partial: commands" .

### DIFF
--- a/nservicebus/messaging/routing.md
+++ b/nservicebus/messaging/routing.md
@@ -34,7 +34,7 @@ As described in [messages, events and commands](/nservicebus/messaging/messages-
 
 partial: commands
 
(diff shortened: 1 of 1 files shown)


danielmarbach:
@Particular/nservicebus-maintainers I sent in https://github.com/Particular/docs.particular.net/pull/3661 which addresses the same problem in the partial. I think it should be done in the partial because the text in the chapter suggests that the overriding behavior applies for all NSB versions

timbussmann:
hey @janv8000 , thanks a lot for raising this PR, good catch 👍 I agree with @danielmarbach though that we should remove it from the partial instead for now although there are a few more improvements required for that section anyway. I'm going to close this PR as it's been addressed by #3661

PR #3649, opened 2018-03-06
Fix on-premise showcase sample

timbussmann:
* Referenced the correct RC compatible 2.0-rc1 package of NServiceBus.Encryption.MessageProperty
* Updated the Assembly redirects to the newer dependency versions
* Fixed the dependency versions to avoid misaligned redirects on the next patch release. @Particular/docs-maintainers thoughts about that approach? Other samples use wildcard dependencies but in this case there seems to be a big risk in regards to the required assembly redirects?

### DIFF
--- a/samples/show-case/on-premise/Core_6/Store.ECommerce/Store.ECommerce.csproj
+++ b/samples/show-case/on-premise/Core_6/Store.ECommerce/Store.ECommerce.csproj
@@ -39,14 +39,14 @@
     <Reference Include="System" />
     <Reference Include="System.Web" />
     <Reference Include="System.Configuration" />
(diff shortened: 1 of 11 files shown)


bording:
@timbussmann Looks like the cloud-azure showcase sample also needs to be updated.

indualagarsamy:
@bording - I am going to merge this. Since this PR is targeted for On-Premise. I made sure the V6 works as well. It was broken. But one of you @Particular/nservicebus-maintainers can go ahead and create a new PR for the cloud.

indualagarsamy:
Discussed this with @timbussmann. He pointed to me that there is a different issue that's tracking all sample updates and the cloud sample update and the other samples will be handled as part of that issue.

PR #3655, opened 2018-03-09
auto start download and hide button coming from landing page

sergioc:
auto start download and hide button coming from monitoring demo landing page


DavidBoike:
I changed it to hide the container after all, otherwise you have some unexplained margin leftover from the container making the spacing look funny. Also moved the script adjacent (and importantly, _after_) the markup so it's sure to exist in the DOM.

So I think that'll work. I'll merge assuming it comes up green.

PR #3664, opened 2018-03-15
Be explicit about netcore license locations

DavidBoike:
References:

### Windows

Well, mostly LinqPad, but [this is source for both](https://github.com/dotnet/corefx/blob/v2.0.6/src/System.Runtime.Extensions/src/System/Environment.Win32.cs#L65-L70).

### Linux/macOS
* [LocalApplicationData](https://github.com/dotnet/corefx/blob/v2.0.6/src/System.Runtime.Extensions/src/System/Environment.Unix.cs#L139-L147)
* [CommonApplicationData](https://github.com/dotnet/corefx/blob/v2.0.6/src/System.Runtime.Extensions/src/System/Environment.Unix.cs#L101)

### DIFF
--- a/nservicebus/licensing/index_license-management_core_[7,].partial.md
+++ b/nservicebus/licensing/index_license-management_core_[7,].partial.md
@@ -1,23 +1,33 @@
+{{NOTE:
+Depending on the operating system, the paths may be case sensitive.
+
+NServiceBus uses the [`Environment.GetFolderPath(SpecialFolder)`](https://docs.microsoft.com/en-us/dotnet/api/system.environment.getfolderpath) method to determine the locations of some paths on each OS.
+}}
+
+
 ### Application-specific license location
 
 A license located at `{AppDomain.CurrentDomain.BaseDirectory}/license.xml` will be automatically detected.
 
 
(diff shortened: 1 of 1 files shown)


timbussmann:
I think this definitely needs a "by default [...]" when mentioning the specific locations. Maybe even a note or statement that these values might differ depending on the OS configuration.

DavidBoike:
@timbussmann Do you think that .NET Core code will change?

DavidBoike:
I do think we might need a statement about expanding the environment variables, or that the variables for Linux/macOS are presented in bash syntax, but I'm not sure how to word that. I'm not actually that good at bash, I'm leaning on what @bording came up with for our trial extension email that uses the same paths to tell you how to install your license.

bording:
I think we need the wording here to indicate that we're using `SpecialFolder.LocalApplicationData` and `SpecialFolder.CommonApplicationData` in our code, and that those map to the indicated locations based on the OS.

I'm also wondering if we should split the "machine-wide" locations out into a "per-user" section and the actual "machine-global" location.

DavidBoike:
What do you think now @bording?

PR #3665, opened 2018-03-15
.NET Core what isn't supported

DavidBoike:
Connects to https://github.com/Particular/DotNetCoreLaunch/issues/108

### DIFF
--- a/nservicebus/upgrades/supported-platforms.md
+++ b/nservicebus/upgrades/supported-platforms.md
@@ -32,7 +32,7 @@ Some packages cannot currently support .NET Core or running on non-Windows platf
 
 * Transports
   * NServiceBus.AmazonSQS - Blocked from running on Linux due to a [bug in the AWS SDK](https://github.com/aws/aws-sdk-net/issues/796).
-  * NServiceBus.Azure.Transports.WindowsAzureServiceBus - The new Azure Service Bus library isn't fully-featured and contains breaking changes. See the blog post [A new Azure Service Bus transport—but not just yet](https://particular.net/blog/a-new-azure-service-bus-transport-but-not-just-yet).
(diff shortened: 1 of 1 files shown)


nservicebus/upgrades/supported-platforms.md
```diff
@@ -13,15 +13,43 @@ NServiceBus is supported on the .NET Framework and .NET Core.
...
 * **Windows**: All Windows versions according to [.NET Core 2.0 - Supported OS versions](https://github.com/dotnet/core/blob/master/release-notes/2.0/2.0-supported-os.md) are fully supported.
 * **Linux**: All Linux versions according to [.NET Core 2.0 - Supported OS versions](https://github.com/dotnet/core/blob/master/release-notes/2.0/2.0-supported-os.md) are fully supported.
 * **macOS**: macOS is supported only for development purposes.
+
+### Packages not supporting .NET Core
+
+Some packages cannot currently support .NET Core or running on non-Windows platforms:
+
+* Transports
+  * NServiceBus.AmazonSQS - Blocked from running on Linux due to a [bug in the AWS SDK](https://github.com/aws/aws-sdk-net/issues/796).
```
SeanFeldman: I think we have a blocking issue about SQS in PlatDev that needs to be resolved first.

DavidBoike: Don't think this needs to block on a PlatDev issue. This is the state of things as it exists on NuGet right now. When that issue resolves (I'll add a note to the PoA) this should be updated to reflect it.

DavidBoike:
This page only gets average 10 unique views per week. Going to merge it so it exists and is linkable. Will announce so that maintainer groups can review and propose tiny changes rather than holding up the big change.

PR #3672, opened 2018-03-19
Tweak list of packages that don't support .NET core

bording:
FYI @Particular/dotnetcore-launch-squad

### DIFF
--- a/nservicebus/upgrades/supported-platforms.md
+++ b/nservicebus/upgrades/supported-platforms.md
@@ -28,7 +28,7 @@ NServiceBus is supported for applications targeting .NET Core on a variety of op
 
 ### Packages not supporting .NET Core
 
-Some packages cannot currently support .NET Core or running on non-Windows platforms:
+Some packages do not currently support .NET Core or running on non-Windows platforms:
 
 * Transports
   * NServiceBus.AmazonSQS - Blocked from running on Linux due to a [bug in the AWS SDK](https://github.com/aws/aws-sdk-net/issues/796).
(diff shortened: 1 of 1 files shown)


bording:
What about the Azure host packages? This list currently implies that they'll be replaced by templates from `ParticularTemplates`, but that's not currently true.

@Particular/azure-maintainers What's the plan for these?

SeanFeldman:
The are not replaced AFAIK, but deprecated.

bording:
> The are not replaced AFAIK, but deprecated.

What does that mean? What's the alternative?

SeanFeldman:
Customers will be able to upgrade from version 7 to version 8 (from NSB 6 to 7) with warning and error in version 9. Instead of Azure host, we're advising to switch to self hosting ([sample](https://docs.particular.net/samples/azure/self-host/?version=core_7)) in upgrade guide [here](https://docs.particular.net/nservicebus/upgrades/acs-host-7to8)

> The NServiceBus Azure Host will be deprecated as of Version 9 and users are recommended to switch to self hosting for new endpoints. Upgrading existing endpoints is still supported for Version 8.

For the [shared host](https://docs.particular.net/nservicebus/hosting/cloud-services-host/shared-hosting) we've concluded that it would be better to decouple from Azure, but nothing after that (https://github.com/Particular/NServiceBus.Host.AzureCloudService/issues/5).

bording:
@SeanFeldman Instead of just having some code in the upgrade guide, is that something we should consider adding a template for?

SeanFeldman:
Good question. @Particular/azure-maintainers thoughts?

yvesgoeleven:
@bording @SeanFeldman what are these templates you are talking about? project templates?

bording:
> @bording @SeanFeldman what are these templates you are talking about? project templates?

Yes, project templates that can be used via `dotnet new`: https://docs.particular.net/nservicebus/dotnet-templates

DavidBoike:
Merged this, but I think the conversation about the Azure hosts should continue!

PR #3675, opened 2018-03-20
Add extra quickstart download for VS2015

sergioc:
Related to https://github.com/Particular/PlatformDevelopment/issues/1709


sergioc:
Currently using 2 dummy URLs to see if the build passes.

@HEskandari once you have the tutorial sample zip file for VS2015, we'll need to update the final button URLs.

sergioc:
Seems like the hard coded HTML and URLs have passed.

PR #3678, opened 2018-03-21
Port of quickstart for VS2015

HEskandari:
Ports the quickstart solution to VS2015

### DIFF
--- a/tutorials/quickstart/VS2015Solution/Billing/Billing.csproj
+++ b/tutorials/quickstart/VS2015Solution/Billing/Billing.csproj
@@ -1,0 +1,75 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props" Condition="Exists('$(MSBuildExtensionsPath)\$(MSBuildToolsVersion)\Microsoft.Common.props')" />
+  <PropertyGroup>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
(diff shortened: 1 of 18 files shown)


DavidBoike:
Added a commit to remove some of the cruft: Release configuration, empty Properties folder in the solution view, an unneeded app.config.

One thing we still need to figure out is how to include the `.vs/RetailDemo/v14/.suo` file in order for multiple startup projects to work. For a normal sample/tutorial, DocsEngine would do this during build. We need the zipped file to include that file as well. I'm hesitant to just check in the one my VS2015 generates as it might have a lot of other garbage in it too, like which side of the window I have my solution explorer docked to (left) which is not standard.

@adamralph You did this using DocsEngine tech for the workshop, correct? Any recommendations? I would be ok with a one-time process that generates the file (minimally, containing only the startup project information) and then manually committing that once.

adamralph:
@DavidBoike the way we do it in the workshop is slightly different. We have a [build target](https://github.com/Particular/Workshop/blob/e7b775c8e9ce69af877241b306a6eea92700face/build.csx#L78-L93) that does this for us. The build target itself is really simple. It calls into [SetStartupProjects](https://www.nuget.org/packages/SetStartupProjects/) to do all the work.

We then instruct the workshop attendees to run a [batch file which wraps the invocation of that build target](https://github.com/Particular/Workshop/blob/e7b775c8e9ce69af877241b306a6eea92700face/set-startup-projects.cmd).

Would it be possible to invoke SetStartupProjects during the build to create the .suo file? The problem with committing a .suo file is that every time someone edits the solution in VS, they will generate a diff on the .suo file that they will have to remember not to commit (or the reviewers will have to remember not to approve the PR).

DavidBoike:
@adamralph Good to see how Workshop does it. That will be useful, though we don't want that "stuff" in the QuickStart. I don't want the suo in the repository, we just need to (temporarily) generate it whenever we generate the zip file to dump into (possibly) Azure blob storage.

Long-term, if we decide to have both solution styles for all samples, it would be DocsEngine responsible for creating both versions with the suo file.

So given that information, I'm going to merge this PR, as the solution is exactly like we want it. I'm out of time today but perhaps we can make a simple LINQPad script that would use SetStartupProjects and generate the diff that we could use whenever (if ever) we need to redeploy the zip file while we're in this testing phase.

PR #3698, opened 2018-04-03
No nuget warn for outbox

Scooletz:
`NHibernate` requires the `Iesi Collections` package. NuGet complains when restoring the dependent package and finding the closest version that matches the one specified by NH package. 

I thought that it could be useful to do not have a Warning in a sample project.


danielmarbach:
Can you explain why this is necessary?

Scooletz:
@danielmarbach I updated the description of the issue. Could you take a look now?

PR #3702, opened 2018-04-05
Added queue length limitation regarding dynamic subscribe/unsubscribe

ramonsmits:


monitoring/metrics/definitions.md
```diff
@@ -67,6 +67,7 @@ As noted before, the current implementation might produce estimates which signif
  * In [Distributor](/transports/msmq/distributor/)-based deployments there is no queue length metric provided for the distributor node, only for the workers
  * After restarting any component the estimated queue length value can be off until all messages sent before restart are consumed
  *  In [Service Fabric Stateful Services with Partition Affine Routing](/samples/azure/azure-service-fabric-routing/) the queue length cannot be determined. Individual partitions can have a different backlog of messages e.g. one partition has more messages to be processed then another or is faster than another.
+ * Does not work with dynamically subscribe/unsubscribe, assumption is that subscription routes are static.
```
mikeminutillo: This should by dynamic subscribe/unsubscribe or dynamically subscribing/unsubscribing.

As someone not familiar with the inner workings of the platform, I'm not sure I'd even know what this refers to. 

Maybe

> Can report incorrect information when routing changes. Once a link it is established between two endpoints, it is assumed that this link is stable. This link might be broken by one of the following:
>
> - Unsubscribing from an event
> - Moving a message handler to a different endpoint
>
> Once a link is broken, the queue length will still increase when new messages of the linked type are sent/published. As these messages never arrive at the original linked destination, the queue length for the receiver will not decrease. This results in  a steadily rising "floor" in the queue length metric. To break the link between the sender and receiver, restart the ServiceControl Monitoring instance.

SzymonPobiega:
@ramonsmits can you follow up?

mikeminutillo:
Can we close this as we're not pursuing the estimated queue length any further?

danielmarbach:
The PR is outdated, given that we are working on the native queue length feature I'm closing this PR. Feel free to reopen if required.

PR #3703, opened 2018-04-05
Added resource usage for SC Monitoring

ramonsmits:
Based on a support request.

### DIFF
--- a/monitoring/metrics/index.md
+++ b/monitoring/metrics/index.md
@@ -38,6 +38,23 @@ To enable collecting metrics in an environment:
 NOTE: [Monitoring NServiceBus solutions: Getting started](/tutorials/monitoring-setup/) is an in-depth, step-by-step tutorial about installing and configuring everything to get the most out of performance monitoring.
 
 
+## Performance impact on system resources
+
+A ServiceControl monitoring instance is very lightweight compares a regular ServiceControl instance. We would still advice not to host a monitoring instance on the same machine that is hosting any production endpoint instances.
+
+### Wire usage
+
+Each endpoint instance collects performance metrics which are buffered and then send. A single metrics message contains a batch of metrics values in a compact binary format. This makes the sending of performance metrics very lightweight.
+
+### Storage usage
+
+Metrics are processed by a Service Control Monitoring instance. A monitoring instance only requires RAM, it does not store any metrics on disk except the creation of a log file. Metrics data is stored in RAM in aggregated form for a maximum duration of one hour. A 100mb process can easily hold metric state for atleast 100 endpoint instances.
+
+### CPU Usage
+
+The metrics service is only performing simple summing aggregation logic. The CPU resources 
+
+
 ## Reporting metric data to other places
 
 - Metrics can be reported to [Windows Performance Counters](performance-counters.md).


monitoring/metrics/index.md
```diff
@@ -38,6 +38,23 @@ To enable collecting metrics in an environment:
 NOTE: [Monitoring NServiceBus solutions: Getting started](/tutorials/monitoring-setup/) is an in-depth, step-by-step tutorial about installing and configuring everything to get the most out of performance monitoring.
 
 
+## Performance impact on system resources
+
+A ServiceControl monitoring instance is very lightweight compares a regular ServiceControl instance. We would still advice not to host a monitoring instance on the same machine that is hosting any production endpoint instances.
```
Scooletz: > A ServiceControl monitoring instance is very lightweight compares a regular ServiceControl instance

Should be probably

> A ServiceControl monitoring instance is much lighter than a regular ServiceControl instance.

monitoring/metrics/index.md
```diff
@@ -38,6 +38,23 @@ To enable collecting metrics in an environment:
...
+
+### Wire usage
+
+Each endpoint instance collects performance metrics which are buffered and then send. A single metrics message contains a batch of metrics values in a compact binary format. This makes the sending of performance metrics very lightweight.
```
Scooletz: > A single metrics message contains a batch of metrics values in a compact binary format

Could be

> A single metrics message contains a batch of values written in a compact binary format.

Also
> This makes the sending of performance metrics very lightweight.

I'd shorten to:
> It makes reporting very lightweight

monitoring/metrics/index.md
```diff
@@ -38,6 +38,23 @@ To enable collecting metrics in an environment:
...
+
+### Storage usage
+
+Metrics are processed by a Service Control Monitoring instance. A monitoring instance only requires RAM, it does not store any metrics on disk except the creation of a log file. Metrics data is stored in RAM in aggregated form for a maximum duration of one hour. A 100mb process can easily hold metric state for atleast 100 endpoint instances.
```
Scooletz: > A monitoring instance only requires RAM, it does not store any metrics on disk except the creation of a log file. 
To:
> A monitoring instance stores data only in RAM. It does not store any data on disk except the creation of a log file. 

> Metrics data is stored in RAM in aggregated form for a maximum duration of one hour. A 100mb process can easily hold metric state for atleast 100 endpoint instances.

To:

> Metrics data are stored for at most one hour. A 100MB process can easily hold metric state for atleast 100 endpoint instances.

monitoring/metrics/index.md
```diff
@@ -38,6 +38,23 @@ To enable collecting metrics in an environment:
...
+
+### CPU Usage
+
+The metrics service is only performing simple summing aggregation logic. The CPU resources
```
Scooletz: Something missing here

monitoring/metrics/index.md
danielmarbach: For my curiosity. Why wouldn't we recommend running it on the same box as endpoints?

SzymonPobiega:
@ramonsmits ping

monitoring/metrics/index.md
ramonsmits: @danielmarbach @Scooletz That is the advice in general to not host any platform tool components on the same box so that these cannot affect any production resources.

PR #1634, opened 2016-06-24
[WIP] ServiceControl installer updates

johnsimons:
Updated all images with the new UI changes
Referenced the new Maintenance Mode
Removed majority of manual steps to prevent misconfiguration
Reorganised the logging.md to emphasise important sections

PR #3713, opened 2018-04-09
Re-introduce ExposeRavenDB settings to doco

### LINKED ISSUE #3700: Add doco for ExposeRavenDB setting
https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl/Infrastructure/Settings/Settings.cs#L89

so, when debugging, users dont need to troll through the SC code base to reverese engineer

boblangley:
Fixes #3700

We [removed the documentation for this setting](https://github.com/Particular/docs.particular.net/pull/1634/files#diff-bb72f827aebf113f1a6918701701dbadL9) when we introduced Maintenance Mode. This PR re-introduces it in the Settings part of the doco and with a few changes.


boblangley:
@mikeminutillo I do not think so. Maintenance mode doesn't provide this type of warning, and the ServiceControl API does not provide any more security than this, so I am not sure we need to add _special_ warnings just for this setting.

PR #3723, opened 2018-04-11
Fix 'replay' typo in urls

ramonsmits:
- [x] Remaining files in `/nservicebus/sagas`, but cannot rename online.


SzymonPobiega:
@ramonsmits this does not build because you changed the reference to the file but not the file. Plus the redirect would probably not work for the `png` file.

weralabaj:
Seems like other issues have been addressed after Szymon's comment, so I'm marking this as completed and wait for build checks.

PR #3725, opened 2018-04-13
Use NServiceBus/LicensePath for retrieving license

ramonsmits:
Based on https://github.com/Particular/ServiceControl/issues/1132


ramonsmits:
@WilliamBZA I'm not convinced this works, I'll have to try it myself but question it because of:

> In Versions 1.17 and below, a license can be applied to an individual instance rather than using a license installed in the registry. To do this, copy the license.xml file to a license directory under the installation path of the instance.

Source: https://docs.particular.net/servicecontrol/license#instance-licensing

ramonsmits:
Getting the following error that I do not know how to resolve:

```
System.Exception : Some snippets exist that were not used in markdown. Or, the snippet may be used in some versions, but not all, in which case you need to remove the snippet from the version in which it's not needed. These are the unused snippets and the versions in which they weren't used: 
serializercache (>= 1.2.0 && < 2.0.0) 
bondserialization (>= 1.2.0 && < 2.0.0) 
bondserializationdelegates (>= 1.2.0 && < 2.0.0) 
bondcontenttypekey (>= 1.2.0 && < 2.0.0) 
metrics-enable (>= 1.1.0 && < 2.0.0) 
metrics-log (>= 1.1.0 && < 2.0.0) 
metrics-log-info (>= 1.1.0 && < 2.0.0) 
metrics-tracing (>= 1.1.0 && < 2.0.0) 
metrics-custom-function (>= 1.1.0 && < 2.0.0) 
sendmetricdatatoservicecontrol (>= 1.2.0 && < 2.0.0) 
sendmetricdatatoservicecontrolhostid (>= 1.2.0 && < 2.0.0) 
enable-criticaltime (>= 1.1.0 && < 2.0.0) 
enable-sla (>= 1.1.0 && < 2.0.0)
```

DavidBoike:
Stop adding dummy files. Rebase your work, that will rebuild it.

PR #3726, opened 2018-04-13
Remove WebSender from Core sample

bording:
Also includes additional cleanup on package references.

FYI @Particular/callback-maintainers @timbussmann


samples/callbacks/sample_dotnetcore_callbacks_[3,).partial.md
```diff
@@ -0,0 +1 @@
+WARNING: The WebSender project is not included in the .NET Core version of the sample
```
danielmarbach: Would an info be enough?

bording: I suppose so, but there's already another info box, so I was thinking about using warning just to have a color distinction. I could switch it, though.

danielmarbach:
I was just thinking it feels not important enough to be a warning but it is
just a gut feeling and nothing more.

Brandon Ording <notifications@github.com> schrieb am Fr., 13. Apr. 2018,
22:53:

PR #3731, opened 2018-04-17
Update ServiceControl samples

bording:
This cleans up the ServiceControl samples, which appear to have been missed from other recent rounds of cleanup.


SeanFeldman:
Approving. We should probably smoke test it at some point. As a separate task not to block this one.

bording:
I'm going to go ahead and merge.

PR #3736, opened 2018-04-20
Added note from which version this is added.

ramonsmits:
Instead of using document url `index_unrecoverableexceptions_core_[6.0,).partial.md` it maybe needs to be `index_unrecoverableexceptions_core_[6.2,).partial.md`. Not sure if that is supported.

@Particular/docs-maintainers

### DIFF
--- a/nservicebus/recoverability/index_unrecoverableexceptions_core_[6.0,).partial.md
+++ b/nservicebus/recoverability/index_unrecoverableexceptions_core_[6.0,).partial.md
@@ -1,5 +1,7 @@
 ## Unrecoverable exceptions
 
+Note: This is added to NServiceBus 6.2
+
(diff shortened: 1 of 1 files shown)


SzymonPobiega:
The version-based breakdown of the page is based on the snippet folders so the name of the partial is correct. Unless there is a specific folder with snippets for 6.2 the partial needs to reference to 6.0 as a version.

Adding a note makes sense. What do you think @Particular/nservicebus-maintainers ? Can one of you review this?

andreasohlund:
So we're saying that instead of adding a `[6.2,)` version of the page we're going with a note instead?

SzymonPobiega:
@andreasohlund yeah. The `[6.2,)` in the partial name would only work if there was a `6.2` snippets project in the corresponding snippets sln. Otherwise it won't work as there will be a single `6.0` page generated.

weralabaj:
Why approved but not merged? is it waiting for anything more?

PR #3750, opened 2018-04-27
Clarify ec2 setup guidance

andreasohlund:
To add ec2 for better key word searching


weralabaj:
Why approved but not merged? @mauroservienti is it waiting for anything more?

PR #3753, opened 2018-04-29
add sql native

SimonCropp:


DavidBoike:
@SimonCropp if you’re wondering about that new IntegrityTests build step, it’s a new thing to verify samples and snippets conform to conventions like not multitargeting and only using TargetFramework (not plural) in project files. If you rebase they’ll be in your fork and then the text will execute.

DavidBoike:
That’s why we haven’t made them a required build step yet. Will be adding them to readme before we do.

weralabaj:
So @DavidBoike is that ok if the ingerity tests don't pass or not? How do we fix the build then?

DavidBoike:
It's totally OK. If this is merged and happens to break the integrity tests, we'll find out about it on master and fix it. Guided by the docs principle that things can be a bit broken in prod and it's not a big deal.

Snippets/SqlTransportNative/SqlTransportNative_1/SqlTransportNative_1.csproj
```diff
@@ -0,0 +1,14 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFrameworks>net462;netstandard2.0</TargetFrameworks>
```
DavidBoike: I notice here you're multi-targeting a snippets project. What's your rationale for that? We were going back and forth on whether snippets should be allowed (encouraged? enforced?) to multi-target. To ensure that any APIs used are compiled and validated using both targets?

bording: The thinking around enforcing no multi-targeting for snippets as well as samples is that there shouldn't be a public API difference between the various assemblies in the package, so there's no reason to take the overhead of compiling multiple versions of the snippets assemblies.

Validating that the API is the same should happen in the project via some sort of API approval, not the snippets.

SimonCropp: Fair enough. But that conversation should be happening elsewhere. Not on this pr

Can this be merged and when the multi targetting is decided this can be fixed

SimonCropp: @bording some extra context. there are several snippet projects that use explicit unit tests to generate doco. u might want to confirm the multi target changes dont effect those. i dont think it will, but just to be sure...

SimonCropp:
@DavidBoike thanks

transports/sql/sql-native.md
```diff
@@ -0,0 +1,219 @@
...
+#### Single
+
+Sending a single message.
+
+snippet: Send
+
+
+#### Batch
+
+Sending a batch of message.
```
HindyLambrozo: should be messages

transports/sql/sql-native.md
```diff
@@ -0,0 +1,219 @@
...
+#### Single
+
+Sending a single message.
+
+snippet: SendDelayed
+
+
+#### Batch
+
+Sending a batch of message.
```
HindyLambrozo: should be messages

PR #3761, opened 2018-05-04
Ensure project references have lower bound to satisfy NU1604 error

DavidBoike:
NU1604: Project dependency (whatever) does not contain an inclusive lower bound. Include a lower bound in the dependency version to ensure consistent restore results.

Build server must have timed out on a package source and started using cached packages. Result is wrong versions being used that don't have the correct types (Ngpsql in this case) or mismatched versions (ms-logging) and of course when you run it locally it's fine.

The included integrity test will keep it from happening again.

### DIFF
--- a/tests/IntegrityTests/ReferenceVersions.cs
+++ b/tests/IntegrityTests/ReferenceVersions.cs
@@ -1,0 +1,28 @@
+﻿using System.Linq;
+using System.Xml.Linq;
+using System.Xml.XPath;
+using NUnit.Framework;
+
+namespace IntegrityTests
+{
+    public class ReferenceVersions
+    {
+       [Test]
+        public void MustScopeReferencesToAMajorVersion()
+        {
+            new TestRunner("*.csproj", "Package References cannot have Version=\"*\" or restore will sometimes fail and yield old, incorrect, or mismatched versions.")
+                .Run(projectFilePath =>
+                {
+                    var xdoc = XDocument.Load(projectFilePath);
+                    if (xdoc.Root.Attribute("xmlns") != null)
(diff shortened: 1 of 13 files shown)


tests/IntegrityTests/ReferenceVersions.cs
```diff
@@ -0,0 +1,34 @@
...
+                    var xmlnsAtt = xdoc.Root.Attribute("xmlns");
+
+                    if (xmlnsAtt != null)
+                    {
+                        query = "/x:Project/x:ItemGroup/x:PackageReference[@Version='*']";
+                        var xmlns = xmlnsAtt.Value;
+                        nsMgr.AddNamespace("x", xmlns);
+                    }
+
+                    var badPackageRefs = xdoc.XPathSelectElements(query, nsMgr);
```
bording: Seems like it would be worth seeing if LINQ to XML would result in simpler code than all the stuff needed here to get xpath working.

DavidBoike: The LINQ-to-XML is roughly:

```
var xdoc = XDocument.Load(path);
XNamespace ns = xdoc.Root.Attribute("xmlns")?.Value ?? "";
 
var badRefs = xdoc
    .Descendants(ns + "PackageReference")
    .Where(pr => pr.Attribute("Version")?.Value == "*");
```

Save some lines, gain some confusing `?.` and ternary operators, and you're no longer validating that the PackageReference is inside `/Project/ItemGroup` ... it could be anywhere.

Unfortunately XML namespaces are awful and Linq-to-Xml can't get you out of them. :cry:

PR #3766, opened 2018-05-07
Clarify SQLT header columns

SzymonPobiega:


SzymonPobiega:
@SimonCropp can you take a look?

transports/sql/sql-statements.md
```diff
@@ -75,13 +75,17 @@ Retrieves a message from the queue.
 
 snippet: ReceiveTextSql
 
+NOTE: The `CorrelationId`, `ReplyToAddress` and `Recoverable` columns are required for backwards compatibility with Version 1 of the transport. When receiving messages sent by endpoints that use later versions, the values of correlation ID and reply-to address should be read from the headers (`NServiceBus.CorrelationId` and `NServiceBus.ReplyToAddress`) instead. The value `Recoverable` can be ignored as it always is `true`/`1`.
```
SimonCropp: So if not targetting version 1, those columns dont need to exist?

SzymonPobiega: They do need to exist as they are part of the `select` query. They could actually be removed in the next release.

SimonCropp: So it is ok for the values to exist in the headers column, and those explicit columns be null?

SzymonPobiega: Yeah. The values are read here https://github.com/Particular/NServiceBus.SqlServer/blob/develop/src/NServiceBus.SqlServer/Queuing/MessageRow.cs#L46 because of the `sequential access` flag. Then https://github.com/Particular/NServiceBus.SqlServer/blob/develop/src/NServiceBus.SqlServer/Queuing/MessageRow.cs#L60 actually overrides the `reply-to` value present in the headers with a value from the column (if set). Note that the correlation ID is not used at all.

SimonCropp: so given v1 is no longer supported shouldnt this page be updated to write nulls to those legacy columns?

SzymonPobiega: That's a good question. I think that the original purpose of this page was to illustrate the actual queries that are used by the transport but then it became sort of scripting guide.

I'd say that the scripting section should definitely be reviewed but this page I am not sure...

SimonCropp: sorry, yes i meant the other page

SimonCropp:
thanks peeps

PR #3771, opened 2018-05-09
Add websocket gateway 3 sample

weralabaj:
rebased the reversed pr


bording:
FYI since #3779 has been merged, this should be updated to use RTM wildcards as well.

PR #3774, opened 2018-05-14
How to use a shared Ninject kernel between instances

ramonsmits:
Added remark on how to use a shared Ninject kernel between multiple NServiceBus endpoint instances

### DIFF
--- a/dummy
+++ b/dummy
@@ -1,0 +1,1 @@
+Dummy to easily force rebuild
--- a/dummy
+++ b/dummy
@@ -1,1 +1,2 @@
 Dummy to easily force rebuild
+Another dummy, as the previous build is gone from TC
--- a/dummy
+++ b/dummy
@@ -1,2 +1,0 @@
-Dummy to easily force rebuild
-Another dummy, as the previous build is gone from TC
--- a/nservicebus/dependency-injection/ninject.md
+++ b/nservicebus/dependency-injection/ninject.md
@@ -27,6 +27,10 @@ snippet: Ninject_Existing
 
 partial: uow
 
+### Multi hosting
+
+Sometimes you want to host multiple endpoints in a single process. You cannot share a single Ninject kernel with multiple endpoint instances. You need to create a child container. Ninject supports this with the [Ninject.Extensions.ChildKernel](https://github.com/ninject/Ninject.Extensions.ChildKernel) extension. Before you pass the kernel to NServiceBus you need to do `new ChildKernel(parentKernel)` and pass this new kernel instance.
+
 
 ### DependencyLifecycle Mapping
 
--- a/nservicebus/dependency-injection/ninject.md
+++ b/nservicebus/dependency-injection/ninject.md
@@ -29,7 +29,7 @@ partial: uow
 
 ### Multi hosting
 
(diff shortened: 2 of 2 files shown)


timbussmann:
not really sure we want to document this procedure. In most cases, hosting multiple endpoints in a single process is not the best approach. Introducing a shared DI container increases the smell imo. If you need to share something between endpoints, register the shared instance via the built-in registration API for every endpoint instead so you don't have to introduce container specific solutions.

thoughts @Particular/container-maintainers ?

ramonsmits:
Hosting multiple endpoints within a single process is in my opinion not really a big issue at all. Why wouldn't this be a good approach?

timbussmann:
> Hosting multiple endpoints within a single process is in my opinion not really a big issue at all. Why wouldn't this be a good approach?

because the driver for that most often is that these endpoints need to share something. if this isn't some sort of a bridging endpoint, then two endpoints shouldn't have shared dependencies and shouldn't share a any instances?

WilliamBZA:
I can confirm that child containers still share the same instance of singletons. We should probably mention that somewhere?

timbussmann:
> I'm not sure I agree with that. Different endpoints can be hosted together in a single process without there needing to be coupling. It could simply be a deployment configuration (e.g. in dev run all the endpoints in one process so you don't need to have 30 console apps open, but in prod deploy them separately)

in that case I see no reason to share a kernel between the endpoints?

> I can confirm that child containers still share the same instance of singletons. We should probably mention that somewhere?

Do you think is is counter-intuitive? I can only speak from my perspective, but this seems to be exactly what a singleton is supposed to do?

ramonsmits:
> in that case I see no reason to share a kernel between the endpoints?

That is just not possible because NServiceBus cannot register the same interfaces for different endpoint instances.

timbussmann:
> That is just not possible because NServiceBus cannot register the same interfaces for different endpoint instances.

I don't understand what you're referring to?

ramonsmits:
> I don't understand what you're referring to?

@timbussmann You said, why not use a single container / kernel for all instances. That is not possible as that would result in duplicate registrations.

timbussmann:
@ramonsmits what I was trying to say is that I think that the described scenario is **not** a valid reason to share a kernel between endpoints. Hosting multiple endpoints in one process is one thing, but sharing instances smells like strong coupling between those endpoints and that they can't be hosted independently.

danielmarbach:
marked as WIP until discussion is held

timbussmann:
@ramonsmits any updates on this?

ramonsmits:
@timbussmann I need to fix content verification issues. I still would like to merge this. Do you think we need a discussion?

I'm not sure what the issue is to not share singletons between multiple endpoints instance or even prevent coupling between endpoints.

ramonsmits:
My stance is that this is a valid construct. It is a problem that once in a while pops up. Even if you only host a single endpoint but combine it with a web API or any other kind of communication mechanism where a specific resource is shared between all api facades.

Another option is when you would have multiple instances in a single process where you would have a form or priority queues.

So I do agree with @timbussmann that is the endpoints are logically different that you probably shouldn't need to share a parent container but there are many cases that this is a valid setup.

PR #3776, opened 2018-05-14
Document the EnableV1CompatibilityMode setting for SQS

boblangley:
For the 3.3.0 and 4.0 release

Connects to https://github.com/Particular/NServiceBus.AmazonSQS/issues/220


transports/sqs/configuration-options_v1compatibilitymode_sqs_[3,).partial.md
```diff
@@ -0,0 +1,11 @@
+## V1 Compatibility Mode
+
+**Optional**
+
+**Default**: Disabled.
+
+This enables serialization of the `TimeToBeReceived` and `ReplyToAddress` in the message envelope in addition to the headers for compatibility with endpoints using version 1 of the transport.
```
SeanFeldman: Not a request to change, but would this be easier to digest:

>This option enables serialization of the `TimeToBeReceived` and `ReplyToAddress` message headers in the message envelope for compatibility with endpoints using version 1 of the transport.

PR #3792, opened 2018-05-22
Draft info about pessimistic concurrency

SzymonPobiega:


persistence/sql/saga-concurrency_pessimistic_[4.1,).partial.md
```diff
@@ -0,0 +1 @@
+In addition to that the persister also uses pessimistic concurrency to avoid excessive optimistic concurrency conflict for highly congested sagas. The pessimistic concurrency is implemented using `SELECT ... FOR UPDATE` or equivalent construct.
```
weralabaj: I think right now it's not clear when persister uses optimistic vs pessimistic concurrency, we should clarify, e.g how about listing operations for pessimistic?

SzymonPobiega: I think that 

> implemented using `SELECT ... FOR UPDATE`

implies that the pessimistic concurrency is used when querying.

PR #3797, opened 2018-05-24
Proper aspnetcore shutdown

andreasohlund:


andreasohlund:
Update: Seems to be an issue with IISExpress https://stackoverflow.com/a/43844804/236004

Do we keep the code? Switch to Kestrel?

andreasohlund:
Works when selecting kestrel instead of IISExpress

image


Wonder if there is a way to make the the default? cc @bording

andreasohlund:
>Wonder if there is a way to make the the default?

Removing the launchSettings file did the trick

DavidBoike:
I'm 100% OK with this. I don't see the point of keeping IIS Express around when aspnetcore is all about Kestrel.

bording:
Not to mention that kestrel is cross-platform, but IIS Express isn't!

andreasohlund:
This one should be good to go, please review and merge

samples/web/asp-mvc-application/Core_7/AsyncPagesMVC.Core/AsyncPagesMVC.Core.csproj
```diff
@@ -18,4 +18,8 @@
     <ProjectReference Include="..\Shared\Shared.Core.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
```
bording: Don't think this should be here

samples/web/send-from-aspnetcore-webapi/Core_7/WebApplication/WebApplication.Core.csproj
```diff
@@ -5,7 +5,10 @@
     <LangVersion>7.1</LangVersion>
   </PropertyGroup>
   <ItemGroup>
-    <Folder Include="wwwroot\" />
+    <Compile Remove="wwwroot\**" />
```
bording: Don't think this should be here

samples/web/send-from-mvc-controller/Core_7/WebApplication.Core/Startup.cs
```diff
@@ -18,6 +18,7 @@ public Startup(IConfiguration configuration)
 
         public IConfiguration Configuration { get; }
 
+
```
bording: 🔥

samples/web/send-from-mvc-controller/Core_7/WebApplication.Core/WebApplication.Core.csproj
```diff
@@ -6,8 +6,8 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.All" Version="2.0.6" />
-	<PackageReference Include="Autofac.Extensions.DependencyInjection" Version="4.2.2" />    
-    <PackageReference Include="NServiceBus.Autofac" Version="7.*" />   
+	<PackageReference Include="Autofac.Extensions.DependencyInjection" Version="4.2.2" />
```
bording: Is there a tab here, or is the spacing off?

andreasohlund:
@bording should be good to go now

PR #3799, opened 2018-05-24
Removed SendLocal from sender

dvdstelt:
The RabbitMQ Native sample contained a `SendLocal` in `Program.cs` with an empty message, causing every start of the receiver an empty message to be received.
This was confusing and has now been removed.

Discussion in Slack: https://particularsoftware.slack.com/archives/C06CXSNF4/p1527156607000068


bording:
I'm not sure why that was in there, either. Must have been a mistake.

dvdstelt:
I think Andreas initially created it? He mentioned that it was probably a copy-n-paste issue. I suspect a test to see if the endpoint was properly working, etc.

PR #3801, opened 2018-05-24
Netcore sample list

DavidBoike:


DavidBoike:
We can iterate on making the full list of samples nicer, but this is a good starting point that checks all boxes.

bording:
@DavidBoike Would it make sense to modify the sample? Probably not...

DavidBoike:
Changed to sniff project files for `netcoreapp2.0` and discovered a couple broken samples in the process.

Assuming this goes green I'll merge.

PR #3817, opened 2018-05-30
Document integrity tests before making them required on TeamCity

DavidBoike:
The integrity test experiment has been very useful during the .NET Core launch, for example, providing an easy way to make sure that we removed the correct prerelease.txt files at launch and ensuring that they'll stay correct.

But that won't work if we don't make them required to merge PRs.

This PR documents existing integrity tests. Assuming we merge this, we can then require the checks on PRs in TeamCity, and then we would also reject any PR that modified the integrity tests without also documenting them in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -878,11 +878,13 @@ A [number of integrity tests](tests/IntegrityTests) validate that samples and sn
 
 The integrity tests include:
 
-* [Samples should not be multi-targted](tests/IntegrityTests/ProjectFrameworks.cs) using the `<TargetFrameworks>` (plural) element, and instead have separate solutions and projects to support .NET Framework and .NET Core, all using the same underlying code, i.e.:
+* [Samples should not be multi-targeted](tests/IntegrityTests/ProjectFrameworks.cs) using the `<TargetFrameworks>` (plural) element. This is because a visitor who downloads the solution without .NET Core installed will not be able to run a multi-targeted project even on the .NET Framework. Instead a sample should have separate solutions and projects to support .NET Framework and .NET Core, all using the same underlying code, i.e.:
     * `SampleName.sln` references `ProjectName.csproj` targeting `net462`
     * `SampleName.Core.sln` references `ProjectName.Core.csproj` targeting `netcoreapp2.0`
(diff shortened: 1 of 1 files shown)


DavidBoike:
P.S. After merging and making the TC tests mandatory I would make a company-wide announcement.

DavidBoike:
@Particular/docs-maintainers I would still like your review here.

README.md
```diff
@@ -866,6 +867,18 @@ This is a temporary state and once a stable is released it is changed back to th
 ```
 
 
+## Integrity Tests
+
+A [number of integrity tests](tests/IntegrityTests) validate that samples and snippets conform to conventions that make the documentation easier to maintain. PRs that fail these tests will not be mergeable.
+
+The integrity tests include:
+
+* [Samples should not be multi-targted](tests/IntegrityTests/ProjectFrameworks.cs) using the `<TargetFrameworks>` (plural) element, and instead have separate solutions and projects to support .NET Framework and .NET Core, all using the same underlying code, i.e.:
```
weralabaj: I think it'd be useful here to explain why we use this approach, as this may change in the future with updated tooling support and it'll be easier to catch that rules are outdated.

README.md
```diff
@@ -866,6 +867,18 @@ This is a temporary state and once a stable is released it is changed back to th
...
+    * `SampleName.sln` references `ProjectName.csproj` targeting `net462`
+    * `SampleName.Core.sln` references `ProjectName.Core.csproj` targeting `netcoreapp2.0`
+* [Package references cannot use a wildcard-only version](tests/IntegrityTests/ReferenceVersions.cs`) using `Version="*"` as this can cause a package restore operation to sometimes fail and yield old, incorrect, or mismatched versions.
+* Versioned sample/snippet directories (i.e. `Core_7`) must contain a `prerelease.txt` file **only** when the contained projects use a prerelease version of that component's NuGet package.
```
weralabaj: Provide explanation how it's verified, e.g. how analyzer verifies that pre-release version is used.

DavidBoike:
@weralabaj Addressed your comments in the integrity tests readme doco.

weralabaj:
Thanks @DavidBoike !

PR #3819, opened 2018-05-31
Fix bond snippets

andreasohlund:
To hopefully make master green


SimonCropp:
replaced by https://github.com/Particular/docs.particular.net/pull/3821

PR #3820, opened 2018-05-31
Updated to not start the instance when running installers

ramonsmits:
- [x] Update samples for other core versions besides 6
- [x] ~~Document DisableInstallers for v7~~

### DIFF
--- a/Snippets/Core/Core_5/Installers.cs
+++ b/Snippets/Core/Core_5/Installers.cs
@@ -35,7 +35,7 @@ namespace Core5
                 busConfiguration.EnableInstallers();
                 // This will run the installers but not start the instance.
                 Bus.Create(busConfiguration);
-                Environment.Exit(0); // Make sure we stop so we do not accidentally start consuming messages
+                Environment.Exit(0);
             }
             
             #endregion
--- a/Snippets/Core/Core_5/Installers.cs
+++ b/Snippets/Core/Core_5/Installers.cs
@@ -33,8 +33,9 @@ namespace Core5
             if (runInstallers)
             {
                 busConfiguration.EnableInstallers();
+                // This will run the installers but not start the instance.
                 Bus.Create(busConfiguration);
-                return;
+                return; // Exit application
             }
         }
 
--- a/Snippets/Core/Core_5/Installers.cs
+++ b/Snippets/Core/Core_5/Installers.cs
@@ -1,4 +1,5 @@
-﻿namespace Core5
+﻿// ReSharper disable RedundantJumpStatement
+namespace Core5
 {
     using System;
     using System.Linq;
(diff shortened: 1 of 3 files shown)


ramonsmits:
Looking at the code for versions before core v5 and it doesn't seem that you cannot create but not start an endpoint. @Particular/nservicebus-maintainers is that correct?

ramonsmits:
@Particular/nservicebus-maintainers In V6.4 `DisableInstallers` was introduced

- https://docs.particular.net/nservicebus/operations/installers?version=core_6

What exactly is its use? I always thought installers were never run unless explicitly via `EnableInstallers()`.

timbussmann:
> What exactly is its use? I always thought installers were never run unless explicitly via EnableInstallers().

@ramonsmits that is not the case. See the v5 version of the installer doco page https://docs.particular.net/nservicebus/operations/installers?version=core_5 which states:

> By default, installers are only run when the application starts with a debugger attached. This includes applications being debugged remotely.

ramonsmits:
@timbussmann thanks for the clarification, so DisableInstallers results in never to run the installers also when debugging prior to 7 and from 7 onwards we do not invoke the installers during debugging. Is that correct? That doco is so very difficult to understand differences between major versions.

timbussmann:
> DisableInstallers results in never to run the installers also when debugging

that's correct. Note that the `DisableInstallers` API was added in 6.4 IIRC.

> from 7 onwards we do not invoke the installers during debugging.

that's correct, unless you explicitly specify `EnableInstallers`

ramonsmits:
@timbussmann As you already provided help, could you review?

Snippets/Core/Core_5/Installers.cs
```diff
@@ -32,6 +33,9 @@ public static void Main(string[] args)
             if (runInstallers)
             {
                 busConfiguration.EnableInstallers();
+                // This will run the installers but not start the instance.
+                Bus.Create(busConfiguration);
+                return; // Exit application
```
timbussmann: could we maybe just remove this and get rid of the resharper warning comment at the top of the file? or use Environment.Exit or so?

Snippets/Core/Core_6/Installers.cs
```diff
@@ -27,13 +28,17 @@ class SwitchInstallersWithCommandline
...
         {
             var runInstallers = args.Any(x => x.ToLower() == "/runInstallers");
 
             if (runInstallers)
             {
                 endpointConfiguration.EnableInstallers();
+                // This will run the installers but not start the instance.
+                await Endpoint.Create(endpointConfiguration)
+                    .ConfigureAwait(false);
+                return; // Exit application
```
timbussmann: same here as the comment above

PR #3834, opened 2018-06-05
Fail integrity tests when WARN used instead of WARNING

danielmarbach:


DavidBoike:
I'd personally be fully OK with using WARN in addition to WARNING in the engine. (It already allows NOTE or INFO for blue boxes.) @Particular/docs-maintainers any thoughts?

DavidBoike:
Argument for accepting WARN in the engine would be that just writing something down that seems right and having it work the first time will be more productive long term than either (1) not realizing it's a thing, having your PR blow up, and having to fix it and commit again, or (2) knowing it's a thing but not being able to remember which one is right, so you're forced to take time to look it up so as to avoid the build blowing up.

danielmarbach:
Ok will send in a docs engine PR

PR #3835, opened 2018-06-05
Compliance documentation for GDPR

WojcikMike:


dbelcham:
@WojcikMike not sure what I've done is complete, but take a look and see if you think it's heading in a good direction

dbelcham:
@WojcikMike not sure where this really belongs in the bigger doco menu structure. Adding it under NServiceBus | Security seems somewhat appropriate though.

WojcikMike:
@Particular/docs-maintainers do you have any opinion on where this page should show in the menu?

DavidBoike:
Does NServiceBus need to have a new subcategory called "Compliance" as the last option, after Operations?

dbelcham:
@DavidBoike I liked the idea of creating a whole new section for Compliance. I suspect we could flush that out with advice for many things like HL7, banking, insurance, etc. The only thing I think is weak for it is the `index` landing page. What other things do you think I could/should mention in it?

DavidBoike:
Yeah the index would be weak. I don't think that's a big deal. "This page contains any compliance information related to using NServiceBus in production systems." then UL of topics. Does the job.

PR #3838, opened 2018-06-06
WARN can be used as well

danielmarbach:


weralabaj:
I've made the same change on master before looking the content of this PR, I assumed it's someting else :)

PR #3841, opened 2018-06-06
Use ReceiveOnly transaction mode to prevent usage of TransactionScope in ASB

SzymonPobiega:
This does not change the behavior but is easier to configure.

### DIFF
--- a/samples/azure/azure-service-bus-msmq-bridge/ASB_7/Bridge/Program.cs
+++ b/samples/azure/azure-service-bus-msmq-bridge/ASB_7/Bridge/Program.cs
@@ -1,6 +1,5 @@
 ﻿using System;
 using System.Threading.Tasks;
-using System.Transactions;
 using NServiceBus;
 using NServiceBus.Bridge;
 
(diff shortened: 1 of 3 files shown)


samples/azure/azure-service-bus-msmq-bridge/ASB_7/Bridge/Program.cs
```diff
@@ -22,6 +21,8 @@ class Program
             .Between<MsmqTransport>("Bridge-MSMQ")
             .And<AzureServiceBusTransport>("Bridge-ASB", transport =>
             {
+                //Prevents ASB from using TransactionScope
+                transport.Transactions(TransportTransactionMode.ReceiveOnly);
```
SeanFeldman: Is my understanding correct that despite downgrading transport transaction mode to `RecieveOnly` we're not going to see ghost messages? I.e. the bridge (in both directions) will function as if it was configured `SendsAtomicWithReceive`?

SeanFeldman: I just want to make sure that's the case. OK to merge if that's the case, just have no idea how's that accomplished 😃

SzymonPobiega: Everything will be as it used to be because previously we were suppressing the TransactionScope anyway. So there might be cases where the message will be sent to the other side of the bridge while the ASB message ACK fails but it should be resolved by retry.

PR #3844, opened 2018-06-07
Microsoft.Extensions.Logging.Abstractions

SimonCropp:


nservicebus/logging/microsoft.md
```diff
@@ -10,7 +10,7 @@ related:
  - samples/logging/mslogging-custom
 ---
 
-Support for writing all NServiceBus log entries to [Microsoft.Extensions.Logging](https://github.com/aspnet/Logging).
+Support for writing all NServiceBus log entries to [Microsoft.Extensions.Logging.Abstractions](https://github.com/aspnet/Logging).
```
danielmarbach: So this was changed to indicate that we support the abstraction and not the concrete implementation, correct?

SimonCropp: yeah that is the underlying nuget that is supported. but re-reading in the light of your question i think Microsoft.Extensions.Logging is sufficient since it points to the doco.

so i will close

PR #3847, opened 2018-06-07
Sql Native v2

SimonCropp:


transports/sql/sqlnative-not-node.include.md
```diff
@@ -0,0 +1,3 @@
+## Licensing
```
danielmarbach: @SimonCropp The PR looks good. Not sure about this note though. I need to verify this internally.

danielmarbach: According to the license, a node is defined as

> A "node" is a single installation of an operating system running on a machine (virtual or physical) that will utilize the Software. The "node" is counted when used by Developers, Testers and in Production.

since when this extension is used "the software" doesn't run it never counts as a node. I'm removing this note while merging

danielmarbach:
Replaced by https://github.com/Particular/docs.particular.net/pull/3852

transports/sql/sqlnative-not-node.include.md
SimonCropp: I had already confirmed this with @andreasohlund

SimonCropp: @andreasohlund no nservicebus/licensing is not enough. this is important information for anyone using native integration. trusting that they will stumble accross it when reading about licensing is likely to result in a certain number of customers paying for nodes that they shouldnt be. 

Also sql native is new. so an existing user discovering it now, would likely have already have read the licensing page and not expect it to have changed.

andreasohlund: We're having some internal discussion on how to sort this out, will let you know once we have more info

danielmarbach: We discussed this internally together with the developer success team and we don't think a component (even more a community component) should explicitly mention the licensing on the docs site. The licensing is part of the licensing description on the website. If customers have questions about licensing they can always reach out to our developer success team

PR #3849, opened 2018-06-08
SqlP supporting core sagas

DavidBoike:
Connects to https://github.com/Particular/PlatformDevelopment/issues/2020


persistence/sql/sqlsaga.md
```diff
@@ -0,0 +1,47 @@
+---
```
weralabaj: Based on the recent discussion, I understood that we're going to support both SqlSaga and core saga as equally good options, until things change (e.g. core catches up with mapping api etc.).
So based on that assumption I think we should describe both on the same page instead of moving SQL Saga away, make two sections on the same level "Saga" and "SqlSaga" (in that order), and move the SqlSaga-specific sections one level deeper in the page.

> Which API to use is largely a matter of personal preference.

IMO that'll be more confusing for the user, so not lowering the entry barrier, which is our goal here. We need a specific recommendation.

I think it's ok to say "go with core saga by default for new projects" and then say when to use SQL saga, i.e. edge cases such as derived classes for IL inspection, preference for more concise syntax, etc. But it's our thing to guide user in making that decision, especially a person new to NServiceBus.

persistence/sql/sqlsaga.md
```diff
@@ -0,0 +1,47 @@
...
+title: SqlSaga Base Class
+component: SqlPersistence
+related:
+ - persistence/sql/saga
+reviewed: 2018-06-08
+---
+
+`SqlSaga<T>` is an alternate, experimental saga base class for use with SQL Persistence that offers a less verbose mapping API than `NServiceBus.Saga<T>`. It is possible (but not guaranteed) that in the future, the NServiceBus saga class will evolve to look more like `SqlSaga`. Which API to use is largely a matter of personal preference.
```
weralabaj: > `SqlSaga<T>` is an alternate, experimental saga base class

If I understood correctly, for now we're going to fully support both as equal citizens, but this phrase indicates a different thing and may raise unnecessary concerns among people who currently are using SqlSagas.

> It is possible (but not guaranteed) that in the future, 

Skip this part since we cannot make any promises here.

persistence/sql/sqlsaga_required-in-some-versions_sqlpersistence_[,4).partial.md
```diff
@@ -0,0 +1 @@
+INFO: In order to use the normal `Saga<T>` class for sagas, upgrade to at least SQL Persistence 3.2.
```
weralabaj: > the normal `Saga<T>` class

the NServiceBus Core `Saga<T>` class or the NServiceBus Core `Saga<T>` base class

weralabaj: > upgrade to at least SQL Persistence 3.2.

upgrade SQL Persistence to Version 3.2 or above or use SQL Persistence Version 3.2 or above

persistence/sql/sqlsaga_required-in-some-versions_sqlpersistence_[4,5).partial.md
```diff
@@ -0,0 +1 @@
+INFO: In order to use the normal `Saga<T>` class for sagas, upgrade to at least SQL Persistence 4.2.
```
weralabaj: > upgrade to at least SQL Persistence 4.2.

upgrade SQL Persistence to Version 4.2 or above or use SQL Persistence Version 4.2 or above

weralabaj: > the normal `Saga<T>` class

the NServiceBus Core `Saga<T>` class or the NServiceBus Core `Saga<T>` base class

persistence/sql/index.md
```diff
@@ -30,7 +30,7 @@ SQL Persistence can store NServiceBus data in multiple different relational data
 * Sagas are:
   * Stored using [Json.NET](http://www.newtonsoft.com/json) to serialize complex data structures, with no need to manage complex table structures
   * Built to be [version-aware](saga.md#json-net-settings-custom-settings-version-type-specific-deserialization-settings) with support for data evolution
-  * Built to [allow changing the `CorrelationId` over time](saga.md#correlation-ids-correlation-and-transitional-ids)
+  * Built to [allow changing the `CorrelationId` over time](saga.md#correlation-ids)
```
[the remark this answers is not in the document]
DavidBoike: This is just the intro I wrote awhile back to give a visitor a sense of the features made SqlP attractive before getting into the technical stuff. The URL had to change because of the complexities of having older majors (v1 which is out of support and v2 which runs out before the end of the year) that don't support core sagas. The user still gets led to the same content.

persistence/sql/sqlsaga.md
DavidBoike: Made some changes that you suggested, however it's nearly impossible to present it all together on one page. I tried. The weight of the old versions (especially v1-2) makes it really difficult. I had to go with presenting the Core API as default because that's what a new visitor would be aware of, and presenting SqlSaga as another thing you could go look at with its own page. The separate page makes a different pool with lower overall complexity making the versioning possible to pull off there.

persistence/sql/sqlsaga_required-in-some-versions_sqlpersistence_[,4).partial.md
DavidBoike: I made this change. I don't know why GitHub is still showing it this way!

persistence/sql/sqlsaga_required-in-some-versions_sqlpersistence_[4,5).partial.md
DavidBoike: I made this change. I don't know why GitHub is still showing it this way!

DavidBoike: I made this change. I don't know why GitHub is still showing it this way!

DavidBoike:
@SzymonPobiega Have a look. Everything is ready to go except the alpha/beta packages would need to be replaced with the release packages (once they're built) which would also take of the integrity test that's mad because there's no prerelease.txt even though prerelease packages are being used.

DavidBoike:
I don't know what's going on with this PR. GitHub is showing the diff to be contrary to what's actually in the commits.

Going to close this one, squash the commits, and open a brand new PR.

DavidBoike:
Ok, I take it all back! Turns out that phrasing was in 4 partials and I thought there were only 2, which I had updated, and was then looking at the wrong ones.

So I go back to my previous statement:

> @SzymonPobiega Have a look. Everything is ready to go except the alpha/beta packages would need to be replaced with the release packages (once they're built) which would also take of the integrity test that's mad because there's no prerelease.txt even though prerelease packages are being used.

PR #3850, opened 2018-06-08
Mention SourceLink in 6to7 upgrade guide

DavidBoike:
Connects to https://github.com/Particular/DotNetCoreLaunch/issues/164


bording:
SourceLink is actually something implemented by the compiler, so maybe it would make sense to link to https://github.com/dotnet/designs/blob/master/accepted/diagnostics/source-link.md instead?

We're currently using the tooling provided by https://github.com/ctaggart/SourceLink, but we'll probably be switching to Microsoft's packages in the future (in fact core already has).

The `SourceLink.Copy.PdbFile` package he created and linked to in his Known Issues is a way to workaround that problem until it gets fixed for real...

PR #3857, opened 2018-06-12
Nested encryption upgrade guide under Message Property Encryption

ramonsmits:


ramonsmits:
Getting build server error:
```
System.Exception : Failed to parse menu.yaml. Error:
 Object reference not set to an instance of an object.
  ----> System.NullReferenceException : Object reference not set to an instance of an object.
   at MenuParser.Start() in C:\BuildAgent\work\ca82fba321bd227a\src\ParticularDocsCore\Navigation\MenuParser.cs:line 26
```

Not sure how to fix, so removed that change from the PR only leaving the missing article that was not linked in the menu yet.

PR #3869, opened 2018-06-19
Azure Servicebus standard documentation

yvesgoeleven:


transports/azure-service-bus-netstandard/compatibility.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Backward compatibility
+
+The Azure Service Bus .Net Standard transport is backward compatible with the Azure Service Bus transport under a certain set of preconditions.
+
+### Preconditions
+
+#### Forwarding topology
+
+The Azure Service Bus .Net Standard transport only supports the forwarding topology, an entity layout where a shared set of topics is used for publishing between the endpoints.
```
SeanFeldman: > a shared set of topics is used for publishing between the endpoints.

Single topic only. ASB has removed "bundling" and by default it's one topic only.

transports/azure-service-bus-netstandard/compatibility.md
```diff
@@ -0,0 +1,30 @@
...
+
+#### Single namespace
+
+The Azure Service Bus .Net Standard transport only supports a single namespace at the moment.
```
SeanFeldman: Please remove "at the moment". We don't want to write anythign we can't commit to. I'd rather say less with what we've got, than promise more of what we might or might not have.

transports/azure-service-bus-netstandard/compatibility.md
```diff
@@ -0,0 +1,30 @@
...
+
+#### Topic prefix must be matching
```
SeanFeldman: ASB's feature is called ["bundle prefix"](https://docs.particular.net/transports/azure-service-bus/configuration/full#configuring-the-topology-forwarding-topology). The overall result (topic path) must match.

transports/azure-service-bus-netstandard/compatibility.md
```diff
@@ -0,0 +1,30 @@
...
+
+Both transports must be configured using the same topic prefix for publishing to work properly.
```
SeanFeldman: For ASBS we'll probably just call it something like "publishing topics" or similar. It won't be just a prefix, but the whole path.

transports/azure-service-bus-netstandard/compatibility.md
```diff
@@ -0,0 +1,30 @@
...
+
+#### Namespace alias not used
+
+The Azure Service Bus .Net Standard transport currently doesn't support namespace aliases.
```
SeanFeldman: > The Azure Service Bus .Net Standard transport ~~**currently**~~ doesn't support namespace aliases.

transports/azure-service-bus-netstandard/configuration.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Configuration
+summary: Explains the configuration options
+component: ASBS
+tags:
+ - Azure
+reviewed: 2018-06-01
+---
+
+some markdown content required for docsengine to work
```
SeanFeldman: 😮 didn't know that

transports/azure-service-bus-netstandard/index.md
```diff
@@ -0,0 +1,21 @@
...
+title: Azure Service Bus .Net Standard Transport
+component: ASBS
+tags:
+ - Azure
+reviewed: 2018-06-19
+---
+
+## Azure Service Bus .Net Standard Transport
+
+The Azure Service Bus .Net Standard transport, ASBS in short, is the successor to the [Azure Service Bus transport](transports/azure-service-bus/). It leverages the .Net Standard compatible [Microsoft.Azure.ServiceBus](https://www.nuget.org/packages/Microsoft.Azure.ServiceBus/) client SDK.
```
SeanFeldman: > ASBS is the successor to the ASB

and 

> It leverages the .Net Standard compatible client SDK.

are negating each other. It is **not** a successor IMO. It's a completely new and different transport. Just like the new ASB client is not a successor to the old one. I think we better off positioning it as a brand new transport to set expectations right.

transports/azure-service-bus-netstandard/index.md
```diff
@@ -0,0 +1,21 @@
...
+
+[Azure Service Bus (ASB)](https://azure.microsoft.com/en-us/services/service-bus/) is a messaging service hosted on the Azure platform, that allows for exchanging messages between various applications in a loosely coupled fashion. ASB Queues offer <a href="https://en.wikipedia.org/wiki/FIFO_(computing_and_electronics)">"First In, First Out" (FIFO)</a> guaranteed message delivery, and support a range of standard protocols (REST, AMQP, WS*) and APIs (to put messages on and pull messages off the queue). ASB Topics deliver messages to multiple subscribers and facilitate use of the fan-out pattern to deliver messages to downstream systems.
```
SeanFeldman: > ASB Queues offer guaranteed message delivery

Is this necessary? NSB doesn't allow message ordering, so why would we mention at all native feature we don't use?

transports/azure-service-bus-netstandard/compatibility.md
yvesgoeleven: How will ASBS expose this? As prefix? As path? As name?

yvesgoeleven: I see it's answered below

transports/azure-service-bus-netstandard/index.md
yvesgoeleven: maybe the term replacement is better?

transports/azure-service-bus-netstandard/index.md
yvesgoeleven: It's the same text we have on ASB doco explaininig what SB is. I don't read guaranteed ordering anywhere. it says guaranteed delivery, which is very important to the nsb message

SeanFeldman: @yvesgoeleven I don't mind that, but FIFO reference has to go 🙂

transports/azure-service-bus-netstandard/index.md
```diff
@@ -0,0 +1,21 @@
...
+title: Azure Service Bus .Net Standard Transport
+component: ASBS
+tags:
+ - Azure
+reviewed: 2018-06-19
+---
+
+## Azure Service Bus .Net Standard Transport
+
+The Azure Service Bus .Net Standard transport, ASBS in short, is the replacement of to the [Azure Service Bus transport](transports/azure-service-bus/) which leverages the .Net Standard compatible [Microsoft.Azure.ServiceBus](https://www.nuget.org/packages/Microsoft.Azure.ServiceBus/) client SDK.
```
SeanFeldman: > The Azure Service Bus .Net Standard transport, ASBS in short, is the replacement

Please remove. We should simply state that ASBS is an Azure Service Bus transport that can run on .NET Core and full framework. Not a replacement. Not a successor. 

If I'm a ASB customer, reading successor or replacement would indicate that I'll get features from ASB in ASBS.

transports/azure-service-bus-netstandard/index.md
```diff
@@ -0,0 +1,21 @@
...
+
+[Azure Service Bus (ASB)](https://azure.microsoft.com/en-us/services/service-bus/) is a messaging service hosted on the Azure platform, that allows for exchanging messages between various applications in a loosely coupled fashion. ASB Queues offer <a href="https://en.wikipedia.org/wiki/FIFO_(computing_and_electronics)">"First In, First Out" (FIFO)</a> guaranteed message delivery, and support a range of standard protocols (REST, AMQP, WS*) and APIs (to put messages on and pull messages off the queue). ASB Topics deliver messages to multiple subscribers and facilitate use of the fan-out pattern to deliver messages to downstream systems.
```
SeanFeldman: FIFO shouldn't be mentioned IMO.

transports/azure-service-bus-netstandard/operational-scripting.md
```diff
@@ -0,0 +1,69 @@
...
+summary: Explains how to create queues and topics using scripting
+component: ASBS
+tags:
+ - Azure
+reviewed: 2018-06-20
+---
+
+## Operational Scripting
+
+In order to provision or deprovision the resources required by an endpoint, the `asb-transport` command line (cli) tool can be used.
```
SeanFeldman: > (cli)

needs to be capitalized

transports/azure-service-bus-netstandard/operational-scripting.md
```diff
@@ -0,0 +1,69 @@
+---
+title: Operational scripting
+summary: Explains how to create queues and topics using scripting
+component: ASBS
+tags:
+ - Azure
+reviewed: 2018-06-20
+---
+
+## Operational Scripting
```
SeanFeldman: We can now add information on how to get the tool

``` c#
dotnet tool install -g NServiceBus.Transport.AzureServiceBus.CommandLine
```

PR #3872, opened 2018-06-20
Add a prerequisite step for ServiceControl

WilliamBZA:
This pre-requisite ensures that the internal data migrations will already be run and completed before the major upgrade is run

### DIFF
--- a/servicecontrol/upgrades/1to2.md
+++ b/servicecontrol/upgrades/1to2.md
@@ -11,7 +11,7 @@ Upgrading ServiceControl from version 1 to version 2 is a major upgrade and requ
 
 ## Pre-requisites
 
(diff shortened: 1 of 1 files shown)


servicecontrol/upgrades/1to2.md
```diff
@@ -9,6 +9,10 @@ reviewed: 2018-06-01
 
 Upgrading ServiceControl from version 1 to version 2 is a major upgrade and requires careful planning. During the upgrade process, the instance of ServiceControl that is being upgraded will no longer be available and will not be ingesting any messages. The upgrade process is not reversible and cannot be reversed once started.
 
+## Pre-requisites
```
mikeminutillo: Can/should this link directly to the 1.41.3 installer (or at the very least the last 1.x installer)? If customers are using the Platform Installer, they don't really have the opportunity to select a version.

WilliamBZA: Added a link to the release

PR #3882, opened 2018-06-21
Added nuget alias for NServiceBus.Metrics.ServiceControl.Msmq

ramonsmits:
Pointing NServiceBus.Metrics.ServiceControl.Msmq to MetricsServiceControl, not sure if that is allowed as NServiceBus.Metrics.ServiceControl already does that.


DavidBoike:
@ramonsmits The bombed out content verification test pretty much shows that's not possible.

That alias file is for mapping an alias (left hand side) to a NuGet (right hand side) so that when we see `MetricsServiceControl_1` as a versioned snippets/sample path, we know what that relates to. I don't think it's for mapping in the opposite direction. (This is my naive understanding anyway, without digging really deep in DocsEngine code.)

What problem are you actually trying to solve?

Should the additional NuGet actually be defined here?
https://github.com/Particular/docs.particular.net/blob/master/components/components.yaml#L428-L436

ramonsmits:
I want to have https://docs.particular.net/nuget/nservicebus.metrics.servicecontrol.msmq not result in a 404

components/nugetAlias.txt
```diff
@@ -48,6 +48,7 @@ MessagePack: NServiceBus.MessagePack
 MsgPack: NServiceBus.MsgPack
 Metrics: NServiceBus.Metrics
 MetricsServiceControl: NServiceBus.Metrics.ServiceControl
+MetricsServiceControl: NServiceBus.Metrics.ServiceControl.Msmq
```
weralabaj: should be MetricsServiceControlMsmq

PR #3888, opened 2018-06-22
add RabbitMQ delayed delivery sample

adamralph:


DavidBoike:
@tmasternak Does the Rabbit delays implementation surface anything you can see from the NServiceBus APIs or is it all hidden in the Rabbit internals? It would be cool to have a loop where you could enter different delays (capped at 60s) and have the handler pump out what the routing key bitwise stuff was.

tmasternak:
@DavidBoike It's all hidden I'm afraid :/. 

The routing key that hold binary encoding of the delay value is not even in the headers (see [this](https://github.com/Particular/NServiceBus.RabbitMQ/blob/1f018c1a9a178eec306fd13216130d058aa3f7bb/src/NServiceBus.Transport.RabbitMQ/Connection/ConfirmsAwareChannel.cs#L56)) and the custom `DelayedDelivery` header is [removed](https://github.com/Particular/NServiceBus.RabbitMQ/blob/1f018c1a9a178eec306fd13216130d058aa3f7bb/src/NServiceBus.Transport.RabbitMQ/Receiving/MessageConverter.cs#L41) for the message on receive. 

That being said we could achieve something similar using https://github.com/rabbitmq/rabbitmq-tracer. What we could do is to add tracer to the sample and make the sample endpoint connect to it instead of the broker. The tracer works as a man in the middle and outputs raw amqp dumps for each message. In the console of the tracer users would be able to see the routing key. Here is a sample screen from my local setup for message delayed by 2 seconds.

image

DavidBoike:
That sounds awesome @tmasternak - how would you recommend adding the tracer to the sample?

Looping in @bording, take a look at @tmasternak's suggestion about rabbitmq-tracer above.

bording:
I'm definitely :-1: on trying to use something like rabbitmq-tracer. It generates far too much noise, and if I understand what correctly what it's doing, you're still not going to see any of the inner workings of the message as it falls through the delay infrastructure. That's all internal to the broker, so it won't be generating additional messages for the tracer to display.

Looking at this sample, I'm questioning the value of having it at all. The whole point of the native delay support is that there's nothing a user has to do, it just works. There's nothing to demonstrate here.

DavidBoike:
@bording That's fair. It would be really difficult to get them to set up the tracer and then to help them make sense out of what they were seeing. Given that let's close this PR.

Take a look at https://github.com/Particular/docs.particular.net/issues/3929 and see if maybe that helps to better illustrate how everything works so we don't need a sample.

PR #3894, opened 2018-06-24
Remove saga from mailer

udidahan:
I can't think of a good reason to have a saga directly send the email. In every circumstance I think it would be best to have the saga *send a message* to some handler which does the email sending.


SimonCropp:
@udidahan it doesnt directly send a mail https://github.com/HEskandari/NServiceBus.Mailer/blob/master/NServiceBus.Mailer/MailSender.cs#L20

udidahan:
Even so, I wouldn't want to guide people to having those kinds of dependencies inside sagas.

PR #3904, opened 2018-06-29
Scripts should have been updated with SqlP 4.1 release

DavidBoike:


DavidBoike:
@Particular/sqlpersistence-maintainers This should have been updated as part of the SqlPersistence 4.1 release so I'm doing it now without review as it can't really break anything, but please still give it a review - it should mirror the script output changes from https://github.com/Particular/NServiceBus.Persistence.Sql/pull/294 with the additions of updlock hints, merge statement edits on subscribe. My only real concern was that perhaps we should provide something for the SagaSelect where clause other than `where 1 = 1`.

PR #3908, opened 2018-07-05
Fix pipeline registrations

### LINKED ISSUE #3871: Feedback: 'Message Throughput Throttling'
Feedback for 'Message Throughput Throttling' https://docs.particular.net/samples/throttling/

Location in GitHub: https://github.com/Particular/docs.particular.net/blob/master/samples/throttling/sample.md

This sample does not have a need in registering custom behaviour with a specific ordering within a stage in the pipeline. Therefore, the sample should prefer registration using [`endpointConfiguration.Pipeline.Register()`](https://docs.particular.net/nservicebus/pipeline/manipulate-with-behaviors#add-a-new-step) that allows to register an instance rather than `RegisterStep`.

In addition to that, the suggested overload allows to pass in arguments into behaviour constructor.

andreasohlund:
To use the simplified syntax. All touched samples have been smoke tested

Closes #3871


timbussmann:
any reason you removed the learning transport config?

SeanFeldman:
@andreasohlund it would be ideal if the sample code would also demonstrate how to pass dependencies such as configurations. From the original issue:

> In addition to that, the suggested overload allows to pass in arguments into behaviour constructor.

Do you think that's possible? Thanks.

andreasohlund:
>any reason you removed the learning transport config?

@timbussmann Only removed the persistence config since we don't need that

andreasohlund:
> it would be ideal if the sample code would also demonstrate how to pass dependencies such as configurations

@SeanFeldman the throttling behavior in the sample doesn't have any dependencies?

We could add this to:

https://docs.particular.net/nservicebus/pipeline/manipulate-with-behaviors#add-a-new-step

is that what you had in mind?

timbussmann:
> Only removed the persistence config since we don't need that

that's what I was assuming. I'm not entirely sure whether this change improves the idea of the samples though. After all, most other transports/configurations will need some persistence configured atm, so having a placeholder ready (although not used) might give a better understanding of the API and required configurations. No strong opinion on this though, it's probably more of a philosophical question ;)

andreasohlund:
I’d say with the trend being that we move towards native timeouts and
native pubsub for all the transport, persistence is more and more becoming
a outbox and saga thing?

andreasohlund:
>Yes @andreasohlund

Will do that in a separate PR to not hold this one up

PR #3911, opened 2018-07-09
Remove ravendb mention

danielmarbach:


danielmarbach:
I would suggest to remove the mentioning of Ravendb. It is irrelevant. We could even replace the word database with storage

PR #3915, opened 2018-07-11
Multi-instance mode in menu

danielmarbach:
For a period of time we decided not to have it in the menu because it is an advanced scenario that not everybody should be using. Still for customers with larger loads, it would make sense that the doco is more discoverable. So add it to the menu?


mikeminutillo:
Now that we have some customers successfully running on it, :+1:

PR #3929, opened 2018-07-18
More explanation around RabbitMQ delayed messages

DavidBoike:
@Particular/rabbitmq-transport-maintainers I was trying to better understand the mechanics behind our RabbitMQ delayed messages, and found the documentation lacking.

This is what I came up with. So:

1. Are my additions accurate?
2. Do my additions help to create understanding?
3. If Yes and Yes, are there any parts of the surrounding text that should be removed as duplicative?

### DIFF
--- a/transports/rabbitmq/delayed-delivery.md
+++ b/transports/rabbitmq/delayed-delivery.md
@@ -110,9 +110,9 @@ Every endpoint that can receive delayed messages will create bindings like `#.En
 
 ### Example
 
-This example illustrates the delay mentioned above, showing a message sent with a delay of 10 seconds to an endpoint called `destination`. For brevity, the number of delay levels has been reduced to 4, so the routing key for this message is `1.0.1.0.destination`.
+This example illustrates the delay mentioned above, showing a message sent with a delay of 10 seconds to an endpoint called `destination`. For brevity, the number of delay levels has been reduced to 4, so the routing key for this message is `1.0.1.0.destination` and it is published first to the Level 3 exchange:
 
-|Exchange|Routing Key Focus|Routing Value|Routed To|Delay|Dead-letter To|
+|Exchange         |Routing Key Segment    |Routing Value|Routed To        |Delay in Queue|Dead-letter To|
 |-----------------|-----------------------|:-----------:|-----------------|:-------:|:---------------:|
 |Level 03         |**1**.0.1.0.destination|     `1`     |Level 03 Queue   |8 seconds|Level 02 Exchange|
 |Level 02         |1.**0**.1.0.destination|     `0`     |Level 01 Exchange|    -    |        -        |
--- a/transports/rabbitmq/delayed-delivery.md
+++ b/transports/rabbitmq/delayed-delivery.md
@@ -95,7 +95,7 @@ class hiddenExchanges hiddenExchanges
 end
 ```
 
-To avoid unnecessary traversal through the delay infrastructure, the message is published to the first applicable exchange by identifying the first level that will route to a queue, or in other words, the first `1` in the routing key. In the 10-second example (binary `1010`), the message would first be published to the `nsb.delay-level-03` exchange, and all of that exchange's bindings are evaluated. Only two bindings exist for the exchange, so either:
+To avoid unnecessary traversal through the delay infrastructure, the message is published to the first applicable exchange by identifying the first level that will route to a queue, or in other words, the first `1` in the routing key. In the 10-second example (binary `1010`), the message would first be published to the `nsb.delay-level-03` exchange, and all of that exchange's bindings are evaluated. Only two bindings exist for the exchange, so:
 
 * If the value is a `1`, the exchange will route the message to the `nsb.delay-level-03` queue, where it will wait until the TTL expires (2^3 seconds) before being forwarded to the `nsb-delay-level-02` exchange.
 * If the value is a `0`, the exchange will route the message to the `nsb.delay-level-02` exchange.
@@ -103,14 +103,14 @@ To avoid unnecessary traversal through the delay infrastructure, the message is 
 
 ### Delivery
 
-At the end of this process, the message is delivered to the `nsb.delay-delivery` exchange. It is at this final step where the message destination is evaluated to determine where the message should be routed to.
+At the end of this process, the message is routed to the `nsb.delay-delivery` exchange. It is at this final step where the message destination is evaluated to determine where the message should be routed to.
 
 Every endpoint that can receive delayed messages will create bindings like `#.EndpointName` to this exchange to control final routing. The exact process depends upon the [routing topology](routing-topology.md) in use. These bindings match any combination of delay values, but only the binding for the correct destination endpoint will match, resulting in the message being delivered only to the correct endpoint.
 
 
 ### Example
 
-This example illustrates the delay mentioned above, showing a message sent with a delay of 10 seconds to an endpoint called `destination`. For brevity, the number of delay levels has been reduced to 4, so the routing key for this message is `1.0.1.0.destination` and it is published first to the Level 3 exchange:
+This example illustrates the delay mentioned above, showing a message sent with a delay of 10 seconds to an endpoint called `destination`. For brevity, the number of delay levels has been reduced to 4, so the routing key for this message is `1.0.1.0.destination`, and it is published to the Level 3 exchange:
 
 |Exchange         |Routing Key Segment    |Routing Value|Routed To        |Delay in Queue|Dead-letter To|
 |-----------------|-----------------------|:-----------:|-----------------|:-------:|:---------------:|
--- a/transports/rabbitmq/delayed-delivery.md
+++ b/transports/rabbitmq/delayed-delivery.md
@@ -125,7 +125,7 @@ This example can be visualized using the following
 ```mermaid
 graph LR
 
-subgraph Example: delay of 5 seconds
+subgraph Example: delay of 10 seconds
 
 exchange3(Level 3)
 exchange2(Level 2)
--- a/transports/rabbitmq/delayed-delivery.md
+++ b/transports/rabbitmq/delayed-delivery.md
@@ -56,32 +56,34 @@ end
 
 The delay levels are connected in this manner, from highest (27) to lowest (0). Each delay level's routing key's add wildcards as needed so that they are looking at the portion of the message's routing key that corresponds to its delay level.
 
-The following table illustrates dead letter exchange and bindings for the delay queues. It's important to note the wildcard rules for RabbitMQ binding expressions:
+The following table illustrates dead letter exchange and bindings for the delay queues. It's important to note the wildcard rules for RabbitMQ binding expressions, where *word* describes a dot-delimited segment:
 
-* `*` substitutes for exactly one segment
-* `#` substitutes for 0 or more segments.
+* `*` substitutes for exactly one word
+* `#` substitutes for 0 or more words
 
-| Exchange/Queue     | x-dead-letter-exchange | Binding                                                     |
-|--------------------|------------------------|-------------------------------------------------------------|
-| nsb.delay-level-27 | nsb.delay-level-26     | `1.#`                                                       |
-| nsb.delay-level-26 | nsb.delay-level-25     | `*.1.#`                                                     |
-| nsb.delay-level-25 | nsb.delay-level-24     | `*.*.1.#`                                                   |
-|         ...        |           ...          | ...                                                         |
-| nsb.delay-level-03 | nsb.delay-level-02     | `*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.1.#`       |
-| nsb.delay-level-02 | nsb.delay-level-01     | `*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.1.#`     |
-| nsb.delay-level-01 | nsb.delay-level-00     | `*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.1.#`   |
-| nsb.delay-level-00 | nsb.delay-delivery     | `*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.1.#` |
+| Exchange & Queue Name | x-dead-letter-exchange | Queue Binding                                               |
+|-----------------------|------------------------|-------------------------------------------------------------|
+| nsb.delay-level-27    | nsb.delay-level-26     | `1.#`                                                       |
+| nsb.delay-level-26    | nsb.delay-level-25     | `*.1.#`                                                     |
+| nsb.delay-level-25    | nsb.delay-level-24     | `*.*.1.#`                                                   |
+| ...                   | ...                    | ...                                                         |
+| nsb.delay-level-03    | nsb.delay-level-02     | `*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.1.#`       |
+| nsb.delay-level-02    | nsb.delay-level-01     | `*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.1.#`     |
+| nsb.delay-level-01    | nsb.delay-level-00     | `*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.1.#`   |
+| nsb.delay-level-00    | nsb.delay-delivery     | `*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.1.#` |
 
-With this system of queues, the message is first delivered to `nsb.delay-level-27` where the routing key will only match the binding if it begins with `1.`. In that case the queue will "accept" the message and wait for something to process it up until the TTL (2^27 seconds) before forwarding to `nsb.delay-level-27`. If the binding does not match the message will be rejected from `nsb.delay-level-27` and forwarded to the dead letter exchange immediately.
+With this system of exchanges and queues, the message is first delivered to the first applicable queue based on the binary value. In the 10-second example (binary `1010`) the message would first be delivered to the `nsb.delay-level-03` exchange, where the matching queue's binding will only match if a `1` exists in the 4th-to-last digit. In that case the queue will "accept" the message into the `nsb.delay-level-03` queue and wait for something to process it up until the TTL (2^3 seconds) before forwarding to the `nsb.delay-level-02` exchange.
 
-In the second delay queue (`nsb.delay-level-26`) the binding will match any value for the first segment (either `0` or `1`) but only match if the second segment is a `1`. In this way, the process continues all the way through `nsb.delay-level-00`.
+When the bindings for exchanges and queues don't match up, i.e. when the bit value is zero, wildcard bindings with 0 values instead of 1 ensure that messages that shouldn't be delayed for that delay level are immediately forwarded to the next delay level. In other words, for each exchange, the bit value in a message's routing key will either match `1` and be forwarded to the queue to be delayed by the TTL, or will match `0` and be forwarded to the next delay level.
+
+This process continues all the way through `nsb.delay-level-00`, where the x-dead-letter-exchange value (for matching routing keys) or the exchange-to-exchange binding (for non-matching routing keys) forward the message to the `nsb.delay-delivery` exchange.
 
 
 ### Delivery
 
 It is at the final step, in the `nsb.delay-delivery` exchange, where the real message destination is evaluated to determine which queue (based on endpoint name) the message should be delivered to.
 
-Each endpoint creates bindings in the `nsb.delay-delivery` exchange, similar to the following:
+Each endpoint creates a binding to the `nsb.delay-delivery` exchange, similar to the following:
 
 | Destination Queue | Routing Key        |
 |-------------------|--------------------|
--- a/transports/rabbitmq/delayed-delivery.md
+++ b/transports/rabbitmq/delayed-delivery.md
@@ -21,6 +21,12 @@ N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.destination
 
 Where `N` is either `0` or `1`, representing the delay value in binary, and `destination` is the name of endpoint the delayed message will be sent to.
 
+As an example, a delay of 10 seconds (`1010` in binary) on a message bound for the `destination` queue would be encoded with a routing key of:
+
+```
+0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.0.1.0.1.0.destination
+```
+
 
 ### Delay levels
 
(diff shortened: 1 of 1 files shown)


WojcikMike:
@DavidBoike I read your changes and I like them and I think they are accurate. With regards to 2nd point I would ask someone who don't know how delays work in Rabbit to read it, as that would be the best indicator.

For the point 3 I don't see duplication, but I will allow for other to chime in as well

transports/rabbitmq/delayed-delivery.md
```diff
@@ -50,10 +56,41 @@ end
...
+| nsb.delay-level-25 | nsb.delay-level-24     | `*.*.1.#`                                                   |
+|         ...        |           ...          | ...                                                         |
+| nsb.delay-level-03 | nsb.delay-level-02     | `*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.1.#`       |
+| nsb.delay-level-02 | nsb.delay-level-01     | `*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.1.#`     |
+| nsb.delay-level-01 | nsb.delay-level-00     | `*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.1.#`   |
+| nsb.delay-level-00 | nsb.delay-delivery     | `*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.1.#` |
+
+With this system of queues, the message is first delivered to `nsb.delay-level-27` where the routing key will only match the binding if it begins with `1.`. In that case the queue will "accept" the message and wait for something to process it up until the TTL (2^27 seconds) before forwarding to `nsb.delay-level-27`. If the binding does not match the message will be rejected from `nsb.delay-level-27` and forwarded to the dead letter exchange immediately.
```
bording: This explanation is inaccurate/incomplete. You've left out the 0 bindings.

transports/rabbitmq/delayed-delivery.md
```diff
@@ -50,10 +56,41 @@ end
...
+
+In the second delay queue (`nsb.delay-level-26`) the binding will match any value for the first segment (either `0` or `1`) but only match if the second segment is a `1`. In this way, the process continues all the way through `nsb.delay-level-00`.
+
 
 ### Delivery
 
-The final delay-level exchange is bound to the delivery exchange instead of another delay level. Every endpoint that can receive a delayed message will have a binding to this exchange with a routing key corresponding to the endpoint's name, so the message will be delivered to the endpoint's queue.
+It is at the final step, in the `nsb.delay-delivery` exchange, where the real message destination is evaluated to determine which queue (based on endpoint name) the message should be delivered to.
+
+Each endpoint creates bindings in the `nsb.delay-delivery` exchange, similar to the following:
```
bording: Better as "Each endpoint creates a binding to the `nsb.delay-delivery` exchange"

transports/rabbitmq/delayed-delivery.md
```diff
@@ -50,10 +56,41 @@ end
...
+
+In the second delay queue (`nsb.delay-level-26`) the binding will match any value for the first segment (either `0` or `1`) but only match if the second segment is a `1`. In this way, the process continues all the way through `nsb.delay-level-00`.
+
 
 ### Delivery
 
-The final delay-level exchange is bound to the delivery exchange instead of another delay level. Every endpoint that can receive a delayed message will have a binding to this exchange with a routing key corresponding to the endpoint's name, so the message will be delivered to the endpoint's queue.
+It is at the final step, in the `nsb.delay-delivery` exchange, where the real message destination is evaluated to determine which queue (based on endpoint name) the message should be delivered to.
```
bording: In the revision here, you've lost the fact that `nsb.delay-level-00` is bound to `nsb.delay-delivery`.

transports/rabbitmq/delayed-delivery.md
```diff
@@ -50,10 +56,41 @@ end
 
 The delay levels are connected in this manner, from highest (27) to lowest (0). Each delay level's routing key's add wildcards as needed so that they are looking at the portion of the message's routing key that corresponds to its delay level.
 
+The following table illustrates dead letter exchange and bindings for the delay queues. It's important to note the wildcard rules for RabbitMQ binding expressions:
+
+* `*` substitutes for exactly one segment
```
bording: "segment" is probably fine, but the RabbitMQ docs use the term "word" instead.

transports/rabbitmq/delayed-delivery.md
DavidBoike: Added in the section above, I think it belongs there.

DavidBoike:
@bording better?

transports/rabbitmq/delayed-delivery.md
```diff
@@ -50,10 +56,43 @@ end
...
+
+This process continues all the way through `nsb.delay-level-00`, where the x-dead-letter-exchange value (for matching routing keys) or the exchange-to-exchange binding (for non-matching routing keys) forward the message to the `nsb.delay-delivery` exchange.
+
 
 ### Delivery
 
-The final delay-level exchange is bound to the delivery exchange instead of another delay level. Every endpoint that can receive a delayed message will have a binding to this exchange with a routing key corresponding to the endpoint's name, so the message will be delivered to the endpoint's queue.
+It is at the final step, in the `nsb.delay-delivery` exchange, where the real message destination is evaluated to determine which queue (based on endpoint name) the message should be delivered to.
+
+Each endpoint creates a binding to the `nsb.delay-delivery` exchange, similar to the following:
```
bording: Getting into this level of detail, you run into the fact that a portion of the delayed delivery implementation is topology-specific. The binding created to the `nsb.delay-delivery` exchange is controlled by the routing topology. For the conventional topology, this is an exchange-to-exchange binding between `nsb.delay-delivery` and the endpoint's exchange, but for the direct routing topology, this binding is made between `nsb.delay-delivery` and the actual queue. And of course, a custom routing topology is free to implement this however it needs to.

The routing topology has to implement the following method: 
`public void BindToDelayInfrastructure(IModel channel, string address, string deliveryExchange, string routingKey)`

so it does get the delivery exchange and routing key value, but is free to use those in whatever way is needed.

Anyway, that's a long of way of saying that "destination queue" in this section isn't entirely correct, but I'm not sure how correct you actually want to be here.

transports/rabbitmq/delayed-delivery.md
```diff
@@ -50,10 +56,43 @@ end
...
+
+| Destination Queue | Routing Key        |
+|-------------------|--------------------|
+| Sales             | `#.Sales`          |
+| Billing           | `#.Billing`        |
+| Shipping          | `#.Shipping`       |
+| Other.Endpoint    | `#.Other.Endpoint` |
+
+These bindings match any combination of delay values, but only the binding for the correct destination queue will match, resulting in the message being delivered to only the correct destination queue.
```
bording: The "These bindings match any combination of delay values" section is a bit confusing. Perhaps being a bit more direct about the fact you're referring to the "binary" portion of a message's routing key not being relevant, and only the destination?

transports/rabbitmq/delayed-delivery.md
```diff
@@ -50,10 +56,43 @@ end
...
+| ...                   | ...                    | ...                                                         |
+| nsb.delay-level-03    | nsb.delay-level-02     | `*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.1.#`       |
+| nsb.delay-level-02    | nsb.delay-level-01     | `*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.1.#`     |
+| nsb.delay-level-01    | nsb.delay-level-00     | `*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.1.#`   |
+| nsb.delay-level-00    | nsb.delay-delivery     | `*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.*.1.#` |
+
+With this system of exchanges and queues, the message is first delivered to the first applicable queue based on the binary value. In the 10-second example (binary `1010`) the message would first be delivered to the `nsb.delay-level-03` exchange, where the matching queue's binding will only match if a `1` exists in the 4th-to-last digit. In that case the queue will "accept" the message into the `nsb.delay-level-03` queue and wait for something to process it up until the TTL (2^3 seconds) before forwarding to the `nsb.delay-level-02` exchange.
```
bording: Probably should drop the first "first": "the message is delivered to the first applicable"

bording: Also, that should be "first applicable exchange" not queue. Messages are never sent to queues!

bording: "In that case the queue will "accept" the message into the `nsb.delay-level-03`"

Messages are published to an exchange, and the all of the bindings to the exchange are evaluated, and all ones that are valid get a copy of the message sent to the destination of the binding. 

So, try to rewrite this from the perspective of the exchange, not the queue. Message is published to `nsb.delay-level-03`, which has two bindings, the 0 binding and 1 binding. If the 1 binding is a match, then the message is delivered to the queue. If the 0 binding matches, the message gets delivered to the `nsb.delay-level-02` exchange.

bording: "before forwarding to the `nsb.delay-level-02` exchange."

This should point out that the TTL expiring causes the message to dead-lettered, so it is republished to the configured exchange, `nsb.delay-level-02` in this case.

transports/rabbitmq/delayed-delivery.md
```diff
@@ -50,10 +56,43 @@ end
 
 The delay levels are connected in this manner, from highest (27) to lowest (0). Each delay level's routing key's add wildcards as needed so that they are looking at the portion of the message's routing key that corresponds to its delay level.
 
+The following table illustrates dead letter exchange and bindings for the delay queues. It's important to note the wildcard rules for RabbitMQ binding expressions, where *word* describes a dot-delimited segment:
+
+* `*` substitutes for exactly one word
+* `#` substitutes for 0 or more words
+
+| Exchange & Queue Name | x-dead-letter-exchange | Queue Binding                                               |
```
bording: This table still doesn't make it clear that there are both 0 and 1 bindings on the exchange.

transports/rabbitmq/delayed-delivery.md
```diff
@@ -21,58 +21,111 @@ N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.destination
...
+```
+
 
 ### Delay levels
 
 Each exchange/queue pair that makes up a level represents one bit of the total delay value. By having 28 of these levels, corresponding to `2^27` through `2^0`, the delay infrastructure can delay a message for any value that can be represented by a 28-bit number. With 28 total levels, the maximum delay value is 268,435,455 seconds, or about 8.5 years.
 
 A delay level is created by declaring a topic exchange that is bound to a queue with a routing key of `1`, and to the exchange corresponding to `level - 1` with a routing key of `0`. The queue for the delay level is declared with an [`x-message-ttl`](https://www.rabbitmq.com/ttl.html) value corresponding to `2^level` seconds. The queue is also declared with an [`x-dead-letter-exchange`](https://www.rabbitmq.com/dlx.html) value corresponding to the `level - 1` exchange, so that when a message in the queue expires, it will be routed to the `level - 1` exchange.
 
+The delay levels are connected in this manner, from highest (27) to lowest (0). Each delay level's routing key's add wildcards as needed so that they are looking at the portion of the message's routing key that corresponds to its delay level.
```
bording: key's -> key

transports/rabbitmq/delayed-delivery.md
```diff
@@ -21,58 +21,111 @@ N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.destination
...
+
+The following diagram illustrates the relationships between exchanges and queues, which are connected by bindings and `x-message-ttl` delay values. It's important to note the wildcard rules for RabbitMQ binding expressions, where *word* describes a dot-delimited segment:
+
+* `*` substitutes for exactly one word
+* `#` substitutes for 0 or more words
+
+<style>
```
bording: What's this for?

transports/rabbitmq/delayed-delivery.md
bording: Seems like there is some pronoun confusion in the second sentence, switching between "they" and "its" later in the sentence.

transports/rabbitmq/delayed-delivery.md
```diff
@@ -21,58 +21,111 @@ N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.destination
...
+    font-family: Consolas, Courier New, monospace;
+}
+</style>
+
 ```mermaid
 
 graph TD
 
-subgraph Delay Level
+subgraph Delay Level Exchanges, Queues, Bindings, and TTLs
```
bording: Should be Delay-Level since it's a compound adjective here, right?

transports/rabbitmq/delayed-delivery.md
```diff
@@ -21,58 +21,111 @@ N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.destination
...
 
-classDef exchangeClass stroke:#000000,stroke-width:2px;
-class exchangeN,exchangeN-1 exchangeClass
+exchange1(Exchange: nsb.delay-level-01)
+queue1(Queue: nsb.delay-level-01)
+exchange0(Exchange: nsb.delay-level-00)
+queue0(Queue: nsb.delay-level-00)
+delivery(Exchange: nsb.delay-delivery)
+
+exchange27 -->|"Binding: 1.#<br/>(No * wildcards)"| queue27
```
bording: Is the listing of how many wildcards there are really needed? Seems like it's just noise and taking up space.

transports/rabbitmq/delayed-delivery.md
```diff
@@ -21,58 +21,111 @@ N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.destination
...
 
 A delay level is created by declaring a topic exchange that is bound to a queue with a routing key of `1`, and to the exchange corresponding to `level - 1` with a routing key of `0`. The queue for the delay level is declared with an [`x-message-ttl`](https://www.rabbitmq.com/ttl.html) value corresponding to `2^level` seconds. The queue is also declared with an [`x-dead-letter-exchange`](https://www.rabbitmq.com/dlx.html) value corresponding to the `level - 1` exchange, so that when a message in the queue expires, it will be routed to the `level - 1` exchange.
 
+The delay levels are connected in this manner, from highest (27) to lowest (0). Each delay level's routing key's add wildcards as needed so that they are looking at the portion of the message's routing key that corresponds to its delay level.
+
+The following diagram illustrates the relationships between exchanges and queues, which are connected by bindings and `x-message-ttl` delay values. It's important to note the wildcard rules for RabbitMQ binding expressions, where *word* describes a dot-delimited segment:
```
bording: "relationships between **the** exchanges and queues"

Perhaps even "relationships between the exchanges and queues in the delay infrastructure" ?

bording: Would be more correct to say they are connected by bindings and dead letter exchange settings. (x-dead-letter-exchange)

transports/rabbitmq/delayed-delivery.md
```diff
@@ -21,58 +21,111 @@ N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.destination
...
+class exchange27,exchange26,exchange25,exchange1,exchange0,delivery exchangeClass
+
+classDef hiddenExchanges font-style:italic,fill:transparent,stroke:transparent;
+class hiddenExchanges hiddenExchanges
 
 end
 ```
 
-The delay levels are connected in this manner, from highest (27) to lowest (0). Each delay level's routing key's add wildcards as needed so that they are looking at the portion of the message's routing key that corresponds to its delay level.
+With this system of exchanges and queues, the message is first delivered to the first applicable exchange based on the binary value. In the 10-second example (binary `1010`) the message would first be delivered to the `nsb.delay-level-03` exchange, and all of that exchange's bindings are evaluated. All matching bindings will receive a copy of the message, but the bindings are arranged so that there are only two possibilities based on the value of the 4th-to-last digit:
```
bording: Instead of "With this system of exchanges and queues" 

What about, "To avoid unnecessary traversal through the delay infrastructure"

bording: Too many "first" in the sentence. Probably keep the second one and drop the first.

bording: "based on the binary value"  

Would it make more sense to say something like "based on the first 1 in the routing key", or "based on identifying the first level that will route to a queue"?

bording: Needs a comma after "(binary `1010`)"

bording: " All matching bindings will receive a copy of the message, but the bindings are arranged so that there are only two possibilities based on the value of the 4th-to-last digit:"

Not sure what the point of this is. There are only going to be two bindings per exchange, so maybe just drop this and replace with something that leads to the two options below?

transports/rabbitmq/delayed-delivery.md
```diff
@@ -21,58 +21,111 @@ N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.destination
...
+
+* If the value is a `1` the exchange will deliver the message to the `nsb.delay-level-03` queue and wait for something to process it up until the TTL (2^3 seconds) before forwarding to the `nsb-delay-level-02` exchange.
```
bording: Needs comma after "`1`"

bording: deliver -> route

bording: "wait for something to process it" We don't actually expect this to happen, so I'd drop this and just say

"will deliver the message to the `nsb.delay-level-03` queue, where it will wait until the TTL expires (2^3 seconds) before being forwarded..."

transports/rabbitmq/delayed-delivery.md
```diff
@@ -21,58 +21,111 @@ N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.destination
...
+* If the value is a `0` the exchange will immediately deliver the message immediate to the `nsb.delay-level-02` exchange.
```
bording: Better as:

If the value is a `0`, the exchange will route the message to the `nsb.delay-level-02` exchange.

transports/rabbitmq/delayed-delivery.md
bording: Putting all the suggestions together:

If the value is a `1`, the exchange will route the message to the `nsb.delay-level-03` queue, where it will wait until the TTL expires (2^3 seconds) before being forwarded to the `nsb-delay-level-02` exchange.

transports/rabbitmq/delayed-delivery.md
```diff
@@ -21,58 +21,111 @@ N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.destination
...
 
 
 ### Delivery
 
-The final delay-level exchange is bound to the delivery exchange instead of another delay level. Every endpoint that can receive a delayed message will have a binding to this exchange with a routing key corresponding to the endpoint's name, so the message will be delivered to the endpoint's queue.
+At the end of this conditional delay process, the message is delivered to the `nsb.delay-delivery` exchange. It is at this final step where the real message destination is evaluated to determine where the message should be delivered to.
```
bording: Maybe drop "conditional delay"

bording: Maybe drop "real"

transports/rabbitmq/delayed-delivery.md
```diff
@@ -21,58 +21,111 @@ N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.destination
...
+
+Every endpoint that can receive delayed messages will create bindings like `#.EndpointName` to this exchange similar to `#.EndpointName`. The exact process depends upon the [routing topology](routing-topology.md) in use. These bindings match any combination of delay values, but only the binding for the correct destination endpoint will match, resulting in the message being delivered to only the correct endpoint.
```
bording: First sentence seems to have some weird duplication going on. In general the bindings are always going to be `#.EndpointName`, but the thing being bound to could change. For the conventional topology, it's an exchange->exchange binding, but for direct it's an exchange->queue binding.

transports/rabbitmq/delayed-delivery.md
bording: "delivered" maybe "routed" instead?

transports/rabbitmq/delayed-delivery.md
bording: The "delivered"s in this section would be more correct as "published"

transports/rabbitmq/delayed-delivery.md
DavidBoike: The style is to get the bindings to use a fixed-width font because otherwise the repeated star dot star dot star looks like garbage...

image

(Image because GitHub will make bold periods otherwise)

transports/rabbitmq/delayed-delivery.md
DavidBoike: Without it, not being in a table or right next to each other, makes it difficult to see that they're getting bigger.

transports/rabbitmq/delayed-delivery.md
```diff
@@ -21,58 +21,111 @@ N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.destination
...
+class exchange27,exchange26,exchange25,exchange1,exchange0,delivery exchangeClass
+
+classDef hiddenExchanges font-style:italic,fill:transparent,stroke:transparent;
+class hiddenExchanges hiddenExchanges
 
 end
 ```
 
-The delay levels are connected in this manner, from highest (27) to lowest (0). Each delay level's routing key's add wildcards as needed so that they are looking at the portion of the message's routing key that corresponds to its delay level.
+To avoid unnecessary traversal through the delay infrastructure, the message is published to the first applicable exchange by identifying the first level that will route to a queue, or in other words, the first `1` in the routing key. In the 10-second example (binary `1010`), the message would first be published to the `nsb.delay-level-03` exchange, and all of that exchange's bindings are evaluated. Only two bindings exist for the exchange, so either:
```
bording: I think we can drop "either" from the end of this. "Either if" doesn't read very well.

transports/rabbitmq/delayed-delivery.md
```diff
@@ -21,58 +21,111 @@ N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.destination
...
+At the end of this process, the message is delivered to the `nsb.delay-delivery` exchange. It is at this final step where the message destination is evaluated to determine where the message should be routed to.
+
+Every endpoint that can receive delayed messages will create bindings like `#.EndpointName` to this exchange to control final routing. The exact process depends upon the [routing topology](routing-topology.md) in use. These bindings match any combination of delay values, but only the binding for the correct destination endpoint will match, resulting in the message being delivered only to the correct endpoint.
 
 
 ### Example
 
-Using a simplified version of the delay infrastructure that has 4 delay levels (0-3), this is an example of sending a message with a delay of 5 seconds to an endpoint called `destination`:
+This example illustrates the delay mentioned above, showing a message sent with a delay of 10 seconds to an endpoint called `destination`. For brevity, the number of delay levels has been reduced to 4, so the routing key for this message is `1.0.1.0.destination` and it is published first to the Level 3 exchange:
```
bording: Needs a comma after "`1.0.1.0.destination`"

bording: Probably can drop "first" from the second sentence.

transports/rabbitmq/delayed-delivery.md
```diff
@@ -21,58 +21,111 @@ N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.N.destination
...
+
+* If the value is a `1`, the exchange will route the message to the `nsb.delay-level-03` queue, where it will wait until the TTL expires (2^3 seconds) before being forwarded to the `nsb-delay-level-02` exchange.
+* If the value is a `0`, the exchange will route the message to the `nsb.delay-level-02` exchange.
 
 
 ### Delivery
 
-The final delay-level exchange is bound to the delivery exchange instead of another delay level. Every endpoint that can receive a delayed message will have a binding to this exchange with a routing key corresponding to the endpoint's name, so the message will be delivered to the endpoint's queue.
+At the end of this process, the message is delivered to the `nsb.delay-delivery` exchange. It is at this final step where the message destination is evaluated to determine where the message should be routed to.
```
bording: Still think "routed" would be better than "delivered" here since we're still talking about actual RabbitMQ concepts.

transports/rabbitmq/delayed-delivery.md
bording: I guess it's ok to reference Consolas?

transports/rabbitmq/delayed-delivery.md
bording: I'll wait and see how it looks on the page.

PR #3953, opened 2018-08-06
Added doco for sc multi instance cluster communication

SzymonPobiega:


WilliamBZA:
Looks like this has too much indentation?

SzymonPobiega:
You're right @mikeminutillo . I forgot that the REST API is not public. Should not be part of the documentation then. Maybe we can just add an `md` file to the repo in a `doco` folder?

PR #3960, opened 2018-08-07
revert unintentional changes from PR 3946

adamralph:
relates to #3946


dbelcham:
@adamralph good catch and thanks for fixing it.

PR #3967, opened 2018-08-09
when using serilog should use ILogger for logging

SimonCropp:
related to https://github.com/NServiceBusExtensions/NServiceBus.Serilog/issues/18

### DIFF
--- a/samples/logging/serilog-custom/Serilog_2/Sample/MyHandler.cs
+++ b/samples/logging/serilog-custom/Serilog_2/Sample/MyHandler.cs
@@ -1,13 +1,13 @@
 ﻿using NServiceBus;
-using NServiceBus.Logging;
+using Serilog;
 
 public class MyHandler :
     IHandleMessages<MyMessage>
 {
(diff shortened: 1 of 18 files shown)


SzymonPobiega:
@SimonCropp if the user code is not using NSB logging abstraction but the Serilog directly, do you think there is still purpose for keeping this sample? It seems a bit redundant to me -- it's like documenting Serilog's APIs.

SimonCropp:
@SzymonPobiega it is not "documenting how to log in serilog". it is illustrating how to use NSB.Serilog with NSB. that includes the config code. it also gives people a significant leg up if someone wants to spike the use of Serilog with NSB, since they can download a runnable sample

SzymonPobiega:
The build failures are not relevant to this change so merging.

SimonCropp:
@SzymonPobiega u might want to review the other logging doco in the context of this 
https://github.com/NServiceBusExtensions/NServiceBus.Serilog/issues/18#issuecomment-412000989

PR #3979, opened 2018-08-17
SC 2 to 3 upgrade guide

danielmarbach:


servicecontrol/upgrades/2to3.md
```diff
@@ -0,0 +1,56 @@
+---
+title: Upgrade ServiceControl from version 2 to version 3
+summary: Instructions on how to upgrade ServiceControl from version 2 to 3
+isUpgradeGuide: true
+reviewed: 2018-08-17
+---
+
+## Overview
+
+Upgrading ServiceControl from version 2 to version 3 is a major upgrade and requires careful planning. During the upgrade process, the instance of ServiceControl that is being upgraded will no longer be available and will not be ingesting any messages. The upgrade process is not reversible and cannot be reversed once started.
```
WilliamBZA: This upgrade can be reversed. The users can uninstall v3 and drop back to a v2 instance. The irreversible part for the v1->v2 upgrade was because the database changed.

servicecontrol/upgrades/2to3.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Pre-requisites
+
+Before upgrading to ServiceControl version 3 the instance being upgraded must be upgraded to at least [version 2.1.3](https://github.com/Particular/ServiceControl/releases/tag/2.1.3).
+
+For more information how to upgrade from Version 1.x to 2.1.3 consult the [upgrade documentation](/servicecontrol/upgrades/1to2.md).
+
+### Install .NET 4.6.1
+
+The minimum .NET version for ServiceControl Version 3 is [.NET 4.6.1](https://www.microsoft.com/en-us/download/details.aspx?id=49982).
```
boblangley: Didn't we add 4.6.1 to the installer? If so, we should mention it will be installed if missing as part of setup.

danielmarbach: Yes we did add it to the installer and it will be installed.

danielmarbach: @boblangley I updated the description

PR #3980, opened 2018-08-17
add clarification to installers note

### LINKED ISSUE #3956: Feedback: 'Installers'
Feedback for 'Installers' https://docs.particular.net/nservicebus/operations/installers

Location in GitHub: https://github.com/Particular/docs.particular.net/blob/master/nservicebus/operations/installers.md

The sentence 

> Consider running the installers only during the initial setup using elevated permissions

surprised me as it is seldom the case that the application can change permission levels. Is the user advised to make platform dependent system service calls, or does NServiceBus offer a platform independent abstraction for that? Consequently, a reference to some text explaining how this is being done would be helpful.

adamralph:
Closes #3956


nservicebus/operations/installers.md
```diff
@@ -15,7 +15,7 @@ Installers ensure that endpoint-specific artifacts (e.g. queues, directories, da
 
 ## Running installers
 
-NOTE: Running installers requires the process to run with permissions to create/delete tables, queues, folders and other resources. It is recommended to apply the [principle of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege) and prevent running an endpoint process with too many permissions. Consider running the installers only during the initial setup, using elevated permissions.
+NOTE: Running installers requires the endpoint process to run with permissions to create/delete tables, queues, folders and other resources. It is recommended to apply the [principle of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege) and avoid running a process with redundant permissions. Consider running installers only during initial deployment, using elevated permissions. For example: start an endpoint process manually, as an administrator or elevated user, before installing it as a service. The alternative to using installers is to create the required resources before the endpoint is run. The method varies for each transport or persistence package. For more information, see [operations](/nservicebus/operations).
```
bording: Not sure "redundant" is correct here.

bording: "Consider running installers only during initial deployment, using elevated permissions"

Does this need a comma?

bording: "start an endpoint process manually, as an administrator or elevated user, before installing it as a service."

Don't think any of those commas are needed.

bording: "The method varies for each transport or persistence package."

Feels like it needs something like "The method to do this varies"

adamralph: The more I read it, the more I realise that I'm battling against the current structure. I'll attempt a re-write, to see if it comes out better.

nservicebus/operations/installers.md
```diff
@@ -11,11 +11,11 @@ redirects:
...
+NOTE: Installers require permission to administer resources such as database tables, queues, or directories. Following the [principle of least privilege](https://en.wikipedia.org/wiki/Principle_of_least_privilege) it is recommended to run an endpoint with these elevated permissions only during initial deployment. For example: an endpoint can be started manually as an administrator or elevated user to allow installers to run. The required resources will be ready as soon as the endpoint has successfully started. The endpoint can then be shut down and installed as a service. The service can be configured to run as a regular user which does not have permission to modify resources. The alternative to using installers is to create the required resources before the endpoint is run. The method of doing this varies for each transport or persistence package. For more information, see [operations](/nservicebus/operations).
```
bording: " Installers require permission"

Should "permission" be "permissions"?

bording: Needs a comma after "(https://en.wikipedia.org/wiki/Principle_of_least_privilege)"

bording: "For example:"

Should that just be a comma instead of a colon?

bording:
>I'm also wondering whether this is getting too big for a note, and needs to be a section of it's own. It could still have a shorter note, which alerts the reader, and links to the section.

I think moving this to it's own section makes sense.

nservicebus/operations/installers.md
bording: "The endpoint can then be shut down and installed as a service. The service can be configured to run as a regular user which does not have permission to modify resources."

Better as 

The endpoint can then be shut down and installed as a service, which can run as a regular user that does not have permission to modify resources."  ?

adamralph:
I've moved the installers to a new section titled "When to run installers".

PR #3985, opened 2018-08-21
Simplifying ASBS sample and removing NSB reference from shared project

SeanFeldman:


bording:
@SeanFeldman I have pushed the changes we discussed.

samples/azure-service-bus-netstandard/send-reply/ASBS_1/Endpoint2/Endpoint2.csproj
```diff
@@ -14,8 +15,7 @@
   </ItemGroup>
 
   <ItemGroup Label="Transitive dependencies">
-    <!-- TODO: uncomment when we can use an official package -->
```
SeanFeldman: Good catch!

PR #3987, opened 2018-08-22
Remove analyzers removal in Directory.Build.props

SeanFeldman:
When smoke-tested changes for Core v6.5, have used one of the samples.
Because samples remove all analyzers, Particular Analyzer was removed as well, causing confusion and unnecessary "bug" hunting with @timbussmann and @bording.

image

Unless there's a reason to do so, I recommend to remove this task from samples.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,4 +1,3 @@
-<?xml version="1.0" encoding="utf-8"?>
 <Project>
   <PropertyGroup>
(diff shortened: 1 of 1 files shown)


bording:
While you're making changes, can you remove the ` <?xml version="1.0" encoding="utf-8"?>` line from the file?

timbussmann:
ping @Particular/docs-maintainers

PR #3994, opened 2018-08-27
Use built-in DI container

twenzel:
Provide a sample using built-in DI container using the new NServiceBus.MSDependencyInjection package.


danielmarbach:
@twenzel sorry for letting you hang there. Thanks for providing a good headstart! Awesome!

I cherry picked your commit and created a new branch containing additional things that we need to do. I didn't want to bother you with the details of the docs site and making everything work. The PR is WIP since I need to fix a few things in the sample but once it is done would you mind reviewing it?

https://github.com/Particular/docs.particular.net/pull/3997

Thanks!

PR #3997, opened 2018-08-29
WIP MS DependencyInjection

danielmarbach:
Replaces #3994


danielmarbach:
Don't know why the content verification test is failing with

```
System.Exception : Path: 'C:\BuildAgent\work\542e5eb63d1ae8b\samples\netcore.md'. Error: 'Error:'Path does not exist'. Url:'/samples/dependency-injection/aspnetcore/?version=core_7'.'
   at AllPagesRenderer.Start() in C:\BuildAgent\work\3ba50a7be36aeb6b\src\ParticularDocsCore\Rendering\AllPagesRenderer.cs:line 43
```

although I changed the file in 

https://github.com/Particular/docs.particular.net/pull/3997/commits/4cb8590d77fd194cb06a6a02e408e4978200fed3

samples/dependency-injection/msdependencyinjection/sample.md
```diff
@@ -8,23 +8,28 @@ related:
...
+[ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/) comes with an integrated [Dependency Injection (DI) feature](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/dependency-injection). When hosting NServiceBus endpoints inside an Asp.NET Core app, components registered for DI might need to be shared between Asp.NET components and NServiceBus message handlers.
 
-It is possible to use an external container to satisfy both requirements.
+### Configuring an endpoint to use built-in DI
 
-Note: While this sample illustrates the scenario using [Autofac](/nservicebus/dependency-injection/autofac.md), the same can be achieved by using most other [DI libraries](/nservicebus/dependency-injection/). Using Autofac requires adding reference to the `Autofac.Extensions.DependencyInjection` package.
-
-
-### Configuring to use shared DI
+Note: Using this sample requires adding reference to the `NServiceBus.MSDependencyInjection` package.
```
DavidBoike: Might want to point out that it's a community package?

samples/dependency-injection/msdependencyinjection/sample.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Asp.NET Core Dependency Injection Integration
+title: ASP.NET Core Dependency Injection Integration
```
DavidBoike: If you're going to move the URL you need to provide a redirect.

DavidBoike:
I think the build is confused because your component is still Core, but you've changed the versioned snippets directory to `msdependencyinjection_1` which means there is no `version=core_7` anymore.

I think Core is correct, this isn't specifically about the msdependencyinjection package, it just uses that. So I'd rename the `msdependencyinjection_1` directory back to `Core_7` and see if that works.

samples/dependency-injection/msdependencyinjection/sample.md
danielmarbach: Hu? That is just a name. I did not move anything here :)

DavidBoike: I was just trying to get the comment close to the metadata where it needs to be fixed.

By renaming the directory you're moving the URL from `/samples/dependency-injection/aspnetcore/` to `/samples/depencency-injection/msdependencyinjection/` but you're not providing a redirect. The first URL will result in a 404.

Honestly though I don't think you should change the URL at all. Use of the msdependencyinjection package is just an implementation detail, in fact it's showing how to do it with that package OR with autofac. The sample is still showing DI in aspnetcore, is it not? That's what the title still says.

DavidBoike:
The integrity test is now failing because msdependencyinjection being prerelease doesn't matter. Core_7 is not prerelease.

DavidBoike:
@twenzel We follow semantic versioning individually on each of our components since NServiceBus 5. The fact that NServiceBus.Autofac (as a for instance) is also on version 7 is just an accident of having the same number of major versions. NServiceBus.Unity is actually on 9.0. I would not encourage you to artificially bump up the version of yours.

samples/dependency-injection/msdependencyinjection/sample.md
danielmarbach: Ah I missed the fact that we already had such an example. That was an oversight from me. That's why I got confused because I thought there is nothing to redirect to. My thinking was that most of the samples follow a component oriented approach that's why a renamed the folder (under the wrong impression it was newly created). But given we already had it with autofac I agree with you @davidboike that the msdependencyinjection part is just another flavour of it. Will change tomorrow. Thanks for showing me where my brain made weird twists

PR #3998, opened 2018-08-29
Extracting the custom instance mapping into a separate doc

andreasohlund:
We had a live chat user that incorrectly went straight to the advanced way of doing physical mapping so after talking to @timbussmann we decided to make the extensibility a bit less discoverable similar to have [the main routing page](https://docs.particular.net/nservicebus/messaging/routing) just have the extensibility as a "related" page. Also cleaned up the content to make sense.


transports/msmq/routing-extensibility_content_msmqtransport_[1,).partial.md
```diff
@@ -0,0 +1,11 @@
+## Custom instance mapping
```
andreasohlund: @timbussmann what was the better title that you came up with when we discussed this?

timbussmann: something like "instance mapping extensibility" maybe?

PR #4000, opened 2018-08-30
Note on obligatory ASB sanitization rule

tmasternak:


samples/servicecontrol/events-subscription/sample.md
```diff
@@ -20,6 +20,7 @@ This sample shows how to monitor heartbeat and failed message events in ServiceC
  1. Using [ServiceControl Management](/servicecontrol/license.md#servicecontrol-management-app) tool, set up ServiceControl to monitor endpoints using MSMQ transport.
  1. Ensure the `ServiceControl` process is running before running the sample. 
 
+NOTE: When using the [AzureServiceBus Transport](/transports/azure-service-bus/) it is necessary to configure [ValidateAndHashIfNeeded](/transports/azure-service-bus/sanitization.md#automated-sanitization) sanitization strategy for both endpoints in this project. This ensures consistent entity name shortening between the endpoints and ServiceControl.
```
danielmarbach: I think this is good but we might need to link to ASBS configuration page as well. Because in the ASBS world there is no such a strategy. You have to implement RuleNameShortner and SubscriptionShortener. Can you add that as well?

PR #4007, opened 2018-09-06
ASBS documentation review

mauroservienti:


samples/azure-service-bus-netstandard/send-reply/ASBS_1/Endpoint1/Program.cs
```diff
@@ -19,7 +19,7 @@ static async Task Main()
         var connectionString = Environment.GetEnvironmentVariable("AzureServiceBus_ConnectionString");
         if (string.IsNullOrWhiteSpace(connectionString))
         {
-            throw new Exception("Could not read the 'AzureServiceBus.ConnectionString' environment variable. Check the sample prerequisites.");
+            throw new Exception("Could not read the 'AzureServiceBus_ConnectionString' environment variable. Check the sample prerequisites.");
```
seanfarmar: Should this be fixed in the sample code as well?

samples/azure-service-bus-netstandard/send-reply/ASBS_1/Endpoint2/Program.cs
```diff
@@ -17,7 +17,7 @@ static async Task Main()
         var connectionString = Environment.GetEnvironmentVariable("AzureServiceBus_ConnectionString");
         if (string.IsNullOrWhiteSpace(connectionString))
         {
-            throw new Exception("Could not read the 'AzureServiceBus.ConnectionString' environment variable. Check the sample prerequisites.");
+            throw new Exception("Could not read the 'AzureServiceBus_ConnectionString' environment variable. Check the sample prerequisites.");
```
seanfarmar: Should this be fixed in the sample code as well?

samples/azure-service-bus-netstandard/native-integration/sample.md
```diff
@@ -48,7 +48,7 @@ snippet: NecessaryHeaders
 
 NOTE: The `NServiceBus.EnclosedMessageTypes` property must contain the the fully-qualified name of the type expected by the NServiceBus endpoint.
 
-The message itself is defined as an `IMessage` in the `Shared` project.
+The message itself is defined as message using [conventions](/nservicebus/messaging/conventions.md) in the `Receiver` project.
```
seanfarmar: Shall we add a snippet here with 
`endpointConfiguration.Conventions().DefiningMessagesAs(type => type.Name == "NativeMessage");`?

mauroservienti: good point, probably yes we should

samples/azure-service-bus-netstandard/send-reply/ASBS_1/Endpoint1/Program.cs
mauroservienti: the snippet already comes from the sample, so we're good.

samples/azure-service-bus-netstandard/send-reply/ASBS_1/Endpoint2/Program.cs
mauroservienti: the snippet already comes from the sample, so we're good.

samples/azure-service-bus-netstandard/native-integration/sample.md
SeanFeldman: I'm not convinced we should. The focus of sample is about ASBS, not conventions. If we want to _link_ to conventions, we could do so w/o adding extra code snippets and stealing the focus from the main objective. Which is ASBS transport.

For example:
_Note: the sample is using Message Conventions. More about this feature [here](https://docs.particular.net/nservicebus/messaging/conventions)_

mauroservienti:
Link to conventions is already in place. Removing the WIP

PR #4016, opened 2018-09-13
Document how to unit test logging using the TestingLoggerFactory

andreasohlund:


andreasohlund:
@SimonCropp thoughts?

Snippets/Testing/Testing_6/Logging/LoggingTest.cs
```diff
@@ -0,0 +1,28 @@
...
+using NUnit.Framework;
+
+[TestFixture]
+public class LoggingTests
+{
+    #region LoggerTesting
+    [Test]
+    public async Task ShouldLogCorrectly()
+    {
+        var logStatements = new StringBuilder();
```
SimonCropp: would it be better to keep a list of statements?

Snippets/Testing/Testing_6/Testing_6.csproj
```diff
@@ -7,5 +7,6 @@
     <PackageReference Include="NServiceBus.Gateway" Version="2.0.*" />
     <PackageReference Include="NServiceBus.Testing" Version="6.*" />
     <PackageReference Include="NUnit" Version="3.*" />
+    <PackageReference Include="NUnit3TestAdapter" Version="3.*" />
```
SimonCropp: seems like something that should be done on a diff pr

andreasohlund: Then I would have to install R# :) Trying to see if I could cope with out it :)

Snippets/Testing/Testing_6/Logging/LoggingTest.cs
andreasohlund: Not following, do you mean something likt this https://stackoverflow.com/questions/10303315/convert-stringwriter-to-string ?

SimonCropp: as in collect the log statements in a `list<string>`

andreasohlund: Couldn't find a way to get a string writer to populate a list of string? Or are U saying to change the `TestingLoggerFactory` to capture to a list instead of writing to a `TextWriter` ?

Snippets/Testing/Testing_6/Testing_6.csproj
```diff
@@ -7,5 +7,7 @@
...
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.*" />
```
bording: Can you add this to the top of the `ItemGroup` so the package names are sorted properly?

Snippets/Testing/Testing_7/Testing_7.csproj
```diff
@@ -7,5 +7,7 @@
     <PackageReference Include="NServiceBus.Gateway" Version="3.*" />
     <PackageReference Include="NServiceBus.Testing" Version="7.*" />
     <PackageReference Include="NUnit" Version="3.*" />
+    <PackageReference Include="NUnit3TestAdapter" Version="3.*" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.*" />
```
bording: Can you add this to the top of the `ItemGroup` so the package names are sorted properly?

andreasohlund:
Done

Snippets/Testing/Testing_6/Logging/LoggingTest.cs
SimonCropp: i think u would need to change `TestingLoggerFactory` to allow that.

andreasohlund: Raised an issue to track https://github.com/Particular/NServiceBus.Testing/issues/129 that potential change

PR #4027, opened 2018-09-27
Add highlights from champs mail

andreasohlund:


andreasohlund:
There are some parts in there about SqlT, we could move them to the SqlT upgrade guide but I think it make sense to keep them here since this is kind of the starting point for a v6 to v7 migration?

DavidBoike:
I agree. Let's dump all of it.

PR #4042, opened 2018-10-08
Correct gateway receive channel docs when wildcards are used

andreasohlund:
Need to have a non wildcard default for replies to be routed properly

### DIFF
--- a/Snippets/Gateway/Gateway_3/Channels/Wildcard.config.xml
+++ b/Snippets/Gateway/Gateway_3/Channels/Wildcard.config.xml
@@ -9,7 +9,7 @@
     <Channels>
       <Channel Address="http://+:25899/RemoteSite/"
                ChannelType="Http"/>               
-      <Channel Address="http://gateway.mycorp.com:25899/RemoteSite/"
+      <Channel Address="http://gateway.mycorp.com:25900/RemoteSite/"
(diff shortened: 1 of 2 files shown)


Snippets/Gateway/Gateway_3/Channels/Wildcard.config.xml
```diff
@@ -8,8 +8,10 @@
   <GatewayConfig>
     <Channels>
       <Channel Address="http://+:25899/RemoteSite/"
+               ChannelType="Http"/>               
+      <Channel Address="http://gateway.mycorp.com:25899/RemoteSite/"
```
timbussmann: wouldn't that conflict with the strong wildcard which also goes to `xyz:25899/RemoteSite`?

andreasohlund: Good catch, will double check and adjust.

Tests seems to suggest that you are right

https://github.com/Particular/NServiceBus.Gateway/blob/develop/src/NServiceBus.Gateway.AcceptanceTests/When_listening_on_wildcard_uri.cs#L43

andreasohlund: Confirmed, different port is needed. Fixed now

PR #4048, opened 2018-10-10
Review NServiceBus tuning page

danielmarbach:


nservicebus/operations/tuning.md
```diff
@@ -12,37 +12,30 @@ tags:
  - performance
  - throttling
  - concurrency
-reviewed: 2016-11-15
+reviewed: 2018-10-10
 ---
 
-NServiceBus uses defaults that ensure good performance in most common cases. While this is usually the preferred mode of operation there are situations where tuning needs to be applied.
-
+NServiceBus uses defaults that ensure good performance in common cases. While this is usually the preferred mode of operation there are situations where tuning might be desired.
```
andreasohlund: >NServiceBus uses defaults

Think it should be "use defaults"

nservicebus/operations/tuning.md
```diff
@@ -12,37 +12,30 @@ tags:
...
 
 ## Tuning concurrency
 
-
 partial: defaults
 
+Define a maximum concurrency setting that will make sure that no more messages than the specified value are ever being processed at the same time. Set this value to `1` to process messages sequentially. If not specified the transport will choose an optimal value.
```
andreasohlund: >Define a maximum concurrency setting that will make sure that no more

Limit maximum concurrency so that no more... ?

andreasohlund: >If not specified the transport will choose an optimal value

Do we need to specify what the transport optimizes for? (maximum throughput while not running out of resources)

nservicebus/operations/tuning_throttling_core_[,5].partial.md
```diff
@@ -1,6 +1,6 @@
 NOTE: By default endpoints are not throttled.
 
-Define a maximum value for the number of messages per second that the endpoint will process at any given time. This will help avoiding the endpoint from overloading sensitive resources that it's using like web-services, databases, other endpoints etc. Some examples would include
+Define a maximum value for the number of messages per second that the endpoint will process at any given time. This will help avoiding the endpoint from overloading sensitive resources that are using ressources like web-services, databases, other endpoints etc. Some examples would include
```
Comment (andreasohlund) [inline nservicebus/operations/tuning_throttling_core_[,5].partial.md]: >Define a maximum value for the number of messages

Limit the number of messages... ?

nservicebus/operations/tuning.md
danielmarbach: Isn't it uses defaults? Unsure though

andreasohlund: Had a faint memory that "only use plural s on the last word" but I can't find any refs on that

https://english.stackexchange.com/questions/53124/multiple-plural-any-rules

only made me more confused :)

danielmarbach:
@andreasohlund I pushed an update

nservicebus/operations/tuning.md
adamralph: It should be "uses defaults", since "NServiceBus" is a singular entity.

It would have to be plural for "use", e.g. "NServiceBus packages use defaults...".

PR #4051, opened 2018-10-10
[WIP] Update publishing-from-web-applications.md

timbussmann:


timbussmann:
I'm really not sure about this page in total tbh. It seems to address very transport specific problems, mostly MSMQ. The biggest part of this article relates to "storage-driven" (a term I've not seen before tbh) transports and how to make them work with scaled out endpoints, not web applications in particular. The whole description of publish-subscribe mechanics feel unnecessary as well. Any thoughts what really should stay in this article?

I think the only relevant part might be to describe the design in the "storage-driven transport topology" as an approach to avoid sender-side distribution or a distributor when scaling web servers is quite common and would require a lot of updating, while the described topology wouldn't need that. This still seems to only affect MSMQ as both SQL server and ASQ can share a queue across instances and don't need this architecture? Thoughts @Particular/docs-maintainers @Particular/nservicebus-maintainers ?

bording:
@timbussmann I agree with your assessment.

timbussmann:
replaced by https://github.com/Particular/docs.particular.net/pull/4065

PR #4055, opened 2018-10-11
add ? to Why use SQS

adamralph:


adamralph:
@Particular/docs-maintainers is it generally OK to add question marks to titles like this when the title is a question? If so, I can do a sweep to see where else this is needed.

PR #4056, opened 2018-10-11
stop using "persistence" as a noun

adamralph:
This is best demonstrated in our pluralisation of the noun. Sometimes we say "persistences" (26 hits), sometimes "persisters" (35 hits). 

Both of them are made up words, but the use of "persistence" as a noun that can be pluralised seems like the greater evil, and the use of "persister" as a noun (at least in plural) is more common.


adamralph:
FYI I rebased the branch.

@kbaley no, it doesn't cover _all_ occurrences. For this PR, I replaced "persistences" with "persisters" and then made sure that the affected documents were internally consistent. A few instances of "persistence" which should be "persister" were swept up in that, but I'm sure there are still plenty lurking around the repo. I guess it will be a gradual process to catch them all.

PR #4061, opened 2018-10-12
Fix snippets build for sqlp

SzymonPobiega:


bording:
This is already being addressed in #4059

PR #4065, opened 2018-10-15
Rework publishing from web applications page

timbussmann:
An attempt to move the publishing from web apps page to the MSMQ doco. Other transport don't really need to consider much of this as they have at least shared queues.

Thoughts @Particular/nservicebus-maintainers @SzymonPobiega @Particular/docs-maintainers ?


transports/msmq/scaling-out-web-apps.md
```diff
@@ -0,0 +1,31 @@
...
+In general, web applications using NServiceBus can be scaled out the same way as any other endpoint using [Sender-side distribution](). However, as web applications can scale up or down more frequently than regular endpoints, this quickly can become a maintenance burden. Consider the following recommendations when using NServiceBus with MSMQ transport on a web application:
+
+## Send-only endpoints
+
+[Send-only endpoints](/nservicebus/hosting/#self-hosting-send-only-hosting) allow to send and publish messages but without the message processing capabilities. As the web application itself does not handle business processes, they rarely need to consume messages themselves. This makes the web application process consuming less resources than using a regular endpoint and other endpoints don't need to be aware of the active instances.
+
+NOTE: Send-only endpoint also are not capable of receicing event subscription messages when using MSMQ. This means that send-only endpoints with MSMQ transport cannot publish messages except when using the [Storage-driven transport topology](#storage-driven-transport-topology).
+
+
+## Storage-driven Transport topology
```
SzymonPobiega: I am not sure if I like this name. Makes me confused when I read it. I'd rather mention `Subscription Manager Endpoint` directly here.

transports/msmq/scaling-out-web-apps.md
```diff
@@ -0,0 +1,31 @@
...
+reviewed: 2018-10-15
+---
+
+In general, web applications using NServiceBus can be scaled out the same way as any other endpoint using [Sender-side distribution](). However, as web applications can scale up or down more frequently than regular endpoints, this quickly can become a maintenance burden. Consider the following recommendations when using NServiceBus with MSMQ transport on a web application:
+
+## Send-only endpoints
+
+[Send-only endpoints](/nservicebus/hosting/#self-hosting-send-only-hosting) allow to send and publish messages but without the message processing capabilities. As the web application itself does not handle business processes, they rarely need to consume messages themselves. This makes the web application process consuming less resources than using a regular endpoint and other endpoints don't need to be aware of the active instances.
+
+NOTE: Send-only endpoint also are not capable of receicing event subscription messages when using MSMQ. This means that send-only endpoints with MSMQ transport cannot publish messages except when using the [Storage-driven transport topology](#storage-driven-transport-topology).
```
SzymonPobiega: receicing -> receiving

transports/msmq/scaling-out-web-apps.md
SzymonPobiega: Also, I wonder if we should not document this pattern separately. I remember one of our customers on SQL transport used it because he didn't want to remember who's publishing what. So he configured all events as published by some designated endpoint. It didn't matter because all endpoints used the same subscription storage.

timbussmann: I don't like it either, I just took the existing name but happy to find some better name.

timbussmann: > I remember one of our customers on SQL transport used it because he didn't want to remember who's publishing what. 
not sure we really want this to make an "official pattern"

SzymonPobiega: In SQL Server scenario I don't see any drawbacks in having a designated "subscription manager" endpoint and the advantage is less config

timbussmann: sounds like something we could fix anyway by making SQL handling pubsub instead of using message driven pubsub?

timbussmann:
this has gotten too old, going to kill this PR.

PR #4072, opened 2018-10-19
Update headers.md

robvdlv:
Change seemingly inadvertent use of `Correlation Id` to `Message Id`.

### DIFF
--- a/nservicebus/messaging/headers.md
+++ b/nservicebus/messaging/headers.md
@@ -73,7 +73,7 @@ A [unique ID for the current message](/nservicebus/messaging/message-identity.md
 
 NServiceBus implements the [Correlation Identifier](http://www.enterpriseintegrationpatterns.com/patterns/messaging/CorrelationIdentifier.html) pattern by using a `Correlation Id` header.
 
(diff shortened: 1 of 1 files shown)


mikeminutillo:
@robvdlv The original text is correct. `Correlation Id` is sticky. So once it is set it gets copied from incoming to outgoing message. It's only the very first message where the `Correlation Id` get's set to the `Message Id`.

mikeminutillo:
@robvdlv The original text is correct. `Correlation Id` is sticky. So once it is set it gets copied from incoming to outgoing message. It's only the very first message where the `Correlation Id` get's set to the `Message Id`.

mikeminutillo:
@robvdlv The original text is correct. `Correlation Id` is sticky. So once it is set it gets copied from incoming to outgoing message. It's only the very first message where the `Correlation Id` get's set to the `Message Id`.

robvdlv:
I opened the PR as I thought it would be more clear to say that the correlation id of the reply message will have the same value of the message id of the request message.

Once I realized the message id and correlation id of a request message will have an equal value for the request message, I understood why the original text is correct.

Thanks.

robvdlv:
I opened the PR as I thought it would be more clear to say that the correlation id of the reply message will have the same value of the message id of the request message.

Once I realized the message id and correlation id of a request message will have an equal value for the request message, I understood why the original text is correct.

Thanks.

PR #4079, opened 2018-10-24
Remove incorrect information

SeanFeldman:
This doesn't appear to be true. An attempt to retrieve `Notifications` from container will always throw. Instead, it's only possible to get it via `Settings` (separate PR?).


timbussmann:
is this really correct? see https://github.com/Particular/docs.particular.net/issues/4080#issuecomment-433316915

SeanFeldman:
> is this really correct? see [#4080 (comment)](https://github.com/Particular/docs.particular.net/issues/4080#issuecomment-433316915)

Yes, I've verified that. See https://github.com/Particular/docs.particular.net/issues/4080#issuecomment-433561500

PR #4090, opened 2018-10-29
Attempt to highlight difference between MEM, V6+ and native pubsub

### LINKED ISSUE #4078: Feedback: 'Using ServiceControl Events'
Feedback for 'Using ServiceControl Events' https://docs.particular.net/servicecontrol/contracts

Location in GitHub: https://github.com/Particular/docs.particular.net/blob/master/servicecontrol/contracts.md

ping @Particular/servicecontrol-maintainers

How would you do the mappings in netcore? There is no app.config. Is there a way to do it in code perhaps?

danielmarbach:
Closes #4078

### DIFF
--- a/Snippets/ServiceControlContracts/ServiceControlContracts_1/MessageFailedHandler.cs
+++ b/Snippets/ServiceControlContracts/ServiceControlContracts_1/MessageFailedHandler.cs
@@ -1,4 +1,4 @@
-﻿namespace Core6.Monitoring
+﻿namespace Contracts
 {
     using System;
     using System.Threading.Tasks;
--- a/Snippets/ServiceControlContracts/ServiceControlContracts_1/MessageFailedHandler.cs
+++ b/Snippets/ServiceControlContracts/ServiceControlContracts_1/MessageFailedHandler.cs
@@ -1,4 +1,4 @@
-﻿namespace Core5.Monitoring
+﻿namespace Core6.Monitoring
 {
     using System;
     using System.Threading.Tasks;
--- a/servicecontrol/contracts.md
+++ b/servicecontrol/contracts.md
@@ -42,10 +42,16 @@ To subscribe to the `MessageFailed` event:
 
  * Create an [NServiceBus endpoint](/nservicebus/hosting/nservicebus-host/).
  * Install the [ServiceControl.Contracts NuGet package](https://www.nuget.org/packages/ServiceControl.Contracts/).
- * Add the message mapping in the `UnicastBusConfig` section of the endpoint's app.config so that this endpoint will subscribe to the events from ServiceControl as shown:
+ * For transports using the [message-driven publish-subscribe](/nservicebus/messaging/publish-subscribe/) mechanism, either
+   * For NServiceBus Version 6 and higher, register the publisher by using the [routing config] code API(/nservicebus/messaging/routing.md#event-routing-message-driven).
+
+snippet: ServiceControlPublisherConfig
+
+   * or for NServiceBus Version 5 and below, add the message mapping in the `UnicastBusConfig` section of the endpoint's app.config so that this endpoint will subscribe to the events from ServiceControl as shown:
 
 snippet: ServiceControlEventsXmlConfig
 
+ * Transports that natively support publish and subscribe, see [multicast-enabled transports](/transports/#types-of-transports-multicast-enabled-transports), do not require to map the contracts.
  * Customize the endpoint configuration to use `JsonSerializer` as the message published by ServiceControl uses JSON serialization.
  * Customize the endpoint configuration so that the following conventions are used, as the `MessageFailed` event that is published by ServiceControl does not derive from `IEvent`.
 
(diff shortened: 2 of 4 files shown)


Snippets/ServiceControlContracts/ServiceControlContracts_1/MessageFailedHandler.cs
```diff
@@ -1,4 +1,4 @@
-﻿namespace Core5.Monitoring
+﻿namespace Core6.Monitoring
```
bording: Not that it really impacts anything, but this namespace seems really strange. What does "Monitoring" have to do with this? Also, why is "Core" in here at all?

servicecontrol/contracts.md
```diff
@@ -42,10 +42,16 @@ To subscribe to the `MessageFailed` event:
 
  * Create an [NServiceBus endpoint](/nservicebus/hosting/nservicebus-host/).
  * Install the [ServiceControl.Contracts NuGet package](https://www.nuget.org/packages/ServiceControl.Contracts/).
- * Add the message mapping in the `UnicastBusConfig` section of the endpoint's app.config so that this endpoint will subscribe to the events from ServiceControl as shown:
+ * For transports using the [message-driven publish-subscribe](/nservicebus/messaging/publish-subscribe/) mechanism, either
```
bording: This entire section doesn't make sense to be an unordered list, and adding conditional items seems awkward.

This is a set of steps to follow, in order. I think this section should be restructured, which will allow the new parts to fit in better.

servicecontrol/contracts.md
```diff
@@ -42,10 +42,16 @@ To subscribe to the `MessageFailed` event:
...
+   * For NServiceBus Version 6 and higher, register the publisher by using the [routing config] code API(/nservicebus/messaging/routing.md#event-routing-message-driven).
+
+snippet: ServiceControlPublisherConfig
+
+   * or for NServiceBus Version 5 and below, add the message mapping in the `UnicastBusConfig` section of the endpoint's app.config so that this endpoint will subscribe to the events from ServiceControl as shown:
 
 snippet: ServiceControlEventsXmlConfig
 
+ * Transports that natively support publish and subscribe, see [multicast-enabled transports](/transports/#types-of-transports-multicast-enabled-transports), do not require to map the contracts.
```
bording: "do not require to" isn't quite the right way to phrase that.

Something like "do not require any additional configuration" flows better.

servicecontrol/contracts.md
danielmarbach: What is exactly awkward about it? I've made those new items subitems for a good flow. How would you propose it to be changed? Even we create an order list with numbers we would have a conditional in it.

bording: Overall I don't feel like using bullet points at all is the right way to organize this. I'm not sure I'd want to just see an ordered list either, though. I feel like it would be better structured as more of a prose section with paragraphs, which could handle the different conditional paths better.

However, that's just my opinion, perhaps others are fine with it as-is.

danielmarbach: I thought about it and was thinking that it would not be enough content for sections. Happy to give it a try tough. Sometimes when you see something instead of just visualizing it in your mind it makes things clearer. I send in a proposal tomorrow

danielmarbach:
@bording have a look again. So I decided to restructure the flow a bit. Now the ordered list contains all the things that are common for all the transports and the publisher configuration is now its own section that can be skipped for transports using native pubsub. What do you think?

bording:
@Particular/servicecontrol-maintainers What do you think about the note and warning on this page? Should we remove them?

mikeminutillo:
@bording Yeah, I don't think that warning is accurate once we have a version of ServiceControl that runs on top of those other topologies. It's fairer to say that the subscribing endpoint needs to run on the same transport/topology as the ServiceControl instance it is subscribing to.

bording:
@mikeminutillo And what about the note:

>  NOTE: Integration endpoints should _not_ use the same `error` and `audit` queue as business endpoints since this risks failures in the integration endpoint, causing an infinite feedback loop. Using the same `audit` queue will cause the integration messages to be included in search results in ServiceInsight. This will confuse users searching for given failure since both the failure and the failure notification will be shown. See also: [Recoverability](/nservicebus/recoverability/) and [Audit Queue Settings](/nservicebus/operations/auditing.md).

Is that actually relevant for this page? If so, is the current location the right spot for it? Seems out of place to me.

mikeminutillo:
@bording where else would you put it? This is the page that explains how to set up these integration endpoints. If you have a mistake in your handler for the `MessageFailed` event then you're going to go into an infinite loop.

bording:
I guess maybe I don't understand how that note is connected to that page, then. I read the page and don't have any idea why that note is on the page. It uses the term "integration endpoint", but that doesn't seem to be defined anywhere on the page.

mikeminutillo:
Yeah, that's fair. What if we replaced the term "integration endpoint" with "endpoints that subscribe to ServiceControl events"?

bording:
@mikeminutillo That would be perfect. With that change, I think it makes it clear why the note is on this page.

I'll go ahead and apply that tweak to this PR since I know @danielmarbach is off until next week.

bording:
@mikeminutillo I remove the warning and tweaked the wording and position of the note. If you're okay with the changes, please merge the PR!

Issue #4098, opened 2018-11-01
Feedback: 'MSMQ Transport connection strings'


stevebroshar:
Feedback for 'MSMQ Transport connection strings' https://docs.particular.net/transports/msmq/connection-strings

Location in GitHub: https://github.com/Particular/docs.particular.net/blob/master/transports/msmq/connection-strings.md

What is this connection string? We have an nsb connection string to a database. That connection string specifies a database -- using mssql format for the string. I can't tell if that's the same or a different thing this page is talking about. Either I can add the indicated parameters to my existing conn string or this is talking about a different conn string that must be named "nservicebus/transport". I think either option is strange/unexpected/confusing. 

As with many other pages, this info is not clear. I'm left with experimenting to find the answer :(

Also, the default value for timeToReachQueue is not specified. A linked page indicates infinite. But all the other settings on this page indicate the default. Would be nice to be consistent.

ramonsmits:
@stevebroshar This is a connection string for the MSMQ transport. As MSMQ is always local you don't have so specify any sort of identifier.

The bread crumbs at the top identify where this content is part of.

FYI, our current version does not use a connection string anymore as MSMQ doesn't use that concept and caused confusion.  NServiceBus has the concept of a transport and the concept of a persister and both have different connection configuration requirements. 

Regarding time to reach queue, that linked info is false. The default usually is 4 days but the .net property will use that system default if the property value is `Infinite`. I'll add the 4 days to our docs as that is useful information.

ramonsmits:
Created a docs PR #4101 , closing this

PR #4101, opened 2018-11-01
Added MSMQ default value of 4 days for TimeToReachQueue

ramonsmits:
Based on feedback https://github.com/Particular/docs.particular.net/issues/4098


ramonsmits:
While busy with this, shall we move the note that the connection string is optional to the top of the document?

dvdstelt:
Why not backport the doco for previous versions than 6?
I'm okay with moving the note to the top.

ramonsmits:
@dvdstelt This is for all MSMQ core versions that use connection strings

PR #4111, opened 2018-11-09
Add link to Managed identities

markgould:
One of the ways to access Azure resources is to use Managed identities for Azure resources (previously known as Managed Service Identity). Added a link on how to use this with AzureServiceBus


danielmarbach:
Thanks @markgould for this contribution!

PR #4124, opened 2018-11-19
Update routing_content_core_6.partial.md

seanfarmar:


transports/msmq/routing_content_msmqtransport_[1,).partial.md
```diff
@@ -1,5 +1,9 @@
 Mapping the logical destination to the physical address containing the queue and host names is the responsibility of physical routing. 
 
+
```
andreasohlund: Notpick but if the note is the same in all 3 places we should pull it into an "include"

https://github.com/Particular/docs.particular.net#markdown-includes

transports/msmq/routing_content_core_[,6).partial.md
```diff
@@ -1,5 +1,8 @@
+
+NOTE: When using instance mapping, the settings will have no effect on **audit/error** queue and **publish/subscribe** (except for subscription messages, subscription messages will be sent to all listed instances)
```
timbussmann: I think there is no instance mapping file in NSB < v6, so this note wouldn't make much sense here, does it?

transports/msmq/routing_content_core_6.partial.md
```diff
@@ -3,6 +3,9 @@ Mapping the logical destination to the physical address containing the queue and
 The preferred way of configuring the physical routing is via the instance mapping file.
 
 
+NOTE: When using instance mapping, the settings will have no effect on **audit/error** queue and **publish/subscribe** (except for subscription messages, subscription messages will be sent to all listed instances)
```
timbussmann: shouldn't this be within the instance mapping file section?
should it be spelled out "audit and error queues"?
do we need to mention pubsub?

SzymonPobiega: I get why we mention it does not include audit/error but why mentioning pub/sub here?

andreasohlund: Doesn't it make sense to clarify that when publishing to only the instances that have subscribed will get the event regardless what's present in the mapping file?

timbussmann: That way of wording makes more sense than stating that pubsub is t affected except for subscription messages.

transports/msmq/routing_content_core_[,6).partial.md
seanfarmar: Good point, so keep the partial and have it for 6 and up?

seanfarmar:
I've removed it from `routing_content_core_[,6).partial.md` and moved the note into the 
`Instance mapping file` section.

And changed the wording... Hope this is better...

What do you think?

transports/msmq/routing_content_msmqtransport_[1,).partial.md
```diff
@@ -5,6 +5,11 @@ The preferred way of configuring the physical routing is via the instance mappin
 
 ## Instance mapping file
 
+NOTE: When using instance mapping: 
+* The settings will have no effect on **audit and error queues**.
+* **publish/subscribe**: The publisher will be publishing messages only to the endpoint instances that have subscribed to the events, ignoring the settings in the mapping file (the adress of the subscriber that was provided in the subscription messgae will be used).
```
andreasohlund: **publish/subscribe** => **publishing**

transports/msmq/routing_content_msmqtransport_[1,).partial.md
```diff
@@ -5,6 +5,11 @@ The preferred way of configuring the physical routing is via the instance mappin
...
+* **publish/subscribe**: The publisher's instances should be entered in the Instance mapping file (so the subscribers can send a subscription message to them)
```
andreasohlund: **publish/subscribe** => subscriptions
```suggestion
* **subscribing**: Subscription messages will be sent to all publisher instances listed in the Instance mapping file
```

seanfarmar:
Updated

transports/msmq/routing_content_core_6.partial.md
```diff
@@ -2,9 +2,13 @@ Mapping the logical destination to the physical address containing the queue and
 
 The preferred way of configuring the physical routing is via the instance mapping file.
 
-
 ## Instance mapping file
 
+NOTE: When using instance mapping: 
+* The settings will have no effect on **audit and error queues**.
```
timbussmann: does the Note-section allow for bullet points?

seanfarmar: I'm not sure...

mauroservienti:
What’s missing here to get this in?

mauroservienti:
@seanfarmar ping

seanfarmar:
@mauroservienti I think it's all done if we have reviewers happy?

transports/msmq/routing_content_core_6.partial.md
DavidBoike: @timbussmann If you do it like this https://github.com/Particular/docs.particular.net#multi-line

mauroservienti: Updated to make the note multi-line.

timbussmann: thanks @DavidBoike and @mauroservienti ❤️

PR #4128, opened 2018-11-20
Update dotnet new templates page

bording:


Snippets/Templates/Templates.sln
```diff
@@ -6,15 +6,26 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Templates_1", "Templates_1\
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Templates_2", "Templates_2\Templates_2.csproj", "{00F35CB1-EDB4-4E47-82CA-EF604E212816}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Templates_3", "Templates_3\Templates_3.csproj", "{1A32DC47-08E1-4FAA-9EE7-7829CEF0BC48}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
+		Release|Any CPU = Release|Any CPU
```
mauroservienti: Do we need the release config in the snippets? Or a single one is good enough?

Snippets/Templates/Templates_3/NServiceBusDockerContainer.cs
```diff
@@ -0,0 +1,107 @@
+using System;
+using System.Diagnostics;
+using System.Threading.Tasks;
+using NServiceBus;
+using NServiceBus.Logging;
+
+namespace Templates_2
```
mauroservienti: namespace doesn't match, should be `Templates_3`

bording: Forgot to update those. Will do.

Snippets/Templates/Templates.sln
bording: We don't need them, but I didn't think about removing them. I think there's actually a script that will go through and clean that stuff up, but I'll go ahead and remove them.

PR #4130, opened 2018-11-22
Legacy doco review for RabbitMQ callbacks

SzymonPobiega:


transports/rabbitmq/callbacks_queue_rabbit_[,3].partial.md
```diff
@@ -1,6 +1,8 @@
-When scaling out an endpoint, any of the endpoint instances can consume messages from the same shared broker queue. However, this behavior can cause problems when dealing with callback messages because the reply message for the callback needs to go to the specific instance that requested the callback.
+When sending a message, a callback can be registered that will be invoked when a response arrives.
 
-Callbacks are enabled by default, and the transport will create a separate callback receiver queue, named `{endpointname}.{machinename}`, to which all callbacks are routed.
+When scaling out an endpoint, any of the endpoint instances can consume messages from the same shared broker queue. However, this behavior can cause problems when dealing with reply messages for which a callback has been registered. The reply message needs to go to the specific instance that registered the callback.
```
Comment (mikeminutillo) [inline transports/rabbitmq/callbacks_queue_rabbit_[,3].partial.md]:
```suggestion
When scaling out an endpoint, any of the endpoint instances can consume messages from the same shared broker queue. This behavior can cause problems when dealing with reply messages for which a callback has been registered. The reply message needs to go to the specific instance that registered the callback.
```

transports/rabbitmq/callbacks_queue_rabbit_[,3].partial.md
```diff
@@ -1,6 +1,8 @@
-When scaling out an endpoint, any of the endpoint instances can consume messages from the same shared broker queue. However, this behavior can cause problems when dealing with callback messages because the reply message for the callback needs to go to the specific instance that requested the callback.
+When sending a message, a callback can be registered that will be invoked when a response arrives.
```
Comment (mikeminutillo) [inline transports/rabbitmq/callbacks_queue_rabbit_[,3].partial.md]: Is this different from the callbacks package? Why do we have both? Should we differentiate them in doco?

transports/rabbitmq/callbacks.md
```diff
@@ -2,12 +2,12 @@
 title: Callback support
 reviewed: 2016-08-30
 component: Rabbit
-reviewed: 2017-01-13
+reviewed: 2018-11-22
 versions: '[2,]'
 related:
  - nservicebus/messaging/callbacks
 redirects:
  - nservicebus//rabbitmq/callbacks
```
mikeminutillo: Does this `//` work?

SzymonPobiega: That was a typo ❤️

transports/rabbitmq/callbacks_queue_rabbit_[,3].partial.md
Comment (SzymonPobiega) [reply transports/rabbitmq/callbacks_queue_rabbit_[,3].partial.md]: Nope. I copied the header from that callback page to have a nice intro here. I think the old one suggested that `callback` *is* the message while the main callbacks page https://docs.particular.net/nservicebus/messaging/callbacks makes it clear that the callback is registered for handling a response.

SzymonPobiega:
@mikeminutillo ping

PR #4132, opened 2018-11-22
Legacy review of RabbitMQ message id strategy

SzymonPobiega:


transports/rabbitmq/message-id-strategy.md
```diff
@@ -11,8 +11,10 @@ redirects:
...
+
+For integration scenarios where the sender is not controlled, the receiver might need to employ a custom strategy that extracts a message identity from a custom header or part of the message body. This custom strategy can be configured by calling:
 
 snippet: rabbitmq-config-custom-id-strategy
 
-WARNING: It is extremely important to use a uniquely identifying property of the message in a custom message identity strategy. If the value for a message were to change (for example, if attempting to use `Guid.NewGuid().ToString()`) then message retries would break, as the infrastructure would be unable to determine that it was processing the same message repeatedly.
\ No newline at end of file
+WARNING: It is extremely important to ensure that the custom strategy is deterministic i.e. that it returns the same value when invoked multiple times for the same message, and that no two messages are associated the same value.
```
mikeminutillo:
```suggestion
WARNING: It is extremely important that the custom strategy is deterministic (it returns the same value when invoked multiple times for the same message), and that no two messages are assigned the same value. 
```

PR #4135, opened 2018-11-22
improve transport decision flow chart and add header before chart

sergioc:
Made transport flowchart easier to read:

- make flow chart less dense and better organized
- remove redundant node
- use SVG instead of PNG


adamralph:
This change was cherry-picked directly on the branch for #4100

PR #4136, opened 2018-11-22
Update documentation and samples based on new templates

mauroservienti:


bording:
@mauroservienti I pushed up some changes.

The one thing I'm wondering about still is should we add something to windows-service.md that shows using the `--run-as-service` parameter?

mauroservienti:
>I pushed up some changes.

Cool, thanks.

>The one thing I'm wondering about still is should we add something to windows-service.md that shows using the `--run-as-service` parameter?

Right. Since it is not mentioned anywhere we should probably do 2 things:

1. Augment the [templates doco](https://docs.particular.net/nservicebus/dotnet-templates) with a couple of snippets detailing the result of using a `WindowsService template`. So that there will be the aforementioned command line argument mentioned somewhere.
1. Add a note in ` windows-service.md` that if you are using the templates (link to the doco) to generate your code skeleton then the parameter is needed at install time too

bording:
I made a change for `--run-as-service`. What do you think?

bording:
One other thing I noticed, the windows service page still has a "Compared to NServiceBus Host" section which seems out of place now. Should we move that, or just drop it?

mauroservienti:
> I made a change for --run-as-service. What do you think?

I like it.

> One other thing I noticed, the windows service page still has a "Compared to NServiceBus Host" section which seems out of place now. Should we move that, or just drop it?

it sounds off, I'd drop it.

bording:
> it sounds off, I'd drop it.

Done!

Is this ready to merge now?

mauroservienti:
> Is this ready to merge now?

I think so. Removing the WIP mark.

PR #4162, opened 2018-12-04
Fix MsLogging sample/snippets from 2.2.0 obsoletes

DavidBoike:
Hey @SimonCropp,

Microsoft.Extensions.Logging 2.2.0 made some APIs obsolete and wants them declared on an ILoggingBuilder instead of ILoggerFactory, the problem is how to generate that. In a previous commit I temporarily fixed the build by locking the versions to 2.1.*, but this is my attempt to fix it for real.

Does this look OK?

### DIFF
--- a/Snippets/MsLogging/MsLogging_1/ProgramService.cs
+++ b/Snippets/MsLogging/MsLogging_1/ProgramService.cs
@@ -2,6 +2,7 @@
 using System.ComponentModel;
 using System.ServiceProcess;
 using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using NServiceBus;
 using NServiceBus.Logging;
@@ -12,7 +13,7 @@ class ProgramService :
     ServiceBase
 {
     IEndpointInstance endpointInstance;
-    LoggerFactory loggerFactory;
+    Microsoft.Extensions.Logging.ILoggerFactory loggerFactory;
 
     static void Main()
     {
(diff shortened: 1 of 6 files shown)


Snippets/MsLogging/MsLogging_1/ProgramService.cs
```diff
@@ -37,8 +38,16 @@ protected override void OnStart(string[] args)
...
+        var serviceCollection = new ServiceCollection();
+        serviceCollection.AddLogging(loggingBuilder =>
+        {
+            loggingBuilder.AddFilter(level => level >= Microsoft.Extensions.Logging.LogLevel.Information);
+            loggingBuilder.AddConsole();
+        });
+
+        var serviceProvider = serviceCollection.BuildServiceProvider();
+
+        loggerFactory = serviceProvider.GetService<Microsoft.Extensions.Logging.ILoggerFactory>();
```
SimonCropp: this looks correct. but wow that is so much more verbose. can u point me to where u got this from? is there some ms logging upgrade guide?

DavidBoike: It was totally fumbling around in the dark just trying to make something work. The only thing the obsolete message says is to use ILoggingBuilder instead of ILoggerFactory. In an aspnetcore app it’s a lot easier because the IServiceCollection is provided, like in this:

https://github.com/Particular/docs.particular.net/commit/1caea8fb262655198e5d85fd4255ef61d95ef0af#diff-54f6c8671674829fafe0446884e7a16c

But outside that context it appears to be downright painful.

SimonCropp: in that case it looks good to me. i will do some research on the side and attempt to work out the underlying thinking behind the change

PR #4164, opened 2018-12-05
Move transport selection section up

mauroservienti:
Moves the transport selection section right after the list of supported transports, as it sounds more important than community transports and WebSphere


adamralph:
👍 good idea @mauroservienti

PR #4176, opened 2018-12-11
ASB Legacy upgrade guides section title

SeanFeldman:
Upgrade guides for the legacy ASB needs to call out that it's the old transport.


SeanFeldman:
@DavidBoike we've missed this one 😃
Any other location you can think of "legacy" clarification would be required? Any marketing stuff?

PR #4189, opened 2018-12-17
Describe learning transport convention

danielmarbach:


transports/learning/index.md
```diff
@@ -45,7 +45,7 @@ Note: Production transports will run with higher concurrency setting by default
 
 ### Storage Directory
 
-By default all data is stored in a `.learningtransport` directory that exists at the solution root.
+By default all data is stored in a `.learningtransport` directory that exists at the solution root or in any `.learningtransport` directory that exists in the parent directory hierarchy of the endpoint using the learning transport.
```
timbussmann: that sentence is quite long. Can we split it into something like

By default all data is stored in a `.learningtransport` directory. The endpoint will traverse the folder hierarchy upwards in search for a .learningtransport directory or create one at the solution root folder if no matching folder has been found before.

danielmarbach: I think the current text is a bit more "accurate" because it treats the solution directory as a first priority. We could make it less long by rewording to

> By default all data is stored in a `.learningtransport` directory that exists at the solution root or in any `.learningtransport` directory that exists in the parent directory hierarchy.

timbussmann: > I think the current text is a bit more "accurate" because it treats the solution directory as a first priority

what would you say makes my proposal less accurate in regards to the behavior (and not the implementation detail)?

danielmarbach: My intent was to capture the implementation with the precedence as well. That's why I claimed mine is more accurate. Given that you are part of the maintainer group that owns this piece of the code I'll let you throw the dice and make the choice which text you find the better variant

timbussmann: >  because it treats the solution directory as a first priority. 

but is that true? If there is a data directly deeper in the folder hierarchy, this would take precedence to my understanding?

danielmarbach: I meant at the same level. You are right if it is deeper down the hierarchy it would take precedence

Issue #4076, opened 2018-10-23
Entity Framework integration should be changed to do cross db communication


SimonCropp:
It would be unusual for multiple endpoints to all use isolated dbs.

instead both endpoints should be separate the business db from the nsb transport db

Feedback for 'Entity Framework integration' https://docs.particular.net/samples/entity-framework/

Location in GitHub: https://github.com/Particular/docs.particular.net/blob/master/samples/entity-framework/sample.md

ping @Particular/nservicebus-maintainers

timbussmann:
@SimonCropp are you referring to the fact that the sample uses both  a `NsbSamplesEfUowNh`  and a `NsbSamplesEfUowSql` database?

SimonCropp:
@timbussmann pretty much. in most cases i have seen endpoints have their own db for business data (and sometime co located with persistence data), but they all communicate on the sql transport via the one db. and, as i understand it, it should still work without DTC https://docs.microsoft.com/en-us/sql/database-engine/availability-groups/windows/transactions-always-on-availability-and-database-mirroring?view=sql-server-2017

timbussmann:
@SimonCropp I tend to agree, however my understanding of the sample is, that these are actually two independent samples on that solution. One for nhibernate and one for sql persistence. Those sample do not communicate with each other so it's fine for them to have separated dbs. It would be a valid change to split each of the "samples" so that they would actually use dedicated databases for transport and persistence but that might be a bit overkill?

SimonCropp:
perhaps it should be a diff sample?

timbussmann:
reworked the sample in #4195

PR #4195, opened 2018-12-19
Remove NHibernate project from EF integration sample

timbussmann:
relates to #4076 

Removes the NHibernate part of the sample which shows how to integrate NHibernatePersistence with EntityFramework. Whenever possible, users should use SQLP instead of NHPersistence when trying to integrate with EF, there was just a historical reason to provide samples for both and I think we don't really need that any longer.


timbussmann:
thoughts @Particular/nhibernate-persistence-maintainers @Particular/developer-experience-squad ?

PR #4213, opened 2019-01-04
Tweak SQL passthrough sample instructions

andreasohlund:


andreasohlund:
@SzymonPobiega removed the url since I've  tweaked the start url in https://github.com/Particular/docs.particular.net/pull/4214/files

PR #4215, opened 2019-01-04
Ambient transactions are supported in Core 2.1

SzymonPobiega:


transports/sql/transactions_ambient-core-warning_sqltransport_[4,).partial.md
```diff
@@ -1,7 +1,9 @@
 
-#### Transaction scope mode is not supported in .NET Core 2.0
+#### Distributed transactions are not supported in .NET Core
 
-The default transaction mode, i.e. transaction scope, is currently not supported in .NET Core 2.0. Any attempt to use it will result in the following exception:
+Although the .NET Core SQL Server driver supports the ambient transactions, it does not support the Distributed Transaction Coordinator so any attempt to use more than one transactional resource within an ambient transaction will cause an error.
```
andreasohlund:
```suggestion
Although the .NET Core SQL Server driver supports enlisting in ambient transactions, it does not support participating in distributed transactions so any attempt to use more than one transactional resource within the same ambient transaction will cause an error.
```

transports/sql/transactions_ambient-core-warning_sqltransport_[4,).partial.md
```diff
@@ -1,7 +1,9 @@
...
+
+Prior to Version 2.1, the .NET Core SQL Server driver did not support the ambient transactions and any attempt to use it resulted in the following exception:
```
andreasohlund:
```suggestion
Prior to Version 2.1, the .NET Core SQL Server driver did not support enlisting in ambient transactions and any attempt to do so resulted in the following exception:
```

SzymonPobiega:
I applied your suggestions @andreasohlund

PR #4218, opened 2019-01-07
Extended support policy enumeration with table

DavidBoike:
Resolves Particular/CustomerSuccess#1137


HEskandari:
This would resolve issue https://github.com/Particular/CustomerSuccess/issues/1137

PR #4223, opened 2019-01-09
Tagging ext support

HEskandari:
flag the extended support versions

PR #4224, opened 2019-01-09
Extended support notations for ALL packages

DavidBoike:
Builds on (and if accepted, replaces) #4223.

Hides the NuGet searcher behind a class that caches the responses, and after all the raw data is gathered, scans the dependency chains using cached package information to find the NServiceBus reference. (Same concept as used by the coreDependencies.linq script.) That gives us relative confidence on the associated Core assembly version, which we can then use to determine the extended support status by only providing an array of ints describing major core versions.

It turns out we've shipped some assemblies that only have a transitive dependency on an NServiceBus version, for example [NServiceBus.Metrics.ServiceControl 2.0.0](https://www.nuget.org/packages/NServiceBus.Metrics.ServiceControl/2.0.0), which has a dependency on NServiceBus.Metrics 2.x ... transitive to NServiceBus 6.x. Cycling through the dependencies using the cached package metadata allows us to discover those.

There are some really old versions of things don't have ANY NServiceBus dependencies, such as [NServiceBus.Autofac 3.2.8](https://www.nuget.org/packages/NServiceBus.Autofac/3.2.8). These are all V3 so I think what they really mean is that the major version is 3, but since that's not covered by extended support I don't think we need to worry.

This is a lot more complicated though, so this needs @bording 	to look at it as well as @HEskandari.


bording:
I haven't looked over the code changes, but the output from the new script looks great!

PR #4228, opened 2019-01-14
Fix typo on word successful

evolvedlight:
This just replaces succesful with successful.


mauroservienti:
thanks for raising this @evolvedlight

PR #4238, opened 2019-01-24
Update ContextualLoggerUsage.cs

JorgIanFloren:
as per your own sample project


janovesk:
Hmm, I don't think so. 

The [serilog logger](https://docs.particular.net/nservicebus/logging/serilog) enriches the logger context with this information automatically through a behavior so you don't have to do that all over your codebase. See (https://docs.particular.net/nservicebus/logging/serilog#tracing-contextual-logger-handler-enrichment)

```When a handler is invoked, a new logger is forked from the above enriched physical logger with a new enriched property named Handler that contains the the FullName of the current handler.```

JorgIanFloren:
I see, so the code base of the sample should be updated rather than the documentation. Both notations work, but I don't see an immediate use case where one should want to override the type of the handler -only in logging- Because that's all the property does.

PR #4244, opened 2019-02-06
Change huge CTA to link to Get Started

dvdstelt:
There's a huge CTA on the doco homepage, linking to a page with almost no information, but a million links.
It would be so much better to link this to the Getting Started page and immediately show code.

I press this button a million times because it's the **biggest**, **baddest** CTA button out there.
This was discussed quite a few times already, why not just merge it?! o:-)


sergioc:
The current label doesn't reflect the new link you propose. Also, you're removing the only way to access documentation topics from the home page.

PR #4245, opened 2019-02-06
Typo

aleksandr-samila:
I have a few more things to say. Will do this in comments

### DIFF
--- a/samples/non-durable-messaging/sample.md
+++ b/samples/non-durable-messaging/sample.md
@@ -42,7 +42,9 @@ snippet: message-definition
 
 ## Running the solution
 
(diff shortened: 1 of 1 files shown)


aleksandr-samila:
**1.** 
image
Should we indicate that ***Receiver*** at the beginning should be run first? Otherwise ***Sender*** will fail with exception
image
For experienced programmers this may be obvious, but what about the rest?

aleksandr-samila:
**2.**
image
Using built-in Viewer I can't find this property
image

HEskandari:
@aleksandr-samila I think your points makes sense. I'll have the PR updated.

PR #4251, opened 2019-02-07
Pass in the publish options created to the Publish method

jbogard:
It wasn't actually using the same transaction/connection.


SeanFeldman:
Thank you @jbogard !

SzymonPobiega:
@jbogard thanks ❤️ What a nasty bug!

PR #4261, opened 2019-02-12
Fix docker sample

danielmarbach:
- Remove network new because it is unnecessary
- Specify hostname rabbitmq and connection port to allow other containers to connect to broker

### DIFF
--- a/samples/hosting/docker/Core_7/docker-compose.yml
+++ b/samples/hosting/docker/Core_7/docker-compose.yml
@@ -5,8 +5,6 @@ services:
         build:
             context: .
             dockerfile: ./Sender/Dockerfile
-        networks:
-            - new
         depends_on:
             rabbitmq:
                 condition: service_healthy
@@ -15,8 +13,6 @@ services:
         build:
             context: .
             dockerfile: ./Receiver/Dockerfile
-        networks:
-            - new
         depends_on:
             rabbitmq:
                 condition: service_healthy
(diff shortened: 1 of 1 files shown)


danielmarbach:
I prepared a client demo and found out that the docker sample is not working. The issue is that the default connection port is not exposed as well as the hostname is not set. This causes the sender and the receiver to not be able to connect to the broker

samples/hosting/docker/Core_7/docker-compose.yml
```diff
@@ -15,20 +13,16 @@ services:
...
             rabbitmq:
                 condition: service_healthy
     rabbitmq:
         image: "rabbitmq:3-management"
+        hostname: "rabbitmq"
         ports:
             - "15672:15672"
-        networks:
-            - new
+            - "5672:5672"
```
dbelcham: This seems to be the only change that is required to make the demo work.

danielmarbach: Based on my reading the network piece is not necessary and thus can be removed.

About the host name change I can check if that is really required

dbelcham: My understanding is that `network` is a way to establish a virtual subnet that the different containers in a composition reside within. With it you won't need the containers to know things like host names for each other. Docker-compose should just take care of making those work for you.

That said, `new` is a horrible name for a network. It probably should be something like `demo-net`

danielmarbach: By default Compose sets up a single network for your app. Each container for a service joins the default network and is both reachable by other containers on that network, and discoverable by them at a hostname identical to the container name.

https://docs.docker.com/compose/networking/

Ergo this meant for me no network definition is necessary
About the hostname if thought rabbiz it binds by default to localhost only. By specifying the hostname it binds to that hostname as well. Will check later.

danielmarbach: Hostname is not necessary. Double checked and removed

PR #4262, opened 2019-02-13
Remove legacy ASB transport FAQ page

timbussmann:
The FAQ page doesn't look well maintained and the content isn't written in FAQ style. I've tried to extract useful content into other pages and removed the FAQ page.

Thoughts @Particular/azure-maintainers ?


danielmarbach:
I'm 👍 for this change. The comparison is outdated and not necessary.

transports/azure-service-bus/legacy/transaction-support.md
```diff
@@ -46,3 +46,8 @@ When transactions are disabled in NServiceBus then the transport uses the Azure
 The message is deleted from the queue directly after the receive operation completes, before it is processed, meaning that it's not possible to retry that message in case of processing failures. As transient exceptions occur regularly when integrating with online services, disabling retries when in unreliable mode is not recommended. This mode should only be used in very specific situations, when message loss is acceptable.
 
 NOTE: For a full explanation of the transactional behavior in Azure, refer to [Understanding internal transactions and delivery guarantees](understanding-transactions-and-delivery-guarantees.md).
+
+
+## Duplicate Detection
+
+Using the supported transaction levels, the transport only provides an **`at-least-once`** delivery model. Azure Service Bus provides a [Duplicate detection feature](https://docs.microsoft.com/en-us/azure/service-bus-messaging/duplicate-detection) which will ensure the message is received exactly once if enabled. However, Duplicate Detection comes at the expense of throughput and is time constrained.
```
SeanFeldman: Thing is, our Delayed retries will block customers from using native deduplication feature. If we're adding this section here, I'd add a note about that as well.

timbussmann: interesting. How does it block customers? or does the native deduplication "swallow" delayed retries?

Sounds like we should clearly state the drawbacks or even remove this section altogether too?

SeanFeldman: > does the native deduplication "swallow" delayed retries?

Exactly that. I'm 👍 to note that as a feature conflicting with NSB's Delayed Retries.

timbussmann:
@SeanFeldman I added a warning about the outbox compatibility. Please have a look.

SeanFeldman:
> WARN: The duplicate detection feature is not compatible with the [outbox](/nservicebus/outbox). Enabling duplicate detection and the outbox together can cause message loss.

Is problematic.
- "The duplicate detection feature" refers to the native/ASB service feature. Not sure that's clear enough to someone that is not well versed in ASB/NSB. 
- The warning is not addressing what we've discussed, SLR not working with ASB's dedup.

Sorry to such a PITA...

timbussmann:
err, no idea how I got to the outbox :D will fix it.

timbussmann:
have another look @SeanFeldman

SeanFeldman:
Content verification fails with
```
System.Exception : Path: 'C:\BuildAgent\work\a48e3341ae42e493\transports\azure-service-bus\legacy\transaction-support.md'. Error: 'Error:'Path does not exist'. Url:'recoverability/#delayed-retries'.'
```
Should it be `recoverability.md#delayed-retries`?

timbussmann:
> Should it be `recoverability.md#delayed-retries`?

not sure, the recoverability document is index.md inside the recoverability folder.

PR #4267, opened 2019-02-15
Update retries.md

timbussmann:


transports/azure-service-bus/legacy/retries.md
```diff
@@ -23,7 +23,7 @@ Azure Service Bus supports a `MaxDeliveryCount` at the entity level, which defin
 
 To implement Immediate Retries, the native Azure Service Bus `MaxDeliveryCount` property of the `BrokeredMessage` is used to track the processing attempts by the endpoint instances for a specific message. If the number of attempts exceeds the `MaxRetries` setting, then the message will be forwarded to the delayed retries if configured, and eventually to to the error queue in case error persists. If immediate retries are disabled for the endpoint, the transport will default `MaxDeliveryCount` to 10 attempts.
```
timbussmann: > If immediate retries are disabled for the endpoint, the transport will default `MaxDeliveryCount` to 10 attempts.

that's a weird behavior, should that be in a Note or even Warning?

SeanFeldman: The legacy transport always defaulted max delivery count to a specific value.

A note perhaps, but not a warning.

timbussmann: normally, if you specify 0 retries you do that because you don't want retries, not because you want 10 retries, so this seems to be completely out of sync with the "expected" and documented behavior?

SeanFeldman: Setting retries to zero would not be translated to zero delivery counts. That's invalid.
Also, given the nature of the underlying service, setting it to a low number such as 1 would be too risky, hence the default was chosen as 10.

Note that the default would be applied if FLR/SLR would not yield a proper number that could be used for delivery attempts. And for error and audit queues.

timbussmann: > Setting retries to zero would not be translated to zero delivery counts.

no but 1 (assuming no network related errors).

> Also, given the nature of the underlying service, setting it to a low number such as 1 would be too risky

why not warn about this then?

> Note that the default would be applied if FLR/SLR would not yield a proper number that could be used for delivery attempts. And for error and audit queues.

since FLR has a default value, this is only the case if the user explicitly sets it to 0? How does FLR configuration setting affect this? What do you mean by error and audit queues? NServiceBus only write to those queues, I don't understand how this setting is relevant for this?

SeanFeldman: > no but 1 (assuming no network related errors).

Very dangerous assumption when working with ASB. Also, it would never be 1, as it would mean a single delivery that went rough would dead-letter messages.

> why not warn about this then?

This is a legacy transport we're reviewing here. IMO the amount of investment should be minimal. Those that already use it are well aware. Those that will be using ASB transport, will be using the new one.

> since FLR has a default value, this is only the case if the user explicitly sets it to 0? How does FLR configuration setting affect this? What do you mean by error and audit queues? NServiceBus only write to those queues, I don't understand how this setting is relevant for this?

Forget my mumbling. I haven't touch that code in a long time. We've changed it to decouple from the number of retries and [all queues get `int.MaxValue`](https://github.com/Particular/NServiceBus.AzureServiceBus/blame/master/src/Transport/Creation/AzureServiceBusQueueCreator.cs#L34) for `MaxDeliveryCount`. Unless [overridden by customers](https://github.com/Particular/NServiceBus.AzureServiceBus/blame/master/src/Transport/Creation/AzureServiceBusQueueCreator.cs#L46).

timbussmann: > Very dangerous assumption when working with ASB. Also, it would never be 1, as it would mean a single delivery that went rough would dead-letter messages.

isn't that the expected behavior when setting retries to 0? (except that the message should go to the error queue and not the DLQ)

> This is a legacy transport we're reviewing here. IMO the amount of investment should be minimal. Those that already use it are well aware. Those that will be using ASB transport, will be using the new one.

Fair point. Still confirms me that we should make this behavior very explicit in doco.

> Forget my mumbling. I haven't touch that code in a long time. We've changed it to decouple from the number of retries and [all queues get `int.MaxValue`](https://github.com/Particular/NServiceBus.AzureServiceBus/blame/master/src/Transport/Creation/AzureServiceBusQueueCreator.cs#L34) for `MaxDeliveryCount`. Unless [overridden by customers](https://github.com/Particular/NServiceBus.AzureServiceBus/blame/master/src/Transport/Creation/AzureServiceBusQueueCreator.cs#L46).

is this requiring any doco changes regarding the described behavior then?

SeanFeldman: > isn't that the expected behavior when setting retries to 0? (except that the message should go to the error queue and not the DLQ)

I'd disagree. There's a chance a message would be successfully processed, but failing to be completed due to some network related issue. That message would never be re-delivered if delivery count was bumped, causing it to go to the DLQ on the broker. Nothing we'd be able to do about it.

> is this requiring any doco changes regarding the described behavior then?

We changed so that retries/processing attempts would never cause a message to go to the DLQ. I.e., a message would only be dead-lettered if
- It was explicitly dead-lettered by the transport
- Something went terribly wrong and a message has gone through 2,147,483,647 number of retries w/o going into error queue.

PR #4271, opened 2019-02-18
Clarification of behavior for pubsub with MSMQ sender side distribution

ramonsmits:
Based on a livechat/email

### DIFF
--- a/transports/msmq/sender-side-distribution.md
+++ b/transports/msmq/sender-side-distribution.md
@@ -69,7 +69,7 @@ Each subscriber endpoint instance will at start send a subscription message to e
 
 Events:
 
-When an event is published the event will be send to only one of the endpoint instances. Which instance depends on the [distribution strategy](#message-distribution)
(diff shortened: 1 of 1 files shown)


transports/msmq/sender-side-distribution.md
```diff
@@ -61,6 +61,17 @@ partial: select-destination
...
+
+Scaled out publisher:
+
+Subscriptions are routed to all configured publisher instances. The subscriber endpoint will at start send a subscription message to each configured publisher instance.
+
+Scaled out subscriber:
+
+A scaled out subscriber will result in each subscriber instance to send a subscription message to the publisher. When an event is published the event will be send to only one of the instances.
+
+Info: An event will not be send to all instance of a logical endpoint
```
SzymonPobiega: In don't think we need this. The above paragraph already states the same.

ramonsmits: I've rewritten and shortened it.

PR #4272, opened 2019-02-18
Namespace instead of fully qualified name

ramonsmits:
The namespace must always be present AFAIK but the assembly does not. This depends I think on the configured serializer.

### DIFF
--- a/samples/azure-service-bus-netstandard/native-integration/sample.md
+++ b/samples/azure-service-bus-netstandard/native-integration/sample.md
@@ -46,7 +46,7 @@ For a native message to be processed, NServiceBus endpoints using the Azure Serv
 
 snippet: NecessaryHeaders
 
-NOTE: The `NServiceBus.EnclosedMessageTypes` property must contain the namespace and name of the type expected by the NServiceBus endpoint. Some serializers might also require the [Assembly Qualified Name](https://docs.microsoft.com/en-us/dotnet/api/system.type.assemblyqualifiedname) which also includes the assembly of the type.
+NOTE: The [`NServiceBus.EnclosedMessageTypes` header property](/nservicebus/messaging/headers#serialization-headers-nservicebus-enclosedmessagetypes) must contain the namespace and name of the type expected by the NServiceBus endpoint. Some serializers might also require the [Assembly Qualified Name](https://docs.microsoft.com/en-us/dotnet/api/system.type.assemblyqualifiedname) which also includes the assembly of the type.
 
 The message itself is defined using [conventions](/nservicebus/messaging/conventions.md) in the `Receiver` project.
 
--- a/samples/azure-service-bus-netstandard/native-integration/sample.md
+++ b/samples/azure-service-bus-netstandard/native-integration/sample.md
@@ -46,7 +46,7 @@ For a native message to be processed, NServiceBus endpoints using the Azure Serv
 
 snippet: NecessaryHeaders
 
-NOTE: The [`NServiceBus.EnclosedMessageTypes` header property](/nservicebus/messaging/headers.md#serialization-headers-nservicebus-enclosedmessagetypes) must contain the type FullName (namespace and name) expected by the NServiceBus endpoint. Not to be confused with [Assembly Qualified Name](https://docs.microsoft.com/en-us/dotnet/api/system.type.assemblyqualifiedname) which also includes the assembly and version of the type.
+NOTE: The [`NServiceBus.EnclosedMessageTypes` header property](/nservicebus/messaging/headers.md#serialization-headers-nservicebus-enclosedmessagetypes) must atleast contain the type FullName (namespace and name) expected by the NServiceBus endpoint but can contain the [Assembly Qualified Name](https://docs.microsoft.com/en-us/dotnet/api/system.type.assemblyqualifiedname) which also includes the assembly and version of the type. The Assembly Qualified Name is the default NServiceBus uses.
 
 The message itself is defined using [conventions](/nservicebus/messaging/conventions.md) in the `Receiver` project.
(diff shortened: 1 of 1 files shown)


ramonsmits:
Based on livechat about ASB native integration sample

samples/azure-service-bus-netstandard/native-integration/sample.md
```diff
@@ -46,7 +46,7 @@ For a native message to be processed, NServiceBus endpoints using the Azure Serv
 
 snippet: NecessaryHeaders
 
-NOTE: The `NServiceBus.EnclosedMessageTypes` property must contain the the fully-qualified name of the type expected by the NServiceBus endpoint.
+NOTE: The `NServiceBus.EnclosedMessageTypes` property must contain the namespace and name of the type expected by the NServiceBus endpoint. Some serializers might require also require the fully-qualified name including the assembly of the type.
```
andreasohlund: Namespace + Type is the fully qualified name

https://docs.microsoft.com/en-us/dotnet/api/system.type.fullname?view=netframework-4.7.2

When assembly is needed as well it’s called 

https://docs.microsoft.com/en-us/dotnet/api/system.type.assemblyqualifiedname?view=netframework-4.7.2


Lets use those names and link to the MS doco?

timbussmann: assembly qualified types would only be required if the message type hasn't already been loaded for some reason (unlikely unless you do weird stuff with inheritance)

ramonsmits: @timbussmann Could be, I don't know. Shouldn't scanning already loaded the assembly?

andreasohlund: @bording isn't this the stuff you investigated recently?

bording: This is definitely related to it, yes.

timbussmann: > Shouldn't scanning already loaded the assembly?

not necessarily in case you use unobtrusive message conventions

ramonsmits: @bording @andreasohlund @timbussmann So should it work without the assembly info? As stated, I've seen deserialization not working due to missing assembly info but it is too long ago to remember the context like which major core version and which serializer.

If you guys state that this is not required then I remove that last sentence.

andreasohlund: It definitely should work, eg. our Heartbeats use that trick for the heartbeat Message 
https://github.com/Particular/NServiceBus.Heartbeat/blob/develop/src/NServiceBus.Heartbeat/ServiceControlBackend.cs#L25

ramonsmits: @andreasohlund ok, I'll adjust

ramonsmits:
@andreasohlund I've made an adjustment based on the conversation. Please let me know if this is fine now.

samples/azure-service-bus-netstandard/native-integration/sample.md
```diff
@@ -46,7 +46,7 @@ For a native message to be processed, NServiceBus endpoints using the Azure Serv
 
 snippet: NecessaryHeaders
 
-NOTE: The `NServiceBus.EnclosedMessageTypes` property must contain the the fully-qualified name of the type expected by the NServiceBus endpoint.
+NOTE: The [`NServiceBus.EnclosedMessageTypes` header property](/nservicebus/messaging/headers.md#serialization-headers-nservicebus-enclosedmessagetypes) must contain the type FullName (namespace and name) expected by the NServiceBus endpoint. Not to be confused with [Assembly Qualified Name](https://docs.microsoft.com/en-us/dotnet/api/system.type.assemblyqualifiedname) which also includes the assembly and version of the type.
```
andreasohlund: It can contain both the FullName and the assembly qualified name, we handle both. When we send a message we ([IMO incorrectly](https://github.com/Particular/DeveloperExperience/issues/269)) use the Asm qualified name.

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Pipeline/Outgoing/SerializeMessageConnector.cs#L58

Does this make sense? (Then this need to be reworded)

ramonsmits: @andreasohlund ok, please review again. I reworded it that it atleast must contain the FullName, but that is can contain the Assembly Qualified Name.

Keep in mind that this page is about creating a native integration where a sender must prep the header values to be compatible with NServiceBus.

bording:
FYI I have pushed up an additional tweak/grammar pass to the text:

NOTE: The [`NServiceBus.EnclosedMessageTypes` header property](/nservicebus/messaging/headers.md#serialization-headers-nservicebus-enclosedmessagetypes) must at least contain the [FullName](https://docs.microsoft.com/en-us/dotnet/api/system.type.fullname) of the message type expected by the NServiceBus endpoint, but can instead contain the [AssemblyQualifiedName](https://docs.microsoft.com/en-us/dotnet/api/system.type.assemblyqualifiedname), which also includes the assembly name and version that the type was loaded from. The assembly-qualified name is the value NServiceBus puts into the header by default.

andreasohlund:
The page also needs to explain that message type can be infered from the
payload (eg json with $type). I’ll add that in a new PR

bording:
@andreasohlund Since this is just on a sample's page, I wonder if this should be somewhere else instead.

andreasohlund:
Yea, I will move all this to a dedicated page and link it. Stay tuned

PR #4277, opened 2019-02-22
Updates for SQS 4.3.0 release

kbaley:


transports/sqs/troubleshooting.md
```diff
@@ -0,0 +1,56 @@
...
+related:
+ - nservicebus/sqs/index
+ - samples/sqs/simple
+tags:
+- AWS
+---
+
+## Viewing the message Id on the SQS message
+
+To assist with troubleshooting, the NServiceBus message Id is propagated to the underlying SQS message as a message attribute. That way, even if message deserialization fails, the NServiceBus message Id is still available to help diagnose problems.
```
danielmarbach: Should we link to https://docs.particular.net/nservicebus/messaging/headers#messaging-interaction-headers-nservicebus-messageid or at least the message headers documentation? Should we align with the naming and use ID?

kbaley: I'll add the link. As for the naming convention, I did check this first. We have about 520 instances of "ID" in the docs and about 880 instances of "Id" which is why I went with the latter.

kbaley: Hmmm... just noticed that if I limit the search to .md files (as opposed to code files), the numbers switch: 128 for ID and 67 for Id. I'll switch it and happily too because I prefer the former.

PR #4278, opened 2019-02-23
Add Nancy netcore sample

mikedevbo:


samples/web/send-from-nancy-module/Core_7/WebApplication.Core/Startup.cs
```diff
@@ -0,0 +1,39 @@
...
+namespace WebApplication.Core
+{
+    public class Startup
+    {
+        public void Configure(IApplicationBuilder app, IApplicationLifetime applicationLifetime)
+        {
+			// Configure endpoint
+			var endpointConfiguration = new EndpointConfiguration("Samples.Nancy.Sender");
+			var transport = endpointConfiguration.UseTransport<LearningTransport>();
+			endpointConfiguration.UsePersistence<LearningPersistence>();
```
mauroservienti: Given that it’s a SendOnly endpoint there is need to define any persistence.

mauroservienti:
@andreasohlund this sample is `.NET Core` only. Do we want to have a full framework version as well as we do for all the others, if supported?

I'm fine with sticking with a `.NET Core` only version.

andreasohlund:
There is already a full FX version https://docs.particular.net/samples/web/send-from-nancy-module/

andreasohlund:
Thank you so much @mikedevbo 👍

PR #4286, opened 2019-03-06
Add note about duplicate entries in subscription queue

ramonsmits:
Based on forum post https://discuss.particular.net/t/msmq-persistence/1014/21

### DIFF
--- a/persistence/msmq/subscription.md
+++ b/persistence/msmq/subscription.md
@@ -7,6 +7,8 @@ related:
  - samples/msmq/persistence
 ---
 
+Note: The subscription queue can contain duplicate entries and is by design. This does not result in events to be published multiple times. After a publisher restarts the subscription queue state will be rewritten and deduplicated.
+
 partial: config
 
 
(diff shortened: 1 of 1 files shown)


persistence/msmq/subscription.md
```diff
@@ -7,7 +7,9 @@ related:
  - samples/msmq/persistence
 ---
 
+Note: The subscription queue can contain duplicate entries and is by design. This does not result in events to be published multiple times. After a publisher restarts the subscription queue state will be rewritten and deduplicated.
```
mauroservienti:
```suggestion
Note: The subscription queue can contain duplicate entries and is by design. This does not result in events to be published multiple times as publishers deduplicates at publishing time. Each time a publisher is restarted the subscription queue state will be rewritten and deduplicated.
```

ramonsmits: @mauroservienti it is incorrect, it doesn't dedupe during publishing, it dedupes during reading the subscriptions and when receiving a subscription request from a subscriber. Let me see if I can rewrite.

PR #4287, opened 2019-03-06
SQL Persistence multi-tenant doco

DavidBoike:
For changes in https://github.com/Particular/NServiceBus.Persistence.Sql/pull/337


persistence/sql/multi-tenant.md
```diff
@@ -0,0 +1,81 @@
...
+
+snippet: MultiTenantWithHeaderName
+
+In more complex situations, where the tenant id must be calculated by consulting multiple headers, or where a transition from an old header name to a new header name is occurring, a `Func` can be provided that captures the tenant id from the incoming message.
+
+In this example, the header `NewTenantHeaderName` is consulted first, with `OldTenantHeaderName` as a backup.
+
+snippet: MultiTenantWithFunc
+
+NOTE: A null tenant id probably indicates a failure to propagate the tenant id from an older message, rendering the message invalid. However, it is safe to return null in this `Func` because SQL Persistence will throw an exception if a null tenant id is encountered, rather than trying to process a message without a tenant id.
```
mauroservienti: Thinking out loud: is this always the case? Or is it safer to assume that in a multi-tenant system there might be messages that are not tenant-specific?

persistence/sql/multi-tenant.md
```diff
@@ -0,0 +1,81 @@
+---
+title: Multi-tenant support
+summary: Highlights differences in the database schema between the two persisters
```
SzymonPobiega: Not sure I understand that summary

persistence/sql/multi-tenant.md
```diff
@@ -0,0 +1,81 @@
...
+reviewed: 2019-03-06
+component: SqlPersistence
+versions: '[4.5,)'
+related:
+ - nservicebus/pipeline/manipulate-with-behaviors
+ - samples/multi-tenant/propagation
+---
+
+When working in a multi-tenant system, the data for each customer/client (tenant) is stored in an independent database identified by a message header passed along with each message flowing through the system. With SQL Persistence running in multi-tenant mode, [saga data](saga.md) and [outbox data](outbox.md) are both stored in the same database as the tenant data, requiring only one database connection and transaction for the duration of the message handler.
```
SzymonPobiega: `by a message header` -> `by a message header(s)`

persistence/sql/multi-tenant.md
```diff
@@ -0,0 +1,81 @@
...
+
+[Timeout data](timeouts.md) and [subscription data](subscriptions.md) do not belong to any specific tenant and are stored in a master database in cases where the message transport does not provide native timeouts or native publish/subscribe capabilities.
```
SzymonPobiega: `master` -> `shared`?

persistence/sql/multi-tenant.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Specifying connections per tenant
+
+If the tenant information is propagated in a single header that does not change, multi-tenancy can be enabled by specifying the header name and a `Func` to create a connection given a tenant id:
+
+snippet: MultiTenantWithHeaderName
+
+In more complex situations, where the tenant id must be calculated by consulting multiple headers, or where a transition from an old header name to a new header name is occurring, a `Func` can be provided that captures the tenant id from the incoming message.
```
SzymonPobiega: a `Func` -> a callback

persistence/sql/multi-tenant.md
SzymonPobiega: I would say that an endpoint is either in a multi-tenant mode and can only process multi-tenant messages or in a single tenant mode. I'd rather not mix these in a single endpoint.

Also I'd remove `probably` from the sentence above.

SzymonPobiega:
Nice! I like it @DavidBoike . Now also the multi tenant sample needs to be adjusted. Should we do it in this PR or do it after we release 4.5?

persistence/sql/multi-tenant.md
mauroservienti: Fair enough. One scenario, though, could be: if system is currently a home made multi tenant system as it was the only way to build it. Using a message property to transport the tenant id, there is no way to smoothly transition to a header based tenant id as all in flight messages will fail.

Not sure it’s a scenario we need to support. Thoughts?
-

SzymonPobiega: The connection for the outbox is needed way before the body is deserialized so we are only talking here about a scenario where the user does not use the outbox.

In that case the fist time the `ConnectionManager` is used to build the connection is when creating the Synchronized Storage Session right before executing a handler. In that case I think it should be possible to introduce a behavior in the incoming logical stage that gets the tenant ID from the deserialized message body and sticks it into a header of the `IncomingMessage` object.

mauroservienti: Like it 👍

DavidBoike:
@SzymonPobiega 	& @mauroservienti:

Updated the sample, and one thing I noticed was in https://github.com/Particular/docs.particular.net/blob/sqlp-multi-tenant/samples/multi-tenant/persistence-running.include.md it includes a warning:

> WARNING: Timeouts are stored in the shared database so make sure to not include any sensitive information. Keep such information in saga data and only use timeouts as notifications.

But this is a Learning Transport sample. In fact I ended up removing the shared database - there's no subscriptions, no timeouts, so it was getting created but no tables were.

It looks like this include is also used in the Raven and NH multi-tenant samples. Should it be refactored out of the include and put in both of those sample.md files instead?

BTW the integrity test failure is only because there's no prerelease.txt file on the project files using the prerelease package. But since we have no intention of merging this until after 4.5 is released and the PR is adjusted to use 4.* again.

By the way, I also tried (locally) to change the sample to MSMQ, and then I *did* have to have the shared database, configured by the normal ConnectionBuilder, and everything worked quite nicely!

DavidBoike:
I updated that warning to:

> WARNING: If used with a message transport that does not support native timeouts, timeout data is stored in a shared database so make sure to not include any sensitive information. Keep such information in saga data and only use timeouts as notifications.

Then it makes sense in all the samples.

This is ready. I say we release SqlP 4.5 on Monday, adjust the references here, and then merge.

SzymonPobiega:
Regarding that warning. I had some second thoughts about it when I initially worked on the samples. The thing is, if you are using a transport that does support the native timeouts, then they are also shared (in a shared queue in case of SQL Server transport) so it does not really matter.

Timeouts, as well as messages, are shared in all supported scenarios. Not sure if it is good or bad, though.

PR #4297, opened 2019-03-19
Add snippet and doco for discard recoverability action

HEskandari:
This would document how discard recoverability action can be used.

This won't compile until [this](https://github.com/Particular/NServiceBus/pull/5327/files) gets merged in.


nservicebus/recoverability/custom-recoverability-policy_fullconfig_core_[7,).partial.md
```diff
@@ -0,0 +1,7 @@
+The snippet below shows a fully custom policy that does the following:
+
+ * For unrecoverable exceptions such as `MyBusinessException` failed messages are immediately moved to a custom error queue. 
+ * For exceptions where retrying does not make sense such as `MyBusinessTimedOutException` failed message is *not* moved to an error queue and is discarded.
```
danielmarbach: should we change this a bit to

> an error queue but is discarded.

?

danielmarbach:
One comment. Other than that LGTM

HEskandari:
@DavidBoike Can you please have a look and see why this fails? Basically:

here’s what happens:

- The commit `cd0aca1` fails stating the snippet for v7.2 is not used in _some_ versions.
- I tried to mark 7.2 snippets specific to that version (by adding [7.2,) to the key) in commit `7a8ce35`, but then the build fails saying: Error: Could not use ‘[7.2,)’ since directory is flagged as Prerelease.

Also tried tweaking other things to no effect.

DavidBoike:
Tried some things with Daniel but then TeamCity started freaking out on us and wouldn't even attempt to build. 🤷‍♂️

DavidBoike:
Might feel like a hack but that's probably the "most right" (or at least the "least wrong") way to do it. With the new feature in 7.2, you could argue it's essentially a different snippet, and should have a different name, as it's not showing the exact same thing but just on different versions. Maybe?

In any case, good work!

PR #4298, opened 2019-03-19
Document access to RabbitMQ message details

andreasohlund:


andreasohlund:
Will make this a real PR since TC doesn't like drafts

PR #4301, opened 2019-03-21
Doco on how to access native RabbitMQ message details

andreasohlund:


andreasohlund:
@bording I've tweaked the sample to demo access to the native message. This should be ready for final review now

bording:
Changes LGTM, but we should get the new transport version on MyGet so we can smoke test the sample before merging.

PR #4305, opened 2019-03-22
Update connection string statement

WilliamBZA:


bording:
TeamCity seems to have decided not report status back to this PR for some mysterious reason, but the build is green if you go look, so I'm going to merge this.

PR #4306, opened 2019-03-22
Allow package versions >9

timbussmann:
investigating why the build fails on this PR: https://github.com/Particular/docs.particular.net/pull/4302 it seems the regex can't match for packages with version number > 9. With my limited regex skills, my only idea was to allow multiple digits per version segment. 🤷‍♂️

### DIFF
--- a/tests/IntegrityTests/ValidatePrereleaseTxt.cs
+++ b/tests/IntegrityTests/ValidatePrereleaseTxt.cs
@@ -74,7 +74,7 @@ namespace IntegrityTests
             while (dirPath.Length >= TestSetup.DocsRootPath.Length)
             {
                 string dirName = Path.GetFileName(dirPath);
-                if (Regex.IsMatch(dirName, @"_(All|\d(\.\d)?)$"))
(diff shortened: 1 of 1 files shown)


tests/IntegrityTests/ValidatePrereleaseTxt.cs
```diff
@@ -74,7 +74,7 @@ public void VerifyProjectUsingPrereleaseComponentMarkedWithPrereleaseTxt()
             while (dirPath.Length >= TestSetup.DocsRootPath.Length)
             {
                 string dirName = Path.GetFileName(dirPath);
-                if (Regex.IsMatch(dirName, @"_(All|\d(\.\d)?)$"))
+                if (Regex.IsMatch(dirName, @"_(All|\d+(\.\d+)?)$"))
```
timbussmann: not really understanding too much about this code tbh but I guess this does skip patch versions intentionally?

DavidBoike: We don't (and shouldn't) create versioned directories for patch releases.

PR #4313, opened 2019-03-27
Updates to peek SQL for SQL transport

kbaley:
Relates to https://github.com/Particular/NServiceBus.SqlServer/issues/481


transports/sql/sql-statements.md
```diff
@@ -53,6 +53,12 @@ Checks if there are messages in the queue.
 
 snippet: PeekTextSql
 
+Prior to version 4.1.2 the following SQL was used instead: 
+
+snippet: PeekTextSql411
+
+NOTE: Due to performance issues in the earlier SQL, it is recommended that users update to version 4.1.2 or higher to avoid them.
+
```
DavidBoike: You'll have to put this section in a partial so that it only appears for the correct version - that's what's causing the Missing Snippet error on the build server now.

PR #4320, opened 2019-04-02
Inline code snippets for old migration guides

timbussmann:
* [x] 3to4
* [x] 4to5


timbussmann:
should be ready now @DavidBoike

PR #4322, opened 2019-04-03
fix string comparison in Installers sample code

jjanuszkiewicz:
Fix the comparison, as it always evaluated to false.
Also use OrdinalIgnoreCase comparison instead of culture-specific ToLower().

### DIFF
--- a/Snippets/Core/Core_5/Installers.cs
+++ b/Snippets/Core/Core_5/Installers.cs
@@ -28,7 +28,7 @@ namespace Core5
         {
             #region InstallersRunWhenNecessaryCommandLine
                 
-            var runInstallers = Environment.GetCommandLineArgs().Any(x => x.ToLower() == "/runInstallers");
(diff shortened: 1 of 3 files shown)


jjanuszkiewicz:
I can see that the same code exists in the samples for other NSB versions, it would be great if someone fixed that as well.

jjanuszkiewicz:
Fixed in examples for other NSB versions as well.

dvdstelt:
Not sure how we missed that for quite a while! :)

There are conflicting files atm though. We _just removed_ all v3 and v4 snippets. Could you remove those commits, so we can possibly merge this?

jjanuszkiewicz:
No problem, remove those 2 files.

dvdstelt:
Thanks, @jjanuszkiewicz.
I could've done it myself, but I was 'out of the office' and disconnected for a few days.

PR #4329, opened 2019-04-08
Inline unnecessary partials

timbussmann:
inline partials which are no longer necessary because they are no longer needed due to v3/v4 docs being removed.


nservicebus/hosting/critical-errors.md
```diff
@@ -20,11 +20,6 @@ NServiceBus has built-in [recoverability](/nservicebus/recoverability/) but in c
 partial: default
 
 
-### Logging of critical errors
```
timbussmann: I removed this section as the statements don't seem to be true (the partial states that logging is done as part of the default action and users would have to make sure they log in case they overwrite the default critical error action. NSB does always log though, even when using a custom critical error action)

PR #4333, opened 2019-04-11
Added warning that disabling auditing impacts the platform

ramonsmits:


nservicebus/operations/auditing.md
```diff
@@ -13,11 +13,13 @@ redirects:
  - nservicebus/auditing-with-nservicebus
 ---
 
+WARNING: Disabling auditing impacts the functionality of the platform tools. ServicePulse can work without auditing enabled but will show incorrect retry states on messages, and ServiceInsight will only show failures instead of complete conversations.
```
boblangley:
```suggestion
NOTE: Auditing is required for optimum functionality of the platform tools. ServicePulse can work without auditing enabled but will show incorrect retry states on messages, and ServiceInsight will only show failures instead of complete conversations.
```

I would also put this below before Message headers, somewhere near line 24, instead of at the top. Enabling auditing also has a performance impact on ServiceControl which sometimes means we recommend not auditing some of your endpoints.

ramonsmits:
@boblangley I added a section on the performance impact

PR #4341, opened 2019-04-17
Added warning making sure to setup/init logging before resolving logger

ramonsmits:
Based on live chat question

### DIFF
--- a/nservicebus/logging/usage.md
+++ b/nservicebus/logging/usage.md
@@ -12,7 +12,7 @@ Set up a single static field to an `ILog` in the classes, and then use it in all
 
 snippet: UsingLogging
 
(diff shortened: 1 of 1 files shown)


nservicebus/logging/usage.md
```diff
@@ -12,6 +12,8 @@ Set up a single static field to an `ILog` in the classes, and then use it in all
 
 snippet: UsingLogging
 
+WARNING: Make sure that logging is correctly setup/initialized before resolving a logger instance. Not doing this can result a logger using an incorrect configuration
```
kbaley:
```suggestion
WARNING: Make sure that logging is correctly initialized before resolving the `ILog` instance. Not doing so can result in a logger using an incorrect configuration
```

ramonsmits: @kbaley Its a common problem, but that doesn't mean we should't warn them. Its based on support and this isn't the first time I had this issue.I like to see it added.

PR #4352, opened 2019-04-30
Spell out SqlP in the EF integration sample for better SEO

andreasohlund:


andreasohlund:
Only the CLA acting up since this was raised before this PR was raised. Will digg into that. Using admin rights to merge for now

PR #4357, opened 2019-05-06
not all sagas can be correlated by >1 attribute

mauroservienti:
Makes clear that is the saga in this specific sample that is correlated by more than 1 attribute. I read the original sentence:

> A saga can be correlated with messages using more than one property

as: sagas in general can be correlated by multiple properties. But this is not the case. What I'm trying to make clear is that it's the saga in this specific sample that needs to be correlated by multiple properties.


bording:
@mauroservienti Looks like someone else touched these files. Conflicts!

samples/saga/sagafinder-process.include.md
```diff
@@ -1 +1 @@
-It is important to note that the saga is not sending the order ID to the payment processor. Instead, it is sending a payment transaction ID. A saga can be correlated with messages using more than one property. For example, `OrderId` and `PaymentTransactionId`. This requires both of these properties to be treated as unique.
\ No newline at end of file
+It is important to note that the saga is not sending the order ID to the payment processor. Instead, it is sending a payment transaction ID. This saga needs to be correlated by more than one attribute. For example, `OrderId` and `PaymentTransactionId`. This requires both of these properties to be treated as unique.
```
bording: There is a mix of "attribute" and "property" here. What term do we use for these things in the rest of the saga documentation? We should be consistent.

mauroservienti: property is the correct one. Fixed.

udidahan:
I haven't looked at the code but this sounds like a case where the saga does simple request/response with another endpoint, is that correct? If so, NServiceBus handles the correlation automatically not requiring any saga finding logic.

PR #4360, opened 2019-05-08
Move interface operations to the interface page

andreasohlund:
Also show how `IMessageCreator` is used

Related to https://github.com/Particular/NServiceBus/pull/5363

Replaces https://github.com/Particular/docs.particular.net/pull/4359


bording:
@andreasohlund I pushed up some revisions, including some grammar cleanup that wasn't related to changes in this PR.

nservicebus/sagas/reply-replytooriginator-differences.md
```diff
@@ -16,7 +16,7 @@ The following diagram details a scenario where two sagas and an integration endp
 
 ![Sample sequence diagram](reply-replytooriginator-differences.png)
 
-The reason a call to `Reply<ShipOrder>()` sends a message to the `Shipment Gateway` is that it is invoked in the context of handling the `ShipmentReserved` message, and the return address of `ShipmentReserved` is `Shipment Gateway`.
+The reason a call to `Reply(new ShipOrder())` sends a message to the `Shipment Gateway` is that it is invoked in the context of handling the `ShipmentReserved` message, and the return address of `ShipmentReserved` is `Shipment Gateway`.
```
bording: There were a couple places where you made changes from something like `Reply<ShipOrder>()` to `Reply(new ShipOrder())`. Why make this change?

andreasohlund:
To not promote the lambda syntax since it should only be used when emitting
interfaces, ie an advanced use case. Does that make sense?

PR #4381, opened 2019-05-29
Remove struct limitation

andreasohlund:
Figured that making this version dependant is overkill?


andreasohlund:
@bording good point, I put the note back with some rewording, please take a look?

PR #4383, opened 2019-05-30
RavenDB 4 support

DavidBoike:


DavidBoike:
@mauroservienti 	& @WojcikMike 	please review?

Merging would create a prerelease snippets version for RavenP 6 (Raven4) on the site targeting the beta that's currently on MyGet, so I consider it non-WIP.

Snippets/Raven/Raven.sln
```diff
@@ -12,21 +10,34 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Raven_All", "Raven_All\Rave
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Raven_5", "Raven_5\Raven_5.csproj", "{A5FA02A8-8FB0-4FCA-A8A7-57E2303962B5}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Raven_6", "Raven_6\Raven_6.csproj", "{495186E6-37F3-4F3C-BFC7-12C3478113D9}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
+		Release|Any CPU = Release|Any CPU
```
mauroservienti: Guess the release configuration is not needed

Snippets/Raven/Raven_6/MultiTenantSupport.cs
```diff
@@ -0,0 +1,19 @@
+﻿using NServiceBus;
+
+namespace Raven_5
```
WojcikMike: Is there any reason why Snippets for our new package RavenDB v6 that supports RavenDB 4 uses the namespace Raven_5 as a name?

persistence/upgrades/ravendb-5to6.md
```diff
@@ -0,0 +1,174 @@
...
+isUpgradeGuide: true
+upgradeGuideCoreVersions:
+ - 7
+---
+
+This update supports upgrading to RavenDB 4.2 while using NServiceBus 7.x. The upgrade to RavenDB 4 is a **major** upgrade. It's not backward compatible, and the [API has a lot of breaking changes](https://ravendb.net/docs/article-page/4.2/csharp/migration/client-api/introduction). Additionally, the data storage format has changed, so databases need to have their [data migrated to a new RavenDB 4 database](https://ravendb.net/docs/article-page/4.2/csharp/migration/server/data-migration).
+
+## Required RavenDB client
+
+RavenDB Persistence requires [RavenDB.Client 4.2.0](https://www.nuget.org/packages/RavenDB.Client/4.2.0) or later.
```
WojcikMike: Do we want to comment why are we starting support from 4.2.0 version? Some people might ask that question.

Given that we don't plan on supporting multi-master clusters I don't think we have a one?

DavidBoike: Well, we _could_ in a minor without then requiring a server/client upgrade as well. Also you can go straight from 3.5 to 4.2 so....why wouldn't you?

But really it's mostly because that's what we started with. What do you think?

Snippets/Raven/Raven_6/MultiTenantSupport.cs
DavidBoike: That's a  whoopsie. I'll fix.

persistence/upgrades/ravendb-5to6.md
WojcikMike: Yesterday I had a look at this page: https://ravendb.net/support and I learned that 4.2 is the only version with LTS. 4.0 is already expired, 4.1 will expire in 5 months. So I am retracting my question, it seems like 4.2 is the first stable and supported version from the 4.X familty.

PR #4385, opened 2019-06-07
Update NServiceBus overview page

kbaley:


platform/index.md
```diff
@@ -1,6 +1,7 @@
 ---
 title: The Particular Service Platform
-reviewed: 2018-12-31
+reviewed: 2019-06-07
+suppressRelated: true
```
adamralph: What does this do?

platform/nservicebus.include.md
```diff
@@ -1,5 +1,68 @@
...
+NServiceBus is the heart of a distributed system and the Particular Service Platform. It helps create systems that are scalable, reliable, and flexible.
 
-Most queuing technologies try to guarantee 'at least once', or even 'exactly once', message delivery. They usually fall short of this promise. NServiceBus solves intermittent delivery problems by  automatically retrying messages using various strategies. If all else fails, it forwards messages to an error queue. From there, [ServiceControl](/servicecontrol/) and [ServicePulse](/serviceinsight/) expose the problems for human intervention.
+At its core, NServiceBus works by routing _messages_ between _endpoints_. [Messages](/nservicebus/concept-overview.md#message) are plain C# classes that contain meaningful data for the business process that is being modeled.
 
-NServiceBus is modular in design and highly extensible. There are many options to choose from for queueing, storage, serialization, logging, and more. Many elements of the system can be replaced with custom implementations for specific scenarios.
+```csharp
+public class ProcessOrder : IMessage
```
adamralph: I would consider showing unobtrusive here, which should probably be more strongly advertised as the default throughout our docs. I know for a fact that some developers find inheriting from a base type for a message highly off-putting, and it may not be obvious that it's optional.

```suggestion
public class ProcessOrder
```

platform/nservicebus.include.md
```diff
@@ -1,5 +1,68 @@
...
+}
+```
+
+Endpoints can be running in different processes, on different machines, even at different times. NServiceBus makes sure that each message reaches its intended destination and is processed.
+
+NServiceBus accomplishes this by providing an abstraction over [existing queuing technologies](/transports/). While it's possible to work directly with queuing systems, NServiceBus provides extra features to make applications more reliable and scalable.
+
+## Reliable
+
+NServiceBus offers different ways of ensuring information is not lost due to failures in a system. First, NServiceBus provides native transaction support (e.g. DTC) for the underlying queuing technologies that support it, as well as its own transaction guarantees through the implementation of the [Outbox pattern](/nservicebus/outbox).
```
adamralph: > e.g. DTC

This may not be the best thing to mention here. Ultimately, NServiceBus will use transactions appropriately, and in many cases they do not elevate to the DTC. E.g. when using a SQL DB for transport, persistence, and business data. Those that do elevate to the DTC would probably only involve MSMQ and SQLT, which is a scenario which is probably diminishing in importance.

platform/nservicebus.include.md
```diff
@@ -1,5 +1,68 @@
...
+
+In other cases, NServiceBus has [built-in recoverability](/nservicebus/recoverability) that can automatically adapt to common failures in a system. For intermittent failures, such as network outages, messages can be retried at regular intervals. For more serious errors, messages are set aside in a separate error queue so that they can be investigated at a later time without impacting the overall performance of the system.
+
+## Scalable
+
+NServiceBus is designed to handle a large number of messages. Endpoints are configured for high performance by default, handling multiple messages in parallel. Depending on the workload, the number of messages handled concurrently can be [increased to improve message throughput](/nservicebus/operations/tuning.md).
+
+In high volume scenarios, where there are more messages being produced than a single endpoint can handle, each endpoint can be [scaled out across multiple instances](/transports/scale-out.md) running on different machines, sharing the load.
```
adamralph:
```suggestion
In high volume scenarios, where there are more messages being produced than a single physical endpoint can handle, the logical endpoint can be [scaled out across multiple physical instances](/transports/scale-out.md) running on different machines, sharing the load. 
```

platform/nservicebus.include.md
```diff
@@ -1,5 +1,68 @@
...
+
+Each endpoint tracks [key performance metrics](/monitoring/metrics/definitions.md) that can be exposed as [Windows Performance Counters](/monitoring/metrics/performance-counters.md) or [collected into a central dashboard](/monitoring/metrics/in-servicepulse.md). [The monitoring demo](/tutorials/monitoring-demo/) demonstrates how to find performance bottlenecks and identify endpoints that are ready to scale out.
+
+## Simple and testable
+
+NServiceBus is designed with simplicity in mind. Message handlers don't need additional code to manage logging, exception handling, serialization, transactions, or the specifics of a queueing technology. This allows message handler code to focus on business logic.
+
+Long-running business workflows can be modeled in NServiceBus using a [saga](/nservicebus/sagas/). A saga is a C# class that handles a number of different messages over time, persisting its state between each step in the workflow. NServiceBus makes sure that each saga instance only processes a single message at a time, keeping its internal state consistent. The [NServiceBus sagas: Getting started](/nservicebus/sagas/) tutorial provides more details.
```
adamralph:
```suggestion
Long-running business workflows can be modeled in NServiceBus using [sagas](/nservicebus/sagas/). A saga is a C# class that handles a number of different messages over time, persisting its state between each step in the workflow. NServiceBus makes sure that each saga instance only processes a single message at a time, keeping its internal state consistent. The [NServiceBus sagas: Getting started](/nservicebus/sagas/) tutorial provides more details.
```

platform/nservicebus.include.md
```diff
@@ -1,5 +1,68 @@
...
+
+Message handlers and sagas can be [tested in isolation](/nservicebus/testing/). Simulating an incoming message is as simple as creating a new C# message object and passing it to the appropriate handler or saga. The framework includes a suite of testing tools that capture the behavior of message handlers and sagas under test, allowing assertions to be made about them.
+
+## Flexible
+
+NServiceBus endpoints can be hosted anywhere code can be executed, such as in a Windows Service, a Docker container, or in the cloud with Azure or AWS. NServiceBus is compiled against .NET Standard 2.0 and endpoints can be run on [a variety of platforms](/nservicebus/upgrades/supported-platforms.md).  
+
+NServiceBus works with many different technology stacks, offering choices for [transport](/transports/) and [persistence](/persistence/). Out of the box, defaults are provided for [serialization](/nservicebus/serialization/), [dependency injection](/nservicebus/dependency-injection/), and [logging](/nservicebus/logging/). These defaults can be overridden if a sepcific technology is required.
```
adamralph:
```suggestion
NServiceBus works with many different technology stacks, offering choices for [transport](/transports/) and [persistence](/persistence/). Out of the box, defaults are provided for [serialization](/nservicebus/serialization/), [dependency injection](/nservicebus/dependency-injection/), and [logging](/nservicebus/logging/). These defaults can be overridden if a specific technology is required. 
```

platform/nservicebus.include.md
```diff
@@ -1,5 +1,68 @@
...
+
+When a message fails to be processed, even after a number of retry strategies have been attempted, NServiceBus will forward the message to an [error queue](/nservicebus/recoverability/configure-error-handling.md) for manual investigation. Messages sent to the error queue are instrumented with headers containing details about the failure including a full exception stack trace. ServiceControl picks up messages from the error queue and makes them availible in [ServicePulse](/servicepulse/). Once the root cause of the failure has been found and corrected, all messages caused by the same problem can be retried at once.
```
adamralph:
```suggestion
When a message fails to be processed, even after a number of retry strategies have been attempted, NServiceBus will forward the message to an [error queue](/nservicebus/recoverability/configure-error-handling.md) for manual investigation. Messages sent to the error queue are instrumented with headers containing details about the failure including a full exception stack trace. ServiceControl picks up messages from the error queue and makes them available in [ServicePulse](/servicepulse/). Once the root cause of the failure has been found and corrected, all messages caused by the same problem can be retried at once. 
```

platform/index.md
DavidBoike: It suppresses the default list of "Related Articles" formed by directory traversal that makes up the majority of https://docs.particular.net/nservicebus/ right now. On a page like this, that's what the menu is for.

platform/nservicebus.include.md
```diff
@@ -1,5 +1,68 @@
...
+NServiceBus is the heart of a distributed system and the Particular Service Platform. It helps create systems that are scalable, reliable, and flexible.
 
-Most queuing technologies try to guarantee 'at least once', or even 'exactly once', message delivery. They usually fall short of this promise. NServiceBus solves intermittent delivery problems by  automatically retrying messages using various strategies. If all else fails, it forwards messages to an error queue. From there, [ServiceControl](/servicecontrol/) and [ServicePulse](/serviceinsight/) expose the problems for human intervention.
+At its core, NServiceBus works by routing _messages_ between _endpoints_. [Messages](/nservicebus/concept-overview.md#message) are plain C# classes that contain meaningful data for the business process that is being modeled.
```
YifatS: Do we need to provide a link to explain what is endpoint?

mikeminutillo: There is one at the top of the next paragraph. The idea was to define a message, show one, then define an endpoint, and show one.

PR #4393, opened 2019-06-17
MS DI: configure a custom shared container

mauroservienti:


mauroservienti:
Ah, I’ll fix it. I wrote it from the web UI.

PR #4394, opened 2019-06-19
Platform Sample description and installation guidance

dvdstelt:


platform/installation.md
```diff
@@ -0,0 +1,59 @@
...
+title: Platform Installation
+summary: Guidance on the different ways to install and use the platform
+reviewed: 2019-06-15
+---
+
+The Particular Service Platform can be used in different ways.
+
+- Monitoring production systems
+- Debugging during development
+- Presenting from within Visual Studio
```
DavidBoike: Presenting??? What does that mean? Before scrolling down I have no idea.

I'm not sure if this enumeration is really adding any value though. What about replacing the entire intro with:

> The Particular Service Platform can either be installed to a server for use with production systems, or on a workstation during development. In addition, a portable version is available for use in samples or demonstrations.

platform/installation.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Monitoring production systems
+
+The Platform Installers requires internet connectivity and performs several actions like verifying MSMQ and MSDTC installation. It is possible to [achieve the same result](/platform/installer/offline.md) without requiring internet connectivity.
```
DavidBoike:
```suggestion
The Platform Installer requires Internet connectivity. It is possible to [achieve the same result](/platform/installer/offline.md) without requiring internet connectivity.
```

DavidBoike: I don't believe it's necessary to muddy the waters about MSMQ stuff since it's totally optional.

platform/installation.md
```diff
@@ -0,0 +1,59 @@
...
+
+For usage on production environments it is possible and recommended to install the different components individually. Each component has individual installation instructions.
+
+### ServiceControl
```
DavidBoike: I think this is too much information to split up.

I would not link to the GitHub releases. In fact, for Monitoring the releases/latest doesn't even include compiled artifacts, only source zips. 

It would probably be best to only provide https://particular.net/downloads as a source of downloading any of this stuff, as it already takes the most recent version into account.

I really don't think it's necessary to do anything except say there's the 1) platform installer, 2) instructions for offline, and 3) individual installers on the downloads page.

platform/installation.md
```diff
@@ -0,0 +1,59 @@
...
+
+The latest download is [available on GitHub here](https://github.com/Particular/serviceinsight/releases/latest).
+
+ServiceInsight has the ability to connect to multiple ServiceControl instances using the [Endpoint Explorer](/serviceinsight/#endpoint-explorer).
+
+## Debugging during development
+
+For optimal installation of the Platform on a development machine the [Platform Installer](/platform/installer/) can be used.
+
+## Presenting from within Visual Studio
```
DavidBoike: I'd call this "portable mode" as we've done on samples/tutorials where we've used the platform sample

platform/installation.md
```diff
@@ -0,0 +1,59 @@
...
+
+The Service Platform has the ability to be hosted from within Visual Studio for the benefit of not having to install anything. This can be useful during presentations, to provide samples to others or to demonstrate the usage of the platform from within an existing solution. For this feature the Platform Package has been developed.
+
+There are additional [installation instructions](/platform/platform-sample-package.md) for the Platform Package.
```
DavidBoike: Prefer:

```md
For more information, see [Platform Sample Installation](/platform/platform-sample-package.md).
```

platform/nservicebus.include.md
```diff
@@ -66,3 +66,5 @@ NServiceBus is designed to work with the rest of the Particular Service Platform
...
 
 Additionally, each endpoint can send [heartbeat](/monitoring/heartbeats/), [health check](/monitoring/custom-checks/), and [performance metrics](/monitoring/metrics/) through the platform for visualization in ServicePulse, making it easy to see which endpoints are offline, which are ready to scale out, and which require manual intervention.
+
+The [real-time monitoring demo](https://particular.net/real-time-monitoring) provides the ability to experience the Service Platform in action. The [Platform Sample](/platform/platform-sample-package.md) provides the ability to demonstrate the Service Platform from within any Visual Studio solution, without the need to install anything.
```
DavidBoike:
```suggestion
The [real-time monitoring demo](https://particular.net/real-time-monitoring) provides the ability to experience the Service Platform in action. The [Platform Sample package](/platform/platform-sample-package.md) provides the ability to demonstrate the Service Platform from within any Visual Studio solution, without the need to install anything. 
```

Since there are more than one platform samples, need to add "package" as the noun.

platform/platform-sample-package.md
```diff
@@ -0,0 +1,67 @@
+---
+title: Platform Sample Installation
+summary: Installation instructions for using the platform sample for demo purposes.
+component: PlatformSample
+reviewed: 2019-06-16
+---
+
+The Platform Sample can be used to demonstrate the platform without actually fully installing it. This is the easiest way to get some experience to learn the Particular Service Platform. It is also possible to include and show the platform in presentations or samples. It is also possible to (temporarily) include it in a custom Visual Studio solution for demonstration purposes.
```
DavidBoike:
```suggestion
The [Platform Sample package](https://www.nuget.org/packages/Particular.PlatformSample) is a portable version of the Particular Service Platform which can be used to demonstrate the platform without installing it. This is the easiest way to experience the Particular Service Platform. It is also possible to include and show the platform in demos or samples.
```

platform/platform-sample-package.md
```diff
@@ -0,0 +1,67 @@
...
+
+DANGER: The Platform Sample is not for production usage and should only be used for demonstrating it during presentations or in samples.
```
DavidBoike:
```suggestion
WARNING: The Platform Sample is not for production usage and should only be used for demonstrating it during presentations or in samples.
```

platform/platform-sample-package.md
```diff
@@ -0,0 +1,67 @@
...
+
+The Platform Sample package includes the binaries of ServiceControl and ServiceControl.Monitoring and a self-hosted version of ServicePulse. All are configured to use the LearningTransport and don't support any other transports.
```
DavidBoike:
```suggestion
The Platform Sample package includes ServiceControl, ServiceControl.Monitoring, and ServicePulse. All are configured to use the [Learning Transport](/transports/learning/) and don't support any other transports.
```

platform/platform-sample-package.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Sample
+
+The [Message Replay Tutorial](/tutorials/message-replay) is a sample that includes the Platform Sample and is used to view and retry failed messages. It can be used to verify how the Platform Sample works from within an existing Visual Studio solution.
+
+To have ServiceInsight use the sample, [download](https://github.com/Particular/serviceinsight/releases/latest) and install ServiceInsight. Then connect to `http://localhost:49200/api` from the [Endpoint Explorer](/serviceinsight/#endpoint-explorer) to connect to the running Service Control.
+
+Note: If port `49200` is in use by another process, the Platform Sample console application will start by outputting the used port.
```
DavidBoike: It just finds a range of ports. If 49199 alone is taken, then 49200 will still not be what you expect. Might be better to say to look through the output and note the correct port.

platform/platform-sample-package.md
```diff
@@ -0,0 +1,67 @@
...
+reviewed: 2019-06-16
+---
+
+The Platform Sample can be used to demonstrate the platform without actually fully installing it. This is the easiest way to get some experience to learn the Particular Service Platform. It is also possible to include and show the platform in presentations or samples. It is also possible to (temporarily) include it in a custom Visual Studio solution for demonstration purposes.
+
+DANGER: The Platform Sample is not for production usage and should only be used for demonstrating it during presentations or in samples.
+
+The Platform Sample package includes the binaries of ServiceControl and ServiceControl.Monitoring and a self-hosted version of ServicePulse. All are configured to use the LearningTransport and don't support any other transports.
+
+## Sample
```
DavidBoike: Shouldn't this section come after installation? Seems like that should be the focus of this article.

platform/platform-sample-package.md
```diff
@@ -0,0 +1,67 @@
...
+
+The [Message Replay Tutorial](/tutorials/message-replay) is a sample that includes the Platform Sample and is used to view and retry failed messages. It can be used to verify how the Platform Sample works from within an existing Visual Studio solution.
+
+To have ServiceInsight use the sample, [download](https://github.com/Particular/serviceinsight/releases/latest) and install ServiceInsight. Then connect to `http://localhost:49200/api` from the [Endpoint Explorer](/serviceinsight/#endpoint-explorer) to connect to the running Service Control.
+
+Note: If port `49200` is in use by another process, the Platform Sample console application will start by outputting the used port.
+
+## Installation
+
+Within an existing Visual Studio solution a new console application needs to be added. Within that project, the following steps need to be executed:
```
DavidBoike: You hid instructions in the preface. Creating a console app should be one of the numbered instructions.

platform/platform-sample-package.md
```diff
@@ -0,0 +1,67 @@
...
+
+1. Add a reference to the `Particular.PlatformSample` NuGet package.
+1. Add the following code
+
+```CSharp
```
DavidBoike:
```suggestion
```cs
```

platform/platform-sample-package.md
```diff
@@ -0,0 +1,67 @@
...
+{
+    Particular.PlatformLauncher.Launch();
+}
+```
+
+The package will perform the following actions upon execution:
+
+1. Copy platform binaries into the project output directory.
+1. Find available ports for all the platform tools.
+1. Find .learningtransport directory and folders for logging.
```
DavidBoike: Is all this detail necessary?

platform/platform-sample-package.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Options
+
+The `PlatformLauncher.Launch()` API has very limited options available as optional parameters, which can be mixed as needed.
+
+### Showing console output
+
+By default the console outputs of ServiceControl and ServiceControl.Monitoring are suppressed. To view them for purposes of debugging or curiosity, use the `showPlatformToolConsoleOutput` parameter:
+
+```CSharp
```
DavidBoike:
```suggestion
```cs
```

platform/platform-sample-package.md
```diff
@@ -0,0 +1,67 @@
...
+Particular.PlatformLauncher.Launch(showPlatformToolConsoleOutput: true);
```
DavidBoike: Also should be a legit snippet.

platform/platform-sample-package.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Installation
+
+Within an existing Visual Studio solution a new console application needs to be added. Within that project, the following steps need to be executed:
+
+1. Add a reference to the `Particular.PlatformSample` NuGet package.
+1. Add the following code
+
+```CSharp
+static void Main(string[] args)
```
DavidBoike: You made a component, right? Shouldn't this be a legit snippet? Otherwise if we change the API we won't break anything.

(Granted the project probably needs APIApprovals and doesn't have it, but still...)

platform/platform-sample-package.md
```diff
@@ -0,0 +1,67 @@
...
+```
+
+### ServicePulse default route
+
+Some samples benefit from opening ServicePulse to a specific view, rather than the Dashboard.
+
+For example, to open ServicePulse to the Monitoring view:
+
+```CSharp
+Particular.PlatformLauncher.Launch(servicePulseDefaultRoute: "/monitoring");
```
DavidBoike: legit snippet

servicepulse/host-config.md
```diff
@@ -23,11 +23,14 @@ The default value for ImagePath is:
 
 Change the value after `--url=` and restart the service.
 
+As ServicePulse is installed as a Windows Servce, optionally multiple ServicePulse instances can be installed. The default ports can be changed by providing [installation parameters](/servicepulse/installation.md#installation-available-installation-parameters).
 
 ## Default connection to ServiceControl and ServiceControl Monitoring
 
 ServicePulse will by default, attempt to connect to ServiceControl and ServiceControl Monitoring using the URLs `http://localhost:33333/api/` and `http://localhost:33633/`, respectively.
 
+The configuration is stored inside a javascript file located in the folder ServicePulse is installed in. By default this is `C:\Program Files (x86)\Particular Software\ServicePulse\app\js`.
```
DavidBoike: Isn't this app.constants.js - shouldn't you mention the file name?

platform/platform-sample-package.md
dvdstelt: The plan was to use an existing sample (in my case, even a tutorial, although the same code is probably elsewhere too). It's only a single line of code. And if I'm correct, I cannot reuse snippets from a global location or some other sample/tutorial, correct?

platform/platform-sample-package.md
dvdstelt: Just copied this from the GitHub repo, but I guess not :-)

PR #4397, opened 2019-06-21
Transports features summary

mauroservienti:


boblangley:
I like the idea. Perhaps the main transports page could have this as well with all transports? or is that too much?

mauroservienti:
Maybe all transports by default is too much, but we could have a little "compare" javascript widget that allows to select one or more transports and builds the resulting table, so that people can exclude/include based on their needs.

mauroservienti:
>What doesn't seem to be related to this table is the Scripted deployment.

Agreed. I just put together something to not simply write bla bla. Or maybe the table should not be named features, so that it can have a broader scope.

danielmarbach:
Love this. When will it be done? 🚎

mauroservienti:
If we can agree on the table content, it's an easy addition to all transports and can be done in less than 30 minutes.

I think having `Scripted Deployment` and `Installers` in the table makes sense, even if they are not a transport feature. How can we rename the table column header and the section title to fit?

danielmarbach:
I would go for something like this

|Feature                    |   |  
|:---                       |---
|Transactions |None, ReceiveOnly (Message visibility timeout)
|Pub/Sub                    |Message driven
|Timeouts                   |Native (Requires FIFO Queues)
|Large message bodies       |Native (using S3)
|Scale-out             |Competing consumer
|Scripted Deployment        |Powershell, Terraform, C#
|Installers                 |Optional

danielmarbach:
Me likey likey

mauroservienti:
added the same table for RabbitMQ

transports/rabbitmq/index.md
```diff
@@ -17,6 +17,17 @@ Provides support for sending messages over [RabbitMQ](https://www.rabbitmq.com/)
 
 WARNING: The transport is not compatible with RabbitMQ broker version 3.3.X and below.
 
+## Transport facts
```
danielmarbach:
```suggestion
## Transport at a glance
```

transports/sqs/index.md
```diff
@@ -13,6 +13,18 @@ redirects:
 
 [Simple Queue Service (SQS)](https://aws.amazon.com/sqs/) is a message queue service provided by [Amazon Web Services](https://aws.amazon.com/).
 
+## Transport facts
```
danielmarbach:
```suggestion
## Transport at a glance
```

transports/azure-service-bus/legacy/index.md
```diff
@@ -27,6 +27,18 @@ Note: Publish/Subscribe and Timeouts (including message deferral) are supported
 
 include: azure-transports
 
+## Transport facts
```
danielmarbach:
```suggestion
## Transport at a glance
```

transports/azure-service-bus/index.md
```diff
@@ -14,6 +14,18 @@ The Azure Service Bus transport leverages the .NET Standard [Microsoft.Azure.Ser
 
 [Azure Service Bus](https://azure.microsoft.com/en-us/services/service-bus/) is a messaging service hosted on the Azure platform that allows for exchanging messages between various applications in a loosely coupled fashion. The service offers guaranteed message delivery and supports a range of standard protocols (e.g. REST, AMQP, WS*) and APIs such as native pub/sub, delayed delivery, and more.
 
+## Transport facts
```
danielmarbach:
```suggestion
## Transport at a glance
```

transports/rabbitmq/index.md
```diff
@@ -17,6 +17,17 @@ Provides support for sending messages over [RabbitMQ](https://www.rabbitmq.com/)
...
+
+|Feature                    |   |  
+|:---                       |---
+|Transactions |None, ReceiveOnly
+|Pub/Sub                    |Native
+|Timeouts                   |Native
+|Large message bodies       |via DataBus
```
danielmarbach:
```suggestion
|Large message bodies       |Broker can handle arbitrary message size within available resources, very large messages via DataBus
```

danielmarbach: Maybe this is too many details anyway I wanted to add it for completeness reasons

mauroservienti:
seems that ASQ is the only one missing, right?

danielmarbach:
@mauroservienti I gave it a try

mauroservienti:
@bording can you have a look at the RabbitMQ document and review it?

transports/azure-service-bus/legacy/index.md
```diff
@@ -27,6 +27,18 @@ Note: Publish/Subscribe and Timeouts (including message deferral) are supported
...
+## Transport at a glance
+
+|Feature                    |   |  
+|:---                       |---
+|Transactions |None, ReceiveOnly, SendsWithAtomicReceive
+|Pub/Sub                    |Native
+|Timeouts                   |Native
+|Large message bodies       |via higher tier (i.ex. Premium) or DataBus
+|Scale-out             |Competing consumer
+|Scripted Deployment        |Supported for the Forwarding Topology using `NServiceBus.Transport.AzureServiceBus.CommandLine`, not supported for the Endpoint Oriented Topology
```
SeanFeldman: It's not supported, but not intended. NServiceBus.Transport.AzureServiceBus.CommandLine is only intended for NServiceBus.Transport.AzureServiceBus transport.

transports/azure-service-bus/index.md
```diff
@@ -14,6 +14,18 @@ The Azure Service Bus transport leverages the .NET Standard [Microsoft.Azure.Ser
...
 [Azure Service Bus](https://azure.microsoft.com/en-us/services/service-bus/) is a messaging service hosted on the Azure platform that allows for exchanging messages between various applications in a loosely coupled fashion. The service offers guaranteed message delivery and supports a range of standard protocols (e.g. REST, AMQP, WS*) and APIs such as native pub/sub, delayed delivery, and more.
 
+## Transport at a glance
+
+|Feature                    |   |  
+|:---                       |---
+|Transactions |None, ReceiveOnly, SendsWithAtomicReceive
+|Pub/Sub                    |Native
+|Timeouts                   |Native
+|Large message bodies       |via higher tier (i.ex. Premium) or DataBus
```
SeanFeldman:
```suggestion
|Large message bodies       | with Premium tier or DataBus
```

transports/azure-service-bus/legacy/index.md
mauroservienti: @danielmarbach 👆

danielmarbach: Well but the tool creates all the necessary topology parts of forwarding topology which is independent of the transport. Given that the new transport has to stay compliant with Forwarding topo there is nothing we can break there which means the tool can be used. I get that it was not intended to be used that way but it can that's why I mentioned it here.

danielmarbach: But given this is the legacy transport and it is transport at a glance I'm happy to drop it

SeanFeldman: ☝️ exactly that. We're in the mids of deprecating the legacy ASB transport and having documentation mixing the tool that was designed for the new ASB transport would just add confusion.

SeanFeldman: Based on the conversation above:

```suggestion
|Scripted Deployment        | Not supported
```

SeanFeldman:
ASQ/ASB (both) look good.

mauroservienti:
Thanks folks!

PR #4415, opened 2019-07-03
Wrong code

aperisic:
The above code is technically wrong. The bool parameter passed is false and it means if case sensitive search should be done. (In case the message is from another assembly this above does not work.)

### DIFF
--- a/samples/serializers/change-message-type/Core_7/SamplePhase2/MessageIdentityMutator.cs
+++ b/samples/serializers/change-message-type/Core_7/SamplePhase2/MessageIdentityMutator.cs
@@ -26,7 +26,7 @@ class MessageIdentityMutator :
                 }
                 return Assembly.Load(assemblyName);
             },
-            typeResolver: (assembly, typeName, caseSensitive) =>
+            typeResolver: (assembly, typeName, ignoreCase) =>
             {
                 if (typeName == "CreateOrderPhase1")
                 {
--- a/samples/serializers/change-message-type/Core_7/SamplePhase2/MessageIdentityMutator.cs
+++ b/samples/serializers/change-message-type/Core_7/SamplePhase2/MessageIdentityMutator.cs
@@ -26,13 +26,13 @@ class MessageIdentityMutator :
                 }
                 return Assembly.Load(assemblyName);
             },
-            typeResolver: (assembly, typeName, assemblyPassed) =>
+            typeResolver: (assembly, typeName, caseSensitive) =>
             {
                 if (typeName == "CreateOrderPhase1")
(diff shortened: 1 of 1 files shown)


danielmarbach:
@aperisic Thanks for the contribution! See

https://github.com/Particular/docs.particular.net/pull/4417

is that what you hinted at by saying the code is technically wrong?

aperisic:
Not exactly. I mean you can extend it more if you want but my point is that
existing code has following problems:

// error in this line: parameter assemblyPassed has nothing to do with
passing assembly
// this third parameter is about case sensitive search
typeResolver: (assembly, typeName, assemblyPassed) =>

// and now this parameter is used badly, pulling you into an incorrect
assumption
if (assemblyPassed)
{
return assembly.GetType(typeName);
}



// this is corrected naming, third parameter is named more appropriately
typeResolver: (assembly, typeName, caseSensitive) =>
{
// and now we can check what we can check, and that is if assembly is null
or not
if (assembly != null)
{
return assembly.GetType(typeName);
}

For this example I think we can keep
return assembly.GetType(typeName);

If you want to extend it more, ok, it is up to you, but  if
(assemblyPassed) is nonsensical condition: reason 1. bad named variable,
reason 2. it checks in reality a completely irrelevant thing if search is
case sensitive

btw... the example does not work if I place message in another assembly.

In my opinion, what I said must be fixed, what you are saying does not have
as according to MSDN ignoreCase is always false.

https://docs.microsoft.com/en-us/dotnet/api/system.type.gettype?view=netframework-4.8#System_Type_GetType_System_String_System_Func_System_Reflection_AssemblyName_System_Reflection_Assembly__System_Func_System_Reflection_Assembly_System_String_System_Boolean_System_Type__

typeResolverFunc<Assembly,String,Boolean,Type>

A method that locates and returns the type that is specified by typeName
from the assembly that is returned by assemblyResolver or by standard
assembly resolution. If no assembly is provided, the typeResolvermethod can
provide one. The method also takes a parameter that specifies whether to
perform a case-insensitive search;* false is passed to that parameter.*

So I would go with my version but I would change caseSensitive parameter to
ignoreCase parameter

danielmarbach:
Thanks for the brainstorming and the input. 

Regarding

> btw... the example does not work if I place message in another assembly.

Samples are meant to be just samples and not something that is really production ready, see https://docs.particular.net/samples/#samples-are-not-production-ready

danielmarbach:
@aperisic would you mind sending me a quick email to daniel dot marbach at particular dot net with a shipping address and phone number so that we can send you a small contribution gift?

PR #4416, opened 2019-07-03
Wrong code

aperisic:
The above code is technically wrong. The bool parameter passed is false and it means if case sensitive search should be done. (In case the message is from another assembly this above does not work.)


danielmarbach:
@aperisic Thanks for the contribution! See

https://github.com/Particular/docs.particular.net/pull/4417

is that what you hinted at by saying the code is technically wrong?

PR #4422, opened 2019-07-04
Document rule creation

danielmarbach:


transports/azure-service-bus/operational-scripting.md
```diff
@@ -50,6 +51,8 @@ asb-transport endpoint create [--size]
 
 `-b` | `--subscription`: Subscription name (defaults to endpoint name)
 
+`-e` | `--subscribe-to`: Subscribe to event type full name (e.g. MyNamespace.MyMessage or CustomRuleName,MyNamespace.MyMessage), multiple allowed
```
SeanFeldman: If multiple rules are allowed, are those provided using a delimiter? What delimiter? Should there be an example?

SeanFeldman: Also, `CustomRuleName` rule name is usually needed when the actual event with namespace is too long. Therefore it's more of a `ShortenedRuleName` rule name, to be aligned with our [nomenclature](https://docs.particular.net/transports/azure-service-bus/configuration#entity-creation).

PR #4426, opened 2019-07-08
Review child lifetime doco page

timbussmann:
it felt like this page needs a major overhaul. To start with "child lifetime" felt a bit off, however "child containers" might also not be the best option given not every container uses child containers but other types of scoping mechanisms.

Thoughts?


nservicebus/dependency-injection/child-lifetime.md
```diff
@@ -1,30 +1,27 @@
...
 component: Core
-reviewed: 2017-10-06
+reviewed: 2019-07-08
 redirects:
  - nservicebus/nservicebus-support-for-child-containers
  - nservicebus/containers/child-containers
 ---
 
-Child lifetimes are essentially a snapshot of the main instance where transient instances are treated as singletons within the scope of the child lifetime. This is useful to scope instances for the duration of a web request or the handling of a message in NServiceBus. While this was possible before, child lifetimes bring one more important feature to the table.
+Child containers are essentially a snapshot of the main container providing an additional instance lifetime scope. Instances configured with child container lifetime are treated as singletons within the scope of the child container. This is useful to scope instances for the duration of a web request or the handling of a message in NServiceBus.
```
WilliamBZA:
```suggestion
Child containers are a snapshot of the main container providing an additional instance lifetime scope.
```

nservicebus/dependency-injection/child-lifetime.md
```diff
@@ -1,30 +1,27 @@
...
 
-NOTE: As of [NServiceBus.Spring](https://www.nuget.org/packages/NServiceBus.Spring) Version 6.0.0, child lifetimes are now supported as well. Previous versions don't support child lifetimes so one of the other containers supported by NServiceBus must be used to take advantage of it.
 
+## Child container scope in NServiceBus
 
-## Deterministic disposal
-
-Instance lifetime is usually not tracked by the main instance (Windsor is an exception) and that means that manually disposing is required on any instance that needs deterministic disposal. Child lifetimes solve this issue by automatically disposing all transient objects created within each specific child lifetime.
+NServiceBus creates a child container for each transport message that is received. This means that all child container scoped instances created during message processing are scoped as singletons within the child container lifetime. This supports sharing of database sessions and other resources with lifetimes specific to the processed message.
```
WilliamBZA:
```suggestion
NServiceBus creates a child container for each transport message that is received. This means that all `InstancePerUnitOfWork` scoped instances created during message processing, are scoped as singletons within the context of processing each message. This is helpful to support the sharing of database sessions and other resources with lifetimes specific to the processed message.
```

nservicebus/dependency-injection/child-lifetime.md
```diff
@@ -1,30 +1,27 @@
...
+snippet: InstancePerUnitOfWorkRegistration
 
-When the message finishes processing, the child lifetime and all transient instances are disposed. So if deterministic disposal is required, implement [IDisposable](https://msdn.microsoft.com/en-us/library/system.idisposable.aspx).
 
-When the message is processed, the session is disposed and all resources such as database connections are released.
+## Deterministic disposal
 
-If configuring components using the NServiceBus configure API, it's possible to configure instance lifecycle to be per unit of work, using this:
+Instance lifetime is usually not tracked by the main instance (Windsor is an exception) and that means that manual disposal is required on any instance that needs deterministic disposal. Child containers solve this issue by automatically disposing all instances created within each specific child lifetime. This is useful to managing things like the database sessions.
```
mikeminutillo: The term instance is used three times in this sentence and I suspect (but do not know) that it is referring to at least two different things.

timbussmann:
> Would it be worth adding a note to the Instance per unit of work section on the main DI page about deterministic disposal and retiring this page?

definitely was thinking about this idea as well.

nservicebus/dependency-injection/child-lifetime.md
timbussmann: I've simplified that section, have another look 👍

PR #4453, opened 2019-08-02
Switch to hosted services because it is more idiomatic

danielmarbach:


danielmarbach:
@SimonCropp here we go

danielmarbach:
I did not add any text about the infrastructure. do you think that would be necessary?

SimonCropp:
looks good to me. from my perspective, i dont need any more text. not sure about others.

Although the complexity in getting that pattern correct looks like it should be in a nuget?

samples/web/send-from-aspnetcore-webapi/Core_7/WebApplication/AddNServiceBusServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,37 @@
...
+            var management = provider.GetService<SessionAndConfigurationHolder>();
+            if (management.MessageSession != null)
+            {
+                return management.MessageSession;
+            }
+
+            var timeout = TimeSpan.FromSeconds(30);
+            // SpinWait is here to accomodate for WebHost vs GenericHost difference
+            // Closure here should be fine under the assumption we always fast track above once initialized
+            if (!SpinWait.SpinUntil(() => management.MessageSession != null || management.StartupException != null,
```
davidfowl, line 24: This is pretty crazy

PR #4456, opened 2019-08-12
Use correct transport for TwoWayRouterConfig

janv8000:
Should second transport not be of type RabbitMQTransport?


SzymonPobiega:
@janv8000 yes, that was a silly copy-paste error. Thanks ❤️

PR #4462, opened 2019-08-21
Document saga data migration script

timbussmann:
add a script which allows to migrate the data so that no compatibility mode has to be used.

### DIFF
--- a/persistence/mongodb/migrating-from-tekmaven.md
+++ b/persistence/mongodb/migrating-from-tekmaven.md
@@ -13,21 +13,7 @@ The `NServiceBus.Storage.MongoDB` package was designed to be fully compatible wi
 
 include: migration-warning
 
-## Compatibility mode
-
-The compatibility mode enables the MongoDB persistence to work with existing saga data without modifications to the database.
-
-### Configuration
-
-Use the following compatibility API to configure the package to work with existing saga data:
-
-snippet: MongoDBTekmavenCompatibility
-
-The `VersionElementName` value must match the `BsonDocument` element name used by the previous saga data property decorated with the `[DocumentVersion]` attribute.
-
-include: must-apply-conventions-for-version
-
-### Saga data class changes
+## Saga data class changes
 
 [Saga data classes](/nservicebus/sagas/#long-running-means-stateful) no longer need to provide an `int` version property decorated with a `DocumentVersion`. The version property and attribute may be safely removed from saga data class implementations:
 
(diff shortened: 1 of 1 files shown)


persistence/mongodb/migrating-from-tekmaven.md
```diff
@@ -43,3 +33,39 @@ class MySagaData : IContainSagaData
 ### How Document Versioning Works
 
 include: document-version
+
+
+## Compatibility mode
+
+The compatibility mode enables the MongoDB persistence to work with existing saga data without modifications to the database.
```
kbaley:
```suggestion
Compatibility mode allows the MongoDB persistence to work with saga data created with the `NServiceBus.Persistence.MongoDB` package without modifications to the database.
```

persistence/mongodb/migrating-from-tekmaven.md
```diff
@@ -43,3 +33,39 @@ class MySagaData : IContainSagaData
...
+
+### Configuration
+
+Use the following compatibility API to configure the package to work with existing saga data:
```
kbaley:
```suggestion
Use the following API to configure the package to work with existing saga data:
```

persistence/mongodb/migrating-from-tekmaven.md
```diff
@@ -43,3 +33,39 @@ class MySagaData : IContainSagaData
...
+snippet: MongoDBTekmavenCompatibility
+
+The `VersionElementName` value must match the `BsonDocument` element name used by the previous saga data property decorated with the `[DocumentVersion]` attribute.
+
+include: must-apply-conventions-for-version
+
+
+## Migrating data
+
+Alternatively to the compatibility mode, the existing saga data can be migrated to the data format used by the `NServiceBus.Storage.MongoDB` package. This approach requires the endpoint to be stopped during the migration. Use the `mongo` Shell to connect to the database and execute the following script:
```
kbaley:
```suggestion
As an alternative to compatibility mode, saga data created by the `NServiceBus.Persistence.MongoDB` package can be migrated to the data format used by the `NServiceBus.Storage.MongoDB` package. This approach requires the endpoint to be stopped during the migration. Use the `mongo` shell to connect to the database and execute the following script:
```

kbaley: Technically, they don't need to stop the endpoint though doing it in-flight introduces some challenges. I think we're okay to leave the recommendation as-is but mentioning it in case others disagree.

persistence/mongodb/migrating-from-tekmaven.md
```diff
@@ -43,3 +33,39 @@ class MySagaData : IContainSagaData
...
+    },
+    {
+        $rename: { "Version": "_version" }
+    })
+});
+```
+
+Replace `"Version"` with the name of the version property on the saga data which was previously decorated with the `[DocumentVersion]` attribute.
+
+WARNING: Make sure to create a backup of the database prior to migrating the saga data.
```
kbaley:
```suggestion
WARNING: Be sure to create a backup of the database prior to migrating the saga data.
```

persistence/mongodb/migrating-from-tekmaven.md
timbussmann: yeah I think it's safer to recommend to stop and migrate rather than doing this on a live system. if they are experienced with MongoDB they would figure this out themselves I think?

PR #4469, opened 2019-08-28
Functions sample review

timbussmann:
* added a startup projects file because the trigger project isn't chosen as a startup project by default.
* removed all namespaces as this seems to be the general rule for sample code.


timbussmann:
weird, will have a look.

timbussmann:
fixed

PR #4474, opened 2019-08-30
LogManager.Use TestingLoggerFactory only works for one use

SimonCropp:
since it is a static config, the second time you 

```
 LogManager.Use<TestingLoggerFactory>()
            .WriteTo(new StringWriter(logStatements));
```

it is ignored since the handler has already setup its logger in the static context.

`ShouldLogCorrectlyASecondTime` will fail after `ShouldLogCorrectly` runs first

### DIFF
--- a/Snippets/Testing/Testing_7/Logging/LoggingTest.cs
+++ b/Snippets/Testing/Testing_7/Logging/LoggingTest.cs
@@ -24,5 +24,20 @@ public class LoggingTests
 
         StringAssert.Contains("Some log message", logStatements.ToString());
     }
+    [Test]
+    public async Task ShouldLogCorrectlyASecondTime()
+    {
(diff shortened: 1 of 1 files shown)


danielmarbach:
@SimonCropp, unfortunately, that is by design. So essentially what you pass to the writer can be something that you clear between test runs but it has to be static. Should we clarify this in the doco?

SimonCropp:
@danielmarbach yeah clarification would be good, since atm it only works for the first test it is used in

danielmarbach:
@SimonCropp what do you think about this one https://github.com/Particular/docs.particular.net/pull/4478 ?

danielmarbach:
Replaced by https://github.com/Particular/docs.particular.net/pull/4478

PR #4478, opened 2019-09-02
Hopefully better approach for testing logger

danielmarbach:
Alternative to https://github.com/Particular/docs.particular.net/pull/4474

### DIFF
--- a/nservicebus/testing/index_content_testing_[6,).partial.md
+++ b/nservicebus/testing/index_content_testing_[6,).partial.md
@@ -90,6 +90,6 @@ The following code show how to verify that logging is performed by the message h
 
 snippet: LoggerTestingSetup
 
-The setup fixture above sets the testing logging factory once per assembly because the factory is statically cached for the live time of the application domain. Subsequent test executions then clear the logged statements before every test run as shown below.
+The setup fixture above sets the testing logging factory once per assembly because the factory is statically cached during the lifetime of the application domain. Subsequent test executions then clear the logged statements before every test run as shown below.
 
 snippet: LoggerTesting
--- a/nservicebus/testing/index_content_testing_[6,).partial.md
+++ b/nservicebus/testing/index_content_testing_[6,).partial.md
@@ -88,4 +88,8 @@ To test that logging is performed correctly, use the `TestingLoggerFactory`. The
 
 The following code show how to verify that logging is performed by the message handler.
 
+snippet: LoggerTestingSetup
+
+The setup fixture above sets the testing logging factory once per assembly because the factory is statically cached for the live time of the application domain. Subsequent test executions then clear the logged statements before every test run as shown below.
+
 snippet: LoggerTesting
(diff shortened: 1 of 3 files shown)


nservicebus/testing/index_content_testing_[6,).partial.md
```diff
@@ -88,4 +88,8 @@ To test that logging is performed correctly, use the `TestingLoggerFactory`. The
 
 The following code show how to verify that logging is performed by the message handler.
 
+snippet: LoggerTestingSetup
+
+The setup fixture above sets the testing logging factory once per assembly because the factory is statically cached for the live time of the application domain. Subsequent test executions then clear the logged statements before every test run as shown below.
```
andreasohlund:
```suggestion
The setup fixture above sets the testing logging factory once per assembly because the factory is statically cached during the lifetime of the application domain. Subsequent test executions then clear the logged statements before every test run as shown below.
```

danielmarbach:
@SimonCropp I'm going ahead and merge this. If you feel the current description or sample needs to be improved just ping again here and I'll see what I can do

SimonCropp:
Given handler tests are async and can run in parallel, wont this cause interference between tests?

danielmarbach:
Well, async doesn't really make a change because you are sequentializing access to the handler anyway by awaiting the outcome of a handler. I'd also argue that throwing in concurrency might push you over the border if "unit tests" and might be a different topic. What right though is that for example if you use xunit test parallelization then all the tests that use the static logger cannot be run in parallel. I wonder though if this is a concern that needs to be addressed here because it would significantly blow up the complexity of that documentation snippet

SimonCropp:
> async doesn't really make a change because you are sequentializing access to the handler anyway by awaiting the outcome of a handler.

i dont understand that. two tests can use the same handler at the same time.

> that throwing in concurrency might push you over the border if "unit tests" and might be a different topic

concurrency isnt something that people opt in to for tests, it is something that test frameworks do by default

just to be clear i am not saying the documentation need to be fixed. i think the implementation of that logging recording feature is flawed. shouldnt it use an asynclocal to record logging messages?

danielmarbach:
I'm not sure if it is that easy to say the logic is flawed. I think it is fair to say that the current implementation of the testing lib takes into account that the loggers are statically initialized. All the default logging approaches NServiceBus pushes and internally uses resolve the logger statically. The testing lib assumes that as well. For clear test independence, the testing lib would need to expose a way to scope loggers somehow and that would have to be controllable by the user. Do you want me to raise an issue for you in the testing repo?

danielmarbach:
@SimonCropp what do you think about this approach https://github.com/Particular/NServiceBus.Testing/pull/147

danielmarbach:
@SimonCropp we've released a possibility to scope the logging factory https://discuss.particular.net/t/nservicebus-testing-7-2-minor-release-available/1374

danielmarbach:
WIP docs PR is available at https://github.com/Particular/docs.particular.net/pull/4495

danielmarbach:
Documentation is merged

PR #4482, opened 2019-09-04
Update to the latest Functions (fixed logging and removed FunctionExecutionContext)

SeanFeldman:
To contrast with PR #4483

Specifically, this PR is using not using `FunctionExecutionContext` and accessing `ILog` via `LogManager`:

https://github.com/Particular/docs.particular.net/blob/2539d29fc9bc2876104b32619ba5528ba21eaa8e/samples/azure/functions/service-bus/ASBFunctions_0.1/AzureFunctions.ASBTrigger/AzureServiceBusTriggerFunction.cs#L36-L40

PR #4483, opened 2019-09-04
Update to the latest Functions (fixed logging and re-added FunctionExecutionContext)

SeanFeldman:
To contrast with PR #4482 

Specifically, this PR is using `FunctionExecutionContext` exposing `ILogger`:

https://github.com/Particular/docs.particular.net/blob/3358576500883f66158eb9e798dd578d48e29ba3/samples/azure/functions/service-bus/ASBFunctions_0.1/AzureFunctions.ASBTrigger/AzureServiceBusTriggerFunction.cs#L35-L39


SeanFeldman:
I've uploaded NServiceBus.AzureFunctions.StorageQueues.0.1.0-function-executi0001.nupkg and NServiceBus.AzureFunctions.ServiceBus.0.1.0-function-executi0001.nupkg to MyGet but TC seems to fail to retrieve those.

PR #4485, opened 2019-09-05
Remove backwards compatibility

WilliamBZA:
We say that backwards compatibility is not supported but I see no reason why. I pinged in the support channel to get specifics and no answer given.

In the absence of any real reasons, we should not say it's not supported. If it is in-fact not supported, we should give reasons why and not just say "It might work, but we don't guarantee it". This causes confusion for customers so we should be clearer here.

This PR is an attempt to get consensus on this and make sure the decision is made explicitly.

### DIFF
--- a/nservicebus/gateway/index.md
+++ b/nservicebus/gateway/index.md
@@ -76,6 +76,6 @@ partial: recoverability
 
 ### Version compatibility
 
-The gateway component ensures only forward compatibility for one major version. That means a message sent by the NServiceBus 3.x Gateway can be understood by the NServiceBus 4.x Gateway, and a message sent by the NServiceBus 4.x Gateway can be understood by NServiceBus.Gateway 1.x (the Gateway package targeting NServiceBus 5.x).
+The gateway component ensures only compatibility for one major version. That means a message sent by the NServiceBus 3.x Gateway can be understood by the NServiceBus 4.x Gateway, and a message sent by the NServiceBus 4.x Gateway can be understood by NServiceBus.Gateway 1.x (the Gateway package targeting NServiceBus 5.x).
 
-However, a message sent by the NServiceBus 3.x Gateway will not be understood by NServiceBus.Gateway 1.x (NServiceBus 5.x) as this skips a major version. Likewise, a message sent by NServiceBus.Gateway 2.x will not be understood by NServiceBus.Gateway 1.x, as backwards communication is not supported.
+However, a message sent by the NServiceBus 3.x Gateway will not be understood by NServiceBus.Gateway 1.x (NServiceBus 5.x) as this skips a major version.


nservicebus/gateway/index.md
```diff
@@ -76,6 +76,6 @@ partial: recoverability
 
 ### Version compatibility
 
-The gateway component ensures only forward compatibility for one major version. That means a message sent by the NServiceBus 3.x Gateway can be understood by the NServiceBus 4.x Gateway, and a message sent by the NServiceBus 4.x Gateway can be understood by NServiceBus.Gateway 1.x (the Gateway package targeting NServiceBus 5.x).
+The gateway component ensures only compatibility for one major version. That means a message sent by the NServiceBus 3.x Gateway can be understood by the NServiceBus 4.x Gateway, and a message sent by the NServiceBus 4.x Gateway can be understood by NServiceBus.Gateway 1.x (the Gateway package targeting NServiceBus 5.x).
```
danielmarbach: Is the major version of the core relevant or the Gateway? I'm confused. Especially because of

> and a message sent by the NServiceBus 4.x Gateway can be understood by NServiceBus.Gateway 1.x (the Gateway package targeting NServiceBus 5.x)

WilliamBZA: I have no idea

danielmarbach: Me neither. Because I read this and I was thinking if it is the major of the gateway package shouldn't then this text say:

> and a message sent by the NServiceBus 4.x Gateway can be understood by NServiceBus.Gateway 3.x

WilliamBZA: @janovesk ? @andreasohlund ? Do either of you know? Or perhaps @DavidBoike since he added the [original text](https://github.com/Particular/docs.particular.net/commit/72148d846e93fb7cb0ea5907a9b15895fa87e6eb#diff-07dfcd8614218904bd9fe08bf943086c)?

ramonsmits:
@WilliamBZA I did think you said that you verified compatibility between 1 and 2, 2 and 3 for commands and replies correct?

nservicebus/gateway/index.md
ramonsmits:
```suggestion
Our [support policy guarantees backwards wire compatibility for at least one major version](/nservicebus/upgrades/support-policy#compatibility-guarantees). View our [supported package versions]  (https://docs.particular.net/nservicebus/upgrades/supported-versions#other-packages-nservicebus-gateway) to see which versions of the gateway are still supported.

### Storage compatibility

Gateway storage compatibility is unknown and dependant on gateway persistence implementation. Data migration is only guaranteed one major forward.

### Wire compatibility

The gateway is currently forward and backwards compatible between:
- NServiceBus.Gateway version 1.x and 3.x
- NServiceBus 3.x and 4.x
```

PR #4461, opened 2019-08-16
Document LRU memory usage

timbussmann:

PR #4489, opened 2019-09-09
editorial review of LRU memory usage changes

adamralph:
editorial review of #4461

### DIFF
--- a/nservicebus/recoverability/index.md
+++ b/nservicebus/recoverability/index.md
@@ -104,5 +104,5 @@ partial: exceptionincluded
 
 ## Recoverability memory consumption
 
-MSMQ and SQL Server transport need to cache exceptions in memory for retries. Therefore, exceptions with a large memory footprint can cause high memory usage of the NServiceBus process. NServiceBus can cache up to 1000 exceptions, capping the potential memory consumption to 1000 x `<exception size>`. Refer to [this guide](/nservicebus/recoverability/lru-memory-consumption.md) to resolve problems due to execessive memory consumption.
+MSMQ and SQL Server transport need to cache exceptions in memory for retries. Therefore, exceptions with a large memory footprint can cause high memory usage of the NServiceBus process. NServiceBus can cache up to 1,000 exceptions, capping the potential memory consumption to 1,000 x `<exception size>`. Refer to [this guide](/nservicebus/recoverability/lru-memory-consumption.md) to resolve problems due to execessive memory consumption.
 
--- a/nservicebus/recoverability/lru-memory-consumption.md
+++ b/nservicebus/recoverability/lru-memory-consumption.md
@@ -8,16 +8,16 @@ tags:
  - Error Handling
  - Retry
  - Recoverability
-reviewed: 2019-08-16
+reviewed: 2019-09-09
 ---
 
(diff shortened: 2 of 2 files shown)


nservicebus/recoverability/index.md
```diff
@@ -104,5 +104,5 @@ partial: exceptionincluded
 
 ## Recoverability memory consumption
 
-MSMQ and SQL Server transport need to cache exceptions in memory for retries. Therefore, exceptions with a large memory footprint can cause high memory usage of the NServiceBus process. NServiceBus can cache up to 1000 exceptions, capping the potential memory consumption to 1000 x `<exception size>`. Refer to [this guide](/nservicebus/recoverability/lru-memory-consumption.md) to resolve problems due to execessive memory consumption.
+MSMQ and SQL Server transport need to cache exceptions in memory for retries. Therefore, exceptions with a large memory footprint can cause high memory usage of the NServiceBus process. NServiceBus can cache up to 1,000 exceptions, capping the potential memory consumption to 1,000 x `<exception size>`. Refer to [this guide](/nservicebus/recoverability/lru-memory-consumption.md) to resolve problems due to execessive memory consumption.
```
timbussmann: @adamralph I think we shouldn't use commas here. Commas can be considered a decimal sign in some places and some regions use dots as separators. 

Since this is confusing, there seems to be an ISO standard  (https://en.wikipedia.org/wiki/ISO_31-0#Numbers) stating:
> ISO 31-0 specifies that such groups of digits should never be separated by a comma or point, as these are reserved for use as the decimal sign.

> Numbers consisting of long sequences of digits can be made more readable by separating them into groups, preferably groups of three, separated by a small space. For this reason, ISO 31-0 specifies that such groups of digits should never be separated by a comma or point, as these are reserved for use as the decimal sign.

timbussmann: (that btw. might be something we should state in the doco readme

timbussmann: ok, looking a bit further into this it seems that ISO 31-0 has been replaced by ISO 80000 and I can't find public access to that standard....-.- I'd still vote to not use dots or commas for groupings.

adamralph: As I understand it, our docs site is written in US English, so commas for decimal groupings should be used.

> Commas can be considered a decimal sign in some places and some regions use dots as separators.

Yet we still have to use a decimal sign when we write a non-integer literal. Again, because the docs site is in US English, we use a dot.

timbussmann: > docs site is written in US English

language preference and regional formatting isn't the same thing. I'd vote to eliminate all potential ambiguity when possible as the doco audience is not necessarily able to derive the correct regional style from a website (e.g. we should also clearly avoid ambiguous date formats). E.g. here's an interesting list of the many different applications: https://docs.oracle.com/cd/E19455-01/806-0169/overview-9/index.html

I'd suggest to avoid the grouping entirely in this case, use the "small space" as suggested by ISO 31-0 (I still have no idea what the newer standards describes but that one isn't publicly available) or something which might be less ambiguous like `'`.

adamralph: @timbussmann would you like to raise an issue for this?

timbussmann: I'm raising an issue to be clearer about the goal of avoiding ambiguity but tbh I still feel we should not use the `,` as the default.

adamralph: 👍 OK. Let's discuss it in the issue.

Issue #4491, opened 2019-09-09
Saga-audit article fix required


aleksandr-samila:
Related to [Weekly legacy doco review - 2019-09-09](https://github.com/Particular/DeveloperExperience/issues/531)
and to https://docs.particular.net/servicecontrol/plugins/saga-audit article.

This headers duplication
![image](https://user-images.githubusercontent.com/5980232/64562218-30727880-d355-11e9-9eab-28134f57a3b9.png)
looks a bit weird

aleksandr-samila:
PR with a possible fix was created

SeanFeldman:
Alternative PR: https://github.com/Particular/docs.particular.net/pull/4493

SeanFeldman:
Fix deployed
![image](https://user-images.githubusercontent.com/1309622/64565876-f2079a00-d311-11e9-9a43-f8a1215f37d5.png)

PR #4492, opened 2019-09-09
Saga-audit article. Cosmetic fix

aleksandr-samila:
fixes https://github.com/Particular/docs.particular.net/issues/4491

superseding PR: https://github.com/Particular/docs.particular.net/pull/4493

### DIFF
--- a/servicecontrol/plugins/saga-audit.md
+++ b/servicecontrol/plugins/saga-audit.md
@@ -48,7 +48,6 @@ The following SagaAudit plugin packages have been deprecated and unlisted. If us
 - **ServiceControl.Plugin.Nsb5.SagaAudit**
 - **ServiceControl.Plugin.Nsb6.SagaAudit**
 
-## Configuration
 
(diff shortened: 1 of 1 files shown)


SeanFeldman:
@aleksandr-samila FYI for GitHub to link PR and issue, the description should include [keywords](https://help.github.com/en/articles/closing-issues-using-keywords). I use `fixes #<issue-number>`.

SeanFeldman:
Instead of removing the section title from the main document and replicating it everywhere else, it would be better to keep the Configuration section in the main document and remove it from the partials. The configuration is not going away and is required in all version. If/when that's not the case, it makes sense to move the section name into the partials.

Raised an alternative PR: https://github.com/Particular/docs.particular.net/pull/4493

SeanFeldman:
@aleksandr-samila I'll close this PR but **won't** delete the branch. In case you disagree with the alternative PR, this can be resurrected. Thanks.

PR #4495, opened 2019-09-10
Show async testing overloads and add warning

danielmarbach:


nservicebus/upgrades/all-versions.include.md
```diff
@@ -709,7 +710,8 @@ _This package has been split into NServiceBus.DataBus.AzureBlobStorage and NServ
 
 | Version   | Released       | Supported until   | Notes                             |
 |:---------:|:--------------:|:-----------------:|:---------------------------------:|
-| 1.5.x     | 2019-08-23     | -                 | -                                 |
+| 1.6.x     | 2019-09-06     | -                 | -                                 |
```
timbussmann: why is this changed?

danielmarbach: I had to run the script to update the supported version and it did update this as well. No idea

timbussmann:
conent verification seems to fail but not sure why, the snippet mentioned in the error message looks correctly defined to me. Is this a bug in the docs engine?

danielmarbach:
I think it is an oddity of the docs engine. It probably requires a dedicated snippet csproj in a folder called 7_2

PR #4502, opened 2019-09-16
Tutorials --> monitoring-setup --> sp url fix

aleksandr-samila:
This is the simplest variant of the fix. Just to point user to the list of releases that is sorted by latest version.

Or we should implement something to automatically identify the latest release, to prepare the latest *.exe link. Something like we have for SC on this page. In this case it's better to ignore this PR and to create an issue.


aleksandr-samila:
I pinged you @DavidBoike and @SeanFeldman because I have no idea who should I ask to approve/review this PR.

I noticed this issue just reading an article (not in the range of any existing issue).

PR #4503, opened 2019-09-16
Fix ASP.NET Core samples

bording:
Now that the DocsEngine fix has been deployed, we can properly update the ASP.NET Core samples.

This updates the projects to use the `Microsoft.AspNetCore.App` package.

### DIFF
--- a/samples/gateway/Gateway_3/WebClient.Core/Properties/launchSettings.json
+++ b/samples/gateway/Gateway_3/WebClient.Core/Properties/launchSettings.json
@@ -1,0 +1,27 @@
+{
+  "iisSettings": {
+    "windowsAuthentication": false,
+    "anonymousAuthentication": true,
+    "iisExpress": {
+      "applicationUrl": "http://localhost:64485/",
+      "sslPort": 0
+    }
+  },
+  "profiles": {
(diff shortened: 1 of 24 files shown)


samples/gateway/Gateway_3/WebClient.Core/Properties/launchSettings.json
```diff
@@ -0,0 +1,27 @@
+{
```
andreasohlund: Are we supposed to commit this file into the repo?

bording: I think we are, yes.

bording:
> Should we add some guidance to https://github.com/Particular/docs.particular.net/blob/master/README.md along the lines of "Favour using MS meta packages for samples". Or is this a given?

I don't think we need to do this. There's really no other option. This is how its supposed to be done.

samples/gateway/Gateway_3/WebClient.Core/Properties/launchSettings.json
andreasohlund: Ok, to make it work when you hit F5?

PR #4504, opened 2019-09-17
Add a document to point to queue length inner workings/design

SeanFeldman:
fixes https://github.com/Particular/DeveloperExperience/issues/524


dvdstelt:
@SeanFeldman This new article can't be found unless you're searching for it. Do you think maybe a link from an existing article to this one would add value?

SeanFeldman:
It will be auto-indexed on the main Monitoring page, @dvdstelt, just like the rest of the articles.
If you have an idea where else should it be added, please do. Thank you.

SeanFeldman:
This is exactly the reason I raised an issue rather than "just fixing it".
Others seems to have way more ideas where and how it should be done.

I'm away till next week. This PR will have to wait a little.

DavidBoike:
@SeanFeldman 	Forgot you were away. Mind if I finish it off today?

SeanFeldman:
By all means👍
 And if not, I'll get back to it on Monday.

PR #4509, opened 2019-09-23
Azure Functions bug fix

SeanFeldman:
- Removed the need to specify Service Bus Connection string setting name, following the same model as Storage Queues sample.
- Removed constraint from the [constraints and limitation document](https://docs.particular.net/nservicebus/hosting/azure-functions/) 
> - While the Azure Functions runtime uses the default connection string in the app setting that is named "AzureWebJobsServiceBus" (ASB) or "AzureWebJobsStorage" (ASQ), there's a bug with the Service Bus and the trigger connection has to be specified.

### DIFF
--- a/samples/azure/functions/service-bus/ASBFunctions_0.1/AzureFunctions.ASBTrigger/AzureServiceBusTriggerFunction.cs
+++ b/samples/azure/functions/service-bus/ASBFunctions_0.1/AzureFunctions.ASBTrigger/AzureServiceBusTriggerFunction.cs
@@ -8,7 +8,6 @@ using NServiceBus.AzureFunctions.ServiceBus;
 public class AzureServiceBusTriggerFunction
 {
     private const string EndpointName = "ASBTriggerQueue";
-    private const string ConnectionStringName = "ASBConnectionString";
 
     #region Function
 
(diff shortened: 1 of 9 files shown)


samples/azure/functions/service-bus/ASBFunctions_0.1/AzureFunctions.ASBTrigger/AzureServiceBusTriggerFunction.cs
```diff
@@ -14,7 +14,7 @@ public class AzureServiceBusTriggerFunction
 
     [FunctionName(EndpointName)]
     public static async Task Run(
-        [ServiceBusTrigger(queueName: EndpointName, Connection = ConnectionStringName )]
```
timbussmann: I thought we kept the connection string here because it's non-obvious what the ASB default connection string name is (compared to ASQ where it's already set when setting up a functions account). No strong opinion though, but if we rely on the default value for the sample, shouldn't we remove the `ConnectionStringName`  constant as well or is it used somewhere?

SeanFeldman: We kept it because of the bug, but since it's fixed now, we can keep the sample simpler and coherent with the ASQ. I'll remove the constant. Thank you for catching this, @timbussmann.

PR #4511, opened 2019-09-24
editorial review of saga concurrency

adamralph:


nservicebus/sagas/concurrency.md
```diff
@@ -1,48 +1,50 @@
...
 - persistence/sql/saga-concurrency
 ---
 
-If the endpoint is configured to allow concurrent processing of messages (default) or is scaled out, it is possible that multiple messages will hit the same saga instance simultaneously. To give ACID semantics in this situation, NServiceBus uses the underlying storage to produce consistent behavior, only allowing one of messages to complete. NServiceBus handles most of this automatically but there are some caveats.
+An [endpoint](/nservicebus/concept-overview.md#endpoint) may be configured to allow [concurrent processing of messages](/nservicebus/operations/tuning.md#tuning-concurrency). On a machine with more than one logical CPU, concurrent processing is enabled by default. An endpoint may also be [scaled out](/nservicebus/architecture/scaling.md#scaling-out-to-multiple-nodes) to multiple nodes. In these scenarios, multiple messages may be handled simultaneously by a single logical saga. To ensure consistency between saga state and messages, NServiceBus ensures that only one message handling thread creates, updates, or deletes saga state in the same transaction as receiving the message and sending new messages.
```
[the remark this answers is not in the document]
adamralph: I think I prefer the more explicit version.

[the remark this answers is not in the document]
andreasohlund: Ok, fair enough, just thought it was a wall of text :)

nservicebus/sagas/concurrency.md
```diff
@@ -1,48 +1,50 @@
...
 
-When messages concurrently tries to update the same saga instance the storage will either detect and throw a concurrency exception or serialize access to the instance. Concurrency exceptions will be automatically resolved by the NServiceBus retries will.
+## Completing a saga
 
-Another option is to use a [transaction isolation level](https://msdn.microsoft.com/en-us/library/system.transactions.isolationlevel.aspx) of serializable but that causes [excessive locking](https://docs.microsoft.com/en-us/sql/t-sql/statements/set-transaction-isolation-level-transact-sql) with considerable performance degradation.
+When messages are handled concurrently by the same logical saga, and the handling of at least one message involves completing the saga, NServiceBus ensures that only one message handling thread completes the saga. This is done in the same transaction as receiving the message and sending any new messages.
```
bording: > When messages are handled concurrently by the same logical saga, and the handling of at least one message involves completing the saga, NServiceBus ensures that only one message handling thread completes the saga.

I think this sentence could benefit from some restructuring, similar to what I suggested in the other sections.

bording: Same comments about transactions and threads apply here.

nservicebus/sagas/concurrency.md
```diff
@@ -1,48 +1,50 @@
...
 
-NOTE: While `Serializable` is the default isolation level for TransactionScopes, in NServiceBus Version 4 and higher the isolation level will default to `ReadCommitted`.
+When using [SQL persistence](/persistence/sql/saga-concurrency.md), the messages are handled in serial, one after another.
```
bording: Same comment here about avoiding specific persister details.

nservicebus/sagas/concurrency.md
```diff
@@ -1,48 +1,50 @@
...
 
+For all other persisters, only one message handling thread is allowed to succeed. The saga state is deleted, and sent messages are dispatched. The other message handling threads fail, roll back, and enter [recoverability](/nservicebus/recoverability/).
+
+In both cases, when subsequent messages are handled, the saga state is not found and the [messages are discarded](/nservicebus/sagas/saga-not-found.md).
 
 ### High load scenarios
```
bording: This isn't specific to Completing a saga, so I think it should be `##` instead of `###`.

nservicebus/sagas/concurrency.md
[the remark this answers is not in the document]
adamralph: > On a machine with more than one logical CPU, concurrent processing is enabled by default.

I'm removing this sentence entirely. The details of default settings, etc. are in the linked article.

> > NServiceBus ensures that only one message handling thread creates, updates, or deletes saga state in the same transaction as receiving the message and sending new messages.

> We really shouldn't be talking about threads here since we don't manage threads at all, but instead are looking at concurrent tasks.

I'll rework this (and the rest of the document) to remove references to threads.

> Also, this sentence seems to ignore the the scaled out scenario entirely. Perhaps this paragraph needs to be split to not try and combine both scenarios? They really aren't the same thing an aren't handled the same way.

What is different about the two scenarios? As I understand it, we still guarantee the same level of transactionality, whether it's between two handlers in the same process, or two handlers in different processes.

nservicebus/sagas/concurrency.md
```diff
@@ -1,48 +1,50 @@
...
 
-Concurrent access to saga instances is divided into two scenarios;
+Saga state is created when a saga is started by a message. It may be updated when the saga handles another message. It is deleted when the saga handles another message which completes the saga. In any of these cases, multiple messages may be handled simultaneously.
 
- * Concurrently trying to create the same instance of a new saga.
- * Concurrently trying to update the same instance of an existing saga.
+## Starting a saga
 
-Each saga persister will honor the following semantics, see the specific persister documentation for implementation details.
+When messages are handled concurrently that start the same logical saga, NServiceBus ensures that only one message starts the saga. This is done in the same transaction as receiving the message and sending new messages.
```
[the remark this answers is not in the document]
adamralph: For some downstreams, you have to use the outbox, but that should be contained in the docs for those downstreams. I.e. when I choose that downstream, the docs should tell me that transactionality requires the outbox. But perhaps I'll add a note about it in this doc too.

nservicebus/sagas/concurrency.md
```diff
@@ -1,48 +1,50 @@
...
 
-Sagas are started by the message types that a handled with `IAmStartedByMessages<T>`. If messages mapped to the same saga instance are processed concurrently there is a risk that duplicates of the instance will be created.
+## Changes to saga state
 
-In this case only one message is allowed to complete processing. The others roll back and the built-in retries in NServiceBus kick in. On the next retry, the saga instance is found, the race condition is solved, and that saga instance is updated instead, see below.
+When messages are handled concurrently by the same logical saga, and involve changing the saga state, NServiceBus ensures that only one message handling thread changes the state in the same transaction as receiving the message and sending any new messages.
 
-partial: unique
+When using [SQL persistence](/persistence/sql/saga-concurrency.md), the messages are handled in serial, one after another. In this case, all the message handling threads are allowed to succeed.
```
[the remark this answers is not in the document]
adamralph: I'm going to change it to remove mentions of SQL specifically, and instead link to the persisters section for details. But I'd still like this document to mention both scenarios.

adamralph:
@bording thanks for the review and the excellent comments. I've made edits which I hope address them.

nservicebus/sagas/concurrency_unique_core_[,5].partial.md
```diff
@@ -1 +1 @@
-NOTE: When a `[Unique]` attribute is added to a saga data property, NServiceBus creates the constraint for that property. This works for both the NHibernate and the RavenDB saga persister.
\ No newline at end of file
+WARNING: NServiceBus can only guarantee this behaviour if [uniqueness is defined for the saga data](/nservicebus/sagas/message-correlation.md#uniqueness). Otherwise, more than one saga instance may be created.
```
Comment (bording) [inline nservicebus/sagas/concurrency_unique_core_[,5].partial.md]:
```suggestion
WARNING: NServiceBus can only guarantee this behavior if [uniqueness is defined for the saga data](/nservicebus/sagas/message-correlation.md#uniqueness). Otherwise, more than one saga instance may be created.
```

nservicebus/sagas/concurrency.md
```diff
@@ -1,48 +1,58 @@
...
-Another option is to use a [transaction isolation level](https://msdn.microsoft.com/en-us/library/system.transactions.isolationlevel.aspx) of serializable but that causes [excessive locking](https://docs.microsoft.com/en-us/sql/t-sql/statements/set-transaction-isolation-level-transact-sql) with considerable performance degradation.
+When using other persisters, only one message handler is allowed to succeed. The saga state is deleted, and sent messages are dispatched. The other message handlers fail, roll back, and their messages enter [recoverability](/nservicebus/recoverability/).
 
-NOTE: While `Serializable` is the default isolation level for TransactionScopes, in NServiceBus Version 4 and higher the isolation level will default to `ReadCommitted`.
+In both cases, when subsequent messages are handled, the saga state is not found and the [messages are discarded](/nservicebus/sagas/saga-not-found.md).
 
+See the [documentation for each persister](/persistence/) for details of which method is used.
 
-### High load scenarios
+## High load scenarios
```
bording: Should this be "High-load scenarios" ?

nservicebus/sagas/concurrency.md
```diff
@@ -1,48 +1,58 @@
...
-
-In this case only one message is allowed to complete processing. The others roll back and the built-in retries in NServiceBus kick in. On the next retry, the saga instance is found, the race condition is solved, and that saga instance is updated instead, see below.
+Only one message handler is allowed to succeed. The saga state is created and sent messages are dispatched. The other message handlers fail, roll back, and their messages enter [recoverability](/nservicebus/recoverability/). When those messages are retried, the existing saga state is found. Handling the messages may involve changing the state or completing the saga. See below for how those scenarios are handled.
 
 partial: unique
 
-## Concurrent access to existing saga instances
+## Changes to saga state
+
+When messages are handled simultaneously by the same logical saga, and involve changing the saga state, NServiceBus ensures that only one message handler changes the state in the same transaction as receiving the message and sending any new messages.
```
bording: The "same transaction" language that's used here and other places in the document is sticking out a bit to me. What exactly is trying to be conveyed here? Why bring transactions into it at all?

nservicebus/sagas/concurrency.md
```diff
@@ -1,48 +1,58 @@
...
+When using other persisters, only one message handler is allowed to succeed. The saga state is updated, and sent messages are dispatched. The other message handlers fail, roll back, and their messages enter [recoverability](/nservicebus/recoverability/). When those messages are retried, the process repeats. One or more of those messages may attempt to complete the saga. See below for how that scenario is handled.
+
+See the [documentation for each persister](/persistence/) for details of which method is used.
+
+## Completing a saga
+
+When messages are handled simultaneously by the same logical saga, and the handling of at least one message involves completing the saga, NServiceBus ensures that only one message handler completes the saga. This is done in the same transaction as receiving the message and sending any new messages.
 
-When messages concurrently tries to update the same saga instance the storage will either detect and throw a concurrency exception or serialize access to the instance. Concurrency exceptions will be automatically resolved by the NServiceBus retries will.
+When using some persisters, the messages are handled in serial, one after another.
```
bording: Is "in serial" common usage? It sounds awkward to me.

ramonsmits: Also, its not correct, only saga state is persisted sequentially. Saga handlers can be processed in parallel. 
```suggestion
When using a persister that support pessimistic locking, saga handlers invocations on the same saga instance (not type) are sequential, one after another due.
```

nservicebus/sagas/concurrency.md
```diff
@@ -1,48 +1,58 @@
...
+An [endpoint](/nservicebus/concept-overview.md#endpoint) may be configured to allow [concurrent handling of messages](/nservicebus/operations/tuning.md#tuning-concurrency). An endpoint may also be [scaled out](/nservicebus/architecture/scaling.md#scaling-out-to-multiple-nodes) to multiple nodes. In these scenarios, multiple messages may be handled simultaneously by a single logical saga. To ensure consistency between saga state and messages, NServiceBus ensures that only one message handler creates, updates, or deletes saga state in the same transaction as receiving the message and sending new messages. In this document, "handling" a message refers to the invocation of any saga method that processes a message, such as `IAmStartedByMessages<T>.Handle()`, `IHandleTimeouts<T>.Timeout()`, etc.
```
ramonsmits:
```suggestion
An [endpoint](/nservicebus/concept-overview.md#endpoint) may be configured to allow [concurrent handling of messages](/nservicebus/operations/tuning.md#tuning-concurrency). An endpoint may also be [scaled out](/nservicebus/architecture/scaling.md#scaling-out-to-multiple-nodes) to multiple instances. In these scenarios, multiple messages may be handled simultaneously by a single logical saga instance. To ensure consistency between saga state and messages on the same saga type instance, NServiceBus ensures that only one message handler creates, updates, or deletes saga state.

In this document, "handling" a message refers to the invocation of any saga method that processes a message, such as `IAmStartedByMessages<T>.Handle()`, `IHandleTimeouts<T>.Timeout()`, etc.
```

ramonsmits: I removed the transaction bit as not all persisters support transactions

nservicebus/sagas/concurrency.md
```diff
@@ -1,48 +1,58 @@
...
 
-Concurrent access to saga instances is divided into two scenarios;
+NOTE: When using some transports, the [outbox](/nservicebus/outbox/) must be used in order to provide this level of consistency.
```
ramonsmits: Some persisters do not support outbox and saga consistency cannot be guaranteed unless the saga state itself has outbox data embedded.

nservicebus/sagas/concurrency.md
```diff
@@ -1,48 +1,58 @@
...
 
- * Concurrently trying to create the same instance of a new saga.
- * Concurrently trying to update the same instance of an existing saga.
+Saga state is created when a saga is started by a message. It may be updated when the saga handles another message. It is deleted when the saga handles another message which completes the saga. In any of these cases, multiple messages may be handled simultaneously.
 
-Each saga persister will honor the following semantics, see the specific persister documentation for implementation details.
+## Starting a saga
 
-## Concurrent access to non-existing saga instances
+When messages that start the same logical saga are handled simultaneously, NServiceBus ensures that only one message starts the saga. This is done in the same transaction as receiving the message and sending new messages.
```
ramonsmits:
```suggestion
When messages that start the same logical saga instance are handled simultaneously, NServiceBus ensures that only one message starts the saga by enforcing uniqueness on the saga correlation identifier. First writer wins.
```

nservicebus/sagas/concurrency.md
ramonsmits:
```suggestion
When messages are handled simultaneously by the same logical saga instance, and involve modifying the saga state, NServiceBus ensures that only one message handler succeeds in persisting the state. Any messages to be dispatched only happen if persisting the saga succeeds.
```

nservicebus/sagas/concurrency.md
```diff
@@ -1,48 +1,58 @@
...
+Only one message handler is allowed to succeed. The saga state is created and sent messages are dispatched. The other message handlers fail, roll back, and their messages enter [recoverability](/nservicebus/recoverability/). When those messages are retried, the existing saga state is found. Handling the messages may involve changing the state or completing the saga. See below for how those scenarios are handled.
 
 partial: unique
 
-## Concurrent access to existing saga instances
+## Changes to saga state
+
+When messages are handled simultaneously by the same logical saga, and involve changing the saga state, NServiceBus ensures that only one message handler changes the state in the same transaction as receiving the message and sending any new messages.
+
+When using some persisters, the messages are handled in serial, one after another. In this case, all message handlers are allowed to succeed.
```
ramonsmits:
```suggestion
When using a persister that supports pessimistic locking, saga handlers invocations on the same saga instance (not type) are sequential, one after another. 
```

nservicebus/sagas/concurrency.md
```diff
@@ -1,48 +1,58 @@
...
-NOTE: While `Serializable` is the default isolation level for TransactionScopes, in NServiceBus Version 4 and higher the isolation level will default to `ReadCommitted`.
+In both cases, when subsequent messages are handled, the saga state is not found and the [messages are discarded](/nservicebus/sagas/saga-not-found.md).
 
+See the [documentation for each persister](/persistence/) for details of which method is used.
 
-### High load scenarios
+## High load scenarios
 
-Under extreme high load like batch processing, concurrent access to saga instance might lead to messages being moved to the error queue due to the NServiceBus retries being exhausted.
+Under high load, simultaneous access to saga state may lead to a high number of retries. Those retries may be exhausted, resulting in messages being moved to the error queue.
```
ramonsmits:
```suggestion
Under high load, concurrent write access to saga state may lead to a high number of retries. Those retries may be exhausted, resulting in messages being moved to the error queue.
```

nservicebus/sagas/concurrency.md
adamralph: Probably. I'll change it.

adamralph:
```suggestion
## High-load scenarios
```

nservicebus/sagas/concurrency.md
adamralph: Agreed. "transaction" is a bit overloaded, and could confuse things. I'll find an alternate term/phrase.

nservicebus/sagas/concurrency.md
adamralph: > Is "in serial" common usage? It sounds awkward to me.

Probably not. I'll change it.

adamralph: > Also, its not correct, only saga state is persisted sequentially. Saga handlers can be processed in parallel.

I'm not sure that's correct, and your suggestion doesn't suggest that either. Because pessimistic locking is used, the handler are effectively executed one after another, since the method is only invoked after the data has been read, which doesn't occur until after another handler has finished and the data from that handler is persisted. If that were not the case, and they were executed in parallel, then they would have dirty reads, and the pessimistic locking would not be occurring.

nservicebus/sagas/concurrency.md
adamralph: @ramonsmits I'm not sure what you mean but this. Let's discuss.

PR #4516, opened 2019-09-25
Add information about mongo testing shared transactions

boblangley:


persistence/mongodb/index.md
```diff
@@ -63,6 +63,10 @@ snippet: MongoDBHandlerSharedTransaction
 
 WARN: In order to participate in the shared transaction the MongoDB session must be passed into collection API calls as demonstrated above.
 
+#### Testing
+
+A `TestableMongoSynchronizedStorageSession` class has been provided to facilitate [testing a handler](/nservicebus/testing/) that utilizes the shared transaction feature.
```
bording: Do you think we need to mention the namespace the class is in?

boblangley: I thought about it. It isn't _strictly_ necessary because it should already be referenced when using the testing library anyhow. However, it would hurt to add it either.

If we add it, should I just add it to the class name, or mention the namespace separately?

bording: How about:

The `TestableMongoSynchronizedStorageSession` class in the `NServiceBus.Testing` namespace has been provided to facilitate [testing a handler](/nservicebus/testing/) that utilizes the shared transaction feature.

PR #4527, opened 2019-10-02
Document DisablePublishing API

timbussmann:
related to https://github.com/Particular/NServiceBus/pull/5430

### DIFF
--- a/nservicebus/messaging/publish-subscribe/index_disable-publishing_core_[7.2,).partial.md
+++ b/nservicebus/messaging/publish-subscribe/index_disable-publishing_core_[7.2,).partial.md
@@ -1,0 +1,5 @@
+##### Disabling publishing
+
+By default, NServiceBus requires a subscription persister to be configured. If an endpoint does not publish any events, the publish functionality can be disabled to avoid providing a subscription persister:
+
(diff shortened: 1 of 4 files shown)


nservicebus/messaging/publish-subscribe/index_disable-publishing_core_[7.2,).partial.md
```diff
@@ -0,0 +1,5 @@
+##### Disabling publishing
+
+By default, NServiceBus requires a subscription persister to be configured. If an endpoint does not publish any events, the publish functionality can be disabled to avoid providing a subscription persister:
```
SzymonPobiega:
```suggestion
By default, NServiceBus requires a subscription persister to be configured. If an endpoint does not have access to any supported data store and does not publish any events, the publishing infrastructure can be disabled:
```

timbussmann: I wanted to avoid that statement given that it enables you to not provide a storage in the first place instead of only opting for that in case your storage doesn't support it. In case you don't use sagas, it becomes less and less common to have a storage in the first place?

PR #4529, opened 2019-10-03
Signing sample

DavidBoike:
For inclusion in a story about behaviors.

A ReceivingEndpoint, which receives messages from SignedSender (which works) and an UnsignedSender (which is rejected).


samples/pipeline/message-signing/Core_7/Shared/SharedKeys.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Text;
+
+static class SharedKeys
+{
+    public static byte[] SigningKey { get; }
+
+    static SharedKeys()
+    {
+        SigningKey = Encoding.UTF8.GetBytes("This is not a secure encryption key, but this is just a sample. Seriously, this is not how you should be creating encryption/signing keys.");
```
SeanFeldman:
```suggestion
        SigningKey = Encoding.UTF8.GetBytes("This is not a secure encryption key, but just a sample. For production, do not create encryption/signing keys using this method.");
```

samples/pipeline/message-signing/Core_7/Shared/SignatureVerificationBehavior.cs
```diff
@@ -0,0 +1,38 @@
...
+#region SignatureVerificationBehavior
+
+class SignatureVerificationBehavior :
+    Behavior<IIncomingPhysicalMessageContext>
+{
+    static readonly ILog log = LogManager.GetLogger<SignatureVerificationBehavior>();
+
+    public override Task Invoke(IIncomingPhysicalMessageContext context, Func<Task> next)
+    {
+        using (var hmac = HMAC.Create("hmacsha256"))
```
SeanFeldman: Shouldn't the header from 
https://github.com/Particular/docs.particular.net/pull/4529/files#diff-1ba212019583e2df226e04f2981af784R21 be used here?

PR #4531, opened 2019-10-04
Exception data logging

timbussmann:
documents the changes introduced with https://github.com/Particular/NServiceBus/pull/5437


nservicebus/logging/index_exception-data_core_[7.2,).partial.md
```diff
@@ -0,0 +1,3 @@
+## Additional exception data
+
+Starting from NServiceBus version 7.2, exceptions from failing message handlers might contain additional error information in the `Exception.Data` property. While the default logger exposes those information automatically, other loggers might require additional configuration.
```
andreasohlund: Should we make this more generic since it can contain "Data" from all parts of the pipeline and not just the handlers?

timbussmann: yeah makes sense

andreasohlund: Would it make sense to add a section telling users that they can add their own behaviour to add more data? Eg. adding the "tenant-id" in a multi-tenant solution?

timbussmann:
this one should be good for a re-review

PR #4543, opened 2019-10-14
Document transport seam options in SC

andreasohlund:
Document:
 * [x] RabbitMQ
 * [x] ASB
 * [x] SQS
 * [x] SQL


tmasternak:
@andreasohlund I've added ASB changes to this PR as well. I propose to add SQS and SQL options as well. 

Thoughts?

andreasohlund:
Makes sense, can you re-title the PR to reflect this?

tmasternak:
@andreasohlund I can take the SQL and SQS.

PR #4546, opened 2019-10-15
MsmqP does not support scaled-out publishers

DavidBoike:


persistence/msmq/subscription.md
```diff
@@ -7,6 +7,8 @@ related:
  - samples/msmq/persistence
 ---
 
+WARN: MSMQ Persistence does not support scaled-out publishers. This is because the MSMQ storage is local to the machine, and a subscription message will only be delivered to one endpoint instance of a given logical endpoint group, and only that instance will be able to update its information, while other instances remain unaware of the new subscriber.
```
SeanFeldman: Removed the word "group". It confused more than added value.

```suggestion
WARN: MSMQ Persistence does not support scaled-out publishers. This is because the MSMQ storage is local to the machine, and a subscription message will only be delivered to one endpoint instance of a given logical endpoint, and only that instance will be able to update its information, while other instances remain unaware of the new subscriber. 
```

PR #4559, opened 2019-10-24
New support policy for ServiceControl

andreasohlund:
Proposal to change the supported version policy for ServiceControl

## New policy

- Major versions are supported for a period of one year after the release of the next major version.
- Only the latest minor version of a previous major version is supported.

### Only the current major (current rule)

Forcing users to upgrade and take the risk to move to the next major versions when they are in message loss or other similar critical scenarios would hurt our reputation and cause too much stress for the customers. There is a time sensitivity component to critical bugs. In cases like that, we want to avoid needing the customers to take on more risks like upgrading hardware, installing new .NET Framework versions etc.

### 3 years including minors (same as core)

The policies we apply to the Core are not applicable to ServiceControl based on the following:

* The effort of upgrading ServiceControl is less involved from all perspectives (Code, Risk, Testing, APIs and more) than a core upgrade because Coding is not involved. A ServiceControl upgrade has an operational impact only.
* A core minor might require users to recompile. That's why we support minors as well. This does not apply to ServiceControl so updating to the latest minor withing their current major is fine.


DavidBoike:
These notes don't apply to any specific line so I'll leave them here:

* It would be good if https://docs.particular.net/nservicebus/upgrades/support-policy had a clarifier that the support policy listed there doesn't apply to the apps and then linked to the pages this PR is introducing.
* The SI and SP pages seem open to interpretation. Latest major sure, but what does that mean for minors? Patches? Does the text effectively mean that only the latest version is supported?
* I believe if you have a valid license but lapsed upgrade protection, and you try to upgrade SI (to a version built after upgrade protection lapsed), it won't launch. So those customers would be forced to use an unsupported version of SI or renew their upgrade protection. I wanted to verify that this is as intended, and also might want to specifically ask the sales team about that detail if you haven't already.
* These documents should also be in the menu.
* Content verification is complaining because you can't have directories without an index.md file.

andreasohlund:
@DavidBoike 

>It would be good if https://docs.particular.net/nservicebus/upgrades/support-policy had a clarifier that the support policy listed there doesn't apply to the apps and then linked to the pages this PR is introducing.

This already existed, see https://docs.particular.net/nservicebus/upgrades/support-policy#faq

>The SI and SP pages seem open to interpretation. Latest major sure, but what does that mean for minors? Patches? Does the text effectively mean that only the latest version is supported?

We decided to scope SP and SI out and raise follow up issues to deal with them

>I believe if you have a valid license but lapsed upgrade protection, and you try to upgrade SI (to a version built after upgrade protection lapsed), it won't launch. So those customers would be forced to use an unsupported version of SI or renew their upgrade protection. I wanted to verify that this is as intended, and also might want to specifically ask the sales team about that detail if you haven't already.

SI is now scoped out


>These documents should also be in the menu.

Fixed

PR #4563, opened 2019-10-25
ServiceControl 4.4.1

mikeminutillo:
Documentation updates for [ServiceControl 4.4.1](https://github.com/Particular/ServiceControl/milestone/166)


DavidBoike:
@mikeminutillo Is there a config setting for Audit instances as well?

mikeminutillo:
@DavidBoike no. We left Audit instances as they are, so they will shut down and restart. It's less of an issue for an Audit instance to be down.

PR #4566, opened 2019-10-25
Fixes 404 and mentions Guaranteed Messaging

ramonsmits:


ramonsmits:
@dvdstelt good catch :-) I didn't check the 2nd link for its anchor.

PR #4568, opened 2019-10-28
Clarify ServiceControl 3to4 upgrade guide

DavidBoike:
* Note explicitly that there are no lengthy data migrations
* Be more specific about how original SC database can be compacted after 1 retention period has elapsed.

### DIFF
--- a/servicecontrol/upgrades/3to4/index.md
+++ b/servicecontrol/upgrades/3to4/index.md
@@ -30,7 +30,7 @@ ServiceControl Management version 4 cannot be used to edit ServiceControl instan
 Upgrades that include the separate Audit embedded database will increase disk usage since databases are not automatically compacted. The new Audit embedded database will grow to the same peak storage usage as the original ServiceControl instance embedded audit message storage usage unless the original instance database is compacted after the data is removed via the audit retention process. This could result in as much as double the data usage.
 
 {{NOTE:
-After the upgrade is complete, the Audit information contained in the original ServiceControl instance will become read-only, but will continue to be deleted as it passes the time specified by the retention policy. However, since RavenDB does not release storage space back to the OS, the database will continue to be the same size.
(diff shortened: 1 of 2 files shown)


servicecontrol/upgrades/3to4/index.md
```diff
@@ -25,7 +29,11 @@ ServiceControl Management version 4 cannot be used to edit ServiceControl instan
 
 Upgrades that include the separate Audit embedded database will increase disk usage since databases are not automatically compacted. The new Audit embedded database will grow to the same peak storage usage as the original ServiceControl instance embedded audit message storage usage unless the original instance database is compacted after the data is removed via the audit retention process. This could result in as much as double the data usage.
 
-See [Compacting RavenDB](/servicecontrol/db-compaction.md) for instructions on compacting the database of the original ServiceControl instance once old audit messages have been cleaned up by the retention policy.
+{{NOTE:
+After the upgrade is complete, the Audit information contained in the original ServiceControl instance will become read-only, but will continue to be deleted as it passes the time specified by the retention policy. However, since RavenDB does not release storage space back to the OS, the database will continue to be the same size.
```
danielmarbach:
```suggestion
After the upgrade is complete, the Audit information contained in the original ServiceControl instance will become read-only, but will continue to be deleted as it passes the time specified by the [retention policy](/servicecontrol/creating-config-file.md#data-retention). However, since the internal database does not release storage space back to the OS, the database will continue to be the same size.
```

danielmarbach: I tried to get rid of RavenDB mentioning as well

servicecontrol/upgrades/3to4/index.md
```diff
@@ -11,6 +11,10 @@ Upgrading ServiceControl from version 3 to version 4 is a major upgrade and requ
 
 ## Planning
 
+### Time for upgrade
+
+This upgrade does not contain any data migrations, so the size of the database does not have any impact on the time to perform the upgrade.
```
danielmarbach: I think it would be great to add something similar to the 2 to 3 update because that one also doesn't do data migration

PR #4571, opened 2019-10-30
Satellite Gateway notifications behavior

ramonsmits:


nservicebus/recoverability/subscribing-to-error-notifications.md
```diff
@@ -18,6 +18,14 @@ Error notifications are available for several events:
  * When a [delayed retry](/nservicebus/recoverability/#delayed-retries) occurs.
  * When a message fails all retries and is forwarded to the error queue.
 
+## Satellite behavior
+
+Since Version 6 errors happening in satellites will not trigger error notification and requires a [custom retry policy configuration per satellite](/nservicebus/satellites/#implementing-a-satellite-managing-errors).
+
+NOTE: The Gateway is a satellite see the [Gateway Version 1 to 2 upgrade guide](/nservicebus/upgrades/gateway-1to2.md) for more information.
```
andreasohlund: Does it make sense to link to the GW here? Feels off to me, there are other satellites and we don't link to them?

nservicebus/recoverability/subscribing-to-error-notifications.md
```diff
@@ -18,6 +18,13 @@ Error notifications are available for several events:
  * When a [delayed retry](/nservicebus/recoverability/#delayed-retries) occurs.
  * When a message fails all retries and is forwarded to the error queue.
 
+## Satellite behavior
+
+Since Version 6 errors happening in satellites like the [Gateway](/nservicebus/gateway/) and [TimeoutManager](/nservicebus/messaging/timeout-manager) will not trigger error notification and requires a [custom retry policy configuration per satellite](/nservicebus/satellites/#implementing-a-satellite-managing-errors). The gateway has its own [custom gateway retry policy](/nservicebus/gateway/#using-the-gateway-recoverability) API.
```
andreasohlund: >and requires a custom retry policy configuration per satellite

This isn't relevant for users right? They can only do something if we have chosen to expose an api, like we did for the GW? But eg. for the timeoutmanager they can no longer subscribe?

ramonsmits: mmm, seems like it. Now that I'm looking at that doco that seems only possible to do in the context of the feature.

PR #4574, opened 2019-10-31
Show discard with callbacks

danielmarbach:


samples/callbacks/Callbacks_3/Sender/Program.cs
```diff
@@ -11,6 +12,15 @@ static async Task Main()
         endpointConfiguration.UsePersistence<LearningPersistence>();
         endpointConfiguration.UseTransport<LearningTransport>();
 
+        endpointConfiguration.Recoverability().CustomPolicy((config, context) =>
+        {
+            if (context.Exception is InvalidOperationException invalidOperationException && invalidOperationException.Message.StartsWith("No handlers could be found"))
```
ramonsmits:
```suggestion
            if (context.Exception is InvalidOperationException invalidOperationException && invalidOperationException.Message.StartsWith("No handlers could be found", StringComparison.OrdinalIgnoreCase))
```

danielmarbach: Thanks ramon. The PR was only for demonstration purposes and I don't intend to bring it in or do you think this would be a valuable addition?

ramonsmits: @danielmarbach I think this should be the default behavior for callbacks. Can't we make it into an issue for a maintenance release or maybe add this to the callback docs as a snippet?

danielmarbach: I gave it a stab, see https://github.com/Particular/docs.particular.net/pull/4582

PR #4578, opened 2019-11-04
Remove multihosting sample

WilliamBZA:
Removes the multihosting sample as multihosting should be avoided at all costs.

I've added all the people involved with the discussion.


timbussmann:
afaik this has been a valid approach when briding transports. Not sure we should just remove this.

andreasohlund:
>afaik this has been a valid approach when briding transports. Not sure we should just remove this.

agree, that said using the Router would be my preferred approach now?

adamralph:
> multihosting should be avoided at all costs

I'm yet to see an explanation of why this is.

WilliamBZA:
> afaik this has been a valid approach when briding transports

Then how about a sample for that instead of a general approach?

> I'm yet to see an explanation of why this is.

Off the top of my head: code is needed to isolate DI containers, and assembly scanning becomes a pain. Isolating by process solves both of those problems. I'm not sure what other edge cases there might be too.

> I rather see a big fat warning...

We can have a warning on a page somewhere and not have a sample showing people how to do the thing we're warning against. Steer people towards success rather than show them how to get it wrong.

Either way, I guess this has taken more than 30 minutes so I'm withdrawing the PR.

PR #4582, opened 2019-11-06
Showing discard sample

danielmarbach:


ramonsmits:
@danielmarbach Calling `Discard`, what is the behavior? Is it then send to the audit queue with some discard header or something? Can't read what the behavior is from https://docs.particular.net/nservicebus/recoverability/custom-recoverability-policy#implement-a-custom-policy-full-customization

danielmarbach:
Discard means the message is consumed. But no header manipulation is done

We log info

https://github.com/Particular/NServiceBus/blob/ead779f33e8bdec6844ee99a892729cfd7b7f0bc/src/NServiceBus.Core/Recoverability/RecoverabilityExecutor.cs#L29

PR #4583, opened 2019-11-06
SQL Server transport version 5

mikeminutillo:
Doco for https://github.com/Particular/NServiceBus.SqlServer/pull/518


transports/upgrades/sqlserver-4to5.md
```diff
@@ -0,0 +1,118 @@
...
+var transport = endpointConfiguration.UseTransport<SqlServerTransport>();
+
+var pubSubCompat = transport.EnableMessageDrivenPubSubCompatibilityMode();
+pubSubCompat.RegisterPublisher(eventType, publisherEndpoint);
+pubSubCompat.RegisterPublisher(assembly, publisherEndpoint);
+pubSubCompat.RegisterPublisher(assembly, namespace, publisherEndpoint);
+pubSubCompat.SubscriptionAuthorizer(authorizationFunction);
+```
+
+NOTE: The `routing.DisablePublishing()` API has been deprecated and should be removed. This API was created to allow an endpoint to run without a configured subscription persistence. On version 5 and above, a subscription persistence is not required unless the endpoint runs in backwards compatibility mode.
```
mikeminutillo: @SzymonPobiega are there scenarios where DisablePublishing still makes sense when operating in compatibility mode? i.e. We had this before so we still don't need to configure a subscription persistence but we want to subscribe to events coming from other endpoints.

transports/upgrades/sqlserver-4to5.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Configure subscription caching
+
+Subscription information can be cached for a given period of time so that it does not have to be loaded every single time an event is being published. The longer the cache period is, the higher the chance that new subscribers miss some events.
+
+Because of that there is no good default value for the subscription caching period. It has to be specified by the user. In systems where subscriptions are static the caching period can be relatively long. To configure it, use following API:
+
+```csharp
+var transport = endpointConfiguration.UseTransport<SqlServerTransport>();
+var subscriptions = transport.PubSub();
```
mikeminutillo: @SzymonPobiega Should we rename this API to `.SubscriptionSettings()` to match the naming used in SQL Persistence? This would allow us to have `transport.SubscriptionSettings().CacheFor(x)`.

SzymonPobiega: I like `SubscriptionSettings` 👍 . I'll change it and try to push SQLT 5

transports/upgrades/sqlserver-4to5.md
SzymonPobiega: `SubscriptionMigrationMode` in the core does not look at the `DisablePublishing` settings so without changing the core we would not be able to make use of that feature -- in the compatibility mode subscription persistence is mandatory. I think that what you wrote is OK 👍

transports/sql/native-publish-subscribe.md
```diff
@@ -0,0 +1,42 @@
+---
+title: SQL Server Native Publish Subscribe
+summary: Describes the native publish subscribe implementation in the SQL Server transport
+reviewed: 2019-11-08
+component: SqlTransport
+versions: '[5,)'
+---
+
+The SQL Server transport implements the publish-subscribe pattern. On version 4 and below, this feature relies on message-driven pub-sub which requires a separate persistence for storage of subscription information. On version 5 and above, the transport handles subscription information natively and a separate persistence is not required.
```
SzymonPobiega: I think most references in the doco use "In version", not "On version" . I am not an English expert. Is there a subtle difference here?

mikeminutillo: You are correct. I will fix it

mikeminutillo: The endpoint runs _on_ a version of the transport. The features exist _in_ a version of the transport.

PR #4592, opened 2019-11-12
Hosting updates

timbussmann:
Some updates to the hosting documentation in relation to the NServiceBus.Extensions.Hosting package.

I removed [this sample](https://docs.particular.net/samples/web/send-from-mvc-controller/) as it seems redundant by looking at [this sample](https://docs.particular.net/samples/web/asp-mvc-application/)


SeanFeldman:
I wonder if we should add something to the hosting landing page (https://docs.particular.net/nservicebus/hosting/) as well.

SeanFeldman:
Regarding https://github.com/Particular/docs.particular.net/pull/4592#issuecomment-553010592 - no need as the extensions-hosting article will show up in the related items.

PR #4593, opened 2019-11-13
Autofac register Task<IEndpointInstance> to not use null field

ramonsmits:


ramonsmits:
Closed as this should not be the approach taken. Customer should be using the new https://docs.particular.net/nservicebus/dependency-injection/#externally-managed-mode

PR #4595, opened 2019-11-15
Change incoming from header to mutator

WilliamBZA:
Changes the setting of the identity from explicitly loading the value in the handler to setting it via the mutator

### DIFF
--- a/samples/username-header/Core_6/Endpoint2/HandlerUsingCustomHeader.cs
+++ b/samples/username-header/Core_6/Endpoint2/HandlerUsingCustomHeader.cs
@@ -1,4 +1,5 @@
-﻿using System.Threading.Tasks;
+﻿using System.Threading;
+using System.Threading.Tasks;
 using NServiceBus;
 using NServiceBus.Logging;
 
@@ -11,9 +12,8 @@ public class HandlerUsingCustomHeader :
 
     public Task Handle(MyMessage message, IMessageHandlerContext context)
     {
-        var headers = context.MessageHeaders;
-        var usernameFromHeader = headers["UserName"];
-        log.Info($"Username extracted from header: {usernameFromHeader}");
+        var usernameFromThread = Thread.CurrentPrincipal.Identity.Name;
+        log.Info($"Username extracted from Thread.CurrentPrincipal: {usernameFromThread}");
         return Task.CompletedTask;
     }
 }
--- a/samples/username-header/Core_6/Shared/UsernameMutator.cs
+++ b/samples/username-header/Core_6/Shared/UsernameMutator.cs
@@ -1,4 +1,5 @@
-﻿using System.Threading;
+﻿using System.Security.Principal;
+using System.Threading;
 using System.Threading.Tasks;
 using NServiceBus.Logging;
 using NServiceBus.MessageMutator;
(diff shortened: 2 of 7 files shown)


samples/username-header/Core_6/Endpoint2/HandlerUsingCustomHeader.cs
```diff
@@ -11,9 +12,8 @@ public class HandlerUsingCustomHeader :
 
     public Task Handle(MyMessage message, IMessageHandlerContext context)
     {
-        var headers = context.MessageHeaders;
-        var usernameFromHeader = headers["UserName"];
-        log.Info($"Username extracted from header: {usernameFromHeader}");
+        var usernameFromThread = Thread.CurrentPrincipal.Identity.Name;
```
ramonsmits: `Thread.CurrentPrincipal` isn't guaranteed with async and shouldn't be used. Instead, `IPrincipal` should be resolved via DI by having an resolver for `IPrincipal` or by using a funky context extension like `.GetIdentity()`.

samples/username-header/Core_6/Shared/UsernameMutator.cs
```diff
@@ -1,15 +1,26 @@
...
-    IMutateOutgoingTransportMessages
+    IMutateOutgoingTransportMessages,
+    IMutateIncomingTransportMessages
 {
     static ILog log = LogManager.GetLogger("Handler");
 
+    public Task MutateIncoming(MutateIncomingTransportMessageContext context)
+    {
+        log.Info("Reading headers to set Thread.CurrentPrincipal");
+        var identityName = context.Headers["UserName"];
```
ramonsmits: So the approach to use the pipeline is good, but not sure if a mutator works if you would like to use DI. Likely the resolver needs to be initialized in the first stage.

samples/username-header/Core_6/Shared/UsernameMutator.cs
```diff
@@ -1,15 +1,26 @@
...
+        Thread.CurrentPrincipal = new GenericPrincipal(new GenericIdentity(identityName), new string[] { });
```
ramonsmits: This won't work currently with async AFAIK.

samples/username-header/Core_6/Endpoint2/HandlerUsingCustomHeader.cs
danielmarbach: The problem is like @ramonsmits pointed out that `Thread.CurrentPrincipal` behavior is different depending on the runtime. For the .NET Framework floating the principle over the ExecutionContext works. For .NET Core 3 it works as well because the behavior has been aligned as outline in https://github.com/dotnet/corefx/issues/34489#issuecomment-454860866. For pre 3.0 .NET Core the behavior is to not float the principal into the execution context. So if you execute the following code under .NET Core 3.0 

```
            var identity = new ClaimsIdentity();
            identity.AddClaim(new Claim(ClaimTypes.Name, "User1"));
            Thread.CurrentPrincipal = new ClaimsPrincipal(identity);

            Console.WriteLine($"Current thread: {Thread.CurrentThread.ManagedThreadId}");
            // Check the current user
            Console.WriteLine($"Current user: {Thread.CurrentPrincipal?.Identity.Name}");

            // For the method to complete asynchronously
            await Task.Yield();

            Console.WriteLine($"Current thread: {Thread.CurrentThread.ManagedThreadId}");
            // Check the current user after
            Console.WriteLine($"Current user: {Thread.CurrentPrincipal?.Identity.Name}");
```

the result is

```
Current thread: 1
Current user: User1
Current thread: 4
Current user: User1
```

if executed under .NET Core 2.2 the result is

```
Current thread: 1
Current user: User1
Current thread: 4
Current user:
```

danielmarbach: This is one of the reasons why the migration guide here mentions to avoid `Thread.CurrentPrincipal` or `ClaimsPrincipal.Current`

https://docs.microsoft.com/en-us/aspnet/core/migration/claimsprincipal-current

and the guidance has four ways of doing it

https://docs.microsoft.com/en-us/aspnet/core/migration/claimsprincipal-current?view=aspnetcore-3.0#retrieve-the-current-user-in-an-aspnet-core-app

danielmarbach: I will send it a very rough proposal to discuss

WilliamBZA:
Ok, withdrawing then.

danielmarbach:
Alternative raised https://github.com/Particular/docs.particular.net/pull/4596

PR #4598, opened 2019-11-20
UseNServiceBus must be specified before ConfigureWebHostDefaults

timbussmann:
See https://github.com/Particular/NServiceBus.Extensions.Hosting/issues/23

Looking at the implementation, the asp.net web hosting is also using an `IHostedService`. Given those are Started in order and the web host seems to start accepting requests right after that, there is a race condition which might lead to the message session not being available yet (and therefore breaking the lazy). Ensuring that the NServiceBus hosted service starts first should resolve this problem.

Documenting this behavior for now, going to raise a dedicated issue about this in a bit.

### DIFF
--- a/nservicebus/hosting/extensions-hosting.md
+++ b/nservicebus/hosting/extensions-hosting.md
@@ -17,6 +17,7 @@ snippet: extensions-host-configuration
 
 This code will register the endpoint with the hosting infrastructure and automatically start and stop it based on the host's application lifetime.
 
+WARNING: `UseNServiceBus` must be specified before any other service (e.g. `ConfigureWebHostDefaults`) which requires access to the `IMessageSession`.
 
 ## Dependency injection integration
 


nservicebus/hosting/extensions-hosting.md
```diff
@@ -17,6 +17,7 @@ snippet: extensions-host-configuration
 
 This code will register the endpoint with the hosting infrastructure and automatically start and stop it based on the host's application lifetime.
 
+WARNING: `UseNServiceBus` must be specified before any other service (e.g. `ConfigureWebHostDefaults`) which requires access to the `IMessageSession`.
```
danielmarbach: I think that is OK given the MS DI based eco system reintroduces ordering problems everywhere (sorry for my slight rant)

timbussmann: > I think that is OK given the MS DI based eco system reintroduces ordering problems everywhere (sorry for my slight rant)

yeah I'll raise an issue anyway to discuss this (I think we need to avoid the Lazy to not break the whole application at least).

danielmarbach: That's an interesting thought because the solution would essentially be that core needs to decorate the session and start the endpoint on demand if needed and do that in way the allows reentrancy without starting multiple times. But honestly I'm really on the fence whether it should. The whole service collection ecosystem is busted when it comes to ordering, see

> Make your Autofac registrations. Order is important!
https://autofaccn.readthedocs.io/en/latest/integration/netcore.html

the options pattern

> You can add multiple configuration providers. Configuration providers are available from NuGet packages and are applied in the order that they're registered. 

https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-3.0

> An inconsistent options state may exist due to the ordering of service registrations.

https://docs.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-3.0#accessing-options-during-startup

>   // Order added affects options setup order

https://offering.solutions/blog/articles/2017/02/07/the-difference-between-addmvc-and-addmvccore/

> Each IStartupFilter can add one or more middlewares in the request pipeline. The filters are invoked in the order they were added to the service container.

https://docs.microsoft.com/en-us/aspnet/core/fundamentals/startup?view=aspnetcore-3.0#extend-startup-with-startup-filters

> The order that middleware components are added in the Startup.Configure method defines the order in which the middleware components are invoked on requests and the reverse order for the response. The order is critical for security, performance, and functionality.

https://docs.microsoft.com/en-us/aspnet/core/fundamentals/middleware/index?view=aspnetcore-3.0#middleware-order

danielmarbach: So in short the IServiceCollection ecosystem assumes ordering

SimonCropp:
how would this work for a windows service host (using generic host without the web workload) where there is no `ConfigureWebHostDefaults`?

timbussmann:
> how would this work for a windows service host (using generic host without the web workload) where there is no ConfigureWebHostDefaults?

not sure I fully understand. If there is no second hosted service requiring the message session, there is no order to respect and you can just use `UseNServiceBus`.

johnkors:
@timbussmann , what do you mean? Does this not affect other custom (our own) IHostedService registrations beyond the ones registred via the WebHostDefaults builder?

timbussmann:
other hosted services requiring a message session need to be defined/registered before `UseNServiceBus`. When using other hosted services, there is currently a known limitation that you can't inject `IMessageSession` directly into a hosted service's constructor. Instead, you need to inject the service provider and resolve the message session once the service is being started (or at send time). Does that help @johnkors ? Otherwise, can you give me some more details on what you're trying to do?

PR #4610, opened 2019-11-27
Align maintenance mode

danielmarbach:


SeanFeldman:
YAML - because JSON and XML were not enough 😂

PR #4611, opened 2019-11-28
Log and exception messages are not part of public api

janovesk:


boblangley:
I added a commit. First to fix the content verification, and secondly to put this under a new sub-section listing where we have made interpretations or deviate from SemVer.

danielmarbach:
Rebased. I think this should be merged and internally announced

danielmarbach:
Added an FYI to the engineering channel https://particularsoftware.slack.com/archives/C0316LBE3/p1576851131028400

PR #4612, opened 2019-11-29
ASBS Lock duration fixed at 5 minutes

ramonsmits:


transports/azure-service-bus/configuration.md
```diff
@@ -18,7 +18,10 @@ WARNING: Entity creation settings are applied only at creation time of the corre
  * `EnablePartitioning()`: Partitioned entities offer higher availability, reliability, and throughput over conventional non-partitioned queues and topics. For more information about partitioned entities [see the Microsoft documentation on partitioned messaging entities](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-partitioning).
  * `SubscriptionNameShortener(Func<string, string>)`: By default subscription names are derived from the endpoint name, which may exceed the maximum length of subscription names. This callback allows for a replacement name for the subscription. Subscription names must adhere to the limits outlined in [the Microsoft documentation on subscription creation](https://docs.microsoft.com/en-us/rest/api/servicebus/create-subscription).
  * `RuleNameShortener(Func<string, string>)`: By default rule names are derived from the message type's full name, which may exceed the maximum length of rule names. This callback allows for a replacement name for the rule. Rule names must adhere to the limits outlined in [Service Bus quotas](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-quotas).
- 
+
+Fixed defaults:
```
SeanFeldman: Perhaps we should document not fixed defaults, but a bit more:
- why lock duration is set to 5 mins
- default entity size
- no auto lock renewal (and reason)
- no entity modification if one is found
- etc

PR #4625, opened 2019-12-16
Update docker sample netcore31

danielmarbach:
image

Removed the windows workaround in the program files because

1. It seems docker has fixed it, [issue ](https://github.com/moby/moby/issues/25982)is still open but PRs have been merged
1. The rabbitmq image in docker-compose was never available under windows so windows container would not have worked with that sample anyway

Furthermore the new recommendation would anyway be to use `Microsoft.Extensions.Hosting` that takes care of the lifecycle. I figured though it would unnecessarily complicate this example and did not introduce it here


danielmarbach:
Alternative targetting NETCOREAPP 2.2 https://github.com/Particular/docs.particular.net/pull/4626 until the build agents are ready

mauroservienti:
@danielmarbach I solved the conflicts and rerun the build, all is green now. This is good to go IMO.

PR #4633, opened 2019-12-19
Slightly more helpful licensing troubleshooting hint

### LINKED ISSUE #4367: Feedback: 'Licensing'
Feedback for 'Licensing' https://docs.particular.net/nservicebus/licensing/

Location in GitHub: https://github.com/Particular/docs.particular.net/blob/master/nservicebus/licensing/index.md

Question regarding https://docs.particular.net/nservicebus/licensing/?version=core_7.2#troubleshooting-failed-to-initialize-the-license. Specifically:

> The license management code requires write permissions to store metadata. If the process credentials don't have write permissions the following fatal event log item can be generated: (exception quote).
> Ensure that the process has write permissions at the specified location. If necessary, either modify the location, change permissions, or use another process that has appropriate permissions.

The documentation does not specify what API should be used to modify the location, leaving customers that cannot change permissions or use other process w/o any viable option.

danielmarbach:
Closes #4367

### DIFF
--- a/nservicebus/licensing/index.md
+++ b/nservicebus/licensing/index.md
@@ -48,7 +48,6 @@ The license management code requires write permissions to store metadata. If the
 
 ```
 FATAL NServiceBus.Features.LicenseReminder Failed to initialize the license
-System.UnauthorizedAccessException: Access to the path 'C:\Windows\system32\config\systemprofile' is denied.
+System.UnauthorizedAccessException: Access to the path 'some\file\system\path' is denied.
 ```
-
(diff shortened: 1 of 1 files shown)


danielmarbach:
@SeanFeldman did you see how I linked to the licensing management header?

SeanFeldman:
I did. When I raised this issue, it was specifically about that path and the fact that as a user I couldn't do anything at all. While the change is taking a stub at providing a few options when that path (`%system-drive%:\Windows\system32\config\systemprofile`) is logged, the users won't be able to do anything. I can go ahead and merge the PR which will close the issue, but it won't help the customers with that specific error. Unless I'm missing something here.

danielmarbach:
@SeanFeldman have another look

PR #4637, opened 2019-12-20
Fixed missing closing tag

Sebbl22:
The closing tag for the "ServiceControl/AllowMessageEditing" appSetting was missing, which will cause ServiceControl to fail, if you just copy&paste it.


timbussmann:
good catch, thanks for the PR @Sebbl22 👍

PR #4647, opened 2020-01-10
Changed languange of description

shaswat94:
Previously: If batch job fails then what?
Now: but what if batch job fails?


andreasohlund:
Thanks @shaswat94 👍

PR #4649, opened 2020-01-10
Spiking use of MS di adapters for plain endpoints

andreasohlund:


samples/dependency-injection/structuremap/StructureMap_7/Sample/Program.cs
```diff
@@ -12,25 +12,23 @@ static async Task Main()
...
+        registry.For<MyService>()
+                  .Use(new MyService());
 
         #endregion
 
         endpointConfiguration.UsePersistence<LearningPersistence>();
         endpointConfiguration.UseTransport<LearningTransport>();
 
-        var endpointInstance = await Endpoint.Start(endpointConfiguration)
+        var endpoint = ServiceCollectionEndpoint.Create(endpointConfiguration,
```
andreasohlund: @timbussmann @danielmarbach like we discussed yesterday. With a little wrapper we can make it very nice for users to switch to the MS DI adapters instead, what do you think?

PR #4651, opened 2020-01-16
RabbitMQ add warning that messages can only be delayed for roughly 8.5

ramonsmits:


danielmarbach:
Why should this be a warning? The text is already clear enough what the limitations are. I don't see a reason to bug people with more yellow boxes.

adamralph:
I don't think this should be a warning.

ramonsmits:
What do you suggest then? I do want this limitation to be more prominent as I don't think the text is clear. Its a wall of text that nobody reads in detail.

adamralph:
@ramonsmits `NOTE:`?

PR #4660, opened 2020-01-22
NServiceBus.Windsor.Castle 8 Sample

tmasternak:


timbussmann:
can't we just update the v7 sample to use the latest version?

bording:
Why has this been changed to reference the v7 package in the v8 sample?

bording:
Ok I see what's happened here. We're releasing a new minor, not a new major, so we don't need this new sample at all.

The build was breaking, so David has already fixed the v7 sample: https://github.com/Particular/docs.particular.net/pull/4666

I'm going go go ahead and close this as unneeded.

PR #4665, opened 2020-01-24
Delete rabbitmqclient-nuget.md

andreasohlund:
@bording given that we now target the current major of the client just like all other transport I don't think this page makes sense anymore?


bording:
Given that we've convinced them to follow SemVer now, yes it should be safe to remove this.

PR #4666, opened 2020-01-24
Fix Castle.Windsor sample

DavidBoike:


bording:
Removing the reference shouldn't be what we want to do here.

PR #4668, opened 2020-01-28
Clarify that changing conversation id is not supported

andreasohlund:
@indualagarsamy this what was the user mentioned in the support case right?

### DIFF
--- a/nservicebus/messaging/headers.md
+++ b/nservicebus/messaging/headers.md
@@ -107,7 +107,7 @@ snippet: override-conversation-id
 
 partial: conversationid
 
-NOTE: Attempting to override an existing Conversation Id is not supported will produce the following error:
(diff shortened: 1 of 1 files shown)


nservicebus/messaging/headers.md
```diff
@@ -107,8 +107,7 @@ snippet: override-conversation-id
 
 partial: conversationid
 
-Note that attempting to override an existing Conversation Id will produce the following error:
-
+NOTE: Attempting to override an existing Conversation Id is not supported will produce the following error:
```
indualagarsamy: do you think its better as a warning?

>WARNING: Attempting to override an existing `Conversation Id` is not supported and will produce the following error:

andreasohlund: 👍

PR #4669, opened 2020-01-29
Doco on pipeline context bag

ramonsmits:


nservicebus/pipeline/steps-stages-connectors_extra_core_[6.0,).partial.md
```diff
@@ -123,6 +123,12 @@ BD --> Transport
  * Audit: Behaviors in the Audit stage have access to the message to be audited/sent to the audit queue and audit address. Behaviors should use `IAuditContext` to enlist in this stage. This stage is only entered if [Message Auditing](/nservicebus/operations/auditing.md) is enabled. 
  * Forwarding: Behaviors in the Forwarding stage have access to the message to be sent to the forwarding queue and the address of the forwarding queue. Behaviors should use `IForwardingContext` to enlist in this stage. This stage is only entered if [Message Forwarding](/nservicebus/messaging/forwarding.md) is enabled.
 
+## Context bag
+
+The pipeline has a context bag which can be used to used to create, read, update or delete custom data to with a key identifier. For example, this can be used to *set* meta data in an incoming behavior that can be used in a later behavior to retrieve that data. Objects stored via the context bag will be removed once the context bag runs out of scope which is the end of the pipeline.
```
WilliamBZA:
```suggestion
The pipeline has a context bag which can be used to used to create, read, update or delete custom objects to with a key identifier. For example, this can be used to *set* metadata in an incoming behavior that can be used in a later behavior to retrieve that object. Objects stored via the context bag will be removed once the context bag runs out of scope which is the end of the pipeline.
```

danielmarbach: @WilliamBZA it is technically also possible to just store an `int` inside the extension bag which is technically not an object. I think data is a bit more generic and better suited

WilliamBZA: @danielmarbach Data has a flatter connotation to me and implies different things to object. So while object may be technically less correct, the implication of using object feels more correct than data to me

WilliamBZA: As an example: Why did I think to ask if the object had to be serializable? Because of the implication that the word "data" leaned me towards.

danielmarbach: How about state?

WilliamBZA: :+1: Yeah

danielmarbach: have a look

PR #4674, opened 2020-02-04
RavenDB IDocumentStore from container API

tmasternak:
## POC
- [x] Update RavenDB. to 6.1 dependency


timbussmann:
I've adjusted the PR a bit so that we're not adding new sections for the new overload but instead just use the new overload as the default in the existing documentation with slight adjustments to the text.

@mauroservienti @DavidBoike could you have a quick look at this since Tomasz isn't around the next few days?

PR #4675, opened 2020-02-05
Remove Bridge doco and add redirects to Router doco

SzymonPobiega:


SzymonPobiega:
@SeanFeldman I remvoed V7 and V8 versions of the ASB-MSMQ bridge sample because they were using the now obsolete Bridge. I left only V9 and updated it to use the Router. Do you think it is OK?

SeanFeldman:
Bridge and Router are community projects. If Bridge is no longer supported, makes total sense.
Thank you, @SzymonPobiega

PR #4695, opened 2020-02-18
Update container doco to use new extensions package

andreasohlund:


nservicebus/dependency-injection/structuremap.md
```diff
@@ -12,19 +12,7 @@ related:
 ---
 
 
-NServiceBus can be configured to use [StructureMap](https://structuremap.github.io/) for dependency injection.
-
-
-### Default usage
-
-snippet: StructureMap
```
andreasohlund: Decided to remove snippets for config since they just repeat official SM doco and we still have the sample that shows it.

andreasohlund:
@timbussmann I've done the StructureMap sample + doco, thoughts on killing the snippets?

SeanFeldman:
> I've done the StructureMap sample + doco, thoughts on killing the snippets?

The current sample doesn't really add any value. Unless you want to show [property injection](https://github.com/Particular/NServiceBus.Extensions.DependencyInjection/blob/de6ec0ac652f86b7c898b2a3ee3ad6e5b4487eb7/src/NServiceBus.Extensions.DependencyInjection.StructureMap.AcceptanceTests/DefaultServer.cs#L22-L29) with StructureMap. Then the sample could be updated to show exactly that.

andreasohlund:
Talked to @timbussmann about the samples and we believe there is still value for us to keep them as is since it gives both the users and us a starting point for smoke tests etc. It also verified that we don't run into issues with incompatible version ranges etc

andreasohlund:
Replaced by https://github.com/Particular/docs.particular.net/pull/4699

PR #4701, opened 2020-02-24
Supported versions link to nuget packages

### LINKED ISSUE #4002: Feedback: 'NServiceBus Packages Supported Versions'
Feedback for 'NServiceBus Packages Supported Versions' https://docs.particular.net/nservicebus/upgrades/supported-versions

Location in GitHub: https://github.com/Particular/docs.particular.net/blob/master/nservicebus/upgrades/supported-versions.md

While it's possible to determine what's the latest patch version of a given component (by navigating to downstreams doco and then to its NuGet page to find out the required major/minor combo), it would be nice to skip the unnecessary jumpts and navigate streight to NugetPackage with the right version.

I.e. 

image

would look like

image

bringing to the the right major/minor when clicked

image

danielmarbach:
Would close #4002

### DIFF
--- a/nservicebus/upgrades/all-versions.include.md
+++ b/nservicebus/upgrades/all-versions.include.md
@@ -2,32 +2,32 @@
 
 | Version   | Released       | Supported until   | Notes                             |
 |:---------:|:--------------:|:-----------------:|:---------------------------------:|
-| 7.2.x     | 2019-10-23     | -                 | -                                 |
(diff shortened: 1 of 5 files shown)


danielmarbach:
@SeanFeldman I gave this a stab. What do you think?

SeanFeldman:
@danielmarbach, it didn't navigate to the path of a majo.minor, but that's way better than it is today! Great addition, thanks.

Issue #4104, opened 2018-11-02
Feedback: 'Recoverability'


SeanFeldman:
Feedback for 'Recoverability' https://docs.particular.net/nservicebus/recoverability/

Location in GitHub: https://github.com/Particular/docs.particular.net/blob/master/nservicebus/recoverability/index.md

> Depending on the concurrency and transactionality settings of the endpoint immediate retries might happen even if configured to be off. To ensure that no retries is performed the endpoints needs to be configured to be either non transactional or the concurrency restricted to 1.

When is this possible? How setting concurrency to more than 1 will affect immediate retries? Very confusing note that is missing details.

SeanFeldman:
@Particular/nservicebus-maintainers can you shed some light on this? Thank you.

andreasohlund:
I have a faint memory that v4 was exposed to some concurrency issues when "counting" the immediate retries in memory since the increment happened after the message was rolled back to the queue. 

Looking at the current transport seam design I don't think this is an issue, eg. the learning transport increments the count before rolling back

https://github.com/Particular/NServiceBus/blob/develop/src/NServiceBus.Core/Transports/Learning/LearningTransportMessagePump.cs#L280

So does Rabbit - https://github.com/Particular/NServiceBus.RabbitMQ/blob/develop/src/NServiceBus.Transport.RabbitMQ/Receiving/MessagePump.cs#L253

But I wonder with eg. Rabbit/ASB/ASQ what happens when message handling take to long, if you have concurrency > 1 the message will be visible in the queue again and will be reprocessed without the retry count being incremented?

bording:
> But I wonder with eg. Rabbit/ASB/ASQ what happens when message handling take to long, if you have concurrency > 1 the message will be visible in the queue again and will be reprocessed without the retry count being incremented?

That won't be a problem for Rabbit since it doesn't have any sort of visibility timeout.

andreasohlund:
>The Immediate Retry mechanism is implemented by making the message available for consumption again at the top of the queue, so that the endpoint can process it again without any delay. Immediate Retries cannot be used when transport transactions are disabled.

@bording this is not correct for RabbitMQ right? (it does them inmemory)

andreasohlund:
I think the summary is: You can never be 100% sure that the message won't be reprocessed since some transports (eg. RabbitMQ) will potentially requeue messages even in transaction mode `none`. 

Unless we see that a as bug and fix rabbitmq?

bording:
[...]

PR #4708, opened 2020-02-24
Immediate retries guarantees clarified

tmasternak:
Connects to https://github.com/Particular/docs.particular.net/issues/4104

### DIFF
--- a/nservicebus/recoverability/index.md
+++ b/nservicebus/recoverability/index.md
@@ -40,7 +40,7 @@ By default up to five immediate retries are performed if the message processing 
 
 Note: The configured value describes the minimum number of times a message will be retried if its processing consistently fails. Especially in environments with competing consumers on the same queue, there is an increased chance of retrying a failing message more times across the endpoints.
 
-Note: Depending on the concurrency and transactionality settings of the endpoint immediate retries might happen even if configured to be off. To ensure that no retries is performed the endpoints needs to be configured to be either [non transactional](/transports/transactions.md) or the [concurrency restricted to 1](/nservicebus/operations/tuning.md).
+Note: Number of immediate retries is a best effort guarantee. Immediate retries can be performed more than configured amount of time (even if configured to be off).
 
 
 ### Transport transaction requirements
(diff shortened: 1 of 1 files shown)


nservicebus/recoverability/index.md
```diff
@@ -40,7 +40,7 @@ By default up to five immediate retries are performed if the message processing
 
 Note: The configured value describes the minimum number of times a message will be retried if its processing consistently fails. Especially in environments with competing consumers on the same queue, there is an increased chance of retrying a failing message more times across the endpoints.
 
-Note: Depending on the concurrency and transactionality settings of the endpoint immediate retries might happen even if configured to be off. To ensure that no retries is performed the endpoints needs to be configured to be either [non transactional](/transports/transactions.md) or the [concurrency restricted to 1](/nservicebus/operations/tuning.md).
+Note: Number of immediate retries is a best effort guarantee. Immediate retries can be performed more than configured amount of time (even if configured to be off).
```
danielmarbach:
```suggestion
Note: Number of immediate retries is best effort. Immediate retries can be performed more than configured amount of time (even if configured to be off).
```

nservicebus/recoverability/index.md
```diff
@@ -40,7 +40,7 @@ By default up to five immediate retries are performed if the message processing
 
 Note: The configured value describes the minimum number of times a message will be retried if its processing consistently fails. Especially in environments with competing consumers on the same queue, there is an increased chance of retrying a failing message more times across the endpoints.
 
-Note: Depending on the concurrency and transactionality settings of the endpoint immediate retries might happen even if configured to be off. To ensure that no retries is performed the endpoints needs to be configured to be either [non transactional](/transports/transactions.md) or the [concurrency restricted to 1](/nservicebus/operations/tuning.md).
+Note: Number of immediate retries is best effort. Immediate retries can be performed more than configured amount of time (even if configured to be off).
```
SeanFeldman: Do we have specifics of the use cases/edge cases this might happen? Is this unique to one or more transports or is this a combination of several things? I don't know about this issue enough to provide anything else other than raising the questions customers would have after reading this section.

tmasternak: It depends on the transport and the way it's configured. That said I'm not sure we can provide any valuable information in the context of a note. I think it's too little space.

I guess the decision we should make is if that change makes the note better or not. If not I propse to close this PR and handle the changes differently.

ramonsmits: This is information loss, as stated "To ensure no retries blabla set the to non-transactional"

You can also split it in multiple notes.

tmasternak: @ramonsmits I've reintroduced note on the transaction mode. 

I also removed concurrency limit as this does not prevent from multiple processings in competing consumer setup.

PR #4709, opened 2020-02-24
Property injection update

timbussmann:
Moves the property injection documentation to the adapter package documentation and greatly simplifies the documentation by removing unnecessary snippets.

note: this will currently fail due to unused snippets, I'll remove them once we agreed whether that's an acceptable approach to solve this.


nservicebus/dependency-injection/property-injection.md
```diff
@@ -1,23 +0,0 @@
----
-title: Property injection
-reviewed: 2020-01-24
-component: Core
-tags:
- - Dependency Injection
-redirects:
- - nservicebus/property-injection-in-handlers
```
SeanFeldman: Should the redirect for `nservicebus/dependency-injection/property-injection` be set up if the page is removed to avoid the broken cached link?

timbussmann: good point, we can redirect to the new package's documentation I'd say?

timbussmann:
merging this now as we discussed this approach yesterday during the sync already.

PR #4715, opened 2020-02-25
Add Autofac snippets to make the extensions package to show up

andreasohlund:


Snippets/Autofac/Extensions.DependencyInjection_1/Usage.cs
```diff
@@ -0,0 +1,17 @@
+﻿class Usage
+{
+    Usage()
+    {
+        #region Autofac
```
andreasohlund: @DavidBoike Is it enough for them to be here or do I actually have to show them on https://github.com/Particular/docs.particular.net/blob/master/nservicebus/dependency-injection/autofac_usage_extensions.dependencyinjection_%5B1%2C).partial.md?

DavidBoike: If you have a snippet, you have to use it.

if the Extensions.DependencyInjection "major version" doesn't use a specific snippet, then delete the snippet.

From the `//not used` here it looks like this entire file should be deleted. I'm guessing the only snippet you'll need is "this is how to use Autofac using Extensions.DependencyInjection"

timbussmann:
so we forgot to add the snippets for autofac?

nservicebus/dependency-injection/structuremap_usage_extensions.dependencyinjection_[1,).partial.md
```diff
@@ -1 +1,3 @@
 NServiceBus can be configured to use [StructureMap](https://structuremap.github.io/) for dependency injection via the [`NServiceBus.Extensions.Hosting`](/nservicebus/hosting/extensions-hosting) or the [`NServiceBus.Extensions.DependencyInjection`](/nservicebus/dependency-injection/extensions-dependencyinjection.md) package.
+
+snippet: Autofac
```
timbussmann: an autofac snippet in the structuremap partial?

andreasohlund: Duh, fixed now

PR #4734, opened 2020-02-28
Generic host - fix console option

SeanFeldman:


danielmarbach:
Nice catch. Stupid mistake. Thanks!

PR #4735, opened 2020-02-28
Update WebAPI sample to ASP.NET Core 3

timbussmann:


samples/web/send-from-aspnetcore-webapi/Core_7.2/WebApplication/Properties/launchSettings.json
```diff
@@ -8,13 +8,13 @@
...
   },
   "profiles": {
-    "IIS Express": {
-      "commandName": "IISExpress",
-      "launchBrowser": true,
-      "environmentVariables": {
-        "ASPNETCORE_ENVIRONMENT": "Development"
-      }
-    },
+    //"IIS Express": {
```
danielmarbach: Ok so the idea is to leave this around for people to decide whether they want to run in console or via IIS express if I understood correctly

timbussmann: yes, but happy to remove one of the options too, wasn't sure which one is preferrable though? I guess IIS Express strictly requires VS to be installed while the other mode should probably also work from the command line. But the readme states that you need VS19 as a prerequisite so 🤷‍♂

PR #4736, opened 2020-03-02
Scheduler is not recommended for fully-reliable scheduling

tmasternak:


nservicebus/scheduling/index.md
```diff
@@ -13,9 +13,11 @@ The Scheduler is a lightweight/non-durable API that helps schedule a task that n
 
 {{WARNING:
 
-Scheduling a task depends on [delayed delivery](/nservicebus/messaging/delayed-delivery.md). If the delayed delivery operation fails, the Scheduler will be interrupted and a `ScheduledTask` message will be forwarded to the error queue. When this happens the scheduled task will stop executing unless the `ScheduledTask` message is retried.
+Under certain failure scenarios, the scheduled task can stop firing until the endpoint is restarted.
```
ramonsmits: This needs explanation, what is 'certain failure scenarios'?

ramonsmits: You are hinting that it operates as at-most-once triggering, if that is the case we should remove the feature ASAP.

tmasternak: @ramonsmits full context is in https://github.com/Particular/NServiceBus/issues/5591.

Long story short when using Outbox and not using the same db with SQL Transport it's quite possible for tasks to stop executing. The user is hitting that scenario ~1 a week.

ramonsmits: Also, why remove the information that it relies on delayed delivery?

nservicebus/scheduling/index.md
```diff
@@ -13,9 +13,11 @@ The Scheduler is a lightweight/non-durable API that helps schedule a task that n
...
+Under certain failure scenarios, the scheduled task can stop firing until the endpoint is restarted. 
 
-The Scheduler leverages the queuing system to trigger scheduled actions. Under heavy load there may be some disparity between the expected time of a scheduled action and execution time due to the delay between timeout messages being generated and processed. On approach to mitigating this behavior is to run the Scheduler in a dedicated endpoint so that appropriate resource allocation can put in place to ensure timely execution of scheduled actions. Or, alternatively, use a difference scheduling technology, for example:
+The Scheduler leverages the queuing system to trigger scheduled actions. Under heavy load, there may be some disparity between the expected time of a scheduled action and execution time due to the delay between timeout messages being generated and processed. This behavior can be mitigated by running the Scheduler in a dedicated endpoint so that appropriate resource allocation can put in place. 
+
+Whenever a timely and/or fully-reliable scheduling is needed, it is recommended to use a dedicated scheduling technology, for example:
```
ramonsmits: I would like to move this to the top and improve it like this:

```suggestion
This simple scheduler is message based and does not provide any execution history and persistent failures can cause the reoccuring task to stop until either its messages is manually retried or the endpoint instance is restarted.

Whenever execution history, a timely and/or fully-reliable scheduling is needed, it is recommended to use dedicated scheduling technology, for example:
```

nservicebus/scheduling/index.md
```diff
@@ -13,9 +13,11 @@ The Scheduler is a lightweight/non-durable API that helps schedule a task that n
...
 
  * A [.NET Timer](https://msdn.microsoft.com/en-us/library/system.threading.timer.aspx).
  * [Quartz.NET](https://www.quartz-scheduler.net/). See [Quartz.NET Sample](/samples/scheduling/quartz/).
```
ramonsmits:
```suggestion
 * [Quartz.NET](https://www.quartz-scheduler.net/). See [Quartz.NET Sample](/samples/scheduling/quartz/).
 * OS task scheduler, like the Windows task scheduler or Linux cron jobs.
```

nservicebus/scheduling/index.md
tmasternak: I think it makes the message hidden in the details. The messages is:
> Scheduler is not fully-reliable.

nservicebus/scheduling/index.md
```diff
@@ -13,12 +13,17 @@ The Scheduler is a lightweight/non-durable API that helps schedule a task that n
...
+Under certain failure scenarios, the scheduled task can stop firing until the endpoint is restarted. 
 
-The Scheduler leverages the queuing system to trigger scheduled actions. Under heavy load there may be some disparity between the expected time of a scheduled action and execution time due to the delay between timeout messages being generated and processed. On approach to mitigating this behavior is to run the Scheduler in a dedicated endpoint so that appropriate resource allocation can put in place to ensure timely execution of scheduled actions. Or, alternatively, use a difference scheduling technology, for example:
+The Scheduler leverages the queuing system to trigger scheduled actions. Under heavy load, there may be some disparity between the expected time of a scheduled action and execution time due to the delay between timeout messages being generated and processed. This behavior can be mitigated by running the Scheduler in a dedicated endpoint so that appropriate resource allocation can put in place.
```
ramonsmits, line 18:
```suggestion
The Scheduler leverages the queuing system to trigger scheduled actions. Under heavy load, there may be some disparity between the expected time of a scheduled action and execution time due to the delay between timeout messages being generated and processed. This behavior can be mitigated by running the Scheduler in a dedicated endpoint so that appropriate resource allocation can put be in place. 
```

nservicebus/scheduling/index.md
ramonsmits: @tmasternak So the problem is that the schedule message gets moved to the error queue? It isn't actually lost.

Maybe adding a [custom recovery policy](https://docs.particular.net/nservicebus/recoverability/custom-recoverability-policy#implement-a-custom-policy-full-customization) for schedule message helps in to not move it to the error queue?

tmasternak: > So the problem is that the schedule message gets moved to the error queue

@ramonsmits correct, the message gets moved to the error queue and stays there until handled manually. I think that a custom recovery policy could work here.

ramonsmits: Any failures in the scheduler lambda will NOT result in any retry logic thus considered a loss:

- https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Scheduling/DefaultScheduler.cs#L41

The problem is that supplied context is wrong. That should be a context based on `IMessageSession`. With this session, a customer can do its own retry mechanism if required.

- https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Scheduling/DefaultScheduler.cs#L36

The "next" occurence is controlled with a send that happens before the execution of the task. Any failure here would result in transport recoverability

- https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Scheduling/DefaultScheduler.cs#L56

The BIGGEST problem is scale-out, brokers aren't supported.

IMHO, we should mark as Obsolete in our code base ASAP.

nservicebus/scheduling/index.md
```diff
@@ -13,12 +13,19 @@ The Scheduler is a lightweight/non-durable API that helps schedule a task that n
 
 {{WARNING:
 
-Scheduling a task depends on [delayed delivery](/nservicebus/messaging/delayed-delivery.md). If the delayed delivery operation fails, the Scheduler will be interrupted and a `ScheduledTask` message will be forwarded to the error queue. When this happens the scheduled task will stop executing unless the `ScheduledTask` message is retried.
+**Does not support scaled-out endpoints!**
```
ramonsmits, line 16:
```suggestion
** DO NOT USE** Does not support scaled-out endpoints, *any* errors during job execution will NOT be retried.
```

nservicebus/scheduling/index.md
tmasternak: @ramonsmits I agree that we should obsolete the scheduler.

tmasternak:
@ramonsmits I say we merge this. It's already better than what is there now.

PR #4737, opened 2020-03-03
Document how to enforce TLS 1.2 and higher with the host

danielmarbach:


timbussmann:
so does this also use TLS 1.1 when I compile my app against a later .net framework version because the host targets net452, or is this only an issue when I also just target net452?

danielmarbach:
Unfortunately the entry point determines the TLS version if I understood correctly. At first we tried to switch the SKU in the config

```
<configuration>
   <startup>
      <supportedRuntime version="v4.0" sku=".NETFramework,Version=v4.7.2" />
   </startup>
</configuration>
```

but that doesn't do the trick. The only way to do it is by using the above app.config AppContext switch. Of course we could release a major or a minor of the Host targeting 4.7.x and that would solve it as well but I'm not sure we would do that given the host is basically deprecated and we have other options available

timbussmann:
> Unfortunately the entry point determines the TLS version if I understood correctly. 

so this is an issue for the hosting package because the hosting package provides the exe (which is then net452)?

+1 for updating the minimum framework version. Would that be a good opportunity to test the new "major version" strategy? Although I'm not sure what value the hosting package really provides anymore given all the more suitable alternatives 🤷‍♂

danielmarbach:
That's why I felt at least documenting how to enforce TLS 1.2 and higher is a good middle ground

PR #4738, opened 2020-03-03
Ravendb upgrade guide timeout conversion

danielmarbach:


persistence/upgrades/ravendb-5to6.md
```diff
@@ -165,6 +165,59 @@ In NServiceBus versions 3 and 4, RavenDB Persistence was part of the main NServi
 
 If attempting to convert directly from NServiceBus 4 or lower to NServiceBus 7 (which is _not recommended_) timeout data will have to be converted to the new format manually.
 
+### Patching legacy timeouts
+
+The following patch can be used to idempotently update all timeout documents to the new format.
```
mikeminutillo:
```suggestion
The following patch can be used to update all timeout documents to the new format.
```

While I can run the patch multiple times, I wasn't worried about it until you pointed it out. Not sure we need this.

persistence/upgrades/ravendb-5to6.md
```diff
@@ -165,6 +165,59 @@ In NServiceBus versions 3 and 4, RavenDB Persistence was part of the main NServi
...
+    if(!this.Destination.Machine) {
+        this.Destination = this.Destination.Queue
+    }
+    else {
+        this.Destination = this.Destination.Queue + "@" + this.Destination.Machine
+    }
+}
+```
+
+If only timeout data using the new format is present nothing will be changed. If old timeout documents are present they will be converted into the new format.
```
mikeminutillo: I would avoid conditional ("if") style language here. The point is that this is safe to run. Customers don't need to worry about whether this applies to them or not.

```suggestion
Only timeout documents with with the old format are modified. Timeout documents already in the new format are not modified.
```

persistence/upgrades/ravendb-5to6.md
```diff
@@ -165,6 +165,59 @@ In NServiceBus versions 3 and 4, RavenDB Persistence was part of the main NServi
...
+will be patched to
+
+```json
+{
+    "Destination": "Queue@Machine",
+    "SagaId": "00000000-0000-0000-0000-000000000000",
+    ...
+```
+
+the patch can be either applied in the management studio under Documents\Patch as shown in [Documents and Collections](https://ravendb.net/docs/article-page/4.2/csharp/studio/database/documents/documents-and-collections)
```
mikeminutillo:
```suggestion
The patch can be either applied in the management studio under Documents\Patch as shown in [Documents and Collections](https://ravendb.net/docs/article-page/4.2/csharp/studio/database/documents/documents-and-collections)
```

mikeminutillo: Also :+1: on just pointing to the RavenDB documentation for how to apply the patch.

persistence/upgrades/ravendb-5to6.md
```diff
@@ -165,6 +165,59 @@ In NServiceBus versions 3 and 4, RavenDB Persistence was part of the main NServi
...
+ or by using the [client patch API.](https://ravendb.net/docs/article-page/4.2/csharp/client-api/operations/patching/set-based).
+
+If the patch is not applied and the database contains old timeout formats the following exception will be thrown on an endpoint trying to process old timeout formats
```
mikeminutillo:
```suggestion
When the endpoint tries to process a timeout which is stored with the old timeout format the following exception is thrown.
```

persistence/upgrades/ravendb-5to6.md
```diff
@@ -165,6 +165,59 @@ In NServiceBus versions 3 and 4, RavenDB Persistence was part of the main NServi
...
+
+```
+System.InvalidOperationException: Could not convert document TimeoutDatas/Number to entity of type NServiceBus.TimeoutPersisters.RavenDB.TimeoutData ---> Newtonsoft.Json.JsonSerializationException: Cannot deserialize the current JSON object (e.g. {"name":"value"}) into type 'System.String' because the type requires a JSON primitive value (e.g. string, number, boolean, null) to deserialize correctly.
+...
+```
+
+in such a case applying or replying the patch is sufficient to make the timeout processing successful either on the next timeout persister query interval or when the endpoint is restarted (in case the timeout processing needs to be enforced). Failed messages under the group `System.InvalidOperationException: Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateObject(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)` can simply be [archived in ServicePulse](/servicepulse/intro-archived-messages.md).
```
mikeminutillo: I would break this section up. 

Here is what happens in the endpoint and why it's OK. 

Here is a side effect that you might experience using ServiceControl and how you clean it up.

```suggestion
When this exception is thrown, the timeout document is not removed from the database and will be picked up on the next timeout persister query interval. Applying the patch to the database will allow the timeout persister to read the timeout document and correctly process the message. Restarting the endpoint will force the timeout persister to deliver all outstanding timeout messages once the patch has been applied.

Additionally, when the exception is thrown, a copy of the message is sent to the error queue. These are grouped together in ServicePulse under the exception type `System.InvalidOperationException: Newtonsoft.Json.Serialization.JsonSerializerInternalReader.CreateObject(JsonReader reader, Type objectType, JsonContract contract, JsonProperty member, JsonContainerContract containerContract, JsonProperty containerMember, Object existingValue)`. These messages have not been deleted from the endpoint timeout persistence and should not be retried from ServicePulse. Once the patch has been applied to the endpoint timeout persistence, this group of failed messages may be [archived](/servicepulse/intro-archived-messages.md).
```

persistence/upgrades/ravendb-5to6.md
```diff
@@ -165,6 +165,61 @@ In NServiceBus versions 3 and 4, RavenDB Persistence was part of the main NServi
 
 If attempting to convert directly from NServiceBus 4 or lower to NServiceBus 7 (which is _not recommended_) timeout data will have to be converted to the new format manually.
```
ramonsmits, line 166: @danielmarbach I think this statement is incorrect. It doesn't matter is you migrated one version at a time or all at once. RavenDB will not migrated IDLE data AFAIK.

danielmarbach: When I did the research for this we saw that an entity conversion listener was available in previous versions that converted timeouts due into the new format. So why are you know saying this is not correct?

ramonsmits: @danielmarbach The converter is AFAIK only triggered once the timeout expires when read from storage. Meaning, any timeouts that did not yet expire prior to 6 but will after migrating to v6 cannot be converted.

danielmarbach: Exactly and that is what this wording says. Because as long as you run the previous versions no manual conversion is necessary. New timeouts are already stored in the new format and old timeouts are converted upon deserialization like the previous paragraph mentions. I think the text assumes there are no timeouts left anymore because you went through multiple versions of core. 

Maybe the wording is just a bit confusing and should be changed to something like

If timeouts generated by the persistence targeting NServiceBus 4 are known to be due timeout data will have to be converted to the new format manually.

What do you think?

ramonsmits: @danielmarbach I'll create a new PR and add you as a reviewer as I see that you are on leave.

PR #4746, opened 2020-03-06
Switch docker sample to generic hosting based on template

danielmarbach:
image


danielmarbach:
@bording I kept for now some of the TODO's from the template in the sample as well as the previous langversion. We can definitely clean this up a bit but I wanted to get the first version going for now

danielmarbach:
@bording see https://github.com/Particular/docs.particular.net/pull/4748

PR #4748, opened 2020-03-07
Switch to snippets in compose and dockerfile

danielmarbach:
Requires https://github.com/Particular/DocsEngine/pull/453

Also documents the other extensions that are supported

### DIFF
--- a/README.md
+++ b/README.md
@@ -628,6 +628,7 @@ File extensions scanned for snippets include:
  * `.json`
  * `.proto`
  * `.config`
+  * `Dockerfile`
 
 ### Snippets are highlighted using highlightjs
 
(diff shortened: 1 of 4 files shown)


bording:
I don't really like adding snippet comments into the actual file someone will get when the download the sample, but I guess that's not really a new problem since we already do that in the actual code with regions...

danielmarbach:
Exactly. Plus everything we do is samples and not meant to be blindly copied.

bording:
> Exactly. Plus everything we do is samples and not meant to be blindly copied.

Because, you know, that never happens! 😛

danielmarbach:
Requires also https://github.com/Particular/DocsEngine/pull/454

PR #4758, opened 2020-03-26
[Staged] Previews docs

boblangley:
This PR documents the support policy for the upcoming unannounced platform Previews program.

In addition, it establishes "Previews" as a top-level menu item for previews, to be further fleshed out in future PRs.

### DIFF
--- a/previews/support-policy.md
+++ b/previews/support-policy.md
@@ -10,6 +10,15 @@ related:
 
 Product Previews will always be released as a major version 0. Once a product releases version 1 or greater, it is no longer part of the Preview program and does not follow this support policy.
 
+For example:
+
+| Version | Covered by this plicy |
+| --- | --- |
+| 0.1.0 | Yes |
+| 0.2.0 | Yes |
+| 0.2.1 | Yes |
+| 1.0 | No |
+
 WARN: Only the latest minor version of the Preview will be supported.
 
 ## Compatibility guarantees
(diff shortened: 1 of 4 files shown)


previews/support-policy.md
```diff
@@ -0,0 +1,26 @@
...
+title: Support Policy for Previews
+summary: Describes the details for the support policy for previews
+reviewed: 2020-03-26
+related:
+ - nservicebus/upgrades/support-policy
+---
+
+## Supported Versions
+
+Product Previews will always be released as a major version 0. Once a product releases version 1 or greater, it is no longer part of the Preview program and does not follow this support policy.
```
danielmarbach, line 11: Maybe it helps to make an example. I was wondering whether

1.0.0-alpha001 would be considered a preview until I remembered the leading zero is the difference. A small table or a few examples might make this clearer for users like

0.2.1 preview
0.15.6 preview
1.0.1 not a preview
1.0.0-alpha|beta not a preview

danielmarbach:
See also https://github.com/Particular/Exploration/pull/297#discussion_r402100859 comment

danielmarbach:
Fyi

> Propose to switch to preview instead of alpha/beta to avoid touching the GitVersion file

https://github.com/Particular/PlatformDevelopment/issues/2896

This might introduce 1.0.0-preview kind of versioning that could lead to confusion if not spelled out here is a counter example in the recently introduced table?

// cc @andreasohlund

PR #4764, opened 2020-03-31
Change support stance of SagaAudit plugin

WilliamBZA:
The reasons behind why the SagaAudit plugin was not supported in production are no longer completely valid. The SagaAudit plugin put pressure on both the endpoint as well as ServiceControl. ServiceControl has undergone a number of changes that make it less of a concern (specifically: there have been lots of performance improvements, and ServiceControl itself can be scaled out horizontally). Based on this, I propose changing out stance from "Not supported in production" to a warning stating "Use with caution as this will add pressure to your endpoint and will reduce the processing throughput."

### DIFF
--- a/nservicebus/sagas/saga-audit.md
+++ b/nservicebus/sagas/saga-audit.md
@@ -6,7 +6,7 @@ versions: 'SagaAudit:*'
 reviewed: 2019-09-03
 ---
 
-WARN: This plugin will result in a increase in the load placed on ServiceControl and the endpoint it is installed in. Use only if the environment is able to keep up with the increased load.
+WARN: This plugin will result in a increase in the load placed on ServiceControl and the endpoint it is installed in. It should only be used if the environment is able to keep up with the increased load.
 
 The SagaAudit plugin enables the [Saga View feature in ServiceInsight](/serviceinsight/#the-saga-view). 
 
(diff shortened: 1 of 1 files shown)


SzymonPobiega:
@WilliamBZA the state changed saga audit messages are still processed by the main SC instance so all the work we did around scaling out does not really improve the situation here :-(

DavidBoike:
Do we say anything about never-ending sagas? I can't remember if that still results in an ever-lengthening document that is un-page-able for the purposes of viewing in ServiceInsight.

WilliamBZA:
@SzymonPobiega True, but the scaled out instances can continue to ingest audit messages while the main instance handles the audit state change messages.

@DavidBoike I can't find any documentation on never-ending sagas at all.

nservicebus/sagas/saga-audit.md
```diff
@@ -6,7 +6,7 @@ versions: 'SagaAudit:*'
 reviewed: 2019-09-03
 ---
 
-WARN: This plugin will result in a increase in the load placed on ServiceControl and your endpoint. Use only if your environment is able to keep up with the increased load.
+WARN: This plugin will result in a increase in the load placed on ServiceControl and the endpoint it is installed in. Use only if the environment is able to keep up with the increased load.
```
adamralph, line 9:
```suggestion
WARN: This plugin will result in a increase in the load placed on ServiceControl and the endpoint it is installed in. It should only be used if the environment is able to keep up with the increased load.
```

"Use X..." is personal language, it's just that our automation can't detect it.

PR #4772, opened 2020-04-07
HostId with Docker/K8S

DavidBoike:


nservicebus/hosting/override-hostid.md
```diff
@@ -25,3 +25,10 @@ There are scenarios where the rules used by NServiceBus to generate a `hostid` a
 Manual configuration is required when deployments may end up in different paths than previously deployed versions (e.g. using [Octopus Deploy](https://octopus.com/)). The `hostid` must remain the same across restarts unless the physical host has changed.
 
 snippet: HostIdFixer
+
+
+## Stable host identifiers when using Docker
```
andreasohlund, line 30: Do we need to mention this here https://docs.particular.net/monitoring/metrics/install-plugin#configuration-instance-id  as well?

nservicebus/hosting/override-hostid.md
```diff
@@ -25,3 +25,10 @@ There are scenarios where the rules used by NServiceBus to generate a `hostid` a
...
+
+Docker containers hosted in Kubernetes pose unique challenges to creating a stable host identifier, as a [Deployment](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/) will not have a stable, unique identifier. This results in the endpoint having a new, random identity in ServicePulse and ServiceInsight every time the [Pod]() containing the endpoint is (re)scheduled on a different node.
```
ramonsmits, line 32: Pod does not have a link.

nservicebus/hosting/override-hostid.md
ramonsmits: @andreasohlund it copies the ID from core.

DavidBoike:
Thanks for the catch on the link @ramonsmits!

> Good addition to the doco although I think we would be better of with a page dedicated to docker and its nuances. Point to pages like these for customizations.

Agree but that exercise would be much larger. This is the scope I can get away with from a weekly doco review.

PR #4774, opened 2020-04-09
Add warning about outbox cleanup when running RavenDB in multi-tenant mode

andreasohlund:
This makes the users aware that they need to handle outbox cleanup manually when using the multi-tenant support of the RavenDB persister.

### DIFF
--- a/persistence/ravendb/outbox.md
+++ b/persistence/ravendb/outbox.md
@@ -31,6 +31,8 @@ snippet: OutboxRavendBTimeToKeep
 
 partial: disable-cleanup
 
+WARN: When running in [multi-tenant mode](/persistence/ravendb.md#multi-tenant-support) cleanup needs to be handled manually since NServiceBus can't know what databases are in use.
+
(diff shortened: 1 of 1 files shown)


persistence/ravendb/outbox.md
```diff
@@ -31,6 +31,8 @@ snippet: OutboxRavendBTimeToKeep
 
 partial: disable-cleanup
 
+WARN: When running in [multi-tenant mode](/persistence/ravendb.md#multi-tenant-support) cleanup needs to be handled manually since NServiceBus can't know what databases are in use.
```
mauroservienti, line 34:
```suggestion
WARN: When running in [multi-tenant mode](/persistence/ravendb.md#multi-tenant-support), cleanup needs to be manually handled since NServiceBus does not know what databases are in use.
```

timbussmann:
I wish there would be some better guidance on how to do a manual clean up but at least there is a warning about this and we can tackle this as a separate issue

PR #4790, opened 2020-04-21
Saga fix reply to originator and how to set multiple timeouts

ramonsmits:
Pinging you as we did the saga timeouts tutorial. There was a bug in the sample for the last few years :-) and added info on how to set multiple timeouts.


dvdstelt:
Didn't check errors, but code looks legit

PR #4794, opened 2020-04-23
Improve awareness of generating timeout manager duplicates 

ramonsmits:
Based on https://discuss.particular.net/t/duplicate-messages-for-delayed-retries-when-scaled-out-using-sql-transport-with-native-delayed-delivery/1763


SzymonPobiega:
For the sake of completeness we could also mentioned that in the SQT+SQLP/NHP case there is no possibility of ducplicates because the TM re-uses the transport transaction.

PR #4797, opened 2020-04-28
Review Accessing Data page

SzymonPobiega:


nservicebus/handlers/accessing-data.md
```diff
@@ -2,70 +2,87 @@
...
+
+Message processing logic is idempotent if it can be applied multiple times and the outcome is the same as if it was applied once. The outcome includes both the application state changes and the potential outgoing messages sent. Consider the following pseudocode:
+
+snippet: BusinessData-ManualIdempotence
+
+and think about the behavior of the message processing:
+
+ - NServiceBus by default defers sending out messages until message handler has completed so the behavior of the code above is as if the call to `Send` **was after** the call to `ModifyState`
+ - If outgoing messages are sent before state change is committed (e.g if code above used [immediate dispatch](/nservicebus/messaging/send-a-message.md#dispatching-a-message-immediately)) there is a risk of creating *ghost messages* -- messages that carry the state change that has never been made durable
+ - If outgoing messages are sent after the state change is committed there is risk of message loss if the send operation fails. To prevent this, the outgoing messages need to be sent **even if it appears to be a duplicate**
```
TraGicCode, line 85: Should you also talk about the case in which business state changes succeed, outgoing messages are sent successfully, BUT the message in which the current handler is being executed for is failed to be reported back to the transport that processing has succeeded which in turn would cause a retry/redelivery?

Snippets/Core/Core_7/Handlers/DataAccess/ManualIdempotence.cs
```diff
@@ -0,0 +1,38 @@
+namespace Core7.Handlers.DataAccess
+{
+    using System;
+    using System.Threading.Tasks;
+    using NServiceBus;
+
+    public class ManualIdempotence : IHandleMessages<MyMessage>
+    {
+        #region BusinessData-ManualIdempotence
```
TraGicCode, line 9: This example is meant to convey what not to do correct?  I think this should be clarified so users don't try and do this.

SzymonPobiega:
Thanks @TraGicCode for the suggestions ❤️

PR #4802, opened 2020-04-29
ServiceControl customize logging

ramonsmits:


ramonsmits:
@TraGicCode I'm not sure what you mean with "going away" but as the note states any logging settings will be ignored as the custom configuration takes priority. See how the code base handles this:

- https://github.com/Particular/ServiceControl.Monitoring/blob/43f9a6a4532ea9a7251d4f12ed2cbb0c7f079f91/src/ServiceControl.Monitoring/MonitorLogs.cs#L23-L26

PR #4805, opened 2020-05-01
Adding events related information

WojcikMike:


sergioc:
The size of the viewport used for the screenshots is unnecessarily https://github.com/Particular/StaffSuccess/pull/1786 wide and will make the screenshots appear tiny. Better use around 1600px width.

PR #4815, opened 2020-05-05
Fix duplicate snippet content on di-aspnetcore sample

lailabougria:
Removed a duplicate snippet causing the snippet content to appear twice on the website

### DIFF
--- a/samples/dependency-injection/aspnetcore/Core_7.2/SampleAutofac/Startup.cs
+++ b/samples/dependency-injection/aspnetcore/Core_7.2/SampleAutofac/Startup.cs
@@ -29,25 +29,25 @@ public class Startup
             applicationBuilder.UseDeveloperExceptionPage();
         }
 
-        #region RequestHandling
-
-        applicationBuilder.Run(
-            handler: context =>
-            {
-                if (context.Request.Path != "/")
-                {
-                    // only handle requests at the root
-                    return Task.CompletedTask;
-                }
-                var applicationServices = applicationBuilder.ApplicationServices;
-                var endpointInstance = applicationServices.GetService<IMessageSession>();
-                var myMessage = new MyMessage();
-
-                return Task.WhenAll(
-                    endpointInstance.SendLocal(myMessage),
-                    context.Response.WriteAsync("Message sent"));
-            });
-
-        #endregion
+        // #region RequestHandling
+        //
+        // applicationBuilder.Run(
+        //     handler: context =>
+        //     {
+        //         if (context.Request.Path != "/")
+        //         {
+        //             // only handle requests at the root
+        //             return Task.CompletedTask;
+        //         }
+        //         var applicationServices = applicationBuilder.ApplicationServices;
+        //         var endpointInstance = applicationServices.GetService<IMessageSession>();
+        //         var myMessage = new MyMessage();
+        //
+        //         return Task.WhenAll(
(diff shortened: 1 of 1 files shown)


samples/dependency-injection/aspnetcore/Core_7.2/SampleAutofac/Startup.cs
```diff
@@ -29,25 +29,25 @@ public void Configure(IApplicationBuilder applicationBuilder, IWebHostEnvironmen
...
+        //         var applicationServices = applicationBuilder.ApplicationServices;
+        //         var endpointInstance = applicationServices.GetService<IMessageSession>();
+        //         var myMessage = new MyMessage();
+        //
+        //         return Task.WhenAll(
+        //             endpointInstance.SendLocal(myMessage),
+        //             context.Response.WriteAsync("Message sent"));
+        //     });
+        //
+        // #endregion
```
WilliamBZA, line 51: Why commented out and not removed completely?

lailabougria: Totally right I'll kick it out :-)

samples/dependency-injection/aspnetcore/Core_7.2/SampleAutofac/Startup.cs
```diff
@@ -28,26 +28,5 @@ public void Configure(IApplicationBuilder applicationBuilder, IWebHostEnvironmen
...
-                var applicationServices = applicationBuilder.ApplicationServices;
-                var endpointInstance = applicationServices.GetService<IMessageSession>();
-                var myMessage = new MyMessage();
-
-                return Task.WhenAll(
-                    endpointInstance.SendLocal(myMessage),
-                    context.Response.WriteAsync("Message sent"));
-            });
-
-        #endregion
```
WilliamBZA, line 51: Sorry, I should have looked closer the first time round. I think this is required for the sample to work.

"The sample" in this case being the `SampleAutofac.Core` project.

The solution has 2 projects, one `Sample.Core` and the other `SampleAutofac.Core`. Removing this piece of code here prevents the `SampleAutofac.Core` project from behaving correctly. 

I suggest instead renaming this snippet from `RequestHandling` to `RequestHandlingAutofac` (Apparently I can't make a suggestion on deleted lines in Github??). Making this change will mean that only one snippet will correspond to the `RequestHandling` key.

timbussmann:
wouldn't removing this code break the autofac sample? This is a weird sample in the first place because it contains 2 projects doing very similar things. I haven't looked to closely but we might have to keep the existing code and just drop the `#region` part so there are not 2 regions with the same name around?

samples/dependency-injection/aspnetcore/Core_7.2/SampleAutofac/Startup.cs
lailabougria: Since we're not using <code>RequestHandlingAutofac</code> as a snippet anywhere, do we still need to enclose that in a region for some reason? If not I think it's a better idea to just have the code in there in order for the sample to behave correctly without any region.

WilliamBZA: Yeah, I think we can drop the region. The Sample.Core snippet is explained just above the autofac code - there doesn't seem to be a compelling reason to keep the region.

PR #4819, opened 2020-05-07
Convert netcoreapp2.1 solutions to 3.1/2.1 multitargeted

DavidBoike:


bording:
It seems like at least some of these could also have a .NET Framework target. Should we be looking at that as well?

Also, for the projects that can't be multi-targeted, should we be looking at another way to provide a .NET Core 3.1 project?

PR #4828, opened 2020-05-08
Fix samples after update script

bording:


DavidBoike:
Integrity tests are fixed on master, I will merge with admin rights.

PR #4830, opened 2020-05-11
AWSLambda sample to netcoreapp3.1

DavidBoike:
@boblangley Take a look and let me know what you think.

### DIFF
--- a/samples/aws/lambda/sqs/SQSLambda_0.1/AwsLambda.SQSTrigger/Function.cs
+++ b/samples/aws/lambda/sqs/SQSLambda_0.1/AwsLambda.SQSTrigger/Function.cs
@@ -6,7 +6,7 @@ using Amazon.Lambda.SQSEvents;
 using NServiceBus;
 
 // Assembly attribute to enable the Lambda function's JSON input to be converted into a .NET class.
-[assembly: LambdaSerializer(typeof(Amazon.Lambda.Serialization.Json.JsonSerializer))]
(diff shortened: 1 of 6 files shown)


samples/aws/lambda/sqs/SQSLambda_0.1/AwsLambda.SQSTrigger/Function.cs
```diff
@@ -17,7 +17,7 @@ public Function()
     #region FunctionHandler
     public async Task FunctionHandler(SQSEvent evnt, ILambdaContext context)
     {
-        using (var cancellationTokenSource = new CancellationTokenSource(context.RemainingTime.Subtract(DefaultRemainingTimeGracePeriod)))
```
danielmarbach, line 20: @DavidBoike any reasons this was removed? When we created the sample we deliberately added this to show how the execution can be canceled with alignment with the lambda context

danielmarbach: FYI I'm not strictly opposed to remove this. We had a debate whether we are violating the SRP of a sample or not. At the end we added it. I'm more curious to hear if it was intentional and if yes why

DavidBoike: That was totally my bad. I did that to try to get the web app API to work, and forgot to revert that before committing.

DavidBoike:
Since you both looked at it and I fixed the one goof you both identified, I'll go ahead and merge rather than force you to re-review. Can always change it further later.

PR #4854, opened 2020-05-25
Add status and support info to Previews page

sergioc:
_Superceeded by https://github.com/Particular/docs.particular.net/pull/4864_


boblangley:
That table is in the scope of the Azure Functions issue

SeanFeldman:
> That table is in the scope of the Azure Functions issue

Then I don't quite understand why do we need these sections

```
## Upcoming Previews

The following Preview(s) are under initial development and will be available soon for anyone to use:

* Azure Functions support in NServiceBus

## Available to use

* No Previews are currently available for testing
```

They seem to duplicate what the table will have anyways, no? I'd expect to see an empty table for now.

PR #4855, opened 2020-05-27
SNS/SQS permission fix

danielmarbach:


mauroservienti:
I just used the `docstool` to serve this locally on my machine and I love it so much 😄

PR #4856, opened 2020-05-27
Exclude .dependabot/config.yml from repo sync

mauroservienti:
Related to https://github.com/Particular/RepoStandards/pull/22

### DIFF
--- a/.reposync.yml
+++ b/.reposync.yml
@@ -1,0 +1,2 @@
+exclusions:
+- .dependabot/config.yml


andreasohlund:
This will start to sync things like strong naming keys, license etc. Is this what we want? (docs was excluded previously since its not a "platform code" related repo?

mauroservienti:
I later realized that no, we don’t want that. Closing.

PR #4857, opened 2020-05-27
Improved wording on "retired" previews

udidahan:


previews/support-policy.md
```diff
@@ -35,4 +35,4 @@ Previews may break backwards compatibility at any time. This includes, but is no
 Product previews are public for a limited time. When this time has elapsed, they are either:
 
 - Promoted to being part of the Particular Service Platform with a 1.0 release, and supported according to standard support policies.
-- Retired and no further maintenance or support is provided.
+- Converted into community projects with no further guaranteed maintenance or support.
```
SeanFeldman, line 38: Should this explicitly call out that "no further guaranteed maintenance or support" would be in the context of Particular? I.e. 
> Converted into community projects with no further guaranteed maintenance or support provided by Particular Software?

udidahan: Sure, we can do that 👍

PR #4861, opened 2020-05-29
Legacy docs review of customizing error handling

SzymonPobiega:


SzymonPobiega:
@dbelcham I opened it as part of our doco review shift because I found it is wrong while we had support duty. Can you give it a quick review?

PR #4896, opened 2020-06-24
UseNServiceBus should be specified before ConfigureWebHostDefaults

Stallind:
According to https://github.com/Particular/docs.particular.net/pull/4598 this structure is incorrect.


danielmarbach:
Thanks!

PR #4897, opened 2020-06-24
Fix conventions inheritance example

evolvedlight:
@seanfarmar 

Can you confirm if this docs fix is correct?

Example reproduction:

```csharp
using NServiceBus;
using System;

namespace NSBConventions
{
    class Program
    {
        static void Main(string[] args)
        {
            var ourType = typeof(TestNSBClass);

            Console.WriteLine($"NSB Docs way, is assignable: {ourType.IsAssignableFrom(typeof(ICommand))}");
            Console.WriteLine($"My way, is assignable: {typeof(ICommand).IsAssignableFrom(ourType)}");
        }
    }

    public class TestNSBClass : ICommand
    {
    }
}
```


dvdstelt:
Nice PR @evolvedlight ! Thanks!

PR #4898, opened 2020-06-24
Add OPTIONS verbs to web.config

markgould:
OPTIONS is used against the /monitoring and /api/endpoints and needs to be allowed.

I kept getting 401 errors and tracked it down to these verbs not being allowed.


danielmarbach:
Yeah you are right. Verified with ServicePulse. Merging

PR #4914, opened 2020-07-03
Breaking changes related to RabbitMQ.Client 6.1.0

ramonsmits:


bording:
I pushed up some fixes, but I really don't understand some of the changes that have been made here. Why have we touched the non-durable messaging page at all? The information there was largely correct, we just needed to remove the note about publisher confirms.

Why is there now a "Non-durable" section right on the index of the transport page? That is very out of place. There's also no reason to even mention publisher confirms in that section.

ramonsmits:
@bording The info was first only visible on the non-durable page where the info presented was aligned with a specific version of Core. As for V7 we now have multiple majors with different behavior I had to move it to the rabbitmq section and have different docs per major version of the transport.

The texts only got moved, they are not new.

In regards to its location, you say that it is very out of place but do you have an alternative? I do not see a suggestion. I considered creating a dedicated page for it but didn't as that seemed like overkill.

PR #4915, opened 2020-07-06
Update netcore samples

mikeminutillo:
Update the netcore samples script to include samples that target netcore 3.1 and run it.


mikeminutillo:
Why do these links have version numbers? The version numbers lead to duplicate links. 

Shouldn't the latest versions of each of these have a .NET Core version? i.e. It would be odd for a sample to have a .NET Core version in an earlier version but not in the latest version.

DavidBoike:
I'd really like GitHub to notice I force-pushed a rebase and run the checks…

bording:
> I'd really like GitHub to notice I force-pushed a rebase and run the checks…

Something definitely seems off. The PR isn't even showing a rebased commit yet.

PR #4922, opened 2020-07-07
RabbitMQ: Added note on disabling publisher confirms

ramonsmits:


transports/rabbitmq/connection-settings_publisher-confirms-setting_rabbit_[4,].partial.md
```diff
@@ -2,4 +2,6 @@
 
 [Publisher confirms](https://www.rabbitmq.com/confirms.html) are enabled by default, but they can be disabled by using the following:
 
-snippet: rabbitmq-config-disablepublisherconfirms
\ No newline at end of file
+snippet: rabbitmq-config-disablepublisherconfirms
+
+Note: When publisher confirms are disabled send operations will not verify if relevant exchanges and/or queues exists and will not fail.
```
Comment (bording) [inline transports/rabbitmq/connection-settings_publisher-confirms-setting_rabbit_[4,].partial.md:7]:
```suggestion
Note: When publisher confirms are disabled, send operations might not fail with an exception when the destination exchanges/queues don't exist.
```

PR #4924, opened 2020-07-08
Add System.Data.SqlClient package when targeting .NET Core

mauroservienti:


bording:
Why is this needed? If there is a package reference missing somewhere, it seems like we should be solving this somewhere other than samples.

mauroservienti:
This is needed because the sample as is blows up. This might be due to the fact that NHibernate for .NET Core can support either the Microsoft.Data stuff or the System.Data one.

As I’m back from vacation I can further investigate.

bording:
> This is needed because the sample as is blows up. This might be due to the fact that NHibernate for .NET Core can support either the Microsoft.Data stuff or the System.Data one.

If the sample is blowing up then it seems likely any project pulling in our packages would have a problem, so fixing it at the NServiceBus.NHibernate level seems like the right move.

mauroservienti:
Agreed, not had time yet to investigate.

mauroservienti:
I double-checked and there is no problem, AFAIK, in the NHibernate repo. The thing is:

- NHibernate is database agnostic
- At configuration time the dialect and the driver are selected, e.g. https://github.com/Particular/docs.particular.net/blob/2254ddf0fd0e5ddb7a4f9ff3d9671862d1582d8b/samples/nhibernate/simple/NHibernate_8/Server/Program.cs#L22-L23
- At startup, NHibernate initializes the underlying database client based on the above settings

If the hosting framework is the .NET Framework that it just works because the `SqlClient` is available in the base class library if it's .NET Core it blows up. A similar error would have happened if the selected driver and dialect were Oracle.

Base on the above I guess that no changes are required to the NHibernate persistence package, and the changes in this PR are needed to fix the sample when using .NET Core. Not sure if it's worth a note in the documentation as my gut feeling tells me that NHibernate users are well aware of this.

Thoughts?

bording:
@mauroservienti Okay I see what you mean about it being a dialect selection that needs a specific client based on that dialect. Looking at the NServiceBus.NHibernate repo, I can see that the test projects do include a reference to the package.

Based on that, it does make sense to add the `System.Data.SqlClient` package to these samples if we're expecting them to run against SQL Server. However, there is no reason to condition the reference like you've done here. It should be fine for `net48` to also reference the package.

mauroservienti:
my intention was to highlight that it's only needed when targeting something that is not .NET Framework, but I'm fine removing it

PR #4952, opened 2020-08-10
new multi-hosting sample with generic host

HEskandari:


samples/hosting/generic-multi-hosting/Core_7/Sample.MultiHosting/Program.cs
```diff
@@ -0,0 +1,101 @@
...
+        
+        await Task.WhenAll(endpointOneTask, endpointTwoTask);
+
+        #endregion
+    }
+    
+    private static IHostBuilder ConfigureEndpointOne(IHostBuilder builder)
+    {
+        builder.UseConsoleLifetime();
+        builder.ConfigureLogging((ctx, logging) =>
```
andreasohlund, line 29: @danielmarbach this one got me thinking, given that NSB uses static logging won't its be broken since both instances will essentially share loggers?

danielmarbach: Yep last one wins. So as long as the logging configuration is the same it will work but there is no way to differentiate the endpoints in the log other than adding some contextual information to the logger. I think the logger has a dedicated way to [scope things](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loggerextensions.beginscope) but I haven't looked closer into this.

Another option would be that we decorate things internally somehow but I would have to think about it

danielmarbach:
Discussed with @andreasohlund and how about we make this sample the new multi hosting sample replacing

https://github.com/Particular/docs.particular.net/blob/master/samples/hosting/multi-hosting/sample.md

Btw. the old sample even with custom hosting already had the logging override problem

samples/hosting/generic-multi-hosting/Core_7/Sample.MultiHosting/Program.cs
```diff
@@ -0,0 +1,99 @@
...
+    static async Task Main(string[] args)
+    {
+        Console.Title = "Samples.MultiHosting";
+
+        #region multi-hosting-startup
+        
+        var endpointOneBuilder = ConfigureEndpointOne(Host.CreateDefaultBuilder(args)).Build();
+        var endpointTwoBuilder = ConfigureEndpointTwo(Host.CreateDefaultBuilder(args)).Build();
+        
+        var endpointOneTask = endpointOneBuilder.RunAsync();
```
danielmarbach, line 19: Should we also add how to stop those again or is that overkill?

samples/hosting/generic-multi-hosting/sample.md
```diff
@@ -0,0 +1,22 @@
+---
+title: Generic Host multiple endpoint hosting
+summary: Hosting multiple endpoints in a generic host process.
+reviewed: 2020-08-10
+component: Core
+related:
```
danielmarbach, line 6: Are redirects automagically created or do we need to add a redirect that points back to the previous URI we used?

samples/hosting/generic-multi-hosting/Core_7/Sample.MultiHosting/Program.cs
HEskandari: Should we raise an issue for this? or is there anything for it in our backlog (couldn't find any open issue).

andreasohlund: Raised https://github.com/Particular/PlatformArchitecture/pull/578

samples/hosting/generic-multi-hosting/Core_7/Sample.MultiHosting/Program.cs
danielmarbach: I checked and the console lifetime takes care of it

samples/hosting/generic-multi-hosting/sample.md
danielmarbach: Done

PR #4957, opened 2020-08-10
GH Actions checkout troubleshooting

DavidBoike:


DavidBoike:
All good, since the action compares `refs/remotes/pull/####/merge` to `origin/master`, the changes to master are already contained in the merge branch, so the `git diff origin/master..HEAD` is exactly the changes in the PR. Previously (in TeamCity) it had to be `git diff origin/master...HEAD` (3 dots) to go back to the common branch ancestor to do the same thing the merge branch is now doing.

PR #4960, opened 2020-08-12
Remove duplicate in table of total number of retries

oskar:
Remove duplicate in table of total number of retries.


boblangley:
@oskar thank you for this contribution!

PR #4961, opened 2020-08-12
Swap headers in table of total number of retries

oskar:
I'm pretty sure the table headers have been mixed up. The default number of immediate retries is **five** and the default number of delayed retries is **three**. By swapping the table headers, the last row makes much more sense.


oskar:
@boblangley Sure, no problem. I've updated the PR according to your suggestion. The diff looks a bit weird, but I'm pretty sure it's correct 😅

boblangley:
@oskar thanks again for helping us improve our documentation!

Issue #4959, opened 2020-08-11
Feedback: 'Hosting endpoints in Docker Linux containers' - deprecated depends_on argument is used


andreasohlund:
Feedback for 'Hosting endpoints in Docker Linux containers' https://docs.particular.net/samples/hosting/docker/

Location in GitHub: https://github.com/Particular/docs.particular.net/blob/master/samples/hosting/docker/sample.md

Docker v3 [no longer supports conditions for `depends_on`](https://docs.docker.com/compose/compose-file/#depends_on) so our docker compose needs to change.

The recommendation is to [do the checking our self in the container or use tools like `wait-for-it`, `dockerize`, or sh-compatible `wait-for`](https://docs.docker.com/compose/startup-order/).

How should we proceed?

andreasohlund:
@danielmarbach proposed a hosted service like shown here https://github.com/Particular/NServiceBus.Extensions.Hosting/pull/73

andreasohlund:
Restart on failure have also been proposed https://stackoverflow.com/questions/31746182/docker-compose-wait-for-container-x-before-starting-y but would work so nicely when running tests though

timbussmann:
given that our transports/persistences are resilient to outages after starting up successfully initially, shouldn't this be a potential new feature to disable the "availability" requirement at endpoint startup time and let the endpoint start without checking transport/persistence connections (should probably be opt-in given that this can also lead to easily miss configuration issues)

PR #4962, opened 2020-08-12
Adjust docker hosting sample to work with Docker 3

andreasohlund:
Closes https://github.com/Particular/docs.particular.net/issues/4959


samples/hosting/docker/Core_7/Receiver/Program.cs
```diff
@@ -19,6 +20,12 @@ static IHostBuilder CreateHostBuilder(string[] args)
         {
             return Host.CreateDefaultBuilder(args)
                 .UseConsoleLifetime()
+                .ConfigureLogging(logging =>
+                {
+                    logging.AddConsole();
+                    logging.SetMinimumLevel(Microsoft.Extensions.Logging.LogLevel.Information);
```
danielmarbach, line 26: I guess it doens't matter much but I would just add a using above and then use `LogLevel.Information`

samples/hosting/docker/Core_7/Sender/Program.cs
```diff
@@ -20,6 +20,12 @@ static IHostBuilder CreateHostBuilder(string[] args)
         {
             return Host.CreateDefaultBuilder(args)
                 .UseConsoleLifetime()
+                .ConfigureLogging(logging =>
+                {
+                    logging.AddConsole();
+                    logging.SetMinimumLevel(Microsoft.Extensions.Logging.LogLevel.Information);
```
danielmarbach, line 26: I guess it doens't matter much but I would just add a using above and then use `LogLevel.Information`

samples/hosting/docker/Core_7/Receiver/Program.cs
andreasohlund: It was clashing the the nsb logging namespace but I fixed it by removing the critical error action

samples/hosting/docker/Core_7/Receiver/Program.cs
```diff
@@ -39,40 +36,8 @@ static IHostBuilder CreateHostBuilder(string[] args)
 
                     endpointConfiguration.UseSerialization<NewtonsoftSerializer>();
                     endpointConfiguration.EnableInstallers();
-                    endpointConfiguration.DefineCriticalErrorAction(OnCriticalError);
                     return endpointConfiguration;
                 });
         }
-
-        static async Task OnCriticalError(ICriticalErrorContext context)
```
danielmarbach, line 41: I think this needs to stay because we mention that the critical error should be overwritten even here https://docs.particular.net/nservicebus/hosting/docker-host/. Having then our sample not comply feels off or is it ok? Imagine if NSB is shutdown the container would just stay alive. At least `Environment.FailFast` seems to make sense so that the container can be restarted

andreasohlund: Ugh, missed that we recommend to override it, will add it back!

PR #4968, opened 2020-08-17
Show how to register the message session

danielmarbach:


danielmarbach:
We could then also point to the sample in https://docs.particular.net/nservicebus/dependency-injection/#externally-managed-mode-injecting-the-message-session

If you agree we should proceed with this I can then adjust the text as well

SzymonPobiega:
I like the change. Puts users into the pit of success. I think worth changing, even if it will be applicable only for a short while.

samples/dependency-injection/extensions-dependency-injection/Extensions.DependencyInjection_1/Sample/Program.cs
```diff
@@ -12,24 +12,31 @@ static async Task Main()
         #region ContainerConfiguration
 
         var endpointConfiguration = new EndpointConfiguration("Sample");
+        endpointConfiguration.UseTransport<LearningTransport>();
+
+        var serviceCollection = new ServiceCollection();
+        serviceCollection.AddSingleton<MyService>();
```
timbussmann, line 18: what is MyService used for?

samples/dependency-injection/extensions-dependency-injection/Extensions.DependencyInjection_1/Sample/Program.cs
```diff
@@ -12,24 +12,31 @@ static async Task Main()
         #region ContainerConfiguration
 
         var endpointConfiguration = new EndpointConfiguration("Sample");
+        endpointConfiguration.UseTransport<LearningTransport>();
+
+        var serviceCollection = new ServiceCollection();
```
timbussmann, line 17: would it make sense to move all ServiceCollection related code before the endpoint configuration, so that the endpoint configuration is not "interrupted" by ServiceCollection configuration?

samples/dependency-injection/extensions-dependency-injection/Extensions.DependencyInjection_1/Sample/Program.cs
danielmarbach: It was already there, injected into the handler

samples/dependency-injection/extensions-dependency-injection/Extensions.DependencyInjection_1/Sample/Program.cs
danielmarbach: have a look

timbussmann: 👍

PR #4972, opened 2020-08-20
Update DI docs for v8

timbussmann:


Snippets/Core/Core_6/Hosting/Hosting.cs
```diff
@@ -57,21 +57,5 @@ public static void SetInstance(IEndpointInstance endpoint)
             }
         }
         #endregion
-
-        async Task InjectEndpoint()
```
mauroservienti, line 61: @timbussmann why this change to Core 6 snippets?

Snippets/Core/Core_7/Hosting/Hosting.cs
```diff
@@ -57,21 +57,5 @@ public static void SetInstance(IEndpointInstance endpoint)
             }
         }
         #endregion
-
-        async Task InjectEndpoint()
```
mauroservienti, line 61: @timbussmann why this change to Core 7 snippets?

Snippets/Core/Core_6/Hosting/Hosting.cs
timbussmann: @mauroservienti I removed the whole snippet and related doco section as I didn't think it was very clear or useful and it also requires autofac as a dependency just for that snippet. Thoughts?

PR #4979, opened 2020-08-29
data distribution editorial review

adamralph:


nservicebus/messaging/data-distribution.md
```diff
@@ -1,40 +1,40 @@
...
+In the worst-case scenario, some receivers of a data distribution message may succeed in handling the message, and others may fail, causing the message to be moved to the error queue. Until an administrator diagnoses the failure and returns the message to its source queue, transactions may be handled very differently by nodes that had successfully handled the message and those that had not.
 
-These sorts of failures are difficult to track down, as they depend upon more than the contents of a message and the code processing it. The result would also be dependent upon the state of the data being handled by data distribution, which would not be consistent between different servers in a cluster.
+These kinds of failures are difficult to track down. They depend upon more than the contents of a message and the code handling it. The results are also dependent upon the state of the data being handled by data distribution, which will not be consistent across all servers in a cluster.
 
-## Differences from publish/subscribe
+## Differences to publish/subscribe
```
DavidBoike, line 22: I can't back this up, but "difference to" / "different to" sounds unnatural to my ear. (Next paragraph and elsewhere too.) Is this American English vs. English English at play?

nservicebus/messaging/data-distribution.md
```diff
@@ -1,40 +1,40 @@
...
+reviewed: 2020-09-01
 redirects:
 - nservicebus/azure/data-distribution
 - samples/routing/data-distribution
 ---
 
 Data distribution refers to a pattern where asynchronous messages (e.g. using NServiceBus) are used to deliver data to multiple targets in some coordinated fashion.
 
-A common example of a data distribution scenario is having cached data on multiple scaled-out web servers, and then attempting to use messages delivered to all web servers in a cache-busting fashion, so that each web server drops its current cache entries, and then is forced to retrieve fresh data from the database.
+A common example of a data distribution scenario is having cached data on multiple, scaled-out web servers, and attempting to deliver a message to each of them which indicates that they should drop their current cache entries and retrieve fresh data from the database.
```
DavidBoike, line 12: I think comma placement here is an example of the putting-a-comma-where-you-would-pause-when-reading sin, not to actually separate fragments. It's long anyway - how about 2 sentences.
```suggestion
A common example of a data distribution scenario is having cached data on multiple scaled-out web servers and attempting to deliver a message to each of them. The message indicates each server should drop their current cache entries and retrieve fresh data from the database.
```

nservicebus/messaging/data-distribution.md
adamralph: > Is this American English vs. English English at play?

I think it [could be](https://english.stackexchange.com/a/845). I'll revert it to _from_.

nservicebus/messaging/data-distribution.md
adamralph: > I think comma placement here is an example of the putting-a-comma-where-you-would-pause-when-reading sin, not to actually separate fragments.

I thought that the two adjectives should have commas in-between, but it turns out that's only the case when they are of [equal rank](https://www.grammarflip.com/curriculum/commas-between-two-adjectives), which these are not.

> It's long anyway - how about 2 sentences.

👍

PR #4980, opened 2020-09-01
Add details on lambda supported features

WilliamBZA:


Snippets/SQSLambda/SQSLambda_0/Usage.cs
```diff
@@ -61,6 +61,26 @@ static void LicenseFile(AwsLambdaSQSEndpointConfiguration endpointConfiguration,
...
+        {
+            customization.NumberOfRetries(5);
+            customization.TimeIncrease(TimeSpan.FromSeconds(15));
+        });
+
+        #endregion
+
+        #region unrestricted-delayed-delivery
+
+        endpointConfiguration.Transport.UnrestrictedDurationDelayedDelivery();
```
tmasternak, line 79: Looking at the https://docs.particular.net/transports/sqs/delayed-delivery#enable-unrestricted-delayed-delivery
> Unrestricted delayed delivery requires a FIFO queue for each endpoint that receives delayed messages. The transport handles creation of the FIFO queue automatically when installers are enabled.

I think this might not work as we [initialize](https://github.com/Particular/NServiceBus.AwsLambda.Sqs/blob/master/src/NServiceBus.AwsLambda.SQS/TransportWrapper/ManualPipelineInvocationInfrastructure.cs#L10) the AwsEndpoint with NoOpQueueCreator. In other words, I think we should mention that the user needs to create the fifo queue by hand.

previews/aws-lambda-simple-queue-service.md
```diff
@@ -41,6 +41,16 @@ snippet: configure-routing
 
 snippet: custom-diagnostics
 
+### Delayed Retries
+
+[Delayed retries](/nservicebus/recoverability/configure-delayed-retries.md) are disabled by default when using AWS Lambdas. Delayed retries can be configured using the following snippet:
+
+snippet: delayed-retries
+
+If the time increase is expected to be greater than [15 minutes](/transports/sqs/delayed-delivery.md#enable-unrestricted-delayed-delivery), it might be required to enable `UnrestrictedDurationDelayedDelivery` on the endpoint:
```
tmasternak, line 50: This will probably require creating the Fifo queue as well.

tmasternak (changes requested):

Snippets/SQSLambda/SQSLambda_0/Usage.cs
WilliamBZA: Added

previews/aws-lambda-simple-queue-service.md
WilliamBZA: Added

previews/aws-lambda-simple-queue-service.md
```diff
@@ -27,6 +27,10 @@ Since the initial cost of starting an `AwsLambdaSQSEndpoint` endpoint can be hig
 
 snippet: function-definition
 
+### Queue creation
+
+Transport installers are not supported. The required queues will have to be [provisioned manually](/transports/sqs/operations-scripting.md#create-resources).
```
udidahan, line 32: > The required queues will have to be [provisioned manually](/transports/sqs/operations-scripting.md#create-resources).

Maybe better as:

> The creation of the required queues can be scripted via the [CLI](/transports/sqs/operations-scripting.md#create-resources).

tmasternak: Convered in https://github.com/Particular/docs.particular.net/pull/4991

PR #4984, opened 2020-09-01
Stupid change to more easily test ABSDataBus build failure

DavidBoike:


DavidBoike:
Problem was NServiceBus.Azure 6.2.7 on MyGet was different than the version on NuGet, so whenever the MyGet version was retrieved, the build error would occur.

That package version has been deleted from MyGet, as it doesn't really need to be there.

Thanks @bording for the assist!

PR #4986, opened 2020-09-03
Add generic host to the DI upgrade guide

timbussmann:
only added lamar, autofac and unity as the other previously supported containers (structuremap & castle windsor) don't seem to play perfectly with the generic host.


mauroservienti (approved):

adamralph:
@timbussmann @mauroservienti why are the snippets in this PR not in `/Snippets` and compiled, etc?

timbussmann:
@adamralph because those snippets do not compile if you look a bit closer at them. In addition, they would require lots of additional packages to be referenced.

adamralph:
@timbussmann OK, thanks. I guess they are low risk anyway.

PR #5001, opened 2020-09-10
Endpoint should be non-null - SQL assumes this anyway due to PK

### LINKED ISSUE #4993: Feedback: 'SQL Server Transport Upgrade Version 4 to 5'
Feedback for 'SQL Server Transport Upgrade Version 4 to 5' https://docs.particular.net/transports/upgrades/sqlserver-4to5

Location in GitHub: https://github.com/Particular/docs.particular.net/blob/master/transports/upgrades/sqlserver-4to5.md

The template script for creating the `Subscriptions` table is misleading:
The definition of the column `Endpoint` suggests that you can create this column as `Nullable`. Since it is used in a primary key, this does not work for MS Sql Server, tested on 15. When you create the primary key, the column is automatically migrated to `NOT NULL`. 
I would suggest to change the template script, accordingly.

DavidBoike:
Fixes #4993


mikeminutillo (approved):

mikeminutillo:
Raised https://github.com/Particular/NServiceBus.SqlServer/pull/680 to cover the same change in the transport itself.

PR #5008, opened 2020-09-17
editorial review of externally managed container mode sample

adamralph:


samples/dependency-injection/externally-managed-mode/Core_8/Sample/Sample.csproj
```diff
@@ -2,7 +2,6 @@
   <PropertyGroup>
     <TargetFrameworks>netcoreapp3.1;netcoreapp2.1;net48</TargetFrameworks>
     <OutputType>Exe</OutputType>
-    <LangVersion>7.3</LangVersion>
```
adamralph, line 5: As per https://github.com/Particular/RepoStandards/pull/26, we no longer set the language version.

samples/dependency-injection/externally-managed-mode/sample.md
```diff
@@ -1,33 +1,32 @@
...
+title: Externally managed Container Mode Usage
+summary: A sample that uses NServiceBus externally managed container mode to configure a DI container.
 component: Core
-reviewed: 2020-09-04
+reviewed: 2020-09-17
 related:
  - nservicebus/dependency-injection
 ---
 
 ### Configuring an endpoint to use ServiceProvider
```
adamralph, line 10: ServiceProvider is mentioned in this title, but there is no reference to it anywhere in the code or text. It feels like something is missing to connect it with the content.

samples/dependency-injection/externally-managed-mode/sample.md
```diff
@@ -1,33 +1,32 @@
...
 
 ```csharp
 serviceCollection.AddSingleton(p => endpointWithExternallyManagedContainer.MessageSession.Value);
 ```
 
 snippet: InjectingMessageSession
 
-NOTE: The `IMessageSession` can only be resolved from the `IServiceProvider` once the endpoint has been started.
-
+NOTE: The `IMessageSession` may only be resolved from the `IServiceProvider` after the endpoint has been started.
```
adamralph, line 32: Again, ServiceProvider is mentioned here, but there is no reference to it anywhere in the code or text.

samples/dependency-injection/externally-managed-mode/Core_8/Sample/Sample.csproj
adamralph: Hmm... this fails the build. Something feels off with that, but I'll follow up separately. I'll revert this change.

samples/dependency-injection/externally-managed-mode/sample.md
mauroservienti, line 10:
```suggestion
### Configuring an endpoint to use Microsoft's built-in dependency injection container (IServiceProvider)
```

samples/dependency-injection/externally-managed-mode/sample.md
mauroservienti, line 32:
```suggestion
NOTE: The `IMessageSession` may only be resolved from the `IServiceProvider` instance after the endpoint has been started.
```

mauroservienti (commented):
@adamralph I'm not sure I understand. `IServiceProvider` is the default built-in DI seam in .NET Core, what do you think needs to be better explained here?

adamralph:
@mauroservienti I re-read the sample entirely and I think I have a better understanding of it now. I did another pass over the entire sample. I tried to reduce the use of the word service, and I'm not sure we need to mention `IServiceProvider` at all.

What do you think?

mauroservienti (approved):

mauroservienti:
Looks good to me @adamralph

PR #5009, opened 2020-09-17
Correct description of how many Delayed Retries are performed

matthewhermann:
The [documentation for the Default Recoverability Policy](https://docs.particular.net/nservicebus/recoverability/custom-recoverability-policy?) is correct when it describes Immediate Retries:

> If the ImmediateProcessingFailures value is less or equal to the configured MaxNumberOfRetries for Immediate Retries, then the ImmediateRetry action is returned.

[Relevant code](https://github.com/Particular/NServiceBus/blob/c936c938c76f6a2e2e98a2849a9f029bb530e09b/src/NServiceBus.Core/Recoverability/DefaultRecoverabilityPolicy.cs#L30-L33):

```c#
if (errorContext.ImmediateProcessingFailures <= config.Immediate.MaxNumberOfRetries)
{
    return RecoverabilityAction.ImmediateRetry();
}
```

But it incorrectly uses "less or equal" when describing Delayed Retires:

> If the DelayedDeliveriesPerformed is less or equal than MaxNumberOfRetries and the message hasn't reached the maximum time allowed for retries (24 hours), then the DelayedRetry action is returned.

[The code](https://github.com/Particular/NServiceBus/blob/c936c938c76f6a2e2e98a2849a9f029bb530e09b/src/NServiceBus.Core/Recoverability/DefaultRecoverabilityPolicy.cs#L52-L55) actually stops when DelayedDeliveriesPerformed equals MaxNumberOfRetries

```c#
if (delayedDeliveriesPerformed >= config.MaxNumberOfRetries)
{
    return false;
}
```

The snippet for the for the fully customised Recoverability Policy also uses this "less or equal" - this isn't necessarily incorrect, as the snippet is _not_ trying to show the Default Recoverability Policy, but it can lead to confusion.

### DIFF
--- a/nservicebus/recoverability/index_retrycount_core_[6.0,).partial.md
+++ b/nservicebus/recoverability/index_retrycount_core_[6.0,).partial.md
@@ -20,7 +20,7 @@ Given a variety of immediate and delayed configuration values here are the resul
 | 1                | 2              | 6                       |
 | 2                | 2              | 9                       |
 | 1                | 3              | 8                       |
-| **3**            | **5**          | **24  (default)**       |
+| **5**            | **3**          | **24  (default)**       |
 
 ### Scale-out multiplier
 
--- a/Snippets/Core/Core_6/Recoverability/Delayed/CustomPolicies/FullyCustomizedPolicy.cs
+++ b/Snippets/Core/Core_6/Recoverability/Delayed/CustomPolicies/FullyCustomizedPolicy.cs
@@ -42,7 +42,7 @@
             // override delayed retry decision for custom exception
             // i.e. MyOtherBusinessException should do fixed backoff of 5 seconds
             if (context.Exception is MyOtherBusinessException &&
-                context.DelayedDeliveriesPerformed <= config.Delayed.MaxNumberOfRetries)
+                context.DelayedDeliveriesPerformed < config.Delayed.MaxNumberOfRetries)
             {
                 return RecoverabilityAction.DelayedRetry(TimeSpan.FromSeconds(5));
             }
@@ -101,7 +101,7 @@
             // override delayed retry decision for custom exception
             // i.e. MyOtherBusinessException should do fixed backoff of 5 seconds
             if (context.Exception is MyOtherBusinessException &&
-                context.DelayedDeliveriesPerformed <= config.Delayed.MaxNumberOfRetries)
+                context.DelayedDeliveriesPerformed < config.Delayed.MaxNumberOfRetries)
             {
                 return RecoverabilityAction.DelayedRetry(TimeSpan.FromSeconds(5));
             }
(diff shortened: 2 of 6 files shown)


nservicebus/recoverability/index_retrycount_core_[6.0,).partial.md
```diff
@@ -20,7 +20,7 @@ Given a variety of immediate and delayed configuration values here are the resul
 | 1                | 2              | 6                       |
 | 2                | 2              | 9                       |
 | 1                | 3              | 8                       |
-| **3**            | **5**          | **24  (default)**       |
+| **5**            | **3**          | **24  (default)**       |
```
adamralph, line 23: What is the reason for switching these values?

adamralph (commented):
Thanks very much for raising this @matthewhermann.

Your justification for this change makes sense to me. I'd just like to get another pair of eyes on it to confirm before merging.

nservicebus/recoverability/index_retrycount_core_[6.0,).partial.md
matthewhermann: It looks like these were just in the wrong columns - ImmediateRetries is the first column, and DelayedRetries is the second column.

Elsewhere in the page it [mentions the default for Immediate Retries is 5](https://docs.particular.net/nservicebus/recoverability/#immediate-retries):

> By default up to five immediate retries are performed if the message processing results in exception being thrown.

Implies the [default for Delayed Retries is 3](https://docs.particular.net/nservicebus/recoverability/#delayed-retries):

> Delayed Retries introduces another level of retry mechanism for messages that fail processing. Delayed Retries schedules message delivery to the endpoint's input queue with increasing delay, by default first with 10 seconds delay, then 20, and lastly with 30 seconds delay.

And then [states 5 for immediate and 3 for delayed](https://docs.particular.net/nservicebus/recoverability/#total-number-of-possible-retries-scale-out-multiplier):

> When taking the default values for immediate and delayed retries (five and three, respectively) and 6 instances the total number of attempts will be a minumum of (5+1)(3+1)=24 attempts and a maximum of 246=144 attempts.

adamralph: 👍 thanks for the explanation @matthewhermann, that makes sense.

adamralph: Incidentally, the `*` characters in that last sentence are being interpreted as italics markdown. I've fixed that in https://github.com/Particular/docs.particular.net/pull/5013

matthewhermann: Oh yes, I missed that - nice one 👍

tmasternak (approved):

tmasternak:
@matthewhermann thank you very much for the fixes!

PR #5016, opened 2020-09-22
Add some double spaces here and there

SzymonPobiega:


SzymonPobiega:
That was just an automation test

PR #5025, opened 2020-09-29
Add pub/sub to simple RabbitMQ sample

timbussmann:
Splits the single endpoint that used to `SendLocal` into the common Sender/Receiver/Shared structure that is quite common on our samples and let the user choose between sending commands or publishing events.

PR #5028, opened 2020-09-29
Add pub/sub to SQS sample

timbussmann:
adds pub/sub to the SQS simple usage sample. Similar as in #5025


samples/sqs/simple/Sqs_5/Sender/Program.cs
```diff
@@ -0,0 +1,67 @@
...
+        var endpointInstance = await Endpoint.Start(endpointConfiguration);
+
+        await SendMessages(endpointInstance);
+
+        await endpointInstance.Stop();
+    }
+
+    static async Task SendMessages(IMessageSession messageSession)
+    {
+        Console.WriteLine(@"Press
```
SzymonPobiega, line 30: Nitpicking: maybe `c`, `C` and `e`, `E` to be consistent with other samples?

SzymonPobiega (approved):

samples/sqs/simple/Sqs_5/Sender/Program.cs
timbussmann: I was thinking about that, but it felt that whatever chars we select, it's scattered all over the board (not sure about the uppercase idea tbh), so I went with 1-4 so that the options are easy to locate. No strong feelings on this though 👍

PR #5073, opened 2020-10-21
v8 Added Encryption Message Property sample v.3

aleksandr-samila:


andreasohlund (approved):
You need to add a `prerelease.txt` to the sample folder

PR #5080, opened 2020-10-27
one line before headers

adamralph:
I don't really know what "and any other text" refers to, and it may be redundant, but I wanted to stay focused on the two line thing in this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -982,7 +982,7 @@ The first (and all top level) headers in a `.md` page should be a `h2` (i.e. `##
 
 ## Spaces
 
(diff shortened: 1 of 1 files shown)


danielmarbach (approved):

kbaley (approved):
My guess is that two empty lines improves the readability when scanning.

adamralph:
> My guess is that two empty lines improves the readability when scanning.

But then it should be a universal convention, which it is not. I'd even argue that it's distracting, since we only use (inconsistently) in this repo.

PR #5087, opened 2020-10-30
Add documentation, samples and snippets for databus v4+v5

lailabougria:
Disclaimer: Branched off of upgrade-databus to make use of the partials introduced there. This one should be merged before into upgrade-databus.
Both v4 and v5 are pre-release for now;


SeanFeldman:
> Disclaimer: Branched off of upgrade-databus to make use of the partials introduced there. This one should be merged before into upgrade-databus.
> Both v4 and v5 are pre-release for now;

Having both v4 and v5 on a PR for v4 is messy. I suggest we have this one sorted out first and replace v5 with a clean new PR. This way we don't need to worry about cherry-picking, docs tool not handling missing snippets for a version, etc.

PR #5104, opened 2020-11-06
Reflect DI changes in the upgrade guide

danielmarbach:
Properly reflecting https://github.com/Particular/NServiceBus/pull/5741 in the upgrade guide

### DIFF
--- a/nservicebus/upgrades/7to8/dependency-injection.md
+++ b/nservicebus/upgrades/7to8/dependency-injection.md
@@ -21,6 +21,13 @@ The following adapter packages will no longer be provided:
 
 Instead of the container adapter packages, use the [NServiceBus.Extensions.Hosting](/nservicebus/hosting/extensions-hosting.md) package or the [externally managed container mode](/nservicebus/dependency-injection/#externally-managed-mode) to use a third party dependency injection container. See the [migrating to the Generic Host](#microsoft-generic-host) or [migrating to externally managed mode](#externally-managed-container-mode) sections for further information.
 
+## Behavior changes
+
+The behavior has been aligned with the expectations of the `Microsoft.Extensions.DependencyInjection` package. The following changes have been introduced:
+
+- The same component can be registered multiple times
+- The last registration of the same component wins
+
 ## Property injection
 
 Property injection is not covered by `Microsoft.Extensions.DependencyInjection.Abstractions`. Therefore, the NServiceBus default dependency injection container no longer supports property injection. Property injection might still be supported by third party containers.


nservicebus/upgrades/7to8/dependency-injection.md
```diff
@@ -21,6 +21,13 @@ The following adapter packages will no longer be provided:
 
 Instead of the container adapter packages, use the [NServiceBus.Extensions.Hosting](/nservicebus/hosting/extensions-hosting.md) package or the [externally managed container mode](/nservicebus/dependency-injection/#externally-managed-mode) to use a third party dependency injection container. See the [migrating to the Generic Host](#microsoft-generic-host) or [migrating to externally managed mode](#externally-managed-container-mode) sections for further information.
 
+## Behavior changes
```
danielmarbach, line 24: I tried to keep it as short as possible. If you think it would be good to have samples let me know. I felt it would be overkill

mauroservienti (approved):

nservicebus/upgrades/7to8/dependency-injection.md
```diff
@@ -21,6 +21,13 @@ The following adapter packages will no longer be provided:
...
+
+The behavior has been aligned with the expectations of the `Microsoft.Extensions.DependencyInjection` package. The following changes have been introduced:
```
timbussmann, line 26: is "the behavior changed" the right wording here? This is the behavior of using ServiceCollection, which is what changed.  Could we state this e.g. like this "As defined by Microsoft.Extensions.DependencyInjection, IServiceCollection allows to register services multiple times. In such cases, the last registration wins over previous ones. This means that order of registration is relevant".

timbussmann (commented):
I'm ok with the changes, although I think there might be slightly more accurate ways of describing it but I'm fine either way.

nservicebus/upgrades/7to8/dependency-injection.md
danielmarbach: Well this wording is used in an upgrade guide so I think behavior change is the right word because we are comparing v 7 to v8. If that is taken into account I think I'm saying exactly the same as you are suggestion except I'm not further stressing that

> This means that order of registration is relevant

Do you want me to add this?

timbussmann: Yeah I guess it depends a bit on perspective. I'm probably familiar enough to understand the impact of changing to the MS DI containers. It feels a bit that we're documenting the external dependencies behavior but I'm fine with that too in this case (and strictly speaking we have our own ServiceCollection implementation :D )

> Do you want me to add this?

It's probably not that important.

PR #5106, opened 2020-11-06
Storage and transport upgrade guides related to removal of the timeout manager.

ramonsmits:


andreasohlund (approved):

bording:
We also need upgrade guides for the transports, right? Should that be in this PR, or a separate one?

andreasohlund:
Given then title I'd say a separate PR?

ramonsmits:
@bording We didn't change anything behavior or configuration wise for the transports in our TF. I don't think we need to do anything there.

PR #5108, opened 2020-11-09
Explain warn for TransactionScope in netcore

WilliamBZA:
Explanation of https://github.com/Particular/NServiceBus.SqlServer/pull/715

### DIFF
--- a/transports/sql/transactions_ambient-core-warning_sqltransport_[6,).partial.md
+++ b/transports/sql/transactions_ambient-core-warning_sqltransport_[6,).partial.md
@@ -11,3 +11,5 @@ System.NotSupportedException: Enlisting in Ambient transactions is not supported
 ```
 
 To resolve this, choose one of the other transaction modes. Support for `TransactionScope` is expected to be added in future releases of .NET Core.
+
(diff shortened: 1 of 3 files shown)


transports/sql/transactions_ambient-core-warning_sqltransport_[6,).partial.md
```diff
@@ -11,3 +11,5 @@ System.NotSupportedException: Enlisting in Ambient transactions is not supported
 ```
 
 To resolve this, choose one of the other transaction modes. Support for `TransactionScope` is expected to be added in future releases of .NET Core.
+
+It is still possible to use `TransactionScope` in .NET Core but it is not possible to enlist in ambient transactions. This allows for coordination of transactions by ensuring all connections have the same connection string. If `TransactionScope` is used within .NET Core, NServiceBus will log a warning to make sure that this is an explicit decision.
```
tmasternak, line 15:
```suggestion
It is still possible to use `TransactionScope` in .NET Core but promoting local transactions to distributed transactions is no longer supported. It is possible to use multiple connections in the same transaction scope as long as they share the exact same connection string. If `TransactionScope` is used within .NET Core, NServiceBus will log a warning to make sure that this is an explicit decision.
```

tmasternak (commented):
Minor tweak

PR #5120, opened 2020-11-16
Add a sample showing how to truncate a large table at startup

WilliamBZA:
Fixes https://github.com/Particular/NServiceBus.SqlServer/issues/203


samples/sqltransport/startup-purge-behavior/SqlTransport_6/Receiver/TruncateTableStartupTask.cs
```diff
@@ -0,0 +1,17 @@
...
+using Microsoft.Data.SqlClient;
+using NServiceBus;
+
+#region TruncateTableAtStartup
+
+public class TruncateTableAtStartup : INeedInitialization
+{
+    public void Customize(EndpointConfiguration configuration)
+    {
+        var connectionString = @"Data Source=.\SqlExpress;Database=SQLServerTruncate;Integrated Security=True;Max Pool Size=100";
```
mauroservienti, line 11: Is it worth showing how to get the connection string from the `EndpointConfiguration`, if possible?

mauroservienti (approved):
Looks good to me, just a minor comment.

samples/sqltransport/startup-purge-behavior/SqlTransport_6/Receiver/TruncateTableStartupTask.cs
WilliamBZA: I don't know that I know how to do that at this stage of the pipeline, we'd have to actually connect to the database in case someone has provided a connection factory and not a connection string. Any suggestions?

samples/sqltransport/startup-purge-behavior/SqlTransport_6/Shared/SqlHelper.cs
```diff
@@ -0,0 +1,60 @@
...
+    create database [{database}]
+";
+                command.ExecuteNonQuery();
+            }
+        }
+    }
+
+    public static void TruncateMessageTable(string connectionString, string endpointName)
+    {
+        ExecuteSql(connectionString, $@"IF EXISTS (SELECT TOP 1 1 FROM sys.objects WHERE name = 'Samples.SqlServer.TruncateReceiver')
```
tmasternak, line 55: Would it make sense to introduce a parameter for the main input queue. We could also show how to get a transport input queue name based on the endpoint name - this would handle non-default schema and catalog configuration. The only way to do it that I know if is to use `TransportInfrastructure.ToTransportAddress` that is registered by the transport seam in `Settings`. Here is an [example](https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Routing/SubscriptionMigrationMode/SubscriptionMigrationMode.cs#L43) of this being used in Core.

samples/sqltransport/startup-purge-behavior/SqlTransport_6/Receiver/TruncateTableStartupTask.cs
mauroservienti: I think we already do that in the transport somewhere by doing something like:

```
using(var c = connectionFactory())
{
   var connectionString = c.ConnectionString;
}
```

I'm not sure, though, we can do that at this stage. @tmasternak?

samples/sqltransport/startup-purge-behavior/SqlTransport_6/Shared/SqlHelper.cs
WilliamBZA: Done

samples/sqltransport/startup-purge-behavior/SqlTransport_6/Receiver/TruncateTableStartupTask.cs
WilliamBZA: I can't figure out how to do it in a way that doesn't involve reading 2 magic string settings from the settings bag. Doing that makes the sample look more confusing that it needs to be.

tmasternak (approved):

PR #5123, opened 2020-11-17
Update Queue creation documentation

timbussmann:
For some reason, the documentation states that you could register your own implementation of `ICreateQueues`, which is not correct and is part of the transport implementation. I don't remember whether this was different in v5 but given v5 is out of support, it doesn't seem worthwhile to keep the v5 version around any longer either.


andreasohlund (approved):

ramonsmits (changes requested):
I think you misread, it states:

> Transports need to implement a custom queue creator.

Meaning, if you create a transport, you need to have a custom queue creator.

timbussmann:
> Meaning, if you create a transport, you need to have a custom queue creator.

@ramonsmits that doesn't need to be documented, you can't implement the necessary transport classes without doing that. The snippets provided however do not related to the transport and show how to implement a custom implementation and registering it with DI (which won't do anything, because we never resolve this from the container).

ramonsmits:
> how to implement a custom implementation and registering it with DI (which won't do anything, because we never resolve this from the container).

Never have or no longer doing?

timbussmann:
> Never have or no longer doing?

none of the supported versions.

PR #5127, opened 2020-11-19
SqlTransport release supported versions script output

mauroservienti:


DavidBoike (approved):
As long as it doesn't look like the script went off the rails, I would just commit the results of running a script direct to master. In that case the script is the author and you are the reviewer.

mauroservienti:
It was the first time for me, wanted to be sure I wasn't doing anything horribly bad 😄

PR #5132, opened 2020-11-23
Set PowerShell supported version to 5

mauroservienti:


servicecontrol/powershell.md
```diff
@@ -15,6 +15,7 @@ The ServiceControlMgmt module requires:
 
  * Microsoft PowerShell 3.0
 
+ServiceControlMgmt PowerShell module is compatible with PowerShell 5. Versions of PowerShell later than 5 are not supported and might not work as expected.
```
HEskandari, line 18: Do we need to specifically call out PowerShell Core as unsupported?

transports/msmq/distributor/disconnect-workers.md
```diff
@@ -10,6 +10,10 @@ redirects:
 
 The Distributor starts sending messages to a Worker once it is aware of it. A Worker registers itself with a Distributor by sending a message containing a `SessionID` that identifies the current running Worker and the number of messages it can handle concurrently.
 
+## Prerequisites
+
+Prior to installation ensure that PowerShell 2 or higher is installed. NServiceBus PowerShell modules are compatible with PowerShell 5. Versions of PowerShell later than 5 are not supported and might not work as expected.
```
HEskandari, line 15: Again, do we need to call out PSC?

```suggestion
Prior to installation ensure that PowerShell 2 or higher is installed. NServiceBus PowerShell modules are compatible with PowerShell 5. Versions of PowerShell later than 5 (including PowerShell Core) are not supported and might not work as expected.
```

mauroservienti: It doesn't hurt. I'll add it everywhere.

HEskandari (approved):

PR #5133, opened 2020-11-23
Prevent prereleases from using wildcards to avoid samples breaking

danielmarbach:
- [x] Update the csproj that violate the rule


andreasohlund (approved):
https://github.com/Particular/docs.particular.net#temporary-unstables needs to be updated as well

DavidBoike (approved):

PR #5136, opened 2020-11-24
ASP v4 (for core v8) documentation and samples

danielmarbach:


lailabougria (approved):

SeanFeldman (approved):

SeanFeldman (commented):
The new nuget alias is ASTP. But the samples, shouldn't those still be targeting ASP as the component name?

>    System.NullReferenceException : Object reference not set to an instance of an object.  1) Component metadata for component name 'ASTP' was not found.

danielmarbach:
@SeanFeldman The problem was that the prerelease verification tests did not deal with nuget aliases properly. I did a first attempt to solve it although the logic is not 100% bulletproof due to `FirstOrDefault` but it seems to catch what we currently have in docs

PR #5147, opened 2020-12-01
Less getting started

DavidBoike:
So that search results are more focused on what's important


menu/menu.yaml
```diff
@@ -34,9 +34,9 @@
       Title: More tutorials
       Articles:
       - Url: tutorials/monitoring-setup
-        Title: Monitoring NServiceBus solutions
```
kbaley, line 37: I prefer the original. What's the issue with it?

kbaley (approved):

menu/menu.yaml
DavidBoike: Decided to align the names with the demo and make it clear that this is about setting it up, not just information about it.

PR #5153, opened 2020-12-11
ASB Native Pub/Sub Sample

tmasternak:


SeanFeldman (commented):
See comments on the accompanying issue.

samples/azure-service-bus-netstandard/native-integration-pub-sub/ASBS_1/NativeShared/TopologyManager.cs
```diff
@@ -0,0 +1,33 @@
+﻿namespace NativeSender
+{
+    using System;
+    using System.Threading.Tasks;
+    using Microsoft.Azure.ServiceBus;
+    using Microsoft.Azure.ServiceBus.Management;
+
+    public static class TopologyManager
+    {
+        public static async Task CreateSubscription(string connectionString, string name, SqlFilter sqlFilter, string topicPath = "bundle-1", string ruleName = "rule")
```
SeanFeldman, line 10: `name` could benefit from being renamed to `subscriptionName`.

SeanFeldman, line 10: The default value for a rule name in the following way

```c#
ruleName = "rule"
```

is going to break the moment a subscriber has more than one event to subscribe to. While the sample is not a production-grade code, we should strive to lead in the right direction. Making a rule name mandatory is a good way to make readers think about it rather than assume it can be ignored.

samples/azure-service-bus-netstandard/native-integration-pub-sub/ASBS_1/NativeSubscriberA/Program.cs
```diff
@@ -0,0 +1,57 @@
...
+    {
+        var subscriptionName = "Samples.ASB.NativeIntegration.NativeSubscriberA";
+        
+        Console.Title = subscriptionName;
+
+
+        await TopologyManager.CreateSubscription(
+            ConnectionString,
+            subscriptionName,
+        #region SubscriberAFilter
```
SeanFeldman, line 34: Similar to this, I'd suggest adding a snippet for a mandatory rule name and elaborate why is it needed (see my previous comment).

SeanFeldman (changes requested):
That's great. There's one item that has to be addressed. Once that's out of the way, PR is good to be merged.

tmasternak:
@SeanFeldman please have a look now.

SeanFeldman (approved):

PR #5148, opened 2020-12-07
Document IProvidePartitionKeyFromSagaId

lailabougria:

PR #5158, opened 2020-12-15
Review configuration code astp

adamralph:
relates to https://github.com/Particular/docs.particular.net/pull/5148

### DIFF
--- a/Snippets/ASP/ASTP_3/Behaviors.cs
+++ b/Snippets/ASP/ASTP_3/Behaviors.cs
@@ -84,33 +84,38 @@ class ContainerInfoLogicalReceiveContextBehavior
 
 class OrderIdAsPartitionKeyBehavior : Behavior<IIncomingLogicalMessageContext>
 {
-    public OrderIdAsPartitionKeyBehavior(IProvidePartitionKeyFromSagaId partitionKeyFromSagaId)
-    {
+    private static readonly ILog Log = LogManager.GetLogger<OrderIdAsPartitionKeyBehavior>();
+    
+    private readonly IProvidePartitionKeyFromSagaId partitionKeyFromSagaId;
+
+    public OrderIdAsPartitionKeyBehavior(IProvidePartitionKeyFromSagaId partitionKeyFromSagaId) =>
         this.partitionKeyFromSagaId = partitionKeyFromSagaId;
-    }
 
     public override async Task Invoke(IIncomingLogicalMessageContext context, Func<Task> next)
     {
         var correlationProperty = SagaCorrelationProperty.None;
+
         if (context.Message.Instance is IProvideOrderId provideOrderId)
         {
-            var partitionKeyValue = provideOrderId.OrderId;
-            correlationProperty = new SagaCorrelationProperty("OrderId", partitionKeyValue);
+            Log.Debug($"Order ID: '{provideOrderId.OrderId}'");
+
+            correlationProperty = new SagaCorrelationProperty("OrderId", provideOrderId.OrderId);
         }
 
-        await partitionKeyFromSagaId.SetPartitionKey<OrderSagaData>(context, correlationProperty);
+        await partitionKeyFromSagaId.SetPartitionKey<OrderSagaData>(context, correlationProperty)
+            .ConfigureAwait(false);
 
-        if (context.Headers.TryGetValue(Headers.SagaId, out var sagaIdHeader))
+        Log.Debug($"Partition key: {context.Extensions.Get<TableEntityPartitionKey>().PartitionKey}");
+
+        if (context.Headers.TryGetValue(Headers.SagaId, out var sagaId))
         {
-            Log.Debug($"Saga Id Header: {sagaIdHeader}");
+            Log.Debug($"Saga ID: {sagaId}");
         }
 
         if (context.Extensions.TryGet<TableInformation>(out var tableInformation))
         {
-            Log.Debug($"Table Information: {tableInformation.TableName}");
+            Log.Debug($"Table name: {tableInformation.TableName}");
         }
-
-        Log.Debug($"Found partition key '{context.Extensions.Get<TableEntityPartitionKey>().PartitionKey}' from '{nameof(IProvideOrderId)}'");
 
         await next().ConfigureAwait(false);
     }
@@ -121,14 +126,9 @@ class OrderIdAsPartitionKeyBehavior : Behavior<IIncomingLogicalMessageContext>
             base(nameof(OrderIdAsPartitionKeyBehavior),
                 typeof(OrderIdAsPartitionKeyBehavior),
                 "Determines the PartitionKey from the logical message",
-                b => new OrderIdAsPartitionKeyBehavior(b.Build<IProvidePartitionKeyFromSagaId>()))
-        {
+                b => new OrderIdAsPartitionKeyBehavior(b.Build<IProvidePartitionKeyFromSagaId>())) =>
             InsertBefore(nameof(LogicalOutboxBehavior));
-        }
     }
-
-    IProvidePartitionKeyFromSagaId partitionKeyFromSagaId;
-    static readonly ILog Log = LogManager.GetLogger<OrderIdAsPartitionKeyBehavior>();
 }
 #endregion
 
--- a/Snippets/ASP/ASTP_3/Behaviors.cs
+++ b/Snippets/ASP/ASTP_3/Behaviors.cs
@@ -84,10 +84,6 @@ class ContainerInfoLogicalReceiveContextBehavior
 
 class OrderIdAsPartitionKeyBehavior : Behavior<IIncomingLogicalMessageContext>
 {
-    private static readonly ILog Log = LogManager.GetLogger<OrderIdAsPartitionKeyBehavior>();
-    
-    private readonly IProvidePartitionKeyFromSagaId partitionKeyFromSagaId;
-
     public OrderIdAsPartitionKeyBehavior(IProvidePartitionKeyFromSagaId partitionKeyFromSagaId) =>
         this.partitionKeyFromSagaId = partitionKeyFromSagaId;
 
@@ -129,6 +125,9 @@ class OrderIdAsPartitionKeyBehavior : Behavior<IIncomingLogicalMessageContext>
                 b => new OrderIdAsPartitionKeyBehavior(b.Build<IProvidePartitionKeyFromSagaId>())) =>
             InsertBefore(nameof(LogicalOutboxBehavior));
     }
+
+    readonly IProvidePartitionKeyFromSagaId partitionKeyFromSagaId;
+    static readonly ILog Log = LogManager.GetLogger<OrderIdAsPartitionKeyBehavior>();
 }
 #endregion
 
--- a/Snippets/ASP/ASTP_4/Behaviors.cs
+++ b/Snippets/ASP/ASTP_4/Behaviors.cs
@@ -83,35 +83,40 @@ class ContainerInfoLogicalReceiveContextBehavior
 
 #region BehaviorUsingIProvidePartitionKeyFromSagaId
 
-class OrderIdAsPartitionKeyBehavior : Behavior<IIncomingLogicalMessageContext>
+internal class OrderIdAsPartitionKeyBehavior : Behavior<IIncomingLogicalMessageContext>
 {
-    public OrderIdAsPartitionKeyBehavior(IProvidePartitionKeyFromSagaId partitionKeyFromSagaId)
-    {
+    private static readonly ILog Log = LogManager.GetLogger<OrderIdAsPartitionKeyBehavior>();
+    
+    private readonly IProvidePartitionKeyFromSagaId partitionKeyFromSagaId;
+
+    public OrderIdAsPartitionKeyBehavior(IProvidePartitionKeyFromSagaId partitionKeyFromSagaId) =>
         this.partitionKeyFromSagaId = partitionKeyFromSagaId;
-    }
 
     public override async Task Invoke(IIncomingLogicalMessageContext context, Func<Task> next)
     {
         var correlationProperty = SagaCorrelationProperty.None;
+
         if (context.Message.Instance is IProvideOrderId provideOrderId)
         {
-            var partitionKeyValue = provideOrderId.OrderId;
-            correlationProperty = new SagaCorrelationProperty("OrderId", partitionKeyValue);
+            Log.Debug($"Order ID: '{provideOrderId.OrderId}'");
+
+            correlationProperty = new SagaCorrelationProperty("OrderId", provideOrderId.OrderId);
         }
 
-        await partitionKeyFromSagaId.SetPartitionKey<OrderSagaData>(context, correlationProperty);
+        await partitionKeyFromSagaId.SetPartitionKey<OrderSagaData>(context, correlationProperty)
+            .ConfigureAwait(false);
 
-        if (context.Headers.TryGetValue(Headers.SagaId, out var sagaIdHeader))
+        Log.Debug($"Partition key: {context.Extensions.Get<TableEntityPartitionKey>().PartitionKey}");
+
+        if (context.Headers.TryGetValue(Headers.SagaId, out var sagaId))
         {
-            Log.Debug($"Saga Id Header: {sagaIdHeader}");
+            Log.Debug($"Saga ID: {sagaId}");
         }
 
         if (context.Extensions.TryGet<TableInformation>(out var tableInformation))
         {
-            Log.Debug($"Table Information: {tableInformation.TableName}");
+            Log.Debug($"Table name: {tableInformation.TableName}");
         }
-
-        Log.Debug($"Found partition key '{context.Extensions.Get<TableEntityPartitionKey>().PartitionKey}' from '{nameof(IProvideOrderId)}'");
 
         await next().ConfigureAwait(false);
     }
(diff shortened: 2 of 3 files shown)


persistence/azure-table/configuration_code_astp_[3,).partial.md
```diff
@@ -67,23 +67,21 @@ In case installers are enabled, but there's a need to opt out from creating the
...
 
-To atomically store data within the same transaction, all data written as part of the same message handling pipeline needs to share the same partition key.
-When storing sagas, saga IDs are by default deterministically derived from the saga data: the correlation property name and the correlation property value. Therefore, in conversations concerning a saga as part of the message handling pipeline, this deterministically derived ID is a perfect fit to be used as a partition key. Unfortunately, the saga ID assignment happens quite late during the message handling process and is not exposed to the user code.
+During a given message handling pipeline, multiple data storage operations may occur. To commit them atomically, in a single transaction, they must share a partition key. In conversations involving a saga, the saga ID is a good candidate for a partition key. Unfortunately, saga IDs are determined late in the message handling process and are not exposed to user code.
 
 This makes it difficult to:
 
-* Support transactional processing for existing sagas that were stored with the previous version of the persister and may have in-flight messages
```
adamralph, line 77: > and may have in-flight messages

This is redundant. An existing saga that was stored with the previous version of the persister will only be acted upon as a result of a message, and it is irrelevant whether or not there are _other_ "in-flight" messages.

persistence/azure-table/configuration_code_astp_[3,).partial.md
```diff
@@ -67,23 +67,21 @@ In case installers are enabled, but there's a need to opt out from creating the
...
-* Expose the deterministic saga ID early enough to enable business operations to participate in the transaction before the saga has been created
```
adamralph, line 78: > business operations

I believe it's better to avoid the term "business operations" as something distinct from "saga data" operations. Saga data may have just as much, if not more, "business" meaning than "other" data.

persistence/azure-table/configuration_code_astp_[3,).partial.md
```diff
@@ -67,23 +67,21 @@ In case installers are enabled, but there's a need to opt out from creating the
...
 
-To support the above scenarios the `IProvidePartitionKeyFromSagaId` interface implementation can be injected into behaviors in the logical pipeline stage.
+To support the above scenarios, `IProvidePartitionKeyFromSagaId` may be injected into behaviors at the logical pipeline stage:
 
 snippet: BehaviorUsingIProvidePartitionKeyFromSagaId
 
-This component enables users to:
+`IProvidePartitionKeyFromSagaId` enables users to set the partition key based on of the following:
 
-* Derive the table in which business- and saga data will be stored by looking for the `TableInformation` or falling back to the default table
```
adamralph, line 86: This is irrelevant here. While the snippet shows it, it is self explanatory, and it is _not_ something provided by `IProvidePartitionKeyFromSagaId`.

persistence/azure-table/configuration_code_astp_[3,).partial.md
```diff
@@ -67,23 +67,21 @@ In case installers are enabled, but there's a need to opt out from creating the
...
-This component enables users to:
+`IProvidePartitionKeyFromSagaId` enables users to set the partition key based on one of the following:
 
-* Derive the table in which business- and saga data will be stored by looking for the `TableInformation` or falling back to the default table
-* Extract the saga ID header and populate that header value with the partition key for previously created sagas
-  * When the header is not available and the compatibility mode is enabled, do the necessary secondary index lookups. Depending on the configuration, fallback to table scanning. Then populate the found saga ID as the partition key
-  * When no header information is available and the compatibility mode is disabled, precompute the deterministic saga ID based on the provided correlation property information and populate the precomputed value as the partition key
+* The `SagaId` header if present
+* The `SagaId` on the secondary index derived from the correlation property information when migration mode is enabled
+* The deterministic `SagaId` precalculated by using the correlation property information
```
adamralph, line 87: Copied from the [XML docs](https://github.com/Particular/NServiceBus.Persistence.AzureStorage/blob/7fa201a3beff6e31ceb20423a44faed221c4d42c/src/NServiceBus.Persistence.AzureTable/SagaPersisters/IProvidePartitionKeyFromSagaId.cs#L15-L18).

If they are not good enough, they should be improved, and copied back in here again.

persistence/azure-table/configuration_code_astp_[3,).partial.md
danielmarbach: You can shut down and upgrade an endpoint at any time where scenarios like the following are possible:

- Shutdown during the existing inflight message for a saga and might not be handled and rolled back
- During shutdown messages that are destined to go to the saga instance are piling up in the queue to be processed after the endpoint starts
- Messages for the saga instance are already in the queue for the saga instance but not processed yet because the endpoint is shut down

These are the scenarios we summarized under in-flight.

persistence/azure-table/configuration_code_astp_[3,).partial.md
danielmarbach: We used the term "business data" also in the blog post quite intensively

https://particular.net/blog/transactions-with-azure-table-persistence

which passed through editorial review. Do we have any consistency or something we could refer to see what we use in other docs?

persistence/azure-table/configuration_code_astp_[3,).partial.md
danielmarbach: Who says the XML documents need to be the same as this text? I'm not aware of such a rule so I don't see why your proposal of copying the XML docs here is better than the text we have here nor why changing there is a mandate to keep both in sync.

persistence/azure-table/configuration_code_astp_[3,).partial.md
danielmarbach: This edit is incorrect. It is something provided by the helper tool

https://github.com/Particular/NServiceBus.Persistence.AzureStorage/blob/master/src/NServiceBus.Persistence.AzureTable/SagaPersisters/ProvidePartitionKeyFromSagaId.cs#L28-L36

Snippets/ASP/ASTP_4/Behaviors.cs
```diff
@@ -83,36 +83,41 @@ public override async Task Invoke(IIncomingLogicalMessageContext context, Func<T
 
 #region BehaviorUsingIProvidePartitionKeyFromSagaId
 
-class OrderIdAsPartitionKeyBehavior : Behavior<IIncomingLogicalMessageContext>
+internal class OrderIdAsPartitionKeyBehavior : Behavior<IIncomingLogicalMessageContext>
```
danielmarbach, line 86: I'm not aware that we use the visibility modifiers in the snippets. Is this something we introduced recently? If yes then we should also do it in the other snippet https://github.com/Particular/docs.particular.net/pull/5158/files#diff-cf2be1fff701b183297b757e57eae615882ad90da0547fc7996623038de7e479L85. If not then we should remove it

persistence/azure-table/configuration_code_astp_[3,).partial.md
lailabougria: As a developer, I appreciate the documentation providing more insight than just a pure copy of the XML docs. I see this text as an addition to any XML docs. However, if the XML docs are insufficiently clear, we should of course do something about that.

Snippets/ASP/ASTP_3/Behaviors.cs
```diff
@@ -84,34 +84,39 @@ public override async Task Invoke(IIncomingLogicalMessageContext context, Func<T
 
 class OrderIdAsPartitionKeyBehavior : Behavior<IIncomingLogicalMessageContext>
 {
-    public OrderIdAsPartitionKeyBehavior(IProvidePartitionKeyFromSagaId partitionKeyFromSagaId)
-    {
+    private static readonly ILog Log = LogManager.GetLogger<OrderIdAsPartitionKeyBehavior>();
```
lailabougria, line 87: I thought we did not use explicit access modifiers? (I'm all for adding them, as I've trying to retrain my brain *not* to add them everywhere and still fail at that).

Snippets/ASP/ASTP_3/Behaviors.cs
```diff
@@ -84,34 +84,39 @@ public override async Task Invoke(IIncomingLogicalMessageContext context, Func<T
...
     {
         var correlationProperty = SagaCorrelationProperty.None;
+
         if (context.Message.Instance is IProvideOrderId provideOrderId)
         {
-            var partitionKeyValue = provideOrderId.OrderId;
-            correlationProperty = new SagaCorrelationProperty("OrderId", partitionKeyValue);
+            Log.Debug($"Order ID: '{provideOrderId.OrderId}'");
+
+            correlationProperty = new SagaCorrelationProperty("OrderId", provideOrderId.OrderId);
```
lailabougria, line 102: Honestly, I liked the partitionKeyValue as it underlines that it's used as a partition key.

Snippets/ASP/ASTP_3/Behaviors.cs
```diff
@@ -121,14 +126,9 @@ public class Registration : RegisterStep
             base(nameof(OrderIdAsPartitionKeyBehavior),
                 typeof(OrderIdAsPartitionKeyBehavior),
                 "Determines the PartitionKey from the logical message",
-                b => new OrderIdAsPartitionKeyBehavior(b.Build<IProvidePartitionKeyFromSagaId>()))
-        {
+                b => new OrderIdAsPartitionKeyBehavior(b.Build<IProvidePartitionKeyFromSagaId>())) =>
             InsertBefore(nameof(LogicalOutboxBehavior));
-        }
     }
-
```
lailabougria, line 129: Same comment as before, where do we keep the private members? At the bottom or the top? And why is it different in snippets as opposed to our (framework) code?

persistence/azure-table/configuration_code_astp_[3,).partial.md
```diff
@@ -67,23 +67,21 @@ In case installers are enabled, but there's a need to opt out from creating the
 
 snippet: EnableInstallersConfigurationOptingOutFromTableCreation
 
-### Partitioning and compatibility helpers
+### Partitioning and compatibility mode helpers
 
-To atomically store data within the same transaction, all data written as part of the same message handling pipeline needs to share the same partition key.
-When storing sagas, saga IDs are by default deterministically derived from the saga data: the correlation property name and the correlation property value. Therefore, in conversations concerning a saga as part of the message handling pipeline, this deterministically derived ID is a perfect fit to be used as a partition key. Unfortunately, the saga ID assignment happens quite late during the message handling process and is not exposed to the user code.
+During a given message handling pipeline, multiple data storage operations may occur. To commit them atomically, in a single transaction, they must share a partition key. In conversations involving a saga, the saga ID is a good candidate for a partition key. Unfortunately, saga IDs are determined late in the message handling process and are not exposed to user code.
```
lailabougria, line 72: I do like how this is rephrased, however, I think we should keep the information about how the Saga ID is determined that was in the original text

persistence/azure-table/configuration_code_astp_[3,).partial.md
adamralph: @danielmarbach those scenarios are all true, but how does "and may have in-flight messages" add any value to the sentence? The current version (in this PR) says:

> * Enable transactional saga data storage for existing sagas that were stored with the previous version of the persister.

Do we need more than that? This statement is true regardless of whether there are currently messages in the queue for the saga ("in-flight") or whether there will be messages sent to the saga in one year from now.

persistence/azure-table/configuration_code_astp_[3,).partial.md
adamralph: @danielmarbach regardless of what's used elsewhere, does the use of the term add any value to (the new version of) the bullet point? I.e.

> * Allow other data operations to participate in the saga data transaction.

vs

* Allow business operations to participate in the saga data transaction.

I find "other data operations" more precise, and less open to questions like: is X a business operation or not? Ultimately, it doesn't matter. The important point seems to be that other data operations, regardless of what they are, can participate in the saga data transaction.

persistence/azure-table/configuration_code_astp_[3,).partial.md
adamralph: @danielmarbach I see, but I think the way this is laid out here is confusing. The component doesn't _allow users to derive the table_.... The component allows users to pass a value which may or may not be used as the partition key based on an algorithm, and part of that algorithm reads the table name. The confusion is exacerbated by the fact that the code snippet shows the user examining the table without using `IProvidePartitionKeyFromSagaId`:

```C#
if (context.Extensions.TryGet<TableInformation>(out var tableInformation))
{
    Log.Debug($"Table Information: {tableInformation.TableName}");
}
```

Perhaps it's best if we jump on a call next week and thrash this out quickly.

persistence/azure-table/configuration_code_astp_[3,).partial.md
adamralph: > Who says the XML documents need to be the same as this text?

No-one I'm aware of. 🙂 I didn't want to imply that.

> I don't see why your proposal of copying the XML docs here is better than the text we have here

> I appreciate the documentation providing more insight than just a pure copy of the XML docs. I see this text as an addition to any XML docs

It's related to what I mentioned in my previous comment. I don't think this has to be a copy of the XML docs. As I said, probably best to thrash this out over a quick call next week.

Snippets/ASP/ASTP_3/Behaviors.cs
adamralph: I'll put them back at the bottom so we can deal with this separately from this PR.

Snippets/ASP/ASTP_4/Behaviors.cs
adamralph: https://github.com/Particular/docs.particular.net/pull/5158/files#r555686665

persistence/azure-table/configuration_code_astp_[3,).partial.md
danielmarbach: Good arguments @adamralph. I'm 👍 for making the change

Snippets/ASP/ASTP_3/Behaviors.cs
adamralph: Discussed with @danielmarbach and we decided to leave as-is. One guiding factor is that parameter name in `SetPartitionKey` is `correlationProperty`, and it was not always be used as the partition key, so this call would look misleading:

```c#
partitionKeyFromSagaId.SetPartitionKey<OrderSagaData>(context, partitionKeyValue )
```

persistence/azure-table/configuration_code_astp_[3,).partial.md
adamralph: Discussed with @danielmarbach and we agree that that the new bullet is OK.

persistence/azure-table/configuration_code_astp_[3,).partial.md
adamralph: @danielmarbach and I will work out an edit

persistence/azure-table/configuration_code_astp_[3,).partial.md
adamralph: @danielmarbach and I will work out an edit

persistence/azure-table/configuration_code_astp_[3,).partial.md
```diff
@@ -67,23 +67,21 @@ In case installers are enabled, but there's a need to opt out from creating the
...
+* Enable transactional saga data storage for existing sagas that were stored with the previous version of the persister.
+* Allow other data operations to participate in the saga data transaction.
 
-To support the above scenarios the `IProvidePartitionKeyFromSagaId` interface implementation can be injected into behaviors in the logical pipeline stage.
+To support the above scenarios, `IProvidePartitionKeyFromSagaId` may be injected into behaviors at the logical pipeline stage:
 
 snippet: BehaviorUsingIProvidePartitionKeyFromSagaId
 
-This component enables users to:
+`IProvidePartitionKeyFromSagaId` enables users to set the partition key based on one of the following:
```
adamralph, line 83:
```suggestion
`IProvidePartitionKeyFromSagaId` sets the partition key on the `IIncomingLogicalMessageContext` based on one of the following:
```

persistence/azure-table/configuration_code_astp_[3,).partial.md
adamralph, line 87:
```suggestion
* Sets the partition key on the `IIncomingLogicalMessageContext` based on the following algorithm:
  * Use the saga ID header value if present. Otherwise:
  * When compatibility mode is enabled, and the correlation property is not `SagaCorrelationProperty.None`, look up the saga ID either using the secondary index if present, or by table scanning the saga data if that is [enabled](/persistence/azure-table/configuration#saga-configuration). Otherwise:
  * Calculate the saga ID based on the specified correlation property.
* If the table in which all data, including saga data, will be stored is not [already set](persistence/azure-table/configuration#table-name-configuration-and-creation), set it using the saga data name as the table name.
```

persistence/azure-table/configuration_code_astp_[3,).partial.md
adamralph: This is now covered by the explanation of `IProvidePartitionKeyFromSagaId` below.

danielmarbach:
🎆

PR #5165, opened 2020-12-18
Document how to access the native sqs message

lailabougria:
Native SQS message can be accessed by injecting a behavior

### DIFF
--- a/transports/sqs/native-integration.md
+++ b/transports/sqs/native-integration.md
@@ -1,0 +1,15 @@
+---
+title: Native integration
+summary: Considerations when integrating NServiceBus endpoints with native Amazon SQS publishers and consumers.
+reviewed: 2020-12-18
+component: sqs
+versions: '[5.3,]'
+---
+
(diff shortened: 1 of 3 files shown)


danielmarbach (approved):

transports/sqs/native-integration.md
```diff
@@ -0,0 +1,15 @@
...
+title: Native integration
+summary: Considerations when integrating NServiceBus endpoints with native Amazon SQS publishers and consumers.
+reviewed: 2020-12-18
+component: sqs
+versions: '[5.3,]'
+---
+
+This document describes how to consume messages from and send messages to non-NServiceBus endpoints via Amazon SQS in integration scenarios.
+
+### Access to the native Amazon SQS message details
```
danielmarbach, line 11: I was thinking that in theory the native message is also available in the handler and whether it would make sense to show that as well. Then I realized this is an advanced scenario and therefore a behavior as an example is better suited

PR #5185, opened 2021-01-20
Show disabling secondary index

WilliamBZA:
Fixes https://github.com/Particular/docs.particular.net/issues/5176

Shows an example of disabling the secondary index to make it clearer for users how to access the `Compatibility()` function.


samples/rabbitmq/simple/Rabbit_7/Sender/MyCommandHandler.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+using NServiceBus;
+using NServiceBus.Logging;
+
+public class MyCommandHandler : IHandleMessages<MyCommand>
```
mikeminutillo, line 7: Is this related to the same feedback? I'm not sure why this handler is here.

danielmarbach (approved):

PR #5190, opened 2021-01-20
Edit Azure Table upgrade guide section

DavidBoike:
Some adjustments I hope bring greater clarity:

* Removing the "empty" upgrade guide saying "nothing to see here" for Azure Tables 3 to 4. The current convention is not to do this because if made a policy:
  * It would balloon the automatically-generated linkages like on https://docs.particular.net/nservicebus/upgrades/6to7/ to an unusable level.
  * It would create many more articles to review as part of the review process, making it take longer to complete a "full lap" of review.
  * Providing a peace of mind "If you use X don't worry nothing to upgrade there" can be done in markdown in the Core upgrade guide for a major version without causing any of the above.
* Recasting the "NServiceBus.Azure" package as simply "Legacy" in the menu, which can be applied consistently, since we're not that many majors away from having a V6 again which could be confuisng.
* Adjusting the titles to reflect the convention of other packages of mentioning both the From and To versions for upgrade, which could have contributed to the mistake of writing "Upgrading from version 4 to 5" in an article about 3 to 4.


persistence/upgrades/asp-1to2.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Upgrade from Azure Storage persistence Version 1 to Version 2
```
SeanFeldman, line 2: This is incorrect though. Versions 1-3 are a different package and naming it the same way as version 4 would only confuse IMO.

persistence/upgrades/asp-2to3.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: Upgrade from Azure Storage persistence to Version 3
```
SeanFeldman, line 2: Same as above

SeanFeldman (changes requested):

persistence/upgrades/asp-1to2.md
danielmarbach: Agreed with Sean

persistence/upgrades/asp-2to3.md
danielmarbach: Agreed

persistence/upgrades/asp-1to2.md
```diff
@@ -1,5 +1,5 @@
...
+title: Azure Table Storage Upgrade Version 1 to 2
```
danielmarbach, line 2:
```suggestion
title: Azure Storage Persistence Upgrade Version 1 to 2
```

persistence/upgrades/asp-2to3.md
```diff
@@ -1,5 +1,5 @@
...
+title: Azure Table Storage Upgrade Version 2 to 3
```
danielmarbach, line 2:
```suggestion
title: Azure Storage Persistence Upgrade Version 2 to 3
```

danielmarbach (approved):

PR #5192, opened 2021-01-22
Update dotnet-framework-version to 4.7.2

WilliamBZA:


nservicebus/operations/dotnet-framework-version-requirements.md
```diff
@@ -8,6 +8,12 @@ redirects:
  - nservicebus/nservicebus-net-framework-version-requirements
 ---
 
+### NServiceBus Version 8.x
+
+NOTE: NServiceBus Version 8.x is a pre-release version and is subject to change prior to the final release.
+
+Requires [.NET Core 2.1](https://www.microsoft.com/net/core/) (or higher) **or** [.NET Framework Version 4.7.2](https://dotnet.microsoft.com/download/dotnet-framework/net472) (or higher)
```
WilliamBZA, line 15: > Requires [.NET Core 2.1

Is this correct?

mauroservienti: technically speaking no, but effectively yes. The thing is V8 target frameworks are:

- .NET Standard 2.0
- .NET Framework Version 4.7.2

Which makes so that the package can be referenced starting .NET 4.6.x and .NET Core 2.x (IIRC in theory 2.0 too, that should be out of support now). However, the [MSBuild target](https://github.com/Particular/NServiceBus/pull/5870) prevent usage in < 4.7.2.

In essence, I think you're right. @bording?

bording: @mauroservienti That looks correct.

mikeminutillo (approved):

mauroservienti (approved):

PR #5195, opened 2021-01-25
Timeout migration tool: Switch the command order

danielmarbach:


lailabougria (approved):

WilliamBZA (approved):

lailabougria:
Good catch @danielmarbach! Although I did want to bring up if we wanted to keep this in the doco at all, in the sense that the number of possible combinations will grow out of bounds as we work through the rest of the sources and targets. Wouldn't it be sufficient to list the parameters per subcommand? @WilliamBZA @WojcikMike opinion?

danielmarbach:
Might be a good idea to actually clean this up in further iterations of the doco updates. I think we could still bring this in for now

WojcikMike (approved):

danielmarbach:
@lailabougria see https://github.com/Particular/docs.particular.net/pull/5197

PR #5198, opened 2021-01-26
New instance must have different name

yvesgoeleven:


servicecontrol/upgrades/1to2.md
```diff
@@ -49,7 +49,7 @@ The process for a side-by-side upgrade is as follows:
 
 1. Configure the existing version 1 instance of ServiceControl to ingest audit and error messages from a queue named `!disable`
   * This will ensure that the version 1 instance of ServiceControl no longer ingests new messages
-2. Add a new version 2 instance of ServiceControl and configure it to ingest messages from the error and audit queues that the version 1 instance was previously configured to ingest messages from
+2. Add a new version 2 instance of ServiceControl, which has a new name, and configure it to ingest messages from the error and audit queues that the version 1 instance was previously configured to ingest messages from.
```
ramonsmits, line 52:
```suggestion
2. Add a new version 2 instance of ServiceControl, which has a new unique name, and configure it to ingest messages from the error and audit queues that the version 1 instance was previously configured to ingest messages from.
```

ramonsmits (changes requested):
This is not enough, endpoints send heartbeats to the previous SC queue and would need to send them to the new SC queue.

This doesn't seem to be mentioned in the docs too. 

I think it is just better to remove the whole side-by-side section from this upgrade guide as it is very complex and flawed.

servicecontrol/upgrades/1to2.md
```diff
@@ -49,7 +49,7 @@ The process for a side-by-side upgrade is as follows:
 
 1. Configure the existing version 1 instance of ServiceControl to ingest audit and error messages from a queue named `!disable`
   * This will ensure that the version 1 instance of ServiceControl no longer ingests new messages
-2. Add a new version 2 instance of ServiceControl and configure it to ingest messages from the error and audit queues that the version 1 instance was previously configured to ingest messages from
+2. Add a new version 2 instance of ServiceControl, which has a new unique name, and configure it to ingest messages from the error and audit queues that the version 1 instance was previously configured to ingest messages from.
```
WilliamBZA, line 52:
```suggestion
2. Add a new version 2 instance of ServiceControl, with a name different to the version 1 instance name, and configure it to ingest messages from the error and audit queues that the version 1 instance was previously configured to ingest messages from.
```

WilliamBZA (dismissed):

ramonsmits (approved):

PR #5211, opened 2021-02-11
Azure Storage Queues Transport Upgrade v9 to 10

mauroservienti:


SzymonPobiega (approved):

Snippets/ASQ/ASQN_10/Sanitization.cs
```diff
@@ -14,8 +14,10 @@ void CustomSanitization(EndpointConfiguration endpointConfiguration)
     {
         #region azure-storage-queue-sanitization
 
-        var transport = endpointConfiguration.UseTransport<AzureStorageQueueTransport>();
-        transport.SanitizeQueueNamesWith(queueName => queueName.Replace('.', '-'));
+        var transport = new AzureStorageQueueTransport("connection string");
+        transport.QueueNameSanitizer = queueName => queueName.Replace('.', '-');
```
timbussmann, line 18: I'm wondering when we should use object initializer syntax instead?

timbussmann (approved):

Snippets/ASQ/ASQN_10/Sanitization.cs
SzymonPobiega: I didn't use the initialize syntax in RabbitMQ and SQS but I have no strong opinion either way. I guess it would be beneficial if we agree one way or the other ;-)

mauroservienti: I changed to use initializers everywhere to stop discussing with the `.editorconfig`

PR #5213, opened 2021-02-11
IAmStartedByMessage initializes mapped correlation property

Cogax:
It may be obvious to you but I was very confused that it's not explicitly stated, that `IAmStartedByMessage` messages automatically initializes the correlation property on the `SagaData` which is mapped in the `ConfigureHowToFindSaga()` method.


DavidBoike:
Thanks @Cogax! That only happens in NServiceBus 6 and above, and we still have V5 documentation active documentation, so I made a small edit to note that.

PR #5214, opened 2021-02-12
Remove the examples

lailabougria:
@WojcikMike and I discussed this and don't really see the value of keeping the examples.
We explored adding examples for each target keeping the source "undefined" so to speak, but even so, we wonder how meaningful this is. Leaving in a few only helps move forward a part of the users, adding all permutations is noisy.

Thoughts?


danielmarbach (approved):

danielmarbach:
I'm ok with that

WilliamBZA (approved):

PR #5216, opened 2021-02-12
MSMQ v8 transport upgrade

timbussmann:
not included samples due to dependencies
* ~dlq custom checks sample~
* MSMQ to SQL relay sample


Snippets/MsmqTransport/MsmqTransport_2/DisableInstaller.cs
```diff
@@ -0,0 +1,17 @@
...
+
+class DisableInstaller
+{
+    DisableInstaller(EndpointConfiguration endpointConfiguration)
+    {
+        #region disable-installer
+
+        var transport = new MsmqTransport
+        {
+            CreateQueues = false
```
mauroservienti, line 11: More of a general question rather than a comment on the code. Should `CreateQueues` (or maybe something like `CreateInfrastructure)` be on the `TransportDefinition` base class? And then being honored by all transports?

timbussmann: we already bass a createInfrastructure flag on the host settings to the transport. Unfortunately this flag is currently tied to the installers in core so this flag is only needed when you want to disable queue creation while installers are enabled. Most other transports don't really have such an option at the moment and I think its a bit confusing anyway, but I kept it for backwards compatibility.

mauroservienti (approved):

PR #5218, opened 2021-02-14
remove redundant env var in build-projects.yml

adamralph:
has no effect in this workflow


DavidBoike:
Sure is! (Because we're using MSBuild in the PowerShell script, not dotnet anything, becuase some of the projects in Docs still must use older frameworks.)

PR #5221, opened 2021-02-17
Clarify preview support

SzymonPobiega:


previews/support-policy.md
```diff
@@ -17,10 +17,13 @@ For example:
...
 | 0.2.1 | Yes |
-|  ...  | Yes |
 | 1.0.0 | No |
 
-WARN: Only the latest minor version of a preview is supported.
+Note: Only the highest minor version is supported i.e. when `0.2.0` is released, `0.1.0` ceases to be supported.
+
+## Bugfixes
+
+The support policy for preview pacakges guarantees that fixes for critical bugs are released immediately after being fixed. These fixes are available only for the supported version and released as a patch to the latest minor version.
```
sergioc, line 26:
```suggestion
The support policy for preview pacakges guarantees that fixes for critical bugs are released immediately after being fixed. These fixes are available only for the supported version and as such are released as a patch to the latest minor version.
```

SeanFeldman (commented):
I was under the impression that the clarification will also call out that version 1.0 and later (GA-ed versions) will be under the Particular Platform support rather than an unclear "no". I would suggest we have that added.

sergioc:
@SeanFeldman the first paragraph in the document says that. But it's worth emphasizing that in the table too.

SzymonPobiega:
@SeanFeldman please take a look if it looks better now

SeanFeldman (approved):

PR #5225, opened 2021-02-18
Add documentation on how to inject the raven session

lailabougria:
- [x] Point the 6.5 snippet to Myget/nuget version


kbaley:
Is there supposed to be a markdown file to go with this?

lailabougria:
@kbaley No these are snippets, so I'm just adding the snippets that reference the new version, otherwise, the docs engine will be unable to resolve the snippet for the version that we introduce.

mauroservienti (approved):
Does it make sense to craft a sample for this new behavior?

lailabougria:
Personally, I don't think so @mauroservienti . We have the snippets showing how you can inject the session, there's not much more to show than that, so I'm failing to see how a sample adds value in this case.

kbaley (approved):

mauroservienti (approved):

PR #5232, opened 2021-02-24
Update SQS docs and snippets to core V8

SzymonPobiega:


Snippets/Sqs/Sqs_6/Usage.cs
```diff
@@ -0,0 +1,391 @@
...
+
+class Usage
+{
+    void AssumePermissionsOnPolicy(EndpointConfiguration config)
+    {
+        #region assume-permissions
+
+        var transport = new SqsTransport();
+        
+        transport.Policies.SetupTopicPoliciesWhenSubscribing = false;
```
timbussmann, line 22: does that not work with object initializers?

Snippets/Sqs/Sqs_6/Usage.cs
```diff
@@ -0,0 +1,391 @@
...
+        #endregion
+    }
+
+    void MaxTTL(EndpointConfiguration endpointConfiguration)
+    {
+        #region MaxTTL
+
+        var transport = new SqsTransport();
+
+        transport.MaxTimeToLive = TimeSpan.FromDays(10);
```
timbussmann, line 126: in MSMQ I've been using object initializers for these snippets (as that seems to be the code style we should follow now...): https://docs.particular.net/transports/msmq/transportconfig?version=msmqtransport_2

transports/sqs/delayed-delivery_enable_sqs_[6,).partial.md
```diff
@@ -0,0 +1 @@
+In version 6 and higher the unrestricted delayed delivery feature is always enabled but in order to work it has to be enabled in both the sender and the receiver. When communicating with endpoints running older version of the transport make sure to ensure it is enabled.
```
timbussmann, line 1:
```suggestion
In version 6 and higher the unrestricted delayed delivery feature is always enabled by default. In order to work, it has to be enabled in both the sender and the receiver. When communicating with endpoints running older version of the transport make sure to ensure it is enabled.
```

> When communicating with endpoints running older version of the transport make sure to ensure it is enabled.

That sentence reads a bit weird to me. Why shouldn't it be enabled anyway if I want to make use of it? Why is the fact that communicating an older endpoint relevant when I'm supposed to have it enabled anyway?

mauroservienti (approved):
Apart from using object initializers where possible, it looks good to me.

Snippets/Sqs/Sqs_6/Usage.cs
mauroservienti: I don't think so 😞

transports/upgrades/amazonsqs-5to6.md
```diff
@@ -0,0 +1,68 @@
...
+---
+
+## Delayed delivery
+
+The unrestricted delayed delivery is now always enabled so the `UnrestrictedDurationDelayedDelivery()` API has been deprecated.
+
+
+## Configuring the SQS transport
+
+To use the SQS transport for NServiceBus, create a new instance `SqsTransport` and pass it to `EndpointConfiguration.UseTransport`.
```
mauroservienti, line 19:
```suggestion
To use the SQS transport for NServiceBus, create a new `SqsTransport` instance and pass it to `EndpointConfiguration.UseTransport`.
```

transports/upgrades/amazonsqs-5to6.md
```diff
@@ -0,0 +1,68 @@
...
+Use:
+
+```csharp
+var transport = new SqsTransport();
+endpointConfiguration.UseTransport(transport);
+```
+
+## SDK clients
+
+In order to pass customized instances of the SQS and SNS SDK clients to the transport use the `SqsTransport` constructor overload.
```
mauroservienti, line 36:
```suggestion
In order to pass customized instances of the SQS and SNS SDK clients to the transport use the corresponding `SqsTransport` constructor overload.
```

transports/upgrades/amazonsqs-5to6.md
```diff
@@ -0,0 +1,68 @@
...
+
+snippet: 5to6-clients
+
+## S3 configuration
+
+The S3 usage for large messages is configured via `S3` property of the transport. By default the value is `null` which means S3 usage for sending large messages is disabled.
```
mauroservienti, line 42:
```suggestion
The S3 usage for large messages is configured via the `S3` property of the transport. By default, the value is `null` which means S3 usage for sending large messages is disabled.
```

transports/upgrades/amazonsqs-5to6.md
```diff
@@ -0,0 +1,68 @@
...
+
+snippet: 5to6-S3
+
+### Encryption
+
+The encryption is configured via `Encryption` property of the S3 settings object. By default the value is `null` which means the messages are not encrypted.
```
mauroservienti, line 48:
```suggestion
Message payload encryption is configured via the `Encryption` property of the S3 settings object. By default, the value is `null` which means the messages are not encrypted.
```

mauroservienti (changes requested):

Snippets/Sqs/Sqs_6/Usage.cs
SzymonPobiega: Nope :(

mauroservienti (approved):

PR #5241, opened 2021-03-04
Update supported-platforms.md

SeanFeldman:
"Target Framework" and "Platform" are not the right terms to describe frameworks such as ".NET Framework", ".NET Core", or ".NET 5.0". Those are frameworks. "Platform" and "target framework" have different connotations.

For example, "cross-platform" is associated with OS platforms. Windows to Linux would be considered cross-platform. 
When talking about _frameworks_ for .NET, cross-framework would be the right term to describe interoperability between .NET Framework and .NET 5.0.

### DIFF
--- a/nservicebus/upgrades/supported-platforms.md
+++ b/nservicebus/upgrades/supported-platforms.md
@@ -1,7 +1,7 @@
 ---
-title: Supported platforms
-summary: Platforms that are supported by NServiceBus
-reviewed: 2019-07-19
+title: Supported platforms and frameworks
+summary: Platforms and frameworks supported by NServiceBus
+reviewed: 2021-02-04
 related:
  - nservicebus/licensing
  - nservicebus/upgrades/release-policy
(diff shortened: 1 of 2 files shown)


nservicebus/upgrades/supported-platforms.md
```diff
@@ -1,14 +1,14 @@
 ---
-title: Supported platforms
-summary: Platforms that are supported by NServiceBus
-reviewed: 2019-07-19
+title: Supported platforms and frameworks
```
bording, line 2: Since "Frameworks" come first in the table (and the alphabet) how about:

```suggestion
title: Supported frameworks and platforms
```

nservicebus/upgrades/supported-platforms.md
```diff
@@ -1,14 +1,14 @@
...
+summary: Platforms and frameworks supported by NServiceBus
```
bording, line 3:
```suggestion
summary: Frameworks and platforms supported by NServiceBus
```

bording (changes requested):
The menu yaml file also needs to be updated.

bording (approved):

PR #5229, opened 2021-02-23
Add doco for customized outbox isolation level

SzymonPobiega:

PR #5244, opened 2021-03-05
editorial review of recent changes to docs for customized outbox isolation level

adamralph:
relates to https://github.com/Particular/docs.particular.net/pull/5229

### DIFF
--- a/persistence/sql/outbox_modes_sqlpersistence_[6,).partial.md
+++ b/persistence/sql/outbox_modes_sqlpersistence_[6,).partial.md
@@ -34,12 +34,14 @@ In this mode the SQL persistence creates a `TransactionScope` that wraps the who
  - storing the outbox record
  - persisting the application state change applied via `SynchronizedStorageSession`
 
-In addition to that connection managed by NServiceBus, users can open their own database connections in the message handlers. If the underlying database technology supports distributed transactions managed by Microsoft Distributed Transaction Coordinator -- MS DTC (e.g. SQL Server, Oracle or PostgreSQL), the transaction gets escalated to a distributed transaction.
+In addition to that connection managed by NServiceBus, users can open their own database connections in the message handlers. If the underlying database technology supports distributed transactions managed by the Microsoft Distributed Transaction Coordinator (MSDTC) (e.g. SQL Server, Oracle or PostgreSQL), the transaction is escalated to a distributed transaction.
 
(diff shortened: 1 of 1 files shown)


persistence/sql/outbox_modes_sqlpersistence_[6,).partial.md
```diff
@@ -34,12 +34,14 @@ In this mode the SQL persistence creates a `TransactionScope` that wraps the who
...
 
-The `TransactionScope` mode is most useful in legacy scenarios such as when migrating from MSMQ transport to a messaging infrastructure that does not support MSDTC. In order to maintain consistency, the outbox has to be used in place of distributed transport-database transactions. If the legacy database cannot be modified to add the outbox table, the only option is to place the outbox table in a separate database and use distributed transactions between the databases.
+The `TransactionScope` mode is most useful in legacy scenarios such as when migrating from the MSMQ transport to a messaging infrastructure that does not support MSDTC. To maintain consistency, the outbox must be used instead of the distributed transactions which previously included the transport and the database. If the outbox table cannot be added to the legacy database, it may be placed in a separate database, but access to both databases must be included in distributed transactions.
```
mikeminutillo, line 39: > To maintain consistency, the outbox must be used instead of **the distributed transactions** which previously included the transport and the database.

This section (empasis mine) seems a little off and I'm  trying to come up with a clearer way of putting it. I think we need to either drop the **the** or singularize **transactions**.

> To maintain consistency, the outbox must be used instead of **distributed transactions** which previously included the transport and the database.

OR

> To maintain consistency, the outbox must be used instead of **the distributed transaction** which previously included the transport and the database.

Both don't sound quite right either.

mikeminutillo: Maybe flipping it around?

> Instead of using a distrubted transaction which includes the transport and the database, the outbox must be used to maintain consistency,

adamralph: How about this?

```suggestion
The `TransactionScope` mode is most useful in legacy scenarios such as when migrating from the MSMQ transport to a messaging infrastructure that does not support MSDTC. In this scenario, it is no longer possible to use a distributed transaction which includes the transport and the database. To maintain consistency, the outbox must be used instead. If the outbox table cannot be added to the legacy database, it may be placed in a separate database, but access to both databases must be included in distributed transactions.
```

mikeminutillo: Much better :)

PR #5251, opened 2021-03-11
Dynamic strikethrough for unsupported versions

ramonsmits:
Does not require the version script to run only for applying strikethrough styling.

image

### DIFF
--- a/nservicebus/upgrades/supported-versions.js
+++ b/nservicebus/upgrades/supported-versions.js
@@ -1,23 +1,26 @@
-var now = new Date();
-var x = document.getElementsByTagName("tr");
-var i;
-for (i = 0; i < x.length; i++) {
-    var tr = x[i];
-
-    if(tr.childNodes.length>3)
-    {
-        var dateString = tr.childNodes[2].innerText
-        var extended = tr.childNodes[3].innerText
-        var isExtended = extended.startsWith("Extended");
-        if(isExtended)
+'use strict';
+(function () {
+    var now = new Date();
+    var x = document.getElementsByTagName("tr");
+    var i;
+    for (i = 0; i < x.length; i++) {
+        var tr = x[i];
+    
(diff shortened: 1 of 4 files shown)


andreasohlund (approved):

DavidBoike:
Do you intend to also remove the markup strikethrough? Or is this only supposed to handle when the script hasn't run in awhile?

DavidBoike:
Needs to be applied on http://localhost:5000/nservicebus/upgrades/all-versions too.

Why isn't the strikethrough action applied to the first column as well?

nservicebus/upgrades/supported-versions.js
```diff
@@ -0,0 +1,23 @@
+var now = new Date();
```
DavidBoike, line 1: Could you please wrap this whole thing in an immediately-executed function expression (IIFE) so we don't further pollute the global namespace?

```
'use strict';
(function () {
    // All your code
} ());

DavidBoike (changes requested):
To put it all in one review:

* Do the IIFE
* Add to http://localhost:5000/nservicebus/upgrades/support-policy
* Add to http://localhost:5000/nservicebus/upgrades/all-versions
* Probably apply style to first column?

ramonsmits:
> * Probably apply style to first column?

That is a link and for that reason the color isn't applied as links us the link color.

> Do you intend to also remove the markup strikethrough?

The code isn't removing any styling. It only applies in case it detects a row has an unsupported version.

> Or is this only supposed to handle when the script hasn't run in awhile?

It could be a full replacement for the script. The strikethrough is only a visual thing. The dates in the columns are leading.

ramonsmits:
@DavidBoike I removed the `maroon` color to only apply strike through.

DavidBoike (approved):

PR #5274, opened 2021-03-24
ASB native customization not supported with Outbox

timbussmann:
Documents that native message customization is not supported when the Outbox is enabled.

### DIFF
--- a/transports/azure-service-bus/native-message-access.md
+++ b/transports/azure-service-bus/native-message-access.md
@@ -22,3 +22,5 @@ The behavior above uses the native message's `LockedUntilUtc` system property to
 It can also be useful to access the native Service Bus outgoing message from behaviors and handlers for customizations. 
 
 partial: snippets
+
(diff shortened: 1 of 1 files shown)


transports/azure-service-bus/native-message-access.md
```diff
@@ -21,4 +21,6 @@ The behavior above uses the native message's `LockedUntilUtc` system property to
 
 It can also be useful to access the native Service Bus outgoing message from behaviors and handlers for customizations. 
 
-partial: snippets
\ No newline at end of file
+partial: snippets
+
+Note: Native outgoing messages cannot be customized when using the Outbox as customizations are not persistent.
```
SeanFeldman, line 26:
```suggestion
Note: Native outgoing messages cannot be customized when using the [Outbox](/nservicebus/outbox/) as customizations are not persistent.
```

SeanFeldman (approved):
Good addition, @timbussmann. Approving with a suggestion if you want to take it in.

PR #5275, opened 2021-03-24
Add a note about compatibility with outbox

SeanFeldman:


DavidBoike:
Well, not true. You can use Outbox, you'd just have to explicitly set the SendsAtomic transaction level.

SeanFeldman:
> You can use Outbox, you'd just have to explicitly set the SendsAtomic transaction level.

You mean setting it up during endpoint configuration stage?

SeanFeldman:
@DavidBoike, have a look. I've updated the change to reflect that the default transport TX mode can be set explicitly to allow the sample to be used with the outbox feature.

SeanFeldman:
The outbox documentation also has a note on this topic: https://docs.particular.net/nservicebus/outbox/#enabling-the-outbox

PR #5284, opened 2021-03-26
Remove "current status" column from extended support table

andreasohlund:
Since it adds no value and can be forgotten when updating the page. The dates in there already show if it's supported or not. We could in theory make the column dynamic based on those dates but that is YAGNI for now

remains:

- [ ] Make Current Support dynamic

### DIFF
--- a/nservicebus/upgrades/support-policy.md
+++ b/nservicebus/upgrades/support-policy.md
@@ -51,13 +51,12 @@ During the extended support period:
 
 The following table describes the extended support status for all major versions:
 
-|Version|Released|Mainstream Support Expires|Extended Support Expires|
-|:-:|:-:|:-:|:-:|
-|NServiceBus 7|2018-03-30|Current Version|Current Version|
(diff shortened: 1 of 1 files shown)


ramonsmits (approved):

mikeminutillo:
I dislike this change. We're asking the reader to do the datetime math themselves to determine what status a major version is in. It doesn't change that often, the burden of keeping it up to date is minor.

ramonsmits:
@mikeminutillo I do agree with you, it is so low maintenance but the data was incorrect for a long time. 

I've suggested to make that column dynamic just as I recently did for the strikethrough. I'll add that to this commit and revert that "Current Support" column.

DavidBoike:
Or just add a step to the procedure for taking a version off extended support for next time.

Agree, don't like the removal of clarity.

ramonsmits:
> Or just add a step to the procedure for taking a version off extended support for next time.

@DavidBoike a step to which procedure?

PR #5285, opened 2021-03-26
Switch to dotnet build where possible

danielmarbach:


tools/build-samples-and-snippets.ps1
```diff
@@ -93,13 +93,27 @@ foreach($sample in $samples) {
...
+        {
+            dotnet build $sample.Name -nodeReuse:true -verbosity:minimal -restore -property:RestorePackagesConfig=true
+        }
+        
+        if( -not $? ) {
+            $exitCode = 1
+            Write-Output ("::error::Build failed: {0}" -f $sample.FullName)
+            $failedProjects.Add($sample.FullName)
+        }
+    } finally
```
danielmarbach, line 112: Makes also sure when you hit Ctrl+C locally during build it restores the root folder of the docs directory as the current directory

danielmarbach:
Used a marker file to make sure the build engine is switched for the "legacy" projects. It seems there are three

DavidBoike:
Are those really the only solutions containing non-SDK projects?

danielmarbach:
Yep and they use webapplication targets file. Theoretically we could also xpath the csproj but I felt the marker file is simpler

DavidBoike (approved):
For 4 solutions, for sure a marker file makes more sense. Makes it easier to see what current status is by pressing <kbd>T</kbd> on the repo page too.

DavidBoike:
Because the PR build optimizes for changed projects we won't know for sure if everything that now uses dotnet build works for a few hours, but I'm fine with that.

PR #5286, opened 2021-03-29
Clarify that a OCE can be used to abort the receive operation

andreasohlund:


nservicebus/upgrades/7to8/index.md
```diff
@@ -177,4 +177,4 @@ endpointConfiguration.EnableOutbox();
 
 ### AbortReceiveOperation
 
-`ITransportReceiveContext.AbortReceiveOperation` operation is no longer supported
+`ITransportReceiveContext.AbortReceiveOperation` has been deprecated in favour of throwing a [`OperationCancelledException`](https://docs.microsoft.com/en-us/dotnet/api/system.operationcanceledexception)
```
DavidBoike, line 180:
```suggestion
`ITransportReceiveContext.AbortReceiveOperation` has been deprecated in favor of throwing an [`OperationCancelledException`](https://docs.microsoft.com/en-us/dotnet/api/system.operationcanceledexception).

Throwing `OperationCancelledException` will still result in recoverability policies being executed, eventually resulting in the message being sent to the error queue if the message is retried enough times, which is safer than allowing a message to get stuck in an infinite retry loop. This can be overridden using a [custom recoverability policy](/nservicebus/recoverability/custom-recoverability-policy.md) if necessary.
```

DavidBoike (approved):

adamralph:
@andreasohlund are you going to rebase this one?

nservicebus/upgrades/7to8/index.md
```diff
@@ -177,4 +177,6 @@ endpointConfiguration.EnableOutbox();
 
 ### AbortReceiveOperation
 
-`ITransportReceiveContext.AbortReceiveOperation` operation is no longer supported
+`ITransportReceiveContext.AbortReceiveOperation` has been deprecated in favor of throwing an [`OperationCancelledException`](https://docs.microsoft.com/en-us/dotnet/api/system.operationcanceledexception).
```
ramonsmits, line 180: What is the difference with just throwing any exception?

andreasohlund: Good point, throwing any exception would have the same effect. @adamralph @DavidBoike I have a faint memory that we handled an OCE different in the pump early on. Do U remember?

andreasohlund:
Rebased

nservicebus/upgrades/7to8/index.md
ramonsmits: I'm assuming (maybe false assumption) that when using `AbortReceiveOperation` this action would be 'silent'. Meaning, error is logging with stack trace. Potentially only a DEBUG of INFO log entry.

andreasohlund: That is correct, you get full recoverability now which we believe is safer since the old way could result in "silent" infinite loops

andreasohlund: @ramonsmits tweaked the wording, take another look

PR #5295, opened 2021-04-06
Updated ServiceControl hardware recommendations

ramonsmits:


aleksandr-samila (approved):

servicecontrol/servicecontrol-instances/hardware.md
```diff
@@ -9,7 +9,8 @@ ServiceControl as an application can be used to process the entire message load
 ## General recommendations
 
 * Install ServiceControl on a dedicated server in production.
-* 6 GB of RAM minimum
+* Do not host ServiceControl and ServiceControl.Audit instance on the same server, host in isolation
```
HEskandari, line 12: I don't know if this is applicable for customers with low traffic/load?

servicecontrol/servicecontrol-instances/hardware.md
```diff
@@ -9,7 +9,8 @@ ServiceControl as an application can be used to process the entire message load
...
+* 12 GB of dedicated RAM minimum per instance (16GB machine if only hosting ServiceControl).
```
HEskandari, line 13: Not sure what this means, maybe break it down into two lines? e.g. when to use 12GB when to use 16GB?

Just a small improvement:

```suggestion
* A minimum of 12 GB of dedicated RAM per instance (16GB machine if only hosting ServiceControl).
```

servicecontrol/servicecontrol-instances/hardware.md
ramonsmits: I've rewritten that part.

HEskandari (approved):

PR #5316, opened 2021-04-22
Review throttling

danielmarbach:


danielmarbach:
@timbussmann would you mind quickly reviewing my addition to the sample?

samples/throttling/sample.md
```diff
@@ -60,4 +60,6 @@ snippet: RegisterBehavior
 
 Handles the detection of `Octokit.RateLimitExceededException` and defers the message.
 
-snippet: ThrottlingBehavior
\ No newline at end of file
+snippet: ThrottlingBehavior
+
+NOTE: The above example creates a copy of the message and does not copy headers of the original message. Should headers be required they need to be copied from the `context.Headers` into the  `SendOptions`.
```
timbussmann, line 65:
```suggestion
NOTE: The above example creates a copy of the message but does not copy headers of the original message. Should headers be required they need to be copied from the `context.Headers` into the  `SendOptions`.
```

timbussmann, line 65:
```suggestion
NOTE: The above example creates a new message and does not copy headers of the original message. Should headers be required they need to be copied from the `context.Headers` into the  `SendOptions`.
```

timbussmann (approved):
two alternative suggestions to consider as "a copy" would probably be more expected to be a full copy (including headers), therefore using either "but" instead of "and" or avoiding the word copy.

danielmarbach:
Thanks @timbussmann

PR #5318, opened 2021-04-27
editorial review of recent changes to throttling sample

adamralph:
_relates to https://github.com/Particular/docs.particular.net/pull/5316_

### DIFF
--- a/samples/throttling/sample.md
+++ b/samples/throttling/sample.md
@@ -62,4 +62,4 @@ Handles the detection of `Octokit.RateLimitExceededException` and defers the mes
 
 snippet: ThrottlingBehavior
 
-NOTE: The above example creates a new message and does not copy headers of the original message. Should headers be required they need to be copied from the `context.Headers` into the  `SendOptions`.
+NOTE: The above example sends a copy of the original message, but without the headers of the original message. If the headers of the original message are required, they must be copied from `context.Headers` to `SendOptions`.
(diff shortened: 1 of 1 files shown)


samples/throttling/sample.md
```diff
@@ -62,4 +62,4 @@ Handles the detection of `Octokit.RateLimitExceededException` and defers the mes
 
 snippet: ThrottlingBehavior
 
-NOTE: The above example creates a new message and does not copy headers of the original message. Should headers be required they need to be copied from the `context.Headers` into the  `SendOptions`.
+NOTE: The above example sends a copy of the original message, but does not copy the headers of the original message. If the headers of the original message are required, they must be copied from `context.Headers` to `SendOptions`.
```
adamralph, line 65: Does this note apply to the entire sample, or just the _The pipeline behavior_ section?

If the latter, where is the code that copies the message?

danielmarbach: The behavior only

danielmarbach (approved):

samples/throttling/sample.md
adamralph: OK. Which bit of code does that? It may be worth pointing out in the note.

danielmarbach: Technically this part but this can get convoluted and complex very quickly

https://github.com/Particular/docs.particular.net/blob/master/samples/throttling/Core_7/Limited/ThrottlingBehavior.cs#L56

because from the perspective of the code we just pass in `MessageBeingHandled` and then you have to explain what copying means in this context.

PR #5327, opened 2021-05-06
Explain additional nodes

WilliamBZA:


timbussmann (approved):
I guess a snippet would be cool for the v7 version but this is a good start.

andreasohlund:
>I guess a snippet would be cool for the v7 version but this is a good start.

The cluster sample have a snippet for this

andreasohlund (approved):

PR #5343, opened 2021-05-18
Use common APM names with tagged metadata

mikesigs:
We are using NServiceBus and DataDog and found that the original implementation of this integration created far too many metrics and was nearly impossible to create usable dashboards. What wound up really helping is using the common APM names for specific metrics, and then added things liked EndpointName and MessageType as tags on that metric. This allows you to aggregate all of a single metric for all messages combined (e.g. total cluster throughput), as well as break things down by individual endpoints (e.g. endpoint failure rate), or to the fine grained message type (e.g. processed message count by message).

### DIFF
--- a/samples/logging/datadog/Core_7/Endpoint/DatadogFeature.cs
+++ b/samples/logging/datadog/Core_7/Endpoint/DatadogFeature.cs
@@ -1,5 +1,6 @@
-﻿using System;
+using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Threading.Tasks;
 using NServiceBus;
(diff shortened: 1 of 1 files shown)


mikesigs:
As an example. This is a Timeseries Dashboard that shows me the Throughput by Endpoint (msg/s). This kind of graph isn't possible if each message publishes a unique APM metric name.

image

The metric is configured like so `sum:nservicebus.processed{$dashboard_environment} by {endpoint}.as_rate().rollup(sum, 1)`

image

SeanFeldman:
Thank you @mikesigs.

PR #5346, opened 2021-05-20
Add net5 targets

andreasohlund:


andreasohlund:
@DavidBoike I noticed that all the snippets just use `net48` I guess that is on purpose? (Doesn't that create a gap where we could have a snippet, eg using `BinaryFormatter` , that wouldn't work on `.net5`)

WilliamBZA (approved):

DavidBoike:
@andreasohlund The build output of snippets doesn't matter, except for the fact that it _did successfully build_, so previously we haven't seen any reason to multitarget snippets given that doing so will extend the build time in the [BuildProjects workflow](https://github.com/Particular/docs.particular.net/blob/master/.github/workflows/build-projects.yml).

Note that happens in the background and isn't necessary to generate the site, so having that build take longer isn't a horrible thing. It already takes 25-30 minutes, what's a few more?

So in summary, if adding a net5.0 target to snippets provides a check that snippets aren't using something that's broken in .NET 5 then go ahead and do it.

DavidBoike:
Well, that build is on a cron, so it won't happen until merged to master. But we have the history: https://github.com/Particular/docs.particular.net/actions/workflows/build-projects.yml

PR #5347, opened 2021-05-20
Add net5 target to snippets

andreasohlund:


andreasohlund:
Decided to skip this for now

PR #5350, opened 2021-05-26
Cancellation & exception handling

DavidBoike:


nservicebus/cancellation-and-exception-handling.md
```diff
@@ -0,0 +1,85 @@
...
+All NServiceBus components use the [Particular.Analyzers](https://www.myget.org/feed/particular/package/nuget/Particular.Analyzers) package which contains Roslyn analyzers to verify the patterns described here, among others. If developing an extension to NServiceBus, using the same analyzers package is highly recommended.
+
+## Catching `Exception`
+
+One common pattern is to catch and swallow the general `Exception` type in some cases, like this:
+
+```csharp
+try
+{
+    await SomeOperation(cancellationToken);
```
andreasohlund, line 20: I know what we mean but I wonder if users will treat this as "all tokens no matter what" rather than "if this is the token that NServiceBus provides to participate in shutdown", do you think readers will get that or could we make it more clear somehow?

andreasohlund (approved):

nservicebus/cancellation-and-exception-handling.md
DavidBoike: Putting inside a handler looked too "you should catch exceptions inside a handler" and I just couldn't bring myself to do it, so I tried to get the job done with expanded comments in the snippet.

DavidBoike:
Closing in favor of https://github.com/Particular/docs.particular.net/pull/5354

PR #5351, opened 2021-05-27
Warn about binaryformatter on net5

andreasohlund:


nservicebus/messaging/databus/index.md
```diff
@@ -65,6 +65,7 @@ snippet: MessageWithLargePayloadUsingConvention
 
 By default NServiceBus uses a `BinaryFormatter` to serialize and deserialize data bus properties.
 
+WARN: When targeting `.net5.0` a customer serializer must be used (see below) since `BinaryFormatter` [no longer is supported](https://docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/5.0/binaryformatter-serialization-obsolete).
```
WilliamBZA, line 68: Isn't this only in asp.net? Should we make that distinction?

I'm not really sure it makes sense to make the distinction since asp.net will likely only be 1 or 2 endpoints within a larger system. Meaning they'd have to use the same formatter in the other endpoints too.

WilliamBZA (approved):

nservicebus/messaging/databus/index.md
andreasohlund: Its not supported/recommended regardless of hosting but it's only asp.net that is actively preventing it so I'd say keep as is since you shouldn't use it regardless of hosting

dvdstelt (approved):

PR #5352, opened 2021-05-28
Added note about record type

dvdstelt:
Record types can't be deserialized when they're used as immutable properties. Added a note.

### DIFF
--- a/nservicebus/sagas/index.md
+++ b/nservicebus/sagas/index.md
@@ -35,7 +35,7 @@ Using NServiceBus, it is possible to explicitly define the data used for this st
 
 snippet: simple-saga-data
 
-NOTE: With .NET 5 the immutable type `record` was introduced. The type can be used, except for properties that need to be immutable. Serializers will not be able to deserialize data into the object.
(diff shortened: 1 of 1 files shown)


WilliamBZA (approved):

nservicebus/sagas/index.md
```diff
@@ -35,6 +35,8 @@ Using NServiceBus, it is possible to explicitly define the data used for this st
 
 snippet: simple-saga-data
 
+NOTE: With .NET 5 the immutable type `record` was introduced. The type can be used, except for properties that need to be immutable. Serializers will not be able to deserialize data into the object.
```
andreasohlund, line 38:
```suggestion
NOTE: .NET 5 `record` types can't be used for properties that need to be mutable since it would prevent serializers to deserialize the saga data when loaded from storage.
```

dvdstelt:
```suggestion
NOTE: If a saga property is a [record type](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/record), that record type must be mutable so it can be deserialized.
```

PR #5354, opened 2021-05-31
cancellation and catching exceptions

adamralph:
alternative to https://github.com/Particular/docs.particular.net/pull/5350

as discussed with @andreasohlund

### DIFF
--- a/nservicebus/cancellation-and-catching-exceptions.md
+++ b/nservicebus/cancellation-and-catching-exceptions.md
@@ -22,7 +22,7 @@ Note that an `OperationCanceledException` thrown when `CancellationToken.IsCance
 
 Most of the time, when `System.Exception` is caught, the assumption is that the operation has failed, not that it has been canceled. In these cases, the correct way to catch the `Exception` is add a filter which excludes exceptions which represent cancellation:
 
-```c#
+```csharp
 try
 {
     await foo.Bar(cancellationToken).ConfigureAwait(false);
@@ -39,7 +39,7 @@ Note that, in the above example, exceptions that represent cancellation are _not
 
 In most cases, exceptions which represent cancellation should not be caught, and should be allowed to propagate to the caller of the current method. In some cases, it may be necessary to catch these exceptions to take specific actions. The correct way to catch these exceptions is to add a filter which includes only exceptions which represent cancellation:
 
-```c#
+```csharp
 try
 {
     await foo.Bar(cancellationToken).ConfigureAwait(false);
@@ -53,8 +53,8 @@ catch (OperationCanceledException ex) when (cancellationToken.IsCancellationRequ
 }
 catch (Exception ex)
 {
-    // this catch (if it is required) will now catch only exceptions which do NOT represent cancellation,
-    // so it does not require a filter
+    // this catch (if it is required) will now catch only exceptions which do NOT 
+    // represent cancellation, so it does not require a filter
 }
 ```
 
@@ -69,7 +69,7 @@ public static bool IsCausedBy(this Exception ex, CancellationToken cancellationT
 
 Using this method, the `catch` filters are much simpler in both cases:
 
-```c#
+```csharp
 try
 {
     await foo.Bar(cancellationToken).ConfigureAwait(false);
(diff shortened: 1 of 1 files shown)


nservicebus/cancellation-and-catching-exceptions.md
```diff
@@ -0,0 +1,138 @@
...
+
+Note that an `OperationCanceledException` thrown when `CancellationToken.IsCancellationRequested` is `false`, does _not_ represent cancellation. All this means is that `foo.Bar` threw an `OperationCanceledException` for some reason other than cancellation, and the operation should be treated as a failure.
+
+## Catching `System.Exception`
+
+Most of the time, when `System.Exception` is caught, the assumption is that the operation has failed, not that it has been canceled. In these cases, the correct way to catch the `Exception` is add a filter which excludes exceptions which represent cancellation:
+
+```c#
```
DavidBoike, line 25:
```suggestion
```csharp
```

nservicebus/cancellation-and-catching-exceptions.md
```diff
@@ -0,0 +1,138 @@
...
+}
+```
+
+Note that, in the above example, exceptions that represent cancellation are _not_ caught. This is desirable because cancellation should be propagated to the caller of the current method.
+
+## Catching `System.OperationCanceledException`
+
+In most cases, exceptions which represent cancellation should not be caught, and should be allowed to propagate to the caller of the current method. In some cases, it may be necessary to catch these exceptions to take specific actions. The correct way to catch these exceptions is to add a filter which includes only exceptions which represent cancellation:
+
+```c#
```
DavidBoike, line 42:
```suggestion
```csharp
```

nservicebus/cancellation-and-catching-exceptions.md
```diff
@@ -0,0 +1,138 @@
...
+{
+    // foo.Bar was cancelled — take appropriate action
+
+    // re-throw the exception to propagate the cancellation to the caller of the current method
+    throw;
+}
+catch (Exception ex)
+{
+    // this catch (if it is required) will now catch only exceptions which do NOT represent cancellation,
+    // so it does not require a filter
```
DavidBoike, line 57:
```suggestion
    // this catch (if it is required) will now catch only exceptions which do NOT 
    // represent cancellation, so it does not require a filter
```

nservicebus/cancellation-and-catching-exceptions.md
```diff
@@ -0,0 +1,138 @@
...
+If exception handling is widespread, it may be helpful to introduce an `IsCausedBy` extension method:
+
+```csharp
+public static bool IsCausedBy(this Exception ex, CancellationToken cancellationToken) =>
+    ex is OperationCanceledException && cancellationToken.IsCancellationRequested;
+```
+
+Using this method, the `catch` filters are much simpler in both cases:
+
+```c#
```
DavidBoike, line 72:
```suggestion
```csharp
```

nservicebus/cancellation-and-catching-exceptions.md
```diff
@@ -0,0 +1,138 @@
...
+    // re-throw the exception to propagate the cancellation to the caller of the current method
+    throw;
+}
+```
+
+Note that the second example is catching `Exception` rather than `OperationCanceledException` and the `IsCausedBy` method is filtering on the exception type instead of the `catch` clause itself. This results in one extra type comparison (`isinst`) in the resulting IL code, but the performance cost is negligible and the code is simpler to read.
+
+## Inside the message processing pipeline
+
+For code inside the message processing pipeline, such as a message handler or saga, the above considerations are the same. The only difference is that the `CancellationToken` is provided by the `context.CancellationToken` property. For example:
```
DavidBoike, line 101:
```suggestion
For code inside the message processing pipeline, such as a message handler, saga, or pipeline behavior, the above considerations are the same. The only difference is that the `CancellationToken` is provided by the `context.CancellationToken` property.

However, it is generally preferred to not catch exceptions within message handlers and sagas, and instead let exceptions be handled by [recoverability](/nservicebus/recoverability/).
```

nservicebus/cancellation-and-catching-exceptions.md
```diff
@@ -0,0 +1,138 @@
...
+    }
+    catch (Exception ex) when (ex.IsCausedBy(context.CancellationToken))
+    {
+        // foo.Bar was cancelled — take appropriate action
+
+        // re-throw the exception to propagate the cancellation to the caller of the current method
+        throw;
+    }
+}
+```
```
DavidBoike, line 138: Given this is not in the menu and mostly for users of Particular.Analyzers, and we don't want users to catch exceptions in handlers anyway, these examples are just red herrings.
```suggestion
```

PR #5358, opened 2021-06-07
license minification

HEskandari:
Adds a note about license minification with a parameter added to the download link

### DIFF
--- a/nservicebus/licensing/index.md
+++ b/nservicebus/licensing/index.md
@@ -35,6 +35,8 @@ There are several options available for installing the license file.
 
 partial: license-management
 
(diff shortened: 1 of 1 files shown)


nservicebus/licensing/index.md
```diff
@@ -35,6 +35,8 @@ There are several options available for installing the license file.
 
 partial: license-management
 
+When the license content is embedded (for example for Docker deployments of endpoints) the content could be _minified_. This can be done by removing the spaces and line-breaks within the file. When downloading the license file from Particular website, a 'minify=true' parameter can be added to the download URL to achieve the same.
```
adamralph, line 38:
```suggestion
Sometimes the license needs to be embedded in a single line of text. For example, in a command line when deploying an endpoint with Docker. For these purposes, the license can be _minified_ by removing all spaces and line-breaks. This can be done by adding `minify=true` to the query string of the URL used to download a license file from the Particular Software website.
```

PR #5364, opened 2021-06-10
Document SQL Transport retry behavior

DavidBoike:


transports/sql/index.md
```diff
@@ -81,6 +81,12 @@ Each endpoint should use a dedicated SQL Server principal with `SELECT` and `DEL
 
 partial: alwaysencrypted
 
+### Retries
+
+When an exception occurs during message handling, the transaction is already in doubt and must be rolled back. Failure info is stored in memory, and can't influence the next step for that message (immediate retry, delayed retry, or forward to the error queue) until the next time it is processed on that node.
```
ramonsmits, line 86: Wouldn't this only be for sends atomic with receive and transaction scope? When using receive only, and not use connection sharing there would be no need to fully rollback the transaction before it can do an immediate retry.

Not sure how the current receive strategy for receive only behaves.

Maybe all strategies use the same retry model. If so, discard my comment.

transports/sql/index.md
```diff
@@ -81,6 +81,12 @@ Each endpoint should use a dedicated SQL Server principal with `SELECT` and `DEL
...
+
+When an endpoint is scaled out to multiple endpoint instances, this means that more retries may be observed than configured, as each instance accumulates failures until a single node has observed enough retries to escalate to delayed retries. It's also possible for immediate retries to be observed even when [Recoverability](/nservicebus/recoverability/) is configured for zero immediate retries.
```
ramonsmits, line 88:
```suggestion
When an endpoint is scaled out to multiple endpoint instances, this means that more retries may be observed than configured, as each instance accumulates failures until a single node has observed enough retries to escalate to delayed retries. It's also possible for immediate retries to be observed on a different instance even when [Recoverability](/nservicebus/recoverability/) is configured for zero immediate retries.
```

ramonsmits (approved):

transports/sql/index.md
DavidBoike: SendsAtomic and ReceiveOnly use the same receiving strategy and [roll back the transaction on error](https://github.com/Particular/NServiceBus.SqlServer/blob/master/src/NServiceBus.Transport.SqlServer/Receiving/ProcessWithNativeTransaction.cs#L48-L52).

PR #5372, opened 2021-06-21
Try to clarify the diagram on the outbox flow

JeffreyAllenMiller:
Particular Team.  I had to read this too many times for everything to sink in (keep in mind, i might be slow :).  I was constantly asking my self how is it possible for the messages to be safely sent if that is occurring as part of the handler logic.  That is until I realized that there are two different processes occurring.  I realize the diagram in step 6 says that the transaction is committed, but I made the mistake of assuming there was still a parent transaction wrapped around the whole process (i.e. steps 1 through 10).  Subsequently, the diagram, may merit some updates as it does not make it clear that there are two distinct processes at play.  I also added some verbiage to high light that there are multiple process.  @seanfarmar


nservicebus/outbox/index.md
```diff
@@ -93,6 +93,10 @@ More detail on each stage of the process:
...
... sending emails.  At this juncture, you may be asking yourself, “How do the outbound messages actually get sent if they are stuck serialized in the Outbox table”.  The answer to that question, requires understanding what the second process is responsible for.  Essentially, the second process monitors the Outbox table and does the actual sending of the messages and then updates the Outbox to indicate that it was sent so that it is not picked up and sent again.  It is through these two processes that we are able to simulate guaranteed messaging without the use of distributed transactions, with a few caveats.  In the case of the second process it is technically possible that the same message can be physically sent multiple times.  For example, after a message is sent, an exception may be thrown while trying to update the Outbox to indicate that it was sent.  In this scenario the processing of the Outbox record would be retried and the message would be sent again.  Fortunately, this anomaly is handled by endpoint message handers, thanks to the Deduplication feature which ensures that messages with the same Message Id will only be processed one time, thus making the handlers idempotent.
```
seanfarmar, line 96:
```suggestion
The key to understanding the Outbox Design Pattern is realizing that there are two completely distinct processes occurring.  Meaning all of the steps from 1 through 10 are not participating in the same parent transaction or logical process.  In fact, there are two logical processes occurring.  The first process, represents the message handler logic, with the exception that when using the Outbox feature all outbound messages are also serialized and saved in the Outbox table within the same database as the business data.  Which is to imply that the business logic and the serialization and saving of outbound messages occur within the same transaction.  This ensures that the handler logic and all outbound messages are successfully saved or they are all collectively rolled back.  The exception to this would be, actions within handlers that do not enlist in the transient transaction, such as sending emails.  At this juncture, you may be asking yourself, “How do the outbound messages actually get sent if they are stuck serialized in the Outbox table”.  The answer to that question requires understanding what the second process is responsible for.  Essentially, the second process monitors the Outbox table and does the actual sending of the messages and then updates the Outbox to indicate that it was sent so that it is not picked up and sent again.  It is through these two processes that we are able to simulate guaranteed messaging without the use of distributed transactions, with a few caveats.  In the case of the second process, it is technically possible that the same message can be physically sent multiple times.  For example, after a message is sent, an exception may be thrown while trying to update the Outbox to indicate that it was sent.  In this scenario, the processing of the Outbox record would be retried and the message would be sent again.  Fortunately, this anomaly is handled by endpoint message handers, thanks to the Deduplication feature which ensures that messages with the same MessageId will only be processed one time, thus making the handlers idempotent. 
```

nservicebus/outbox/index.md
```diff
@@ -93,6 +93,10 @@ More detail on each stage of the process:
...
... sending emails.  At this juncture, you may be asking yourself, “How do the outbound messages actually get sent if they are stuck serialized in the Outbox table”.  The answer to that question requires understanding what the second process is responsible for.  Essentially, the second process monitors the Outbox table and does the actual sending of the messages and then updates the Outbox to indicate that it was sent so that it is not picked up and sent again.  It is through these two processes that we are able to simulate guaranteed messaging without the use of distributed transactions, with a few caveats.  In the case of the second process, it is technically possible that the same message can be physically sent multiple times.  For example, after a message is sent, an exception may be thrown while trying to update the Outbox to indicate that it was sent.  In this scenario, the processing of the Outbox record would be retried and the message would be sent again.  Fortunately, this anomaly is handled by endpoint message handers, thanks to the Deduplication feature which ensures that messages with the same MessageId will only be processed one time, thus making the handlers idempotent.
```
seanfarmar, line 96:
```suggestion
The key to understanding the Outbox Design Pattern is realizing that there are two completely distinct processes occurring.  Meaning all of the steps from 1 through 10 are not participating in the same parent transaction or logical process.  In fact, there are two logical processes occurring.  The first process, represents the message handler logic, with the exception that when using the Outbox feature all outbound messages are also serialized and saved in the Outbox table within the same database as the business data.  Which is to imply that the business logic and the serialization and saving of outbound messages occur within the same transaction.  This ensures that the handler logic and all outbound messages are successfully saved or they are all collectively rolled back.  The exception to this would be, actions within handlers that do not enlist in the transient transaction, such as sending emails.  At this juncture, the question may be: “How do the outbound messages actually get sent if they are stuck serialized in the Outbox table”.  The answer to that question requires understanding what the second process is responsible for.  Essentially, the second process monitors the Outbox table and does the actual sending of the messages and then updates the Outbox to indicate that it was sent so that it is not picked up and sent again.  It is through these two processes that it is possible to simulate guaranteed messaging without the use of distributed transactions, with a few caveats.  In the case of the second process, it is technically possible that the same message can be physically sent multiple times.  For example, after a message is sent, an exception may be thrown while trying to update the Outbox to indicate that it was sent.  In this scenario, the processing of the Outbox record would be retried and the message would be sent again.  Fortunately, this anomaly is handled by endpoint message handers, thanks to the Deduplication feature which ensures that messages with the same MessageId will only be processed one time, thus making the handlers idempotent. 
```

JeffreyAllenMiller (commented):
Your tweak makes sense to me.

nservicebus/outbox/index.md
```diff
@@ -93,6 +93,10 @@ More detail on each stage of the process:
...
...h as sending emails.  At this juncture, the question may be: “How do the outbound messages actually get sent if they are stuck serialized in the Outbox table”.  The answer to that question requires understanding what the second process is responsible for.  Essentially, the second process monitors the Outbox table and does the actual sending of the messages and then updates the Outbox to indicate that it was sent so that it is not picked up and sent again.  It is through these two processes that it is possible to simulate guaranteed messaging without the use of distributed transactions, with a few caveats.  In the case of the second process, it is technically possible that the same message can be physically sent multiple times.  For example, after a message is sent, an exception may be thrown while trying to update the Outbox to indicate that it was sent.  In this scenario, the processing of the Outbox record would be retried and the message would be sent again.  Fortunately, this anomaly is handled by endpoint message handers, thanks to the Deduplication feature which ensures that messages with the same MessageId will only be processed one time, thus making the handlers idempotent.
```
tmasternak, line 96:
```suggestion
To understanding the Outbox Pattern better it's worth noting that it doesn't rely on a single parent transaction that spans all the operations but rather on two separate processing phases.  

In the first phase (steps 2 to 6), handler logic, and all outgoing messages are captured in a single business store transaction. Messages are not immediately sent but serialized and persisted in the Outbox store together with the business data. This ensures that the business data changes and all outgoing messages are either successfully saved or rolled back.  

NOTE: Atomicity guarantee does not apply to actions within handlers that do not enlist in [an Outbox transaction](https://docs.particular.net/nservicebus/handlers/accessing-data#synchronized-storage-session), such as e.g. sending emails, changing file system, etc.   

Outgoing messages get sent as part of the second phase (steps 7 to 9) in which the messages are handed to the messaging infrastructure. When done the Outbox store is updated to indicate that the send operation was successful. It's worth noting, that due to possible failures any message can be sent multiple times. For example, if an exception is thrown in step 9 (failure when updating Outbox store) the processing of the Outbox record will be retried and the message will be re-sent. As long as the downstream endpoints use the Outbox, such duplicates will be handled by the deduplication step (3).
,,

tmasternak (commented):
I've tweaked the proposed changes to make them align a bit better with the vocabulary that we use in other parts of our documentation. I hope this doesn't change the original intent of the PR too much.

JeffreyAllenMiller (commented):
The primary emphasis was to ensure users understand that it is a two phased approach, and your edits are in line with that and are more direct.  So I am happy with this.

tmasternak:
@JeffreyAllenMiller I've added two minor editor tweaks.

I also removed the link to https://microservices.io/patterns/data/transactional-outbox.html. The main reason is that NSB Outbox does a bit more and in a bit different way:
 * We don't have a separate `Message Relay`. Sending out messages is done as part of the input message processing.
 * NSB Outbox version has a deduplication logic that makes sure we can handle duplicates produced e.g. by the `Message Relay`

I hope this makes sense. Please let me know what you think :).

tmasternak:
@JeffreyAllenMiller I'll merge the PR as it. If any other changes come up we can handle them in a separate PR.

tmasternak:
@JeffreyAllenMiller Thank you very much for this contribution!

PR #5375, opened 2021-06-23
Mention virtual machine hosting explicitly as the only cloud hosting option for SC

WilliamBZA:


WilliamBZA:
It could also go to the [planning](https://docs.particular.net/servicecontrol/servicecontrol-in-practice) page

PR #5376, opened 2021-06-23
Add note about generic message types

timbussmann:
Adds public documentation that generic message types are currently not supported.

### DIFF
--- a/nservicebus/messaging/messages-events-commands.md
+++ b/nservicebus/messaging/messages-events-commands.md
@@ -50,6 +50,8 @@ Messages should:
 
 Note: Prior to version 7.2, messages had to be defined as a `class`. Defing them as a `struct` would result in a runtime exception.
 
+Generic message definitions (e.g. `MyMessage<T>`) are not supported. It is recommended to use dedicated, simple types for each message or to use inheritance to reuse shared message characteristics.
+
 ## Identifying messages
 
 Endpoints will process any message that can be deserialized into a .NET type but requires message contracts to be identified up front to support:


WilliamBZA (approved):
Is there a 1 sentence we could add about why they're not a good idea?

timbussmann:
> Is there a 1 sentence we could add about why they're not a good idea?

you mean from a design perspective? I'd be open for suggestions but this seems to be something that we don't even seem to have a very strong opinionated stance on rather than just "never had enough value over alternatives to drive priority to support it"?

tmasternak (approved):

DavidBoike (approved):

timbussmann:
merging this as a starting point. There might be potential for further documentation based on @WilliamBZA question.

nservicebus/messaging/messages-events-commands.md
```diff
@@ -50,6 +50,8 @@ Messages should:
 
 Note: Prior to version 7.2, messages had to be defined as a `class`. Defing them as a `struct` would result in a runtime exception.
 
+Generic message definitions (e.g. `MyMessage<T>`) are not supported. It is recommended to use dedicated, simple types for each message or to use inheritance to reuse shared message characteristics.
```
sungam3r, line 53: Why so?

timbussmann: @sungam3r we never implemented this as a fully supported feature and therefore don't have any automated tests to guarantee this to be working across all transports and other platform components, that's why do not recommend using generic message types.

If this is a feature you're strongly interested in, we'd be happy to learn more about your use case to review our current stance on this topic.

sungam3r: That is, you want to say that they are not recommended to use, because you just have no enough tests?

timbussmann: > That is, you want to say that they are not recommended to use, because you just have no enough tests?

because we never considered it a supported feature and therefore also don't have a reliable set of tests to confirm whether it would be working or not. There is some indication that it "works by accident" in certain environments but that's not an intentional feature. Therefore we also do not guarantee that this continues to work across different versions of NServiceBus for example.

WilliamBZA: @sungam3r Can you perhaps explain what your use case is for wanting a generic message handler? We'd love to explore what you're trying to achieve.

PR #5383, opened 2021-06-30
Using transactionscope instead of Create/Begin/Commit/Dispose TX methods to resolve strange TX issue

ramonsmits:
When testing with large batches/load testing with different transaction modes the implementation failed. One of the failures is the following and seems to indicate an issue with leaking connections or transactions. This PR relies fully on transactionscope and uses `Required` when the transport is configured to use `TransactionScope` and otherwise uses `RequiresNew` (isolated inner transaction to either store the timeout in SQL or dispatch the timeout to MSMQ).

```txt
[22988] OnError 1b5fb9f7-ba9c-4483-9870-ad56008c3704 System.Exception: Error while storing delayed message ---> System.InvalidOperationException: The connection does not support MultipleActiveResultSets. 
[22988]    at System.Data.SqlClient.SqlInternalConnectionTds.ValidateConnectionForExecute(SqlCommand command) 
[22988]    at System.Data.SqlClient.SqlInternalTransaction.Commit() 
[22988]    at System.Data.SqlClient.SqlTransaction.Commit() 
[22988]    at SqlTimeoutStorage.CommitTransaction(TransportTransaction transaction) in S:\NServiceBus.Transport.Msmq\src\NServiceBus.Transport.Msmq\Timeouts\Sql\SqlTimeoutStorage.cs:line 221 
```

### DIFF
--- a/persistence/ravendb/connection.md
+++ b/persistence/ravendb/connection.md
@@ -7,14 +7,14 @@ redirects:
  - nservicebus/using-ravendb-in-nservicebus-connecting
  - nservicebus/ravendb/connecting
  - nservicebus/ravendb/connection
-reviewed: 2019-06-10
+reviewed: 2021-06-14
 ---
 
 include: dtc-warning
 
 include: cluster-configuration-warning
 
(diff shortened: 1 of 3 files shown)


ramonsmits:
*sigh* pushed the wrong repo :-), closing

PR #5386, opened 2021-07-06
Retrying a single message will invoke all matching handlers again

ramonsmits:
Based on a support case where a customer expected that already successfully invoked handler would not be invoked again.

### DIFF
--- a/nservicebus/handlers/index.md
+++ b/nservicebus/handlers/index.md
@@ -43,6 +43,22 @@ Receiving a message for which there are no message handlers is considered an err
 
 partial: behaviorcaveat
 
+## Multiple handlers
+
+When an endpoint hosts multiple handlers for a single incoming message and one of the handlers fails then the incoming message gets retried. When this happens all matching handlers get invoked again. Also, the handlers that might have been succesfully invoked already.
+
+The incoming message represents a unit of work. If the handlers s
+
+If you have multiple handlers for the same message hosted in the same endpoint then both get invoked as part of the same unit of work which is defined by single incoming message.
+
+This can be resolved by:
+
+- Hosting each handler in its own endpoint (hosting in isolation).
+- When receiving the message do a `SendLocal` so that each task gets represented by its own message and is executed in isolation. (divide and conquer)
+- Send 2 individual commands from endpoint that publishes the event, instead of a single event
+
+Splitting in multiple message also has the benefit that each task can run in parallel where otherwise execution is sequential.
+
 ## Unit testing
 
 Unit testing handlers is supported by [the `NServiceBus.Testing` library](/nservicebus/testing/#testing-a-handler).
--- a/nservicebus/handlers/index.md
+++ b/nservicebus/handlers/index.md
@@ -45,19 +45,18 @@ partial: behaviorcaveat
 
 ## Invocation of multiple matching handlers
 
-When an endpoint hosts multiple matching handlers for a single incoming message and one of the handlers fails then the incoming message gets retried. When the incoming message is retried all matching handlers get invoked again. Invoked handlers include the handlers that might had already been succesfully invoked in previous attempts.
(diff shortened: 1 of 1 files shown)


nservicebus/handlers/index.md
```diff
@@ -43,6 +43,22 @@ Receiving a message for which there are no message handlers is considered an err
 
 partial: behaviorcaveat
 
+## Multiple handlers
```
ramonsmits, line 46:
```suggestion
## Invocation of multiple matching handlers
```

nservicebus/handlers/index.md
```diff
@@ -43,6 +43,22 @@ Receiving a message for which there are no message handlers is considered an err
...
+
+When an endpoint hosts multiple matching handlers for a single incoming message and one of the handlers fails then the incoming message gets retried. When the incoming message is retried all matching handlers get invoked again. Invoked handlers include the handlers that might had already been succesfully invoked in previous attempts.
+
+In general, an incoming message represents a single unit of work. If multiple handlers get invoked it is best that either all succeed or all fail in a single transactional boundary. It is not recommended to run multiple atomic or isolated transactional operations for a single incoming message.
+
+Ensure correct invocation of isolated or shared state by applying one of more of the strategies:
+
+- Ensuring all handlers share the same transactional context to ensure any the shared transaction can rollback any operations (connection sharing, MSDTC, etc.).
```
tmasternak, line 54:
```suggestion
- All handlers share the same transactional context to ensure that all participating operations rollback if any of them fails.
```

nservicebus/handlers/index.md
```diff
@@ -43,6 +43,22 @@ Receiving a message for which there are no message handlers is considered an err
...
+- Ensuring handlers are idempotent and correctly deal with At-least-once Delivery.
```
tmasternak, line 55:
```suggestion
- Making handlers idempotent and correctly dealing with at-least-once delivery.
```

nservicebus/handlers/index.md
```diff
@@ -43,6 +43,22 @@ Receiving a message for which there are no message handlers is considered an err
...
+- Hosting each handler in its own endpoint (hosting in isolation) so each subscribers receive its own own copy to process in isolation (multiple subscibers).
```
tmasternak, line 56:
```suggestion
- Hosting isolation - host each handler in its own endpoint so each subscriber receives its own copy of the message and processes it in isolation.
```

nservicebus/handlers/index.md
```diff
@@ -43,6 +43,22 @@ Receiving a message for which there are no message handlers is considered an err
...
+- When receiving the message do a `SendLocal` with a new message but same payload so that each task gets represented by its own message and can be executed in isolation. (single subscriber, split at the target).
```
tmasternak, line 57:
```suggestion
- Split the message at the target - do `SendLocal` with a new message but the same payload so that each task gets represented by its own message and gets executed in isolation. 
```

nservicebus/handlers/index.md
```diff
@@ -43,6 +43,22 @@ Receiving a message for which there are no message handlers is considered an err
...
+- Send individual commands from the endpoint that publishes the event, instead of a single event (split at the source).
```
tmasternak, line 58:
```suggestion
- Split the message at the source - send individual commands from the endpoint that publishes the event, instead of a single event.
```

nservicebus/handlers/index.md
```diff
@@ -43,6 +43,22 @@ Receiving a message for which there are no message handlers is considered an err
...
+
+In general a generic message gets converted into several commands where each command represents a specific isolated task. Splitting in multiple message also has the benefit that each task can run in parallel where otherwise execution is sequential.
```
tmasternak, line 60:
```suggestion
In general, a generic message gets converted into several commands where each command represents a single isolated task executed by a single handler. This approach has the additional benefit that each task can run in parallel where otherwise execution is sequential. 
```

PR #5394, opened 2021-07-14
Functions release-1.2

SeanFeldman:


Snippets/ASBFunctions/ASBFunctions_1.2/FunctionsHostBuilderUsage.cs
```diff
@@ -0,0 +1,118 @@
+using NServiceBus;
+
+// needs to be top level first defined for C# to compile
+#region endpoint-trigger-function-wireup
+
+[assembly: NServiceBusEndpointName("MyFunctionsEndpoint")]
```
SeanFeldman, line 6: The attribute can be used only once per assembly and I chose to keep this usage rather than using the overload with a customized function trigger name as well.

bording: Yeah, I can't really think of a way to have two attributes in the project for snippets purposes.

bording (approved):

PR #5395, opened 2021-07-15
Docs for NH isolation level configuration

HEskandari:

PR #5396, opened 2021-07-15
Clarification why outbox does not support snapshot isolation

ramonsmits:
I noticed a new section https://github.com/Particular/docs.particular.net/pull/5395 and I think it could use a little argumentation.

### DIFF
--- a/persistence/nhibernate/outbox_transactionisolation_nhibernate_[8.5,).partial.md
+++ b/persistence/nhibernate/outbox_transactionisolation_nhibernate_[8.5,).partial.md
@@ -4,4 +4,4 @@ The transaction isolation level for the outbox operation can be specified:
 
 snippet: OutboxTransactionIsolation
 
-Note: The default isolation level is `Serializable`. The isolation level values of `Chaos`, `ReadUncommitted`, `Snapshot` (outbox relies on pessimistic locking to prevent more-than-once invocation) and `Unspecified` are not allowed.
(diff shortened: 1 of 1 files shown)


HEskandari (approved):

HEskandari:
@ramonsmits Thanks!

persistence/nhibernate/outbox_transactionisolation_nhibernate_[8.5,).partial.md
```diff
@@ -4,4 +4,4 @@ The transaction isolation level for the outbox operation can be specified:
 
 snippet: OutboxTransactionIsolation
 
-Note: The default isolation level is `Serializable`. The isolation level values of `Chaos`, `ReadUncommitted`, `Snapshot` and `Unspecified` are not allowed.
\ No newline at end of file
+Note: The default isolation level is `Serializable`. The isolation level values of `Chaos`, `ReadUncommitted`, `Snapshot` (outbox relies on pessimistic locking to prevent more-than-once invocation) and `Unspecified` are not allowed.
```
HEskandari, line 7: @ramonsmits does it make sense to state it like so? 

```suggestion
Note: The default isolation level is `Serializable`. The isolation level values of `Chaos`, `ReadUncommitted`, `Snapshot` and `Unspecified` are not allowed. Outbox relies on pessimistic locking to prevent more-than-once invocation.
```

ramonsmits: How about adding the word concurrent too:
```suggestion
Note: The default isolation level is `Serializable`. The isolation level values of `Chaos`, `ReadUncommitted`, `Snapshot` and `Unspecified` are not allowed. Outbox relies on pessimistic locking to prevent concurrent more-than-once invocation.
```

PR #5398, opened 2021-07-16
dotnet-support-policy-issue-1475

cquirosj:
Adding text to https://docs.particular.net/nservicebus/upgrades/supported-platforms clarifying the meaning of "supported". Work done as part of issue https://github.com/Particular/DeveloperExperience/issues/1475

### DIFF
--- a/nservicebus/upgrades/supported-platforms.md
+++ b/nservicebus/upgrades/supported-platforms.md
@@ -14,8 +14,9 @@ related:
 | .NET Core | 2.1 (LTS) | [Windows / Linux](https://github.com/dotnet/core/blob/master/release-notes/2.1/2.1-supported-os.md) | Supported | macOS is supported only for development purposes. |
 | .NET Core | 3.1 (LTS) | [Windows / Linux](https://github.com/dotnet/core/blob/master/release-notes/3.1/3.1-supported-os.md) | Supported | macOS is supported only for development purposes. |
 | .NET | 5.0 | [Windows / Linux](https://github.com/dotnet/core/blob/main/release-notes/5.0/5.0-supported-os.md) | Supported | macOS is supported only for development purposes. |
+| .NET | 6.0 Preview | [Windows / Linux](https://github.com/dotnet/core/blob/main/release-notes/6.0/supported-os.md) | [Current Status](Link to discourse announcement goes here)  |  |
 
-Supported frameworks can be used for production workloads with technical support available from Particular Software. Language features in target frameworks newer than NServiceBus is compiled against are not guaranteed to work.
+Supported frameworks can be used for production workloads with technical support available from Particular Software.
 
 ### Packages not supporting .NET Core/ .NET 5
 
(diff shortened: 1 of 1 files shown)


nservicebus/upgrades/supported-platforms.md
```diff
@@ -15,6 +15,8 @@ related:
 | .NET Core | 3.1 (LTS) | [Windows / Linux](https://github.com/dotnet/core/blob/master/release-notes/3.1/3.1-supported-os.md) | Supported | macOS is supported only for development purposes. |
 | .NET | 5.0 | [Windows / Linux](https://github.com/dotnet/core/blob/main/release-notes/5.0/5.0-supported-os.md) | Supported | macOS is supported only for development purposes. |
 
+**Supported:** It means that automated tests are passing and that customer support is available. Not necessarily all new types and language features added in the new version of .NET are supported. Submit suggestions or feature requests to https://discuss.particular.net
```
WilliamBZA, line 18: I'm not entirely happy with this. Support does not mean automated tests are passing. Automated tests passing is a prerequisite for us to be able to offer support. Support means we'll fix problems that come up and help customers if they need it.

DavidBoike: What about…
```suggestion
Supported frameworks can be used for production workloads with technical support available from Particular Software. Language features in target frameworks newer than NServiceBus is compiled against are not guaranteed to work.
```

WilliamBZA: This is better, but I'm still not sure. It's not that language features newer than any date are definitely going to work or not. There are some language features that just don't make sense and will never work. E.g. Record types work for the most part, but not in some scenarios. And in those scenarios, it doesn't make sense for them to work anyway so why would we change our code to support them.

I guess I just don't think there's any reason to mention language features at all. If someone tries to do something with a language feature and it doesn't work, that may be because we have to make changes to support it, or it may just be a silly thing to try and do.

nservicebus/upgrades/supported-platforms.md
```diff
@@ -15,6 +15,8 @@ related:
 | .NET Core | 3.1 (LTS) | [Windows / Linux](https://github.com/dotnet/core/blob/master/release-notes/3.1/3.1-supported-os.md) | Supported | macOS is supported only for development purposes. |
 | .NET | 5.0 | [Windows / Linux](https://github.com/dotnet/core/blob/main/release-notes/5.0/5.0-supported-os.md) | Supported | macOS is supported only for development purposes. |
 
+Supported frameworks can be used for production workloads with technical support available from Particular Software. Language features in target frameworks newer than NServiceBus is compiled against are not guaranteed to work.
```
WilliamBZA, line 18:
```suggestion
Supported frameworks can be used for production workloads with technical support available from Particular Software.
```

nservicebus/upgrades/supported-platforms.md
```diff
@@ -15,6 +15,8 @@ related:
 | .NET Core | 3.1 (LTS) | [Windows / Linux](https://github.com/dotnet/core/blob/master/release-notes/3.1/3.1-supported-os.md) | Supported | macOS is supported only for development purposes. |
 | .NET | 5.0 | [Windows / Linux](https://github.com/dotnet/core/blob/main/release-notes/5.0/5.0-supported-os.md) | Supported | macOS is supported only for development purposes. |
```
WilliamBZA, line 16:
```suggestion
| .NET | 5.0 | [Windows / Linux](https://github.com/dotnet/core/blob/main/release-notes/5.0/5.0-supported-os.md) | Supported | macOS is supported only for development purposes. |
| .NET | 6.0 Preview | [Windows / Linux](https://devblogs.microsoft.com/dotnet/announcing-net-6-preview-6/) | [Current Status](Link to discourse announcement goes here)  |  |
```

DavidBoike: This document https://github.com/dotnet/core/blob/main/release-notes/6.0/supported-os.md already exists for net6, that's probably what we should link the Windows / Linux column to

WilliamBZA (approved):

PR #5381, opened 2021-06-28
Document Gateway SetRepyToUri()

DavidBoike:

PR #5399, opened 2021-07-16
fix sln config for gateway snippets

adamralph:
relates to https://github.com/Particular/docs.particular.net/pull/5381


DavidBoike (approved):
So, I have no idea. AFAIK you could just set everything to Any CPU as well. I have no opinion but I'll throw you a PR Approval with carte blanche to do whatever you want. :wink:

PR #5402, opened 2021-07-19
Add net6 as a build target for samples

WilliamBZA:


DavidBoike:
Only 4 projects fail to build as a result of adding .NET 6, only one important one which is bolded.

* samples\callbacks\Callbacks_3 - Is a project built by MSBuild (current, non-prerelease version) that also contains net6.0 projects, which can't be built by that version of msbuild
* samples\gateway\Gateway_3 - Same
* **samples\pipeline\stream-properties\Core_7 - error SYSLIB0014: `WebClient.WebClient()` is obsolete: WebRequest, HttpWebRequest, ServicePoint, and WebClient are obsolete. Use HttpClient instead.**
* samples\sqltransport-sqlpersistence\Core_7 - A rogue nuget.config file in there is referring to ServiceStack's MyGet which threw a 404 :shrug:

DavidBoike:
Interestingly, that pipeline stream properties samples was broken ANYWAY.

@WilliamBZA @cquirosj The commits I tacked on the end would (should?) make this PR mergeable. Users of the website would be able to download samples for ".NET 6 (Preview)". The only bummer is that all these samples would be multitargeting .NET 6 for all of us, so if we're working in Docs, we can only do that in Visual Studio if we've got a VS preview installed, or from VSCode if you've got a dotnet preview installed.

WilliamBZA:
Validated that the code works with the following samples:

* [sqltransport-sqlpersistence](https://docs.particular.net/samples/sqltransport-sqlpersistence/)
* [SQS Simple](https://docs.particular.net/samples/sqs/simple/)
* [RabbitMQ simple](https://docs.particular.net/samples/rabbitmq/simple/)
* [azure-service-bus-netstandard send-reply](https://docs.particular.net/samples/azure-service-bus-netstandard/send-reply/)
* [cosmosdb simple](https://docs.particular.net/samples/previews/cosmosdb/simple/)
* [azure-table simple](https://docs.particular.net/samples/azure/azure-table/simple/)

PR #5421, opened 2021-08-12
Functions docs improvements

timbussmann:
A bunch of changes to improve/correct the docs

* It looks like some of the snippets are matched to the wrong versions
* use correct attribute name for 1.3 snippet
* provide a snippet for how to write a custom trigger
* move snippets in message consistency section to the snippets project
* list the relevant environment variables for IConfiguration usage

TODO:
* consider removing some duplicate snippets about manual trigger definition or extracting them to a dedicated file
* consider using a different default configuration API in the snippet


HEskandari (approved):

nservicebus/hosting/azure-functions/service-bus_queue-trigger-wiring_ASBFunctions_[1.3,).partial.md
```diff
@@ -0,0 +1,23 @@
+The Azure Function trigger for NServiceBus is auto-generated by specifying the logical endpoint name using a custom assembly attribute:
+
+```csharp
+[assembly: NServiceBusTriggerFunction("MyFunctionsEndpoint")]
+```
+
+The attribute will generate the trigger function required for NServiceBus:
+
+Note: An invalid endpoint name will generate an `NSBFUNC003` error with the message `Endpoint name is invalid and cannot be used to generate trigger function`.
```
HEskandari, line 9: Do we want to elaborate on what is an invalid name?

HEskandari (approved):

nservicebus/hosting/azure-functions/service-bus_queue-trigger-wiring_ASBFunctions_[1.3,).partial.md
timbussmann: looking at the sources, this seems only to be invalid when it's null, empty or whitespace. Not sure this needs to be explicitly spelled out. Happy to further discuss but I'm not sure how to word that right now.

PR #5422, opened 2021-08-12
Attempt to use the EOL concept for renamed packages

DavidBoike:
@adamralph @bording I really think this is wrong, but…

I attempted to deal with the fact that the ASQ and ASP packages got renamed using the endOfLifePackages collection that was currently in the script. This is the result.  I don't think it's right, and what's really boggling my mind is that it's not a PackageA => PackageB, it's PackageA => Reason, and based on that, I can't figure out where this 2021-07-09 expiration date for NServiceBus.Azure.Transports.WindowsAzureStorageQueues 8.2 is coming from.

Do you agree?


DavidBoike:
Brandon helped me out. The July 9 date comes from the last patch of that package. The assumption is the last version is a tombstone package.

This concept will definitely not work here. We need a "renamedPackages" concept.

PR #5423, opened 2021-08-13
Extract custom trigger definition docs page

timbussmann:
Some minor word teaks + extracts the custom trigger documentation into a dedicated page to simplify the main page


HEskandari (approved):

mikeminutillo:
The config partial still includes two snippets with this section in between them:

> The default builder method will look up values such as function name, connection string and license information directly from from the IConfiguration. Fine-grained configuration is also possible:

Does this still make sense? The first snippet will not read anything out of `IConfiguration` because we aren't passing an instance in, and it's not really functionally any different from the 2nd snippet. Should we remove the 2nd snippet and the highlighted text? There's a section below about config that mentions `IConfiguration`

timbussmann:
@mikeminutillo I tried to improve the section you mentioned, can you have another look?

mikeminutillo (approved):

PR #5426, opened 2021-08-17
Assembly scanner docs for azure functions

HEskandari:


nservicebus/hosting/azure-functions/service-bus_assembly-scanner_ASBFunctions_[2.0,).partial.md
```diff
@@ -0,0 +1,3 @@
+### Assembly scanning
+
+The assembly scanner loads assemblies in the `bin` directory of the application, as well as the assemblies that are available at the function's runtime directory. The behavior can be customized by leveragin the [assembly scanner](/nservicebus/hosting/assembly-scanning.md) API.
```
mikeminutillo, line 3:
```suggestion
The assembly scanner loads assemblies in the `bin` directory of the application, as well as the assemblies that are available at the function's runtime directory. The behavior can be customized by leveraging the [assembly scanner](/nservicebus/hosting/assembly-scanning.md) API. 
```

mikeminutillo: > as well as the assemblies that are available at the function's runtime directory

Are these a part of the Azure Functions runtime? It might be worth clarifying what the difference is here. It may even be worth calling out why you might want to control the assembly scanner. Something like:

> The assembly scanner loads assemblies from two different locations:
> - The `bin` directory of the applictation containing function definitions and any dependencies.
> - The Azure Function runtime directory.
>
> If the assembly scanner attempts to load the same assembly  twice (usually because it is present in both locations) then an exception will be thrown.
> (ADD EXAMPLE OF EXCEPTION)
> The [assembly scanner](/nservicebus/hosting/assembly-scanning.md) API can be used to control which assemblies get scanned.

menu/menu.yaml
```diff
@@ -386,6 +386,8 @@
       Articles:
       - Url: nservicebus/hosting/azure-functions/service-bus
         Title: Service Bus
+      - Url: nservicebus/hosting/azure-functions/custom-triggers
```
timbussmann, line 389: I didn't add it to the menu intentionally. If we want to add it, it should be a sub-item of "Service Bus", not on the same level.

timbussmann (commented):
is there any interesting difference in behavior to use user compared to the regular assembly scanner behavior (on "normal" NSB endpoints) that warrants adding dedicated documentation?

menu/menu.yaml
HEskandari: Problem is, that level of nestedness is not supported. How about naming it "Service Bus custom triggers"?

HEskandari:
@timbussmann the interesting behavior (or potentially problem resolution when the user has assembly loading issue) is what's described by Mike's comment. Does that make sense to change the documentation as suggested there?

nservicebus/hosting/azure-functions/service-bus_assembly-scanner_ASBFunctions_[2.0,).partial.md
HEskandari: @mikeminutillo  I reworded this section, see if it reads better now? feel free to suggest further changes.

nservicebus/hosting/azure-functions-service-bus/index_assembly-scanner_ASBFunctions_[2.0,).partial.md
```diff
@@ -0,0 +1,8 @@
+### Assembly scanning
+
+The assembly scanner loads assemblies from two locations:
+
+- The `bin` directory of the function application
+- The Azure Function runtime directory 
+
+Loading an assembly the second time (for example from a different directory), would cause runtime exceptions. Contact [Particular support](https://particular.net/support) for assistance.
```
timbussmann, line 8:
```suggestion
Loading the same assembly from two different locations will cause runtime exceptions. Contact [Particular support](https://particular.net/support) for assistance. 
```

nservicebus/hosting/azure-functions-service-bus/index.md
```diff
@@ -65,6 +65,8 @@ For recoverability to move the continuously failing messages to the error queue
 
 snippet: asb-configure-error-queue
 
+partial: assembly-scanner
```
timbussmann, line 68: I'd vote to put this at the complete end of the documentation.

HEskandari: sounds good. I'll move it.

PR #5449, opened 2021-08-31
Rephrase immutable messages

lailabougria:
The current wording did not make sense to me, reworded for clarity

### DIFF
--- a/nservicebus/messaging/immutable-messages.md
+++ b/nservicebus/messaging/immutable-messages.md
@@ -34,7 +34,7 @@ public class CancelOrder : ICommand
 
 ## Classes with public setters, interfaces with only getters
 
-Using private setters is not supported by all serializers. An alternative is to make use of NServiceBus' support for [multiple inheritance and polymorphic dispatch](/nservicebus/messaging/messages-as-interfaces.md). In this case, a message contract is defined through an interface, which is used by the message handler. An implementation of the message contract, which exposes public setters/getters is used to create the message and pass it to `Send` or `Publish`.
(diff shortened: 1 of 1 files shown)


nservicebus/messaging/immutable-messages.md
```diff
@@ -34,7 +34,7 @@ public class CancelOrder : ICommand
 
 ## Classes with public setters, interfaces with only getters
 
-Using private setters is not supported by all serializers. An alternative is to have a class that contains public setters/getters so that any serializer can deserialize and make use of NServiceBus its support for [multiple inheritance and polymorphic dispatch](/nservicebus/messaging/messages-as-interfaces.md). Interfaces are used only by message handlers and classes only to create the message and pass it to `Send` or `Publish`.
+Using private setters is not supported by all serializers. An alternative is to make use of NServiceBus' support for [multiple inheritance and polymorphic dispatch](/nservicebus/messaging/messages-as-interfaces.md). In this case, a message contract is defined through an interface, which is used by the message handler. An implementation of the message contract, which exposes public setters/getters is used to create the message and pass it to `Send` or `Publish`.
```
WojcikMike, line 37: I think we should state that interface has only getter and the type implementing getter and setter

lailabougria:
```suggestion
Using private setters is not supported by all serializers. An alternative is to make use of NServiceBus' support for [multiple inheritance and polymorphic dispatch](/nservicebus/messaging/messages-as-interfaces.md). In this case, a message contract is defined through an interface containing getters only, which is used by the message handler. An implementation of the message contract, which exposes public setters/getters is used to create the message and pass it to `Send` or `Publish`.
```

WojcikMike (approved):

PR #5459, opened 2021-09-02
Remove file based routing and redirects to new home

andreasohlund:
All deleted assets will be transferred via https://github.com/ParticularLabs/NServiceBus.FileBasedRouting/pull/50 to its new home


SzymonPobiega (approved):

SzymonPobiega:
Looks good but for some reason the docs engine is complaining about the redirect :(

PR #5467, opened 2021-09-07
Update samples to v8

WilliamBZA:
Adds samples for V8


samples/messagemutators/Core_8/Sample/TransportMessageCompressionMutator.cs
```diff
@@ -0,0 +1,52 @@
...
+        }
+        // copy the compressed buffer only after the GZipStream is disposed,
+        // otherwise, not all the compressed message will be copied.
+        context.OutgoingBody = outStream.ToArray();
+        context.OutgoingHeaders["IWasCompressed"] = "true";
+        log.Info($"transportMessage.Body size after compression: {context.OutgoingBody.Length}");
+        return Task.CompletedTask;
+    }
+
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("NServiceBus.Code", "NSB0002:Forward the 'CancellationToken' property of the context parameter to methods", Justification = "<Pending>")]
```
WilliamBZA, line 34: This is here to prevent needing an `#ifdef` for line 45. .NET 4.8 does not have the cancellation token overload while the rest of the runtimes do

SzymonPobiega (approved):

PR #5471, opened 2021-09-08
Isolate snippets, ahead of enforcement by DocsEngine

DavidBoike:
Easiest to review commit-by-commit.

I was not going to go out of my way to fix every snippet for legacy Cloud Service Host paired with legacy versions of ASB/ASQ, resulting in the removal of a bunch of snippets there.


bording (approved):

bording:
Based on the changes here, it looks like to use a snippet on a page, you are now required to specify a component, right? Seems like that change is worth an announcement.

DavidBoike:
Yep, that will be required after https://github.com/Particular/DocsEngine/pull/665

DavidBoike:
And btw, https://github.com/Particular/DocsEngine/pull/665 includes a specific error message for exactly that scenario rather than a misleading "MissingSnippets" error.

PR #5487, opened 2021-09-14
Snippet for SNS client factory

danielmarbach:
I discovered the missing configuration while reading the discussion group

https://discuss.particular.net/t/sqs-sns-msmq-router-not-working-behind-a-proxy/2581

I did not have much time to properly finish it but I figured even a non-compiling PR is already a start


andreasohlund (approved):
Awesome, I was just dealing with a support case from the customer and came to the same conclusion

mauroservienti:
Thanks @danielmarbach

PR #5500, opened 2021-09-28
Update ASQ conn string doco

SzymonPobiega:


SzymonPobiega:
@mikeminutillo can you take a look?

Snippets/ASQ/ASQN_10/Usage.cs
```diff
@@ -15,7 +15,19 @@ void UseTransport(EndpointConfiguration endpointConfiguration)
     {
         #region AzureStorageQueueTransportWithAzure
 
+        var connectionString = "DefaultEndpointsProtocol=https;AccountName=[ACCOUNT];AccountKey=[KEY];";
+
         var transport = endpointConfiguration.UseTransport<AzureStorageQueueTransport>();
+        transport.UseQueueServiceClient(new QueueServiceClient(connectionString));
+
+        //Table and Blob clients don't need to be specified if native delayed delivery is disabled
+        //transport.DelayedDelivery().DisableDelayedDelivery();
```
mikeminutillo, line 24: I think we still need table for native pub-sub (which there is not an API to disable)

mikeminutillo (approved):

PR #5545, opened 2021-11-09
Azure Functions v4 and .net6 disposed container

jenspettersson:
When running the Azure Functions sample on v4 and .net6 exceptions are thrown.

```
System.Private.CoreLib: Exception while executing function: NServiceBusFunctionEndpointTrigger-ASBTriggerQueue. Microsoft.Azure.WebJobs.Script.WebHost: Scope disposed{no name} is disposed and scoped instances are disposed and no longer available.

Microsoft.Extensions.DependencyInjection: Cannot access a disposed object.
Object name: 'IServiceProvider'.
```

Just hit the http-trigger endpoint to reproduce.

### DIFF
--- a/samples/azure-functions/service-bus/ASBFunctions_2/AzureFunctions.ASBTrigger.FunctionsHostBuilder/AzureFunctions.ASBTrigger.FunctionsHostBuilder.csproj
+++ b/samples/azure-functions/service-bus/ASBFunctions_2/AzureFunctions.ASBTrigger.FunctionsHostBuilder/AzureFunctions.ASBTrigger.FunctionsHostBuilder.csproj
@@ -1,12 +1,12 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
-    <AzureFunctionsVersion>v3</AzureFunctionsVersion>
+    <TargetFramework>net6.0</TargetFramework>
+    <AzureFunctionsVersion>v4</AzureFunctionsVersion>
     <LangVersion>7.3</LangVersion>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="NServiceBus.AzureFunctions.InProcess.ServiceBus" Version="2.*" />
-    <PackageReference Include="Microsoft.NET.Sdk.Functions" Version="3.*" />
+    <PackageReference Include="Microsoft.NET.Sdk.Functions" Version="4.*" />
   </ItemGroup>
   <ItemGroup>
     <ProjectReference Include="..\AzureFunctions.Messages\AzureFunctions.Messages.csproj" />
--- a/samples/azure-functions/service-bus-worker/ASBFunctionsWorker_2/AzureFunctions.ASBTrigger.Worker/AzureFunctions.ASBTrigger.Worker.csproj
+++ b/samples/azure-functions/service-bus-worker/ASBFunctionsWorker_2/AzureFunctions.ASBTrigger.Worker/AzureFunctions.ASBTrigger.Worker.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>net5.0</TargetFramework>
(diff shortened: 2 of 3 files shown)


mauroservienti:
@jenspettersson thanks for raising this. We can reproduce the problem. It works if you inject the `IServiceProvider` instance into the message handler and resolve the dependency from there, or if you put a breakpoint in any hot dependency resolution path.

There seems to be some connection with the following issue: https://github.com/Azure/azure-functions-host/issues/5240

/cc @SeanFeldman

SeanFeldman:
The way [NServiceBus is accessing the `UserProperties`](https://github.com/Particular/NServiceBus.AzureFunctions.Worker.ServiceBus/blob/73ae9f242b73a241f027eeaa6b9f185e8319d4c4/src/NServiceBus.AzureFunctions.Worker.ServiceBus/FunctionEndpoint.cs#L47) is probably not how it should be done. It shouldn't be injected, rather accessed via `FunctionContext.BindingContext.BindingData` (see [example](https://github.com/Azure/azure-functions-dotnet-worker/issues/384#issuecomment-816746674)). That way, it can be consistently retrieved from the function method. Alternatively, `UserProperties` should be injected as a `string` and then deserialized.

SeanFeldman:
Here's what I'm talking about: https://github.com/Particular/NServiceBus.AzureFunctions.Worker.ServiceBus/pull/66

samples/azure-functions/service-bus/ASBFunctions_2/AzureFunctions.ASBTrigger.FunctionsHostBuilder/AzureFunctions.ASBTrigger.FunctionsHostBuilder.csproj
```diff
@@ -1,12 +1,12 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
-    <AzureFunctionsVersion>v3</AzureFunctionsVersion>
+    <TargetFramework>net6.0</TargetFramework>
+    <AzureFunctionsVersion>v4</AzureFunctionsVersion>
```
andreasohlund, line 4: @jenspettersson until we fix this I have verified that you can work around this by implementing your own trigger class like this: (you need to comment out `[assembly:NServiceBusTriggerFunction("ASBWorkerEndpoint")]` to avoid the trigger to be autogenerated)

```
public class FunctionEndpointTrigger
{
    IFunctionEndpoint endpoint;

    public FunctionEndpointTrigger(IFunctionEndpoint endpoint)
    {
        this.endpoint = endpoint;
    }


    [Function("NServiceBusFunctionEndpointTrigger-ASBWorkerEndpoint")]
    public async Task Run(
        [ServiceBusTrigger("ASBWorkerEndpoint")] byte[] messageBody,
        IDictionary<string, object> userProperties,
        string messageId,
        int deliveryCount,
        string replyTo,
        string correlationId,
        FunctionContext context)
    {
        await endpoint.Process(messageBody, userProperties.ToDictionary(k => k.Key, k => k.Value.ToString()), messageId, deliveryCount, replyTo, correlationId, context);
    }
}

```

andreasohlund: Note that the change is to use `IDictionary<string, object> userProperties,` instead of `IDictionary<string, string> userProperties,`

jenspettersson: Thanks, I'll take a look when I've got time. Currently occupied elsewhere. Btw, I guess this is the workaround for the second issue, any news about the first with the container being disposed when running on the v4 host InProcess?

andreasohlund:
We've found the issue, not sure how to resolve it yet.

We have raised https://github.com/Particular/NServiceBus.AzureFunctions.InProcess.ServiceBus/issues/359 to track this

PR #5546, opened 2021-11-09
Saga audit retention check

kentdr:
Added documentation for the new Saga audit data retention custom check.

### DIFF
--- a/servicecontrol/upgrades/3to4/index.md
+++ b/servicecontrol/upgrades/3to4/index.md
@@ -23,16 +23,16 @@ For more information how to upgrade from Version 1.x to 3.8.2 consult the [upgra
 
 ### Editing older instances
 
-ServiceControl Management version 4 cannot be used to edit ServiceControl instances until they have been upgraded to version 4. These instances can still be started, stopped, put into maintenance mode, and removed using ServiceControl Managament. Ensure any planned changes have been made to existing ServiceControl instances before installing ServiceControl version 4. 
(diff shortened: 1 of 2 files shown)


servicecontrol/upgrades/3to4/index.md
```diff
@@ -23,16 +23,16 @@ For more information how to upgrade from Version 1.x to 3.8.2 consult the [upgra
...
 
 {{NOTE:
-After the upgrade is complete, the Audit information contained in the original ServiceControl instance will become read-only, but will continue to be deleted as it passes the time specified by the [retention policy](/servicecontrol/creating-config-file.md#data-retention). However, since the internal database does not release storage space back to the OS, the database will continue to be the same size.
+After the upgrade is complete, the Audit information contained in the original ServiceControl instance will become read-only, but will continue to be deleted as it passes the time specified by the [retention policy](/servicecontrol/creating-config-file.md#data-retention). However, since the internal database does not release storage space back to the OS, the database will continue to be the same size. If saga data was being audited in the original instance and no retention policy is configured a custom check will fail, refer to the [troubleshooting guide](/servicecontrol/troubleshooting.md#saga-audit-data-retention-custom-check-failure) to resolve the issue.
```
mikeminutillo, line 33: I'm not sure if this is needed. If the original instance had an audit retention set (which is should have) then that is not removed by the upgrade process. I think this can only happen on a new v4 instance, which never had an audit retention policy set.

mikeminutillo (approved):

PR #5547, opened 2021-11-10
Update Azure Service Bus transport configuration page

mderriey-prospa:
- `CustomTokenProvider(ITokenProvider)` is now `CustomTokenCredential(TokenCredential)`.
- Add note about `Azure.Messaging.ServiceBus` expecting a fully-qualified namespace instead of a connection string when using AAD authentication.


mderriey-prospa:
I'm not familiar with these docs, so please let me know how we can make sure these changes don't apply to the page for v1.9 of the transport: <https://docs.particular.net/transports/azure-service-bus/configuration?version=asbs_1.9>.

Clicking on "Edit" on that page seems to redirect to the same file in GitHub.

transports/azure-service-bus/configuration.md
```diff
@@ -34,5 +34,5 @@ These settings control how the transport connects to the broker.
...
-partial: custom-retry-policy
\ No newline at end of file
+ * `CustomTokenCredential(TokenCredential)`: Allows usage of Azure Active Directory (AAD) authentication instead of the shared secret in the connection string. This opens up additional authentication mechanisms such as [managed identities for Azure resources](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-managed-service-identity). When leveraging AAD authentication, it is worth noting that the underlying library expects a fully-qualified namespace (like `<asb-namespace-name>.servicebus.windows.net`) instead of a connection string (like `Endpoint=sb://<asb-namespace-name>.servicebus.windows.net>;[...]`).
```
tmasternak, line 37:
```suggestion
 * `CustomTokenCredential(TokenCredential)`: Enables using a non-default authentication mechanism such as Azure Active Directory (AAD) or [managed identities for Azure resources](https://docs.microsoft.com/en-us/azure/service-bus-messaging/service-bus-managed-service-identity) instead of the shared secret stored in the connection string. 

Note: AAD authentication requires a fully-qualified namespace usage (e.g. `<asb-namespace-name>.servicebus.windows.net`) instead of a connection string (e.g. `Endpoint=sb://<asb-namespace-name>.servicebus.windows.net>;[...]`).
```

tmasternak:
> I'm not familiar with these docs, so please let me know how we can make sure these changes don't apply to the page for v1.9 of the transport: https://docs.particular.net/transports/azure-service-bus/configuration?version=asbs_1.9.

Once we have the text ready I'll make sure this goes to docs for version 2.0 only.

transports/azure-service-bus/configuration.md
mderriey-prospa: Looks good overall, thanks.

> Enables using a non-default authentication mechanism such as Azure Active Directory (AAD) or managed identities for Azure resources

A nitpick here is that using managed identities is using AAD authentication, so the "or" reads a bit strange to me.
In other words, we're using AAD authentication when we're using managed identities; however, they're not the only way to use AAD authentication, we could use a "regular" service principal (client id / client secret), or the developer's identity when running a solution locally, for example.

I would reword your suggestion as "Enables using Azure Active Directory (AAD) authentication such as managed identities for Azure resources instead of the shared secret in the connection string."

How do you think that reads?

tmasternak: @mderriey-prospa sounds good. I've added your tweaks to the suggestion. Please let me know if that's good and I'll tweak the changes to be version specific.

mderriey-prospa: Looks hood @tmasternak, thanks for that 😉

tmasternak (approved):

tmasternak:
@mderriey-prospa the changes are live. Thank you very much for your contribution!

PR #5550, opened 2021-11-12
Update index.md

HEskandari:


mikeminutillo (approved):
We could mention the setting name here but it's probably pretty obvious if anyone opens that config file.

Also, we might want to mention the version that this setting was introduced in so that someone on an older version doesn't go looking for it.

HEskandari:
Since only the latest version of SI is supported, I thought that'd be a moot point, or worse imply otherwise.

PR #5554, opened 2021-11-17
Added ServiceControl multi major version upgrade guidance

ramonsmits:


servicecontrol/installation.md
```diff
@@ -79,6 +79,20 @@ Clicking the upgrade link will:
 * Run the new binaries to create required queues
 * Start the service
 
+
+To upgrade multiple majors the following approach needs to be taken:
+
+- Upgrade the the latest minor.patch release within a give major
```
lailabougria, line 85:
```suggestion
- Upgrade to the latest minor.patch release within a given major
```

servicecontrol/installation.md
```diff
@@ -79,6 +79,20 @@ Clicking the upgrade link will:
...
+  - When v1.34.0 is installed, then install v1.48.0 and update all instances
+- Then upgrade the latest minor patch release of the **next** major
```
lailabougria, line 87:
```suggestion
- Then upgrade to the latest minor patch release of the **next** major 
```

servicecontrol/installation.md
```diff
@@ -79,6 +79,20 @@ Clicking the upgrade link will:
...
 * Start the service
 
+
+To upgrade multiple majors the following approach needs to be taken:
+
+- Upgrade the the latest minor.patch release within a give major
+  - When v1.34.0 is installed, then install v1.48.0 and update all instances
+- Then upgrade the latest minor patch release of the **next** major 
+  - When 1.48.0 is installed, install 2.1.5 and update all instances.
+- Continue with the last step until the system is update till the latest version
```
lailabougria, line 89:
```suggestion
- Repeat the previous step until the system is up to date with the latest version
```

servicecontrol/installation.md
```diff
@@ -79,6 +79,20 @@ Clicking the upgrade link will:
...
+  - When 2.1.5 is installed, install 3.8.4 and update all instances.
+  - Then install 4.20.3 (current lateste version of this writing) and update all instances
```
lailabougria, line 91:
```suggestion
  - Then install 4.20.3 (current latest version at the time of this writing) and update all instances 
```

servicecontrol/installation.md
```diff
@@ -79,6 +79,20 @@ Clicking the upgrade link will:
...
+- Upgrade the the latest minor.patch release within a give major
+  - When v1.34.0 is installed, then install v1.48.0 and update all instances
+- Then upgrade the latest minor patch release of the **next** major 
+  - When 1.48.0 is installed, install 2.1.5 and update all instances.
+- Continue with the last step until the system is update till the latest version
+  - When 2.1.5 is installed, install 3.8.4 and update all instances.
+  - Then install 4.20.3 (current lateste version of this writing) and update all instances 
+
+
+NOTE: Upgrages *might* take a while to
```
lailabougria, line 94: Something was left out here, what did you mean to add @ramonsmits ?

lailabougria (changes requested):

servicecontrol/installation.md
mikeminutillo: I would maybe even make this the current major

```suggestion
- Upgrade to the latest minor.patch release within a currently installed major
```

servicecontrol/installation.md
```diff
@@ -79,6 +79,20 @@ Clicking the upgrade link will:
 * Run the new binaries to create required queues
 * Start the service
 
+
+To upgrade multiple majors the following approach needs to be taken:
+
+- Upgrade the the latest minor.patch release within a give major
+  - When v1.34.0 is installed, then install v1.48.0 and update all instances
```
mikeminutillo, line 86: I would consider pulling the step to update all instances out as a separate section. Maybe above this. 

```
Each instance of ServiceControl needs to be updated separately. Whenever updating to a new version of ServiceControl:

- Download the new installer
- On each machine with ServiceControl instances:
  - Install the new version
  - Open ServiceControl Management and update each instance
```

mikeminutillo (commented):
I wonder if the number of versions are small enough that we could just got the extra step and list them out. i.e.

A ServiceControl installation should be upgraded one major version at a time. Check the currently installed version of ServiceControl to look up the upgrade path in the table below. For each entry in the upgrade path, install the listed version and upgrade all ServiceControl instances to that version. Once all instaces are upgraded and running, move to the next version in the upgrade path.

| Current ServiceControl Version | Upgrade Path                                       |
|-----------------------------------|------------------------------------------|
| 1.x                                              | 1.48.0<br>2.1.5<br>3.8.4<br>Latest |
| 2.x                                              | 2.1.5<br>3.8.4<br>Latest |
| 3.x                                              | 3.8.4<br>Latest |
| 4.x                                              | Latest |


If we do that, then we can link the version numbers to the downloads and customers can follow the documented process without needing to interpret the rules.

lailabougria:
Isn't it better to take it as a sort of example, instead of linking to the actual version downloads? It's definitely the best experience to have the links, but we'd have to keep that up to date every time we release a new version of SC.

servicecontrol/installation.md
ramonsmits, line 94:
```suggestion
NOTE: Upgrades *might* take a while to run. Account for unavailability of ServiceControl and plan upgrade off hours.
```

ramonsmits: @lailabougria updated the text, I indeed forgot to finish the sentence.

servicecontrol/installation.md
```diff
@@ -79,6 +79,20 @@ Clicking the upgrade link will:
...
+- Upgrade to the latest minor.patch release within a currently installed major
+  - When v1.34.0 is installed, then install v1.48.0 and update all instances
+- Then upgrade to the latest minor patch release of the **next** major 
+  - When 1.48.0 is installed, install 2.1.5 and update all instances.
+- Repeat the previous step until the system is up to date with the latest version
+  - When 2.1.5 is installed, install 3.8.4 and update all instances.
+  - Then install 4.20.3 (current latest version at the time of this writing) and update all instances 
+
+
+NOTE: Upgrades *might* take a while to run. Account for unavailability of ServiceControl and plan upgrade off hours.
```
lailabougria, line 94:
```suggestion
NOTE: Upgrades *might* take a while to run. Account for unavailability of ServiceControl and plan the upgrade off hours.
```

mikeminutillo:
> Isn't it better to take it as a sort of example, instead of linking to the actual version downloads? It's definitely the best experience to have the links, but we'd have to keep that up to date every time we release a new version of SC.

There's two different types of release we might want to do that could affect this page:

1. A major release. This would require adding another row and fixing the end of the v4 line. This is also a pretty rare occurrence.
2. A minor or patch release. This is very unlikely to be for any major other than v4. That means the only thing that changes is what "current" link to. That can be resolved by linking to https://github.com/Particular/ServiceControl/releases/latest

It is _very_ unlikely that we will release a new patch/minor release for v1-3 because 4.20 is the only supported version.

ramonsmits:
> Isn't it better to take it as a sort of example, instead of linking to the actual version downloads? It's definitely the best experience to have the links, but we'd have to keep that up to date every time we release a new version of SC.

@lailabougria I did that intentionally, instead of a static list a better approach would be a dynamically generated list with javascript but not sure that is worth the effort.

I like @mikeminutillo his suggestion to link to 'latest'.

servicecontrol/installation.md
```diff
@@ -79,6 +79,21 @@ Clicking the upgrade link will:
 * Run the new binaries to create required queues
 * Start the service
 
+## Upgrade multiple majors
+
+A ServiceControl installation should be upgraded one major version at a time. Check the currently installed version of ServiceControl to look up the upgrade path in the table below. For each entry in the upgrade path, install the listed version and upgrade all ServiceControl instances to that version. Once all instances are upgraded and running, install the listed version in the next row until all instancen are upgraded to the latest version.
```
mikeminutillo, line 84:
```suggestion
A ServiceControl installation should be upgraded one major version at a time. Check the currently installed version of ServiceControl to look up the upgrade path in the table below. For each entry in the upgrade path, install the listed version and upgrade all ServiceControl instances to that version. Once all instances are upgraded and running, install the listed version in the next row until all instances are upgraded to the latest version.
```

mikeminutillo (approved):

PR #5558, opened 2021-11-25
Repro for Azure Functions

danielmarbach:
Execute the sample with the v3 runtime --> Works as expected
Execute the sample with the v4 runtime --> Object disposed exception


adamralph:
@danielmarbach do you still need the branch?

PR #5561, opened 2021-11-25
Gateway RavenDB Smoke test

danielmarbach:


mauroservienti:
Fixed, it was a typo in the first url

danielmarbach:
Perfect. Then we can release. Thanks!

PR #5566, opened 2021-11-30
Update timeout storage section of upgrade guide

bording:
This updates the ASP 2 to 3 upgrade guide's timeout storage section to not reference NSB 8 at all because both versions are NSB 7 components.

Instead the reason for the removal is listed as both Azure transports natively supporting delays, and directly links to the migration tool if needed.


bording:
Replaced by #5567

PR #5583, opened 2021-12-17
Saga not found troubleshooting

ramonsmits:


nservicebus/sagas/saga-not-found.md
```diff
@@ -29,3 +29,72 @@ The ability to provide an implementation for `IHandleSagaNotFound` is especially
...
+- Use [outbox](/nservicebus/outbox/)
+ 
+### Already completed
+
+A saga that already completed can result in a saga not found.
+
+Mitigation:
+
+- Use a logical complete state
+- Use [saga timeouts](/nservicebus/sagas/timeouts) to postpone physically removing the saga state and invoke [`MarkAsComplete`](/nservicebus/sagas/#ending-a-saga).
```
ramonsmits, line 95:
```suggestion
- Use [saga timeouts](timeouts.md) to postpone physically removing the saga state and invoke [`MarkAsComplete`](/nservicebus/sagas/#ending-a-saga).
```

nservicebus/sagas/saga-not-found.md
```diff
@@ -29,3 +29,67 @@ The ability to provide an implementation for `IHandleSagaNotFound` is especially
...
 The implementation of `IHandleSagaNotFound` should be driven by the business requirements for a specific situation. In some cases the message might be ignored; in others, it might be useful to track whenever that situation happens (e.g. by logging or sending another message). In still other cases, it might make sense to perform a custom compensating action.
+
+## Troubleshooting
+
+Saga not found can occur when:
+
+1. The saga instance does not exist yet
+2. The saga instance already is removed
+
+Sometimes its not always that obvious when saga state will exist or be gone due to race conditions, false assumptions
```
mauroservienti, line 40:
```suggestion
Sometimes it's not always obvious when saga state exists or be gone due to race conditions or false assumptions.
```

nservicebus/sagas/saga-not-found.md
```diff
@@ -29,3 +29,67 @@ The ability to provide an implementation for `IHandleSagaNotFound` is especially
...
+
+### Out of order
+
+If the design assumes message A will create the saga and B updates the saga. If message B is received thus processed before A the saga will not yet exist and will be discarded.
```
mauroservienti, line 44:
```suggestion
If the design assumes message A will create the saga and message B updates the saga. If message B is received and processed before message A, the saga doesn't exist yet, and the message will be discarded.
```

nservicebus/sagas/saga-not-found.md
```diff
@@ -29,3 +29,67 @@ The ability to provide an implementation for `IHandleSagaNotFound` is especially
...
+
+Symptoms:
+
+- Saga are created, expected to be completed but this does not always happen.
+
+Mitigation:
+
+- Create the saga for both message A and B.
```
mauroservienti, line 52: we probably want to expand on this by pointing to documentation about `IAmStartedBy<>`

nservicebus/sagas/saga-not-found.md
```diff
@@ -29,3 +29,67 @@ The ability to provide an implementation for `IHandleSagaNotFound` is especially
...
+
+### Concurrency
+
+Messages might be processed concurrently. If the saga is started by messages 1 but that did not yet complete thus create the saga state and message 2 is processed which depends on the existance of the saga state this message will be discarded with saga not found.
```
mauroservienti, line 56:
```suggestion
Messages might be processed concurrently. If the saga is started by message A but did not complete yet (the saga state is not persisted yet) and message B, which depends on the existence of the saga state, is processed, it will be discarded with a saga not found error.
```

nservicebus/sagas/saga-not-found.md
```diff
@@ -29,3 +29,67 @@ The ability to provide an implementation for `IHandleSagaNotFound` is especially
...
+
+Symptoms:
+
+- Scatter/gather sagas behave unexpectedly and show saga not found.
+
+Mitigation:
+
+- Reduce processing concurrency to 1 to achieve sequential processing. This only works that only a single endpoint instance active.
```
mauroservienti, line 64: or use pessimistic locking where available. Or redesign the saga to make so that all message types can start it

mauroservienti (changes requested):
I reviewed only half of it. I'll have a look at the remaining paragraphs soon.

nservicebus/sagas/saga-not-found.md
```diff
@@ -29,3 +29,67 @@ The ability to provide an implementation for `IHandleSagaNotFound` is especially
 For example, consider a saga that is used for managing the registration process on the website. After a customer registers, they receive an email with a confirmation link. The system will wait for confirmation for a specific period of time, e.g. 24 hours. If the user doesn't click the link within 24 hours, their data is removed from the system and saga is completed. However, they might decide to click the confirmation link a few days later. In this case, the related saga instance can't be found and an exception will be thrown. By implementing `IHandleSagaNotFound` it is possible to handle the situation differently, e.g. redirect the user to the registration website and ask them to fill out the form again.
 
 The implementation of `IHandleSagaNotFound` should be driven by the business requirements for a specific situation. In some cases the message might be ignored; in others, it might be useful to track whenever that situation happens (e.g. by logging or sending another message). In still other cases, it might make sense to perform a custom compensating action.
+
+## Troubleshooting
+
+Saga not found can occur when:
```
timbussmann, line 35: This entrance sentence reads weird if "Saga not found" is just read as a regular part of a sentence.

```suggestion
*Saga not found* Exceptions can occur when:
```

nservicebus/sagas/saga-not-found.md
```diff
@@ -29,3 +29,67 @@ The ability to provide an implementation for `IHandleSagaNotFound` is especially
...
+
+1. The saga instance does not exist yet
+2. The saga instance already is removed
```
timbussmann, line 38:
```suggestion
2. The saga instance has already been removed
```

nservicebus/sagas/saga-not-found.md
```diff
@@ -29,3 +29,67 @@ The ability to provide an implementation for `IHandleSagaNotFound` is especially
 For example, consider a saga that is used for managing the registration process on the website. After a customer registers, they receive an email with a confirmation link. The system will wait for confirmation for a specific period of time, e.g. 24 hours. If the user doesn't click the link within 24 hours, their data is removed from the system and saga is completed. However, they might decide to click the confirmation link a few days later. In this case, the related saga instance can't be found and an exception will be thrown. By implementing `IHandleSagaNotFound` it is possible to handle the situation differently, e.g. redirect the user to the registration website and ask them to fill out the form again.
 
 The implementation of `IHandleSagaNotFound` should be driven by the business requirements for a specific situation. In some cases the message might be ignored; in others, it might be useful to track whenever that situation happens (e.g. by logging or sending another message). In still other cases, it might make sense to perform a custom compensating action.
+
+## Troubleshooting
+
+Saga not found can occur when:
+
+1. The saga instance does not exist yet
```
timbussmann, line 37:
```suggestion
1. The saga instance does not exist (yet)
```

nservicebus/sagas/saga-not-found.md
```diff
@@ -29,3 +29,67 @@ The ability to provide an implementation for `IHandleSagaNotFound` is especially
...
+
+- Scatter/gather sagas behave unexpectedly and show saga not found.
+
+Mitigation:
+
+- Reduce processing concurrency to 1 to achieve sequential processing. This only works that only a single endpoint instance active.
+
+### Initialization
+
+A saga that immediately sends one or more messages at start that will result in a response message could result in the response message not yet be able to be correlated is the init handler is still running.
```
timbussmann, line 68: I don't understand this sentence. Also, how does a saga immediately send messages at startup?

timbussmann (commented):
some of these sentences are hard to read and understand, especially for users less familiar with the details of NServiceBus. Also, there are quite a few spelling errors, I'd recommend going through the whole section with grammarly or something like that first.

nservicebus/sagas/saga-not-found.md
mikeminutillo: I wonder if we can expand on this a little to say something like:

> Any message that are sent from outside of the saga should be able to start a new instance with `IAmStartedBy<>`. Messages which are a result of a saga handler do not need this. i.e. A reply to a request made by the saga instance does not need to be able to create a new instance.

nservicebus/sagas/saga-not-found.md
mikeminutillo: Maybe use the term "bypasses the consistency guarantees" or something like that.

nservicebus/sagas/saga-not-found.md
ramonsmits, line 52:
```suggestion
- Create the saga for multiple message by using [IAmStartedBy<>(https://docs.particular.net/nservicebus/sagas/#starting-a-saga) with multiple types just as `IHandleMessages<>`. This way out-of-order processing will still create the saga.
  - Do keep in mind that the saga contains logic to "buffer" out of order state until it is in the state to continue.
  - Messages which are a result of a saga handler do not need this. i.e. A reply to a request made by the saga instance does not need to be able to create a new instance.
```
@mauroservienti @mikeminutillo how about this?

nservicebus/sagas/saga-not-found.md
ramonsmits: @mauroservienti pessimistic locking 
```suggestion
- Reduce processing concurrency to 1 to achieve sequential processing. This only works that only a single endpoint instance active.
- Use pessimistic locking on the saga persister, this only works if a storage is configured that supports pessimistic locking.
```

nservicebus/sagas/saga-not-found.md
ramonsmits:
```suggestion
A saga that at start immediately sends one or more request messages and already is receiving response messages while the handler is still running will result in a response message not able to be correlated if the init handler is still running and the saga state not yet been persisted.
```

@timbussmann Is this an improvement?

timbussmann: so this is basically just another manifestation of "concurrency", given the examples? From what I read in that sentence, anything that is received concurrently while a saga's first handler is being invoked can run into this issue. Immediate dispatches can cause concurrent processing in various ways, so maybe it would be easier to understand to have a section about the risks of immediate dispatches on saga creation?

nservicebus/sagas/saga-not-found.md
ramonsmits: @timbussmann Based on the docs only Azure and non-durable it seems. RavenDB doesn't support pessimistic locking but we use a "lease" construct. So the effect it the same as long as the saga will not be invoked longer then 60 seconds.

But this applies to pre v7 too for which basically most persisters didn't support pessimistic locking. As least not SQLP and RavenDB.

nservicebus/sagas/saga-not-found.md
mikeminutillo: I think the term "Create the saga" might be confusing. It sounds like the reader needs to do the creation when really we just want them to change the way they declare message handlers. Maybe lead with that?

```suggestion
- Use [IAmStartedBy<>](https://docs.particular.net/nservicebus/sagas/#starting-a-saga) instead of `IHandleMessages<>` for any message type which can originate from outside of the saga. No matter which order messages are delivered in the first one to be processed will create the saga instance.

- Each message handler that can start the saga might need to contain logic to check the saga state and see if it is time to take the next action. 

- Messages which are a result of a saga handler do not need this. i.e. A reply to a request made by saga instance does not need to be able to create a new instance.
```

I wonder if we shouldn't have some examples in this document to make things clearer.

e.g. Imagine a saga that will create schedule shipping after payment has been receieved and stock levels have been checked.

```csharp
class ShippingSchedulePolicy : Saga<ShippingSchedulePolicyData>,
  IAmStartedBy<PaymentReceived>, IAmStartedBy<StockLevelsChecked>
{
  public Task Handle(PaymentReceived message, IMessageHandlerContext context)
  {
    Data.PaymentReceived = true;
    return ScheduleShippingIfReady(context);
  }
  
  public Task Handle(StockLevelsChecked message, IMessageHandlerContext context)
  {
    Data.StockLevelChecked = true;
    return ScheduleShippingIfReady(context);
  }
  
  private async Task ScheduleShippingIfReady(IMessageHandlerContext context)
  {
    if(Data.StockLevelChecked && Data.PaymentReceived)
    {
      await context.Send(new ScheduleShipment { OrderId = Data.OrderId })
        .ConfigureAwait(false);
    }
  }
}
```

Notice that the handler for each message type needs to update the saga state and then check to see if the state is ready before sending the next message. It doesn't matter which order these messages are processed in. When both have been processed, the `ScheduleShipment` message will be sent.

We could do the same for the next bullet point and show a saga that sends a request and handles a response making note of the fact that the response message does not need to be able to start the saga.

nservicebus/sagas/saga-not-found.md
ramonsmits: @timbussmann Yes and no, this is specific to saga creation. The other situations would be caused by missing `IAmStartedBy`. This section is already about Immediate dispatches at creation.

nservicebus/sagas/saga-not-found.md
```diff
@@ -29,3 +29,74 @@ The ability to provide an implementation for `IHandleSagaNotFound` is especially
...
+
+Note: Expect out-of-order delivery if the message flow allows it and expect more-than-once processing.
+
+### Out of order
+
+If the design assumes message A will create the saga and message B updates the saga. If message B is received and processed before message A, the saga doesn't exist yet, and the message will be discarded.
+
+Symptoms:
+
+- Sagas are expected to exist but this does not always seems to be the case. Inspecting storage shows that the saga does exist.
```
mauroservienti, line 50:
```suggestion
- Sagas are expected to exist, but this does not always the case. Inspecting storage shows that the saga does not exist.
```

nservicebus/sagas/saga-not-found.md
```diff
@@ -29,3 +29,74 @@ The ability to provide an implementation for `IHandleSagaNotFound` is especially
...
+
+Mitigation:
+
+- Use [IAmStartedBy<>](/nservicebus/sagas/#starting-a-saga) instead of `IHandleMessages<>` for **any** message type which can originate from outside of the saga. No matter in which order messages are delivered, any message type processed first must create the saga.
+
```
mauroservienti, line 55:
```suggestion
- Use [IAmStartedBy<>](/nservicebus/sagas/#starting-a-saga) instead of `IHandleMessages<>` for **any** message type which can originate from outside of the saga. No matter in which order messages are delivered, any message type processed first must create the saga instance.
```

nservicebus/sagas/saga-not-found.md
```diff
@@ -29,3 +29,74 @@ The ability to provide an implementation for `IHandleSagaNotFound` is especially
...
+- Each message handler that can start the saga might need to contain logic to check the saga state and see if it is time to take the next action. 
+
```
mauroservienti, line 57:
```suggestion
- Each message handler that can start the saga might need to contain logic to check the saga state and see if it is time to take the following action. 
```

nservicebus/sagas/saga-not-found.md
```diff
@@ -29,3 +29,74 @@ The ability to provide an implementation for `IHandleSagaNotFound` is especially
...
+Messages might be processed concurrently which can result in out-of-order processing.
+
+Symptoms:
+
+- Scatter/gather sagas behave unexpectedly and show saga not found exceptions.
+
+Mitigation:
+
+- Reduce processing concurrency to 1 to achieve sequential processing. This only works when only a single endpoint instance is active.
+- Use pessimistic locking on the saga persister, this only works if a storage persister is configured that supports pessimistic locking.
```
mauroservienti, line 71:
```suggestion
- Use pessimistic locking on the saga persister, this only works if the configured storage persister supports pessimistic locking.
```

nservicebus/sagas/saga-not-found.md
```diff
@@ -29,3 +29,74 @@ The ability to provide an implementation for `IHandleSagaNotFound` is especially
...
+
+### Initialization
+
+A saga that at creation immediately sends one or more request messages and already is receiving response messages while the handler is still running will result in a response message not able to be correlated if the init handler is still running as the saga state did not yet been persisted
+
+This can happen when:
+
+- Using immediate dispatch
```
mauroservienti, line 79: Can we link to the doco here?

nservicebus/sagas/saga-not-found.md
```diff
@@ -29,3 +29,74 @@ The ability to provide an implementation for `IHandleSagaNotFound` is especially
...
+- Another `IMessageSession` is used instead of `context` or using an alternative protocol
```
mauroservienti, line 80:
```suggestion
- Another `IMessageSession` is used instead of the `IMessageHandlerContext` instance or using an alternative protocol
```

nservicebus/sagas/saga-not-found.md
```diff
@@ -29,3 +29,74 @@ The ability to provide an implementation for `IHandleSagaNotFound` is especially
...
+
+Mitigation:
+
+- Do not immediately send messages at saga creation, first send an init message to ensure the saga state exists. When a message is sent to a component that will respond the saga state will now exist.
+
+### More-than-once processing
+
+A message can be processed more-than-once because it was physically sent multiple times or a transport its consistency results in the same message becoming available again. More-than-once processing on a handler that completes the saga will now result in a saga not found for every duplicate.
```
mauroservienti, line 88:
```suggestion
A message can be processed more than once because it was physically sent multiple times. The same might occur because the transport consistency is "at-least-once," resulting in the same message being consumed more than once. More-than-once processing on a handler that completes the saga will now result in a saga not found for every duplicate. 
```

nservicebus/sagas/saga-not-found.md
```diff
@@ -29,3 +29,74 @@ The ability to provide an implementation for `IHandleSagaNotFound` is especially
...
+
+- Do not immediately send messages at saga creation, first send an init message to ensure the saga state exists. When a message is sent to a component that will respond the saga state will now exist.
+
+### More-than-once processing
+
+A message can be processed more-than-once because it was physically sent multiple times or a transport its consistency results in the same message becoming available again. More-than-once processing on a handler that completes the saga will now result in a saga not found for every duplicate. 
+
+Mitigation:
+
+- Use [outbox](/nservicebus/outbox/)
```
mauroservienti, line 92:
```suggestion
- Use the [outbox](/nservicebus/outbox/)
```

nservicebus/sagas/saga-not-found.md
```diff
@@ -29,3 +29,74 @@ The ability to provide an implementation for `IHandleSagaNotFound` is especially
...
+ 
+### Already completed
+
+A saga that is already completed can result in a saga not found exception.
```
mauroservienti, line 96:
```suggestion
A saga that is already completed can result in a saga not found exception if messages are dispatched after the completion.
```

nservicebus/sagas/saga-not-found.md
```diff
@@ -29,3 +29,74 @@ The ability to provide an implementation for `IHandleSagaNotFound` is especially
...
+
+Mitigation:
+
+- Use a logical complete state
+- Use [saga timeouts](timeouts.md) to postpone physically removing the saga state and invoke [`MarkAsComplete`](/nservicebus/sagas/#ending-a-saga).
+- Do not send messages and invoke [`MarkAsComplete`](/nservicebus/sagas/#ending-a-saga). Send an additional message to self (local) that will invoke `MarkAsComplete`.
```
mauroservienti, line 102:
```suggestion
- Do not send messages and invoke [`MarkAsComplete`](/nservicebus/sagas/#ending-a-saga). Send an additional message to the same saga instance (send local) that will invoke `MarkAsComplete`.
```

nservicebus/sagas/saga-not-found.md
ramonsmits: > Can we link to the doco here?

```suggestion
- Using [immediate dispatch](/nservicebus/messaging/send-a-message.md#dispatching-a-message-immediately)
```

PR #5588, opened 2021-12-20
Remove single threaded

danielmarbach:
I would suggest not using the phrasing "single threaded". It is not correct. Maybe talk about processing one message at a time, vs processing multiple messages at the same time? I think this is the more precise and yet simpler to understand explanation of what is going on.


SzymonPobiega (approved):

SzymonPobiega:
Thanks @danielmarbach ❤️

PR #5594, opened 2022-01-07
Update operational-scripting.md for clarity about specifying values for certain script command options

dlbromen:
Trying to add clarity that "--dialect" and "--output-dir" require "=value". The parser does not seem to look for arg+1 for the value like some command line tools do.

```sql-persistence script .\bin\...\Endpoint.dll --clean --dialect SqlServer --output-dir C:\Temp\sqlpersistence_out```
> sql-persistence failed: Unrecognized command or argument 'SqlServer'

### DIFF
--- a/persistence/sql/operational-scripting.md
+++ b/persistence/sql/operational-scripting.md
@@ -36,12 +36,12 @@ sql-persistence script <assembly>
 
 #### Options
  
-`-o` | `--output-dir` : Path to the output directory. If not specified, the current directory will be used.
+`-o=<path>` | `--output-dir=<path>` : Path to the output directory. If not specified, the current directory will be used.
 
 `--clean` : Removes existing files in the output directory
 
 `--overwrite`: Overwrites existing files in the output if they match the files to be generated
 
-`--dialect`: Specifies a dialect to generate. Allowed values are: SqlServer, MySql, Oracle, PostgreSql
+`--dialect=<value>`: Specifies a dialect to generate. Allowed values are: SqlServer, MySql, Oracle, PostgreSql
 
 `--verbose`: Verbose logging


DavidBoike:
Thanks @dlbromen!

dlbromen:
> Maybe this should all be in the help output, with this page just saying something like:
> 
> > Run `sql-persistence script --help` for details
> 
> Or if we really want to repeat it in this page, that would be copypasta of the help output.

I could do that I suppose, but I think the Help is currently autogenerated by the McMaster library. This was the least intrusive thing I could do.
I thought if I was going to dive into changing the `NServiceBus.Persistence.Sql.CommandLine` project, then I would go a step further and make the parser work "right" and accept spaces.

Instead of:
`var outputOption = cmd.Option("-o | --output-dir", "Path to the output directory.", CommandOptionType.SingleOrNoValue);`
https://github.com/Particular/NServiceBus.Persistence.Sql/blob/96fdcf87fac297e003c64cd02c353c1c05d0326a/src/CommandLine/Program.cs#L24

Use:
`var outputOption = cmd.Option("-o | --output-dir", "Path to the output directory.", CommandOptionType.Single);`

Doing something like this, combined with setting a value for `option.ValueName` would also generate more accurate help text from what I can see.

If the team is open to that kind of change, let me know.

DavidBoike:
@adamralph I think it's fine to have copypasta, that way Google sees it.

@dlbromen Not personally knowing the McMaster stuff very well, I would be a little nervous about changing in case there would be a script out there with an `=` that would suddenly not work after the change. I'm not sure that's a PR we could merge or release quickly. Personally, I'm just glad the documentation is more clear about how it works now.

adamralph:
> I think it's fine to have copypasta, that way Google sees it

👍 good point

PR #5595, opened 2022-01-10
[Do not merge] Show simple MSMQ-to-ASB routing via router

SzymonPobiega:
Shows the simplest way to forward messages from an MSMQ endpoint to an ASB endpoint

Related to support `00068748`


SzymonPobiega:
Case solved, closing this one.

PR #5609, opened 2022-01-21
Demo body storage in azure blob container

andreasohlund:
@kentdr use this to demo storing audit bodies in azure.

Steps:

1. Create a storage account
2. Add the conn string to a system env variable called `AzureBlobAuditStorage`
3. Change the container name to match what you have configured in ServiceControl
4. Download the nuget packages from the core spike  - https://github.com/Particular/NServiceBus/actions/runs/1723254199
5. Store the nsericebus nuget in a local folder and create a local nuget source pointing to that folder
6. Run the sample and send some messages => audit bodies should now be stored in blob storage and audit messages should only contain headers

### DIFF
--- a/samples/fullduplex/Core_8/Shared/AzureAuditBodyStorageConfiguration.cs
+++ b/samples/fullduplex/Core_8/Shared/AzureAuditBodyStorageConfiguration.cs
@@ -1,0 +1,19 @@
+﻿namespace Shared
+{
+    using Azure.Storage.Blobs;
+    using System;
+    using System.Threading.Tasks;
+
+    public static class AzureAuditBodyStorageConfiguration
+    {
+        public static async Task<BlobContainerClient> GetContainerClient()
+        {
+            var blobClient = new BlobServiceClient(Environment.GetEnvironmentVariable("AzureBlobAuditStorage"));
+
+            var blobContainerClient = blobClient.GetBlobContainerClient("audit-bodies");
+            await blobContainerClient.CreateIfNotExistsAsync().ConfigureAwait(false);
+
+            return blobContainerClient;
+        }
+    }
+}
--- a/samples/fullduplex/Core_8/Shared/StoreAuditBodyInAzureBlobStorageBehavior.cs
+++ b/samples/fullduplex/Core_8/Shared/StoreAuditBodyInAzureBlobStorageBehavior.cs
@@ -1,0 +1,57 @@
+﻿namespace Shared
+{
+    using Azure.Storage.Blobs;
+    using Azure.Storage.Blobs.Models;
+    using Microsoft.Toolkit.HighPerformance;
+    using NServiceBus;
+    using NServiceBus.Pipeline;
+    using NServiceBus.Routing;
+    using System;
+    using System.Collections.Generic;
+    using System.Threading.Tasks;
+
+    public class StoreAuditBodyInAzureBlobStorageBehavior : Behavior<IDispatchContext>
+    {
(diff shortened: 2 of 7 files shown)


samples/fullduplex/Core_8/Shared/StoreAuditBodyInAzureBlobStorageBehavior.cs
```diff
@@ -0,0 +1,57 @@
...
+                        }
+                };
+
+                await blob.UploadAsync(
+                    BinaryData.FromStream(message.Body.AsStream()),
+                    options,
+                    context.CancellationToken
+                    ).ConfigureAwait(false);
+
+                operation.Message.UpdateBody(null); //TODO: Would ReadOnlyMemory<byte>.Empty be better?
```
kentdr, line 49: We may want to consider adding a couple of header values here to indicate the body was stored prior to sending, ServiceControl uses an awkward [`BodyNotStored`](https://github.com/Particular/ServiceControl/blob/3d81d2f1d0f90db387704a856c69ae1e926b03a1/src/ServiceControl.Audit/Auditing/BodyStorage/BodyStorageFeature.cs#L56) flag to know if the body is stored with the headers or in body storage.  Also, maybe consider setting the [`ContentLength`](https://github.com/Particular/ServiceControl/blob/3d81d2f1d0f90db387704a856c69ae1e926b03a1/src/ServiceControl.Audit/Auditing/BodyStorage/BodyStorageFeature.cs#L43) as well since everything will be 0 on the other side.   As it stands, the logic on the ServiceControl end is a bit fuzzy for if something is or isn't stored in body storage. It is generally just assuming if the body is 0 length something else stored it.

samples/fullduplex/Core_8/Shared/AzureAuditBodyStorageConfiguration.cs
```diff
@@ -0,0 +1,19 @@
...
+    using System;
+    using System.Threading.Tasks;
+
+    public static class AzureAuditBodyStorageConfiguration
+    {
+        public static async Task<BlobContainerClient> GetContainerClient()
+        {
+            var blobClient = new BlobServiceClient(Environment.GetEnvironmentVariable("AzureBlobAuditStorage"));
+
+            var blobContainerClient = blobClient.GetBlobContainerClient("audit-bodies");
```
kentdr, line 13: Maybe not worth it for the demo, but if we were to move forward with this, the container name will need to be configured exactly the same on both sides, or, if we think there would possibly be multiple containers, we will need to send it along in a header or something so ServiceControl looks in the right place for the body data.

samples/fullduplex/Core_8/Shared/StoreAuditBodyInAzureBlobStorageBehavior.cs
andreasohlund: Yea, good idea, I'll add something. 

>ServiceControl uses an awkward BodyNotStored flag to know if the body is stored

The `BodyNotStored` feels like an SC internal thing so let go with something like `NServiceBus.BodyStorage` = `AzureBlob`

What more would we add? (Storage account name + container name?)

>Also, maybe consider setting the ContentLength as well since everything will be 0 on the other side

Will do

samples/fullduplex/Core_8/Shared/AzureAuditBodyStorageConfiguration.cs
andreasohlund: yea, I just hacked this together

PR #5614, opened 2022-01-25
2021-01 Quickstart intro restructure

sergioc:


tutorials/quickstart/tutorial.md
```diff
@@ -8,9 +8,33 @@ extensions:
...
+You'll also see how the additional tools in the Particular Service Platform make it easy to manage, monitor, and debug.
+#end-variant
+
+#variant Variant
+
+In this tutorial, you'll see how NServiceBus takes the grunt work out of system design by handling all of the plumbing for you.
+
+You will build part of an e-commerce system that handles receiving an order placed by a user, as well as the billing and shipping that goes along with it. This is what the final solution will look like:
+
+![quickstart solution diagram](https://docs.particular.net/tutorials/quickstart/after.svg)
```
DavidBoike, line 29: This should fix content verification
```suggestion
![quickstart solution diagram](/tutorials/quickstart/after.svg)
```

tutorials/quickstart/tutorial.md
```diff
@@ -8,9 +8,33 @@ extensions:
...
+NServiceBus combines decades of distributed systems design experience and expertise and distills it into one easy-to-use framework. In this tutorial, you'll see how NServiceBus takes all the grunt work out of system design by handling all of the plumbing for you, taking system design best practices like reliability, failure recovery, and extensibility and baking them right into the software, guiding you toward the pit of success.
+
+You'll also see how the additional tools in the Particular Service Platform make it easy to manage, monitor, and debug.
+#end-variant
+
+#variant Variant
+
+In this tutorial, you'll see how NServiceBus takes the grunt work out of system design by handling all of the plumbing for you.
+
+You will build part of an e-commerce system that handles receiving an order placed by a user, as well as the billing and shipping that goes along with it. This is what the final solution will look like:
```
DavidBoike, line 27: In the QuickStart they don't build, per se. They explore, and barely at the very end write some code to extend the system. Is this setting the wrong expectation?

tutorials/quickstart/tutorial.md
```diff
@@ -8,9 +8,33 @@ extensions:
...
+
+![quickstart solution diagram](https://docs.particular.net/tutorials/quickstart/after.svg)
+
+The tutorial is composed of 3 parts - introduction, failure recovery, and extensibility - that demonstrate how NServiceBus helps with design best practices and how the Particular Service Platform helps monitor a distributed system. Each part takes approximately 15-20 minutes to complete.
+
+To run this tutorial you’ll need:
+
+- Experience with C# coding
+- Windows 10 or 11 running an IDE for C# code, for example Visual Studio Code, Visual Studio 2022, or JetBrains Rider.
```
DavidBoike, line 36: VSCode is not optimal for our solutions. We do provide the `.vscode` folder so it IS possible, but should we really lead with that? It's probably best to lead with Visual Studio.

You could also do it on Visual Studio for Mac, or Rider on a Mac, or if you're going to mention VSCode then no reason you can't do it on a Mac, so I'm not sure why we're mandating Windows 10 or 11 here.

tutorials/quickstart/tutorial.md
sergioc:
```suggestion
You will explore part of an e-commerce system that handles receiving an order placed by a user, as well as the billing and shipping that goes along with it. This is what the final solution will look like:
```

tutorials/quickstart/tutorial.md
sergioc: Because SP and SC won't work later on without Windows.

DavidBoike: Ah. I'd be more likely to say then what's actually required, and that one part of the tutorial requires Windows. I don't think you have to worry about versions. Windows 10 has been around for so long I don't think there's many developers out there still trying to rock Windows XP. Honestly we should just defer to Microsoft on this, they already have documented what they [require for .NET Core 3.1](https://github.com/dotnet/core/blob/main/release-notes/3.1/3.1-supported-os.md) so maybe we should just say the ability to build and run a .NET Core 3.1 solution.

tutorials/quickstart/tutorial.md
```diff
@@ -8,9 +8,33 @@ extensions:
   nextText: "Next: Recovering from failure"
   nextUrl: tutorials/quickstart/tutorial-reliability
 previewImage: add-shipping-endpoint.png
+abTestName: Quickstart intro 2022 01
 ---
 
-include: quickstart-tutorial-intro-paragraph
```
adamralph, line 13: @sergioc @soujay @DavidBoike should the now redundant include file be removed as part of this PR?

sergioc: When the test ends it will be needed again, so I'm not sure.

DavidBoike: The include is also used here: https://github.com/Particular/docs.particular.net/blob/master/tutorials/index.md#nservicebus-quick-start

PR #5629, opened 2022-02-09
first draft of dedicated message contracts page

kbaley:


mauroservienti:
the merge conflicts here are huge. I don't understand why git doesn't like doing a proper diff for these files 😭

PR #5647, opened 2022-02-27
editorial review of shared message contracts docs

adamralph:
relates to https://github.com/Particular/docs.particular.net/pull/5634

### DIFF
--- a/nservicebus/messaging/evolving-contracts.md
+++ b/nservicebus/messaging/evolving-contracts.md
@@ -1,5 +1,5 @@
 ---
-title: Evolving Message Contracts
+title: Evolving message contracts
 summary: Guidelines for choosing a strategy for evolving message contracts
 reviewed: 2020-06-16
 component: Core
@@ -12,7 +12,7 @@ Evolving contracts over time is not an easy task and the appropriate strategy sh
 
 This article presents basic guidelines for choosing contracts evolution strategy, avoiding common mistakes, and ensuring that contracts will be easy to evolve over time.
 
-Note: Ensure that message contracts follow the general [messages design guidelines](/nservicebus/messaging/messages-events-commands.md#designing-messages).
+Note: Message contracts should also follow the [general message design guidelines](/nservicebus/messaging/messages-events-commands.md#designing-messages).
 
 ## Common challenges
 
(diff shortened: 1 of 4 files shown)


nservicebus/messaging/evolving-contracts.md
```diff
@@ -12,7 +12,7 @@ Evolving contracts over time is not an easy task and the appropriate strategy sh
 
 This article presents basic guidelines for choosing contracts evolution strategy, avoiding common mistakes, and ensuring that contracts will be easy to evolve over time.
 
-Note: Ensure that message contracts follow the general [messages design guidelines](/nservicebus/messaging/messages-events-commands.md#designing-messages).
+Note: Message contracts should also follow the [general message design guidelines](/nservicebus/messaging/messages-events-commands.md#designing-messages).
```
adamralph, line 15: Note that _{verb} {something}..._ (e.g. _Ensure that message contracts follow..._) is also a form of [personal voice](https://github.com/Particular/docs.particular.net#language-preferences).

kbaley: I recognize that this is personal voice because of the implied _you_ but I often let it go because changing it leads to a lot of passive voice statements, like you have later in the review. E.g.

> Use a connection string

vs.

> A connection string may be used

To me, the former reads more like a direct action we want you to take and the latter reads like a suggestion that can be ignored, not to mention the extra verbosity.

Thoughts?

timbussmann (approved):

mauroservienti (approved):

nservicebus/messaging/evolving-contracts.md
mauroservienti: Far away from being an authority, I agree with Kyle for what matters.

adamralph: @kbaley I assume your comment is general, and not related to this specific change (and I'm merging the PR on that assumption).

Regarding your example, I think I'd have to see an real example, surrounded by context, to form an opinion. But generally, I agree, _{do something}_ and _{something may be done}_ are not equivalent.

kbaley: Yes, my opinion was a general one and to be honest, not super strongly-held. I usually err on the side of using the command form unless it truly doesn't really matter either way.

PR #5662, opened 2022-03-14
Use community packages at your own risk

ramonsmits:


nservicebus/community/index.md
```diff
@@ -51,14 +51,12 @@ redirects:
 
 This is a curated list of community-developed extensions and integrations for the NServiceBus ecosystem.
 
-Community projects are maintained by community members and are **not covered** by the [Particular Software License](https://particular.net/licensing) or the [Particular Software Support Agreement](https://particular.net/supportagreement). Each project is covered by its own license and terms. It's possible that a community contribution may not be updated to support the most recent version of NServiceBus.
-
+WARNING: These are **no recommendations** and are used at your own risk! Community projects are maintained by community members and are **not covered** by the [Particular Software License](https://particular.net/licensing) or the [Particular Software Support Agreement](https://particular.net/supportagreement). Each project is covered by its own license and terms. It's possible that a community contribution may not be updated to support the most recent version of NServiceBus.
```
mikeminutillo, line 54: The term "use at your own risk" violates the "no personal voice" guideline https://github.com/particular/docs.particular.net#language-preferences

If I remove it, then all that is left is "These are **no recommendations**!" which I think is so strong that I would question why we have the page at all. Why list to a collection of community packages if we're telling users not to use them. 

I like highlighting the message with a WARNING box

```suggestion
WARNING: Community projects are maintained by community members and are **not covered** by the [Particular Software License](https://particular.net/licensing) or the [Particular Software Support Agreement](https://particular.net/supportagreement). Each project is covered by its own license and terms. It's possible that a community contribution may not be updated to support the most recent version of NServiceBus.
```

Maybe we could open with something like:

> Particular Software cannot control or guarantee the quality or release lifecycle of community projects.

PR #5665, opened 2022-03-15
Change the word "plumbing" with something more literal

udidahan:
Based on this feedback that I got:

> I would swap out the word “plumbing” here for something more literal. Metaphors in tutorials can be confusing for the reader.


DavidBoike (approved):
I have no objections to the wording change but this falls inside an A/B test variant (which isn't immediately obvious from the GitHub diff view) so I think perhaps @sergioc (I remember he was involved in the A/B test) should weigh in on whether we should merge it or wait.

adamralph (approved):

sergioc:
The test is still running, so this PR should only be merged after the results are analyzed. I'll add processing this PR as an item of the PoA for the [analysis issue](https://github.com/Particular/DeveloperEducation/issues/3720).

udidahan:
Can this go ahead now?

sergioc:
This has been incorporated in #5706.

PR #5668, opened 2022-03-15
Add sample for NServiceBus.Extensions.Logging v2

timbussmann:
Also changes the log level for the handler message to warn since we're using a colored logger which isn't clear at all as long as we're only logging info/debug messages.

### DIFF
--- a/samples/logging/extensions-logging/Extensions.Logging_1/Sample/MyHandler.cs
+++ b/samples/logging/extensions-logging/Extensions.Logging_1/Sample/MyHandler.cs
@@ -9,7 +9,7 @@ public class MyHandler :
 
     public Task Handle(MyMessage message, IMessageHandlerContext context)
     {
-        log.Info("Hello from MyHandler");
+        log.Warn("Hello from MyHandler");
         return Task.CompletedTask;
     }
 }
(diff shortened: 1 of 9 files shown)


samples/logging/extensions-logging/Extensions.Logging_1/Sample/MyHandler.cs
```diff
@@ -9,7 +9,7 @@ public class MyHandler :
 
     public Task Handle(MyMessage message, IMessageHandlerContext context)
     {
-        log.Info("Hello from MyHandler");
+        log.Warn("Hello from MyHandler");
```
HEskandari, line 12: should this be a WARN?

mikeminutillo (approved):
I'm not sure about switching to WARN for visual clarity but not enough to block this

samples/logging/extensions-logging/Extensions.Logging_1/Sample/MyHandler.cs
timbussmann: see the comment on the isse description

timbussmann:
for more context:

With `Info`:
image


with `Warn`:
image

timbussmann:
discussed the log level. Merging this for now since it can always be changed easily.

PR #5671, opened 2022-03-15
Add Datadog sample for NSB v8

timbussmann:
needs some refactoring since the `Feature` isn't required anymore to get to the metrics options.

### DIFF
--- a/samples/logging/datadog/Core_8/Endpoint/LoadSimulator.cs
+++ b/samples/logging/datadog/Core_8/Endpoint/LoadSimulator.cs
@@ -1,0 +1,67 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+using NServiceBus;
+
+// Simulates busy (almost no delay) / quite time in a sine wave
+class LoadSimulator
+{
+    IEndpointInstance endpointInstance;
+    CancellationTokenSource tokenSource = new CancellationTokenSource();
+    TimeSpan minimumDelay;
+    TimeSpan idleDuration;
+    Task fork;
+
+    public LoadSimulator(IEndpointInstance endpointInstance, TimeSpan minimumDelay, TimeSpan idleDuration)
+    {
+        this.endpointInstance = endpointInstance;
+        this.minimumDelay = minimumDelay;
+        this.idleDuration = TimeSpan.FromTicks(idleDuration.Ticks / 2);
+    }
+
+    public Task Start()
+    {
+        fork = Loop();
+        return Task.CompletedTask;
+    }
+
+    async Task Loop()
+    {
+        try
+        {
+            while (true)
+            {
+                await Work()
+                    .ConfigureAwait(false);
+                var delay = NextDelay();
+                await Task.Delay(delay, tokenSource.Token)
+                    .ConfigureAwait(false);
+            }
+        }
+        catch (OperationCanceledException)
+        {
+        }
+    }
+
+    int index;
+
+    TimeSpan NextDelay()
+    {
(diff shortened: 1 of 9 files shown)


samples/logging/datadog/Core_8/Endpoint/LoadSimulator.cs
```diff
@@ -0,0 +1,67 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+using NServiceBus;
+
+// Simulates busy (almost no delay) / quite time in a sine wave
```
mikeminutillo, line 6:
```suggestion
// Simulates busy (almost no delay) / quiet time in a sine wave
```

samples/logging/datadog/sample.md
```diff
@@ -38,7 +38,7 @@ Setup Datadog client. Let' s assume the Datadog agent runs on default address an
 
 snippet: setup-datadog-client
 
-Custom observers need to be registered for the metric probes provided via `NServiceBus.Metrics`. This is configured in the `DatadogFeature`. The registered observers convert NServiceBus.Metric *Signals* to Datadog *Count* and NServiceBus.Metric *Durations* to Datadog *Timer* 
+Custom observers need to be registered for the metric probes provided via the `NServiceBus.Metrics` package. The registered observers convert NServiceBus.Metric *Signals* to Datadog *Count* and NServiceBus.Metric *Durations* to Datadog *Timer*.
```
mikeminutillo, line 41: The v7 version of the sample still uses the feature. Is it worth backporting this change (or making the v8 sample still a feature)?

mikeminutillo (approved):
I think the idea of having it in a feature was so that it was encapsulated and re-usable, not because it was necessary. It does add noise to the test though so maybe it's worth dropping it. It still had hard-coded connection strings inside the feature and it might detract from the sample to show how to set up a configuration extension method to enable the feature and pass connection details around.

samples/logging/datadog/sample.md
timbussmann: I don't see a problem with v7 using a feature, the text doesn't go deeper so I don't think this is a confusing change.

But we can't use a feature unless we take some extra effort to pass everything through the settings to set up the observes in the feature instead. I think we should not use a feature. If we really want, we can encapsulate the setup in a dedicated class but I didn't see that much value in that.

samples/logging/datadog/Core_8/Endpoint/Program.cs
```diff
@@ -0,0 +1,135 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using NServiceBus;
+using StatsdClient;
+
+class Program
+{
+    const string EndpointName = "Samples.Metrics.Tracing.Endpoint";
```
timbussmann, line 10:
```suggestion
    const string EndpointName = "Samples.Metrics.Tracing.Endpoint";

```

samples/logging/datadog/sample.md
mikeminutillo: Fair enough. The feature isn't the point of the sample and people are unlikely to download both versions and compare them, wondering why they're different.

timbussmann:
so btw. in the new relic sample (https://github.com/Particular/docs.particular.net/pull/5672) I've extracted the setup into a dedicated class. Is that making the sample easier to read?

mikeminutillo:
> so btw. in the new relic sample (#5672) I've extracted the setup into a dedicated class. Is that making the sample easier to read?

I think it does. It means that someone who downloads the code will find everything needed for integration in one place that is separate from everything else.

PR #5684, opened 2022-03-22
More snippets

HEskandari:


Snippets/FileShareDataBus/Core_8/CleanupStrategy/Define.cs
```diff
@@ -0,0 +1,17 @@
+namespace Core7.CleanupStrategy
```
mikeminutillo, line 1: Should these namespaces change? Do these even really need namespaces?

mikeminutillo (approved):

Snippets/FileShareDataBus/Core_8/CleanupStrategy/Define.cs
HEskandari: Pushed an update to remove all the unnecessary namespaces.

mikeminutillo (approved):

Snippets/CallbacksTesting/CallbacksTesting_4/Assert.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using System.Threading.Tasks;
+// ReSharper disable UnusedParameter.Global
+// ReSharper disable UnusedTypeParameter
```
timbussmann, line 4:
```suggestion
```

timbussmann (approved):

PR #5690, opened 2022-03-23
Update batch of samples to NServiceBus 8 beta 1

mikeminutillo:


samples/sqltransport-nhpersistence/Core_8/Receiver/OrderLifecycleSagaData.cs
```diff
@@ -2,7 +2,9 @@
 using NServiceBus;
 
 public class OrderLifecycleSagaData :
+#pragma warning disable NSB0012 // Saga data classes should inherit ContainSagaData
```
HEskandari, line 5: Is this a preferred approach? Why not change it inherit the `ContainSagaData` class instead?

mikeminutillo: I wasn't sure if that was possible with NHibernate. I checked and the members are virtual. I will update the sample.

samples/azure/azure-service-fabric-routing/Core_8/Shared/EndpointPartitioning/HardcodeReplyToAddressToLogicalAddressFeature.cs
```diff
@@ -6,7 +6,7 @@ class HardcodeReplyToAddressToLogicalAddressFeature : Feature
     protected override void Setup(FeatureConfigurationContext context)
     {
         var settings = context.Settings;
-        var addressToLogicalAddress = new HardcodeReplyToAddressToLogicalAddress(settings.InstanceSpecificQueue());
+        var addressToLogicalAddress = new HardcodeReplyToAddressToLogicalAddress(context.InstanceSpecificQueueAddress().ToString());
```
timbussmann, line 9: I don't think this is the correct approach here. `ToString` does not generate a valid transport-specific address (not sure what this sample does with the value though). The `QueueAddress` needs to be translated using `ITransportAddressResolver` which is available via DI for users.

Happy to help with this.

timbussmann (changes requested):
(feel free to discard this review if the comment is not relevant)

timbussmann (approved):

PR #5692, opened 2022-03-24
Add troubleshooting guidance for SQL Encrypt breaking change

mikeminutillo:
Microsoft.Data.SqlClient version 4 includes a breaking change https://github.com/dotnet/SqlClient/pull/1210 which causes existing clients to throw an exception when using the default configuration:

```
System.Data.SqlClient.SqlException
  HResult=0x80131904
  Message=A connection was successfully established with the server, but then an error occurred during the login process. (provider: SSL Provider, error: 0 - The certificate chain was issued by an authority that is not trusted.)
  Source=.Net SqlClient Data Provider
```

### DIFF
--- a/servicecontrol/troubleshooting.md
+++ b/servicecontrol/troubleshooting.md
@@ -76,7 +76,7 @@ System.Data.SqlClient.SqlException
   Source=.Net SqlClient Data Provider
 ```
 
-SQL Server uses a certificate to encrypt communication between itself and ServiceControl. Version 4 of the `Microsoft.Data.SqlClient` package includes a [breaking change](https://github.com/dotnet/SqlClient/pull/1210) to set `Encrypt=true` by default (the previous default was `false`) which causes this exception.
+When encyrption is enabled, SQL Server uses a certificate to encrypt communication between itself and ServiceControl. Version 4 of the `Microsoft.Data.SqlClient` package includes a [breaking change](https://github.com/dotnet/SqlClient/pull/1210) to set `Encrypt=true` by default (the previous default was `false`) which causes this exception.
 
 To fix it, [the SQL Server installation must be updated with a valid certificate and the ServiceControl machine must be updated to trust this certificate](https://docs.microsoft.com/en-us/sql/database-engine/configure-windows/enable-encrypted-connections-to-the-database-engine).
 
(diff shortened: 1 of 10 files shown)


HEskandari (approved):

tmasternak (approved):
LGTM

Minor question: can we remove duplication form the guidance files?

mikeminutillo:
Duplication (mostly) removed. I pulled the guidance out of the individual upgrade guides and created a specific upgrade guide for upgrading to SqlClient 4 in both SQL-T and SQL-P. The ServiceControl one has slightly different text so I left it as a "duplicate"

tmasternak (approved):

servicecontrol/troubleshooting.md
```diff
@@ -65,6 +65,23 @@ If ServiceControl fails to start and the logs contain a `Microsoft.Isam.Esent.In
...
+If ServiceControl fails to start and the logs contain the following exception, then ServiceControl is not able to connect to the SQL Server instance.  
+
+```
+System.Data.SqlClient.SqlException
+  HResult=0x80131904
+  Message=A connection was successfully established with the server, but then an error occurred during the login process. (provider: SSL Provider, error: 0 - The certificate chain was issued by an authority that is not trusted.)
+  Source=.Net SqlClient Data Provider
+```
+
+SQL Server uses a certificate to encrypt communication between itself and ServiceControl. Version 4 of the `Microsoft.Data.SqlClient` package includes a [breaking change](https://github.com/dotnet/SqlClient/pull/1210) to set `Encrypt=true` by default (the previous default was `false`) which causes this exception.
```
timbussmann, line 79: > SQL Server uses a certificate to encrypt communication between itself and ServiceControl.

this states that SQL Server already encryptions communication by default, although in the next sentence we state that the default was set to `false`, isn't that contradicting?

timbussmann (approved):
minor nitpick

servicecontrol/troubleshooting.md
mikeminutillo, line 79:
```suggestion
When encyrption is enabled, SQL Server uses a certificate to encrypt communication between itself and ServiceControl. Version 4 of the `Microsoft.Data.SqlClient` package includes a [breaking change](https://github.com/dotnet/SqlClient/pull/1210) to set `Encrypt=true` by default (the previous default was `false`) which causes this exception.
```

PR #5694, opened 2022-03-26
Match the feature complete state to our process

WilliamBZA:
Our branching strategy make it nearly impossible to define a beta as feature complete. This PR updates the release policy to match our branching strategy.

### DIFF
--- a/nservicebus/upgrades/release-policy.md
+++ b/nservicebus/upgrades/release-policy.md
@@ -100,7 +100,6 @@ The release cycle consists of the following quality stages:
 
 ### Beta
 
-* The component is feature complete.
 * Likely to contain a number of known or unknown bugs.
 * Performance and stability testing has not been fully completed.
 * This version does not come with a "go live" license and **is not supported** in production.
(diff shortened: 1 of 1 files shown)


nservicebus/upgrades/release-policy.md
```diff
@@ -100,7 +100,6 @@ The release cycle consists of the following quality stages:
 
 ### Beta
 
-* The component is feature complete.
```
kentdr, line 103: Instead of removing this entirely, would it be better to state something similar to the RC section? Something like:

In this stage of stabilization, all features have been designed and coded but may lack functionality and could contain significant bugs or performance issues.

WilliamBZA: Without enforcing artificial code freezes once a beta starts, we can't enforce that all features have been designed and coded. The only way to enforce that is to branch a release branch at beta

bording (commented):
I don't think I agree with the premise stated in the PR body text, but I do like this change. I've been wanting to get rid of the alpha / beta distinction entirely, and instead move to something like preview / rc instead.

This PR is aligned with that goal.

PR #5701, opened 2022-04-04
Outbox won't handle messages sent using IMessageSession

mauroservienti:


danielmarbach (approved):

danielmarbach:
FYI parts of it is also covered in 

https://docs.particular.net/nservicebus/outbox/#important-design-considerations

> ... sent outside of an NServiceBus message handler (i.e. from a Web API) cannot be deduplicated unless they are sent with the same MessageId

I still it is worth having the additional note

PR #5702, opened 2022-04-05
delayed delivery

HEskandari:


Snippets/Testing/Testing_8/UpgradeGuides/7to8/TestableBehaviorContext.cs
```diff
@@ -0,0 +1,24 @@
...
+    public class TestableBehaviorContext
+    {
+        class TestableBehaviorContextImp : NServiceBus.Testing.TestableBehaviorContext
+        {
+        }
+
+        public void UsingDispatchPropertiesInTestableBehaviorContext()
+        {
+            #region 7to8-adddeliveryconstraint
+            var context = new TestableBehaviorContextImp();
```
mikeminutillo, line 16: I think this should be outside of the snippet. This is not code a customer would need to write. You can avoid this by passing in a `TestableBehaviorContext` in as a method parameter.

```csharp
public void UsingDispatchPropertiesInTestableBehaviorContext(NServiceBus.Testing.TestableBehaviorContext context)
{
  context.Extensions.Set(new DispatchProperties
  {
    DelayDeliveryWith = new DelayDeliveryWith(TimeSpan.FromDays(1))
  });
}
```

HEskandari: @mikeminutillo Updated

PR #5720, opened 2022-05-09
saga batching sample

HEskandari:


samples/saga/batching/Core_8/WorkProcessor/WorkProcessingHandler.cs
```diff
@@ -0,0 +1,16 @@
+namespace NServiceBus.WorkProcessor;
+
+public class WorkProcessingHandler : IHandleMessages<ProcessWorkOrder>
+{
+    public Task Handle(ProcessWorkOrder message, IMessageHandlerContext context)
+    {
+        Console.WriteLine($"Processing work order '{message.WorkOrder}'");
+        Thread.Sleep(Random.Shared.Next(500, 1000));
+
+        return context.Send(new WorkOrderCompleted
```
ramonsmits, line 10: Use Reply instead of Send, and use try..catch for non-transient errors like work order no longer exists or some functional error. See other comment.

samples/saga/batching/sample.md
```diff
@@ -0,0 +1,61 @@
...
+Processing work order '10'
+Processing work order '2'
+...
+Processing work order '431'
+Processing work order '427'
+Processing work order '441'
+Processing work order '420'
+```
+
+Note that while the work requests are generated in sequence, on the processing side the work can happen out of order.
```
ramonsmits, line 51: One of the drawbacks that I think need to be documented is. What when one or more item fails processing? That would result in the 'paged processing' to fail.

I think it would be good to keep in mind that 'paged processing' should not fail due to non-transient errors and that the invoked handler ensures a "Reply"

samples/saga/batching/sample.md
```diff
@@ -0,0 +1,61 @@
...
+reviewed: 2022-04-09
+component: Core
+related:
+- nservicebus/sagas
+---
+
+
+## Introduction
+
+This sample shows how to control the task (work) generation via batching of the work items through Sagas.
```
ramonsmits, line 13: 1. Its missing why someone would want to do this.
2. Its missing alternatives. Like, do processing on an isolated endpoint or even a [sending work items to a specific endpoint instance](https://docs.particular.net/nservicebus/messaging/routing#make-instance-uniquely-addressable) where this endpoint could have a specific lower concurrency or even [processing rate limit](https://github.com/ramonsmits/NServiceBus.RateLimiter).
3. The example isn't stating how to do the paging.
4. A drawback of this solution is that not all work items are processed as fast as possible but I guess that is the whole reason for this overhead to throttle?
5. As critical time can no longer be used to validate to "peek" workload it might be worth stating that it likely is very good idea to capture the time delta between saga start and completion to track how long this "batch operation" takes in total and emit this via a metric framework.
6. An orchestration like alternative could also be based on a routing slip to achieve sequential atomic work item processing without the saga overhead IO to significantly reduced wasted IO and latency.

samples/saga/batching/Core_8/WorkProcessor/WorkProcessingHandler.cs
HEskandari: Reply makes sense, but not sure about Try...Catch. What would you catch in a handler context that makes sense for a sample?

samples/saga/batching/sample.md
HEskandari: I think this would cover that:

>The saga controls the generation of the next batch of work and monitors for the completion signal of all messages in that batch, before starting the next batch.

it wouldn't necessarily fail the paged processing but would stall it and won't move to the subsequent page. If that is not desired, the worker process can send a different kind of 'reply' that contains error info. In other words, this is a basic structure without being too prescriptive as that'd be different based on business requirements.

samples/saga/batching/sample.md
HEskandari: I have updated, and hopefully addressed points 1, 2, and 5. Regarding the other points:

3. The sample is abstract to show the concept. The paging depends on the type of work that needs to be generated.
4. All generated work items are processed as fast as possible. If multiple work items are causing messages to generate, that'd cause massive queue backup which wouldn't happen with this pattern (since we only generate them gradually). E.g. Customers A and B and C generate 3 work tasks. If we generate 10,000 messages per customer, the first customer coming in would cause massive delays for customers B and C.
5. Correct. A feature such as a sequence diagram (if it could show the whole picture) or an activity view via open telemetry is a better fit for distributed systems to show / measure end-to-end time. 
6. Do we have a sample to point to as another alternative?

samples/saga/batching/Core_8/WorkProcessor/WorkProcessingHandler.cs
ramonsmits: Logical errors raised via exceptions would require a try..catch as those would be expected return/response states.

Alternatively, maybe just do 2 different replies not based on a catch but just an imaginary validation. Done in the same message but a different state?

samples/saga/batching/sample.md
```diff
@@ -0,0 +1,68 @@
...
+Processing work order '420'
+```
+
+Notice that while the work requests are generated in sequence, on the processing side the work can happen out of order. The total time it took the process to finish is also logged.
+
+NOTE: Metrics such as Critical Time may not be sufficient to monitor how long the end-to-end process takes. This is something to note and measure to ensure quality of service.
+
+## The Saga
+
+The saga controls the generation of the next batch of work and monitors for the completion signal of all messages in that batch, before starting the next batch.
```
ramonsmits, line 62:
```suggestion
The saga controls the generation of the next batch of work and monitors for the completion signal of all messages in that batch, before starting the next batch. Very long running work items or failed work items could unintentionally stall the subsequent work items. Consider using saga timeouts and or work item failure responses to mitigate this if required.
```

samples/saga/batching/sample.md
ramonsmits: I understand, but I do think its important to highlight issues that could cause harm if not dealt with during the design. I added a suggestion.

samples/saga/batching/sample.md
ramonsmits: > 6\. Do we have a sample to point to as another alternative?

We have a routing slip sample: https://docs.particular.net/samples/routing-slips/

samples/saga/batching/sample.md
```diff
@@ -0,0 +1,68 @@
...
+
+
+## Introduction
+
+This sample shows how to control the task (work) generation via batching of the work items through Sagas. This pattern of throttled message generation can be helpful when the the rate of generation needs to be controlled:
+
+- When lots of messages are generated as part of an incoming message, the amount of messages can further delay other messages in the same queue.
+- When the work as a whole is completed only when all messages are processed. If all messages are generated upfront and the work is considered 'cancelled' or 'failed', generated messages that are already in the queue would still have to be processed (e.g. no-op).
+
+Alternativly to avoid congesting the queue, messages can be generated and routed to a [specific endpoint](https://docs.particular.net/nservicebus/messaging/routing#make-instance-uniquely-addressable) (endpoint per work type), or [lower the concurrency rate](https://docs.particular.net/nservicebus/operations/tuning#configuring-concurrency-limit) to have the messages processed one by one.
```
ramonsmits, line 18: Concurrency limit is different from processing rate.
```suggestion
Alternatively to avoid congesting the queue, messages can be generated and routed to a [specific endpoint](https://docs.particular.net/nservicebus/messaging/routing#make-instance-uniquely-addressable) (endpoint per work type), or [lower the concurrency limit](https://docs.particular.net/nservicebus/operations/tuning#configuring-concurrency-limit) to have the messages processed sequentially.
```

samples/saga/batching/sample.md
```diff
@@ -12,10 +12,10 @@ related:
 
 This sample shows how to control the task (work) generation via batching of the work items through Sagas. This pattern of throttled message generation can be helpful when the the rate of generation needs to be controlled:
 
-- When you need to generate lots of messages as part of one operation, the amount of messages can further delay other messages in the same queue.
+- When lots of messages are generated as part of an incoming message, the amount of messages can further delay other messages in the same queue.
```
ramonsmits, line 15: This isn't really true as the total number of workitems is identical. I think the essence here is that this allows for interleaved processing of very large batches of work items

Maybe it even helps if this is visualised:

Sequential batches due to batches generated massive amounts workitems at start
 
| A       | B       | C       |
|---------|---------|---------|
| 100.000 |         |         |
|         | 100.000 |         |
|         |         | 100.000 |

Paging workitems via saga:

| A      | B      | C      |
|--------|--------|--------|
| 25.000 |        |        |
|        | 25.000 |        |
|        |        | 25.000 |
| 25.000 |        |        |
|        | 25.000 |        |
|        |        | 25.000 |
| 25.000 |        |        |
|        | 25.000 |        |
|        |        | 25.000 |
| 25.000 |        |        |
|        | 25.000 |        |
|        |        | 25.000 |

samples/saga/batching/Core_8/WorkProcessor/WorkProcessingHandler.cs
HEskandari: Used a status to report simulated errors back to the saga.

samples/saga/batching/sample.md
HEskandari: Thanks, added.

samples/saga/batching/sample.md
HEskandari: That's true. I think some visualization will help.

HEskandari: @ramonsmits added a mermaid diagram, please see if it captures this.

samples/saga/batching/Core_7/WorkProcessor/WorkProcessingHandler.cs
```diff
@@ -0,0 +1,46 @@
...
+            await context.Reply(new WorkOrderCompleted
+            {
+                ProcessId = message.ProcessId,
+                WorkOrderNo = message.WorkOrder,
+                Status = WorkStatus.Success
+            });
+        }
+        catch (Exception ex)
+        {
+            await context.Reply(new WorkOrderCompleted
```
ramonsmits, line 25: This would result in committing any existing transaction which is tricky. Likely its better to still throw but to do an immediate send for the reply and to throw an Exception that will cause the message to be either discarded or moved to the error queue.

Its pretty tricky.

HEskandari: The idea was that the receiver can decide to retry the task, continue with the rest of the tasks, or abort. This is a design / business decision and to me, depending on what you want to do, all three are valid.

PR #5725, opened 2022-05-12
SQL Server Transport multi instance sample with Bridge

dvdstelt:


samples/bridge/sql-multi-instance/Bridge_1/Bridge/App.config
```diff
@@ -0,0 +1,18 @@
+﻿<?xml version="1.0" encoding="utf-8"?><configuration>
+  <runtime>
+    <assemblyBinding xmlns="urn:schemas-microsoft-com:asm.v1">
```
andreasohlund, line 3: is this needed?

PR #5733, opened 2022-05-31
Add DotSettings to gitignore

dvdstelt:
There was a comment about .DotSettings in a bridge sample. I removed all 402 files and added the file to .gitignore.
No idea if the file is intentionally in docs, it seems weird that it would not. Maybe @DavidBoike knows why it's there?


adamralph (approved):

DavidBoike (changes requested):
I guess the Engine has never fully gotten on board with our switch from DotSettings to editorconfigs, as far as the generated sample downloads go. The process that creates the sample zip files still [adds the Shared.DotSettings to each solution file](https://github.com/Particular/DocsEngine/blob/master/src/ParticularDocsCore/SolutionZipping/DirectoryZipper.cs#L188-L195). We'll need to get rid of that first or we won't be able to deploy docs changes.

dvdstelt:
I'll close this for now.

PR #5742, opened 2022-06-16
Attempt targeting samples to net7.0-preview5

DavidBoike:


DavidBoike:
Analysis:

* The LINQPad script is bad with Functions, because Functions SDKs require certain framework versions and will blow up if you try to use a newer .NET version. I will do a PR that will ignore Functions projects for the purposes of bumping Docs sample versions.
* The sample (2 versions) in /samples/endpoint-configuration have a `ServiceHelper` class that does some PInvoke. One method returns `IntPtr` but in .NET 7 that's understood to be a `nint` which doesn't fly in C# 7.3 which the sample is pegged to. Not a big deal, this can be fixed when we actually update to .NET 7.
* Sample samples\pipeline\stream-properties\Core_8\PipelineStream.sln uses a `streamReader.ReadToEndAsync()` which in .NET 7 has an additional override that accepts a `CancellationToken`. Once we update to .NET 7 we can silence this with a pragma.

In conclusion, samples should be ready to support .NET 7.

PR #5760, opened 2022-07-04
OpenTelemetry customization sample

mikeminutillo:
Sample that shows how to customize OpenTelemetry traces with custom Activities, Processors, and adding tags to the existing Activities.


samples/open-telemetry/customizing/Core_8/CustomizingTelemetry/Sample/Behaviors/AddOrderIdToTrace.cs
```diff
@@ -0,0 +1,22 @@
...
+class AddOrderIdToTrace : Behavior<IIncomingLogicalMessageContext>
+{
+    public override Task Invoke(
+        IIncomingLogicalMessageContext context,
+        Func<Task> next
+    )
+    {
+        if(context.Message.Instance is ShipOrder shipOrder)
+        {
+            Activity.Current?.AddTag("sample.shipping.order_id", shipOrder.OrderId);
```
timbussmann, line 16: setting message specific properties which is much easier later in the pipeline in this behavior feels weird. Could we add some more generic information, maybe the body size (more appropriate in the physical stage but that should be easy to change)?

timbussmann (commented):
overall LGTM but it feels it could be reduced by showing one example (e.g. in a handler) on how to add custom tags or events and somewhere in the pipeline about how to start a custom span or something like that. In addition to the custom processor that should showcase all extensibility options well enough?

samples/open-telemetry/customizing/Core_8/CustomizingTelemetry/Sample/Sample.csproj
```diff
@@ -0,0 +1,27 @@
...
+    <TargetFramework>net6.0</TargetFramework>
+	<LangVersion>7.3</LangVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="OpenTelemetry" Version="1.3.0" />
+    <PackageReference Include="OpenTelemetry.Exporter.Console" Version="1.3.0" />
+
+	  <!-- Included to prevent a conflict with OpenTelemetry -->
+	<PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="6.0.0" />
```
HEskandari, line 14: Some spacing issues here (makes me sleep better at night knowing this is fixed)

samples/open-telemetry/customizing/sample.md
```diff
@@ -0,0 +1,94 @@
...
+
+snippet: custom-activity-source
+
+The handler for `CreateOrder` includes a custom activity that wraps around the billing section.
+
+snippet: custom-activity-in-handler
+
+This will automatically be created as a child activity of the invoke handler activity created by NServiceBus. The NServiceBus send message activity will treat this custom activity as it's parent.
+
+```
```
HEskandari, line 42: Are these mermaid diagrams? Wouldn't the snippet require `mermaid` after the backticks to render properly?

samples/open-telemetry/customizing/sample.md
```diff
@@ -0,0 +1,94 @@
...
+Send ShipOrder
+  Process ShipOrder <-- custom tag gets added here
+    Invoke ShipOrderHandler
+```
+
+A behavior in the outgoing pipeline adds an order id tag to the send message activity.
+
+snippet: add-tags-from-outgoing-behavior
+
+WARN: `Activity.Current` may be `null` if there is not configured trace listener. Always check if the value is null before calling methods on an `Activity` instance.
```
HEskandari, line 94: Reads a bit funny maybe?

```suggestion
WARN: `Activity.Current` may be `null` if there are no configured trace listeners. Always check if the value is null before calling methods on an `Activity` instance, or use the null-conditional operator (`?.`).
```

HEskandari (commented):
Some small comments

samples/open-telemetry/customizing/sample.md
mikeminutillo: They aren't mermaid diagrams. They just use the codeblock style to preserve spaces. I did mess with trying to get these to render with mermaid using the gannt view but ultimately gave it up as not being that much of an improvement in readibility, and a whole lot harder to maintain.

PR #5772, opened 2022-07-19
Suppress broken CA2252 warning

timbussmann:
The CA2252 code analyzer currently triggers on the `DocumentStore` APIs and also seems to fail with a NRE:

> CSC : error AD0001: Analyzer 'Microsoft.NetCore.CSharp.Analyzers.Runtime.CSharpDetectPreviewFeatureAnalyzer' threw an exception of type 'System.NullReferenceException' with message 'Object reference not set to an instance of an object.'.

Disables the CA2252 warning temporarily since it's breaking the builds and preventing deployments.

If merged, an issue should be raised to further investigate or validate the problem at a later point in time and to remove the workaround again.


timbussmann:
seems to have resolved itself magically somehow 🤷 🧙

PR #5775, opened 2022-07-20
add explicit note to state that SQL-T can only be used with Microsoft SQL Server

adamralph:


danielmarbach (approved):

transports/sql/index.md
```diff
@@ -15,6 +15,8 @@ related:
 
 The SQL Server transport implements a message queuing mechanism on top of [Microsoft SQL Server](https://www.microsoft.com/en-us/sql-server/). It provides support for sending messages using SQL Server tables. It does **not** make use of a [service broker](https://technet.microsoft.com/en-us/library/ms166104.aspx).
 
+NOTE: The SQL Server transport can only be used with [Microsoft SQL Server](https://www.microsoft.com/en-us/sql-server/). It cannot be used with any other database platform.
```
kbaley, line 18: I don't think this is necessary for the number of people that ask about it (the recent support case notwithstanding). But if I'm in the minority, how about saying it's only _supported_ on SQL Server instead of saying it can't be used elsewhere.

kbaley: I meant rather than explicitly saying "The SQL Server can't be used on other databases", how about:

```suggestion
NOTE: The SQL Server transport is supported only on [Microsoft SQL Server](https://www.microsoft.com/en-us/sql-server/). It is not supported on other database platforms.
```

dvdstelt: > I don't think this is necessary for the number of people that ask about it (the recent support case notwithstanding). But if I'm in the minority, how about saying it's only _supported_ on SQL Server instead of saying it can't be used elsewhere.

If you'd broken your first comment up in two comments, I could've just thumbed up the comment. Now I have to write that I agree with this statement. If we merge this PR, why not add to ASB that it can only be used by ASB and not with ASQ, because it has a different transport. And with RavenDb storage that it can't use any other document database or NoSql database. I've never in my life heard anyone ask if SQL Server meant something else than MICROSOFT Sql Server, until this case. If we merge this, what about the other 99% of the readers? Won't they think we're stating the obvious and question how we feel about our customers/readers?

HEskandari: I think @kbaley's suggestion makes sense. Even if it does run on some other dialect of SQL, that's not what we test for and that's not what we support.

adamralph:
It seems there's not enough evidence that we need this, so I'm closing.

PR #5791, opened 2022-07-26
Update RabbitMQ transport 8 documentation

bording:


transports/rabbitmq/connection-settings_add-cluster-node_rabbit_[8,).partial.md
```diff
@@ -0,0 +1,11 @@
+## Connecting to multiple cluster nodes
+
+The connection string for an endpoint allows one host name to be specified. When connecting to a RabbitMQ cluster, it can be useful for the endpoint to know about all of the nodes in the cluster. If a node goes down, the endpoint can attempt to reconnect to a different node.
```
kentdr, line 3:
```suggestion
When connecting to a RabbitMQ cluster, it is beneficial if endpoints are able to connect to any of the nodes in the cluster. For example, if a node goes down, the endpoint can attempt to reconnect to a different node and continue operation. 

Since endpoint connection strings are limited to specifying a single hostname, the `AddClusterNode` API can be used to tell the endpoint about additional cluster nodes:```

kentdr (approved):

PR #5808, opened 2022-08-08
Resolve receiver via context for reliable renewal operation in v2 of ASBS transport

ramonsmits:
Depends on:

- https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/521


DavidBoike:
Sorry I didn’t get to this today and need to look at it tomorrow. I also want to make sure that some misleading stuff (like the idea that you get to specify the lock duration) is gone from the sample as well.

ramonsmits:
> I also want to make sure that some misleading stuff (like the idea that you get to specify the lock duration) is gone from the sample as well.

The only way would then be to query the lock duration from the queue. Or maybe even better to just don't use that and use the LockedUntil property to calculate the next attempt but that is improving the sample instead of fixing and maybe better to do that in a separate PR?

samples/azure-service-bus-netstandard/lock-renewal/sample.md
```diff
@@ -45,11 +45,11 @@ The behavior processes every incoming message and uses [native message access](/
 
 snippet: native-message-access
 
-The request to renew the lock must use the same Azure Service Bus connection object and queue path used to receive the incoming message. These items are available in the `TransportTransaction`:
+The request to renew the lock must the message receiver must be obtained:
```
DavidBoike, line 48: This doesn't make any sense. I'm not sure what you were trying to do here.

samples/azure-service-bus-netstandard/lock-renewal/sample.md
```diff
@@ -45,11 +45,11 @@ The behavior processes every incoming message and uses [native message access](/
...
 
 snippet: get-connection-and-path
 
-With the lock token, connection object, and queue path, an Azure Service Bus `MessageReceiver` object may be created to renew the lock. This is done in a background task, running in an infinite loop until the specified cancellation token is signaled as canceled.
+In the background an infinite loop is started that will renew the lock until the specified cancellation token is signaled as canceled.
```
DavidBoike, line 52:
```suggestion
A background task is started that will renew the lock until the completion of message processing signals the cancellation token.
```

DavidBoike (changes requested):

DavidBoike:
@ramonsmits OK, I'm fine to do that change in another PR, but I left some comments and it looks like you need to release the 2.0.3 package to finish this off.

DavidBoike (approved):
Approved aside from that detail that it won't compile without the new package.

samples/azure-service-bus-netstandard/lock-renewal/ASBS_2/LockRenewal/LockRenewalBehavior.cs
```diff
@@ -4,15 +4,13 @@
 using Azure.Messaging.ServiceBus;
 using NServiceBus.Logging;
 using NServiceBus.Pipeline;
-using NServiceBus.Transport;
 
 class LockRenewalBehavior : Behavior<ITransportReceiveContext>
 {
-    public LockRenewalBehavior(TimeSpan lockDuration, TimeSpan renewLockTokenIn, string queueName)
+    public LockRenewalBehavior(TimeSpan lockDuration, TimeSpan renewLockTokenIn)
```
DavidBoike, line 10: Lock duration is preset on the queue to 5 minutes. This gives users the mistaken impression that they get to set lock duration on a per-message basis. Can't we just change this to:
```suggestion
    public LockRenewalBehavior(TimeSpan renewAfter)
```

Where `renewAfter` must be less than 5 minutes?

ramonsmits: I personally like that approach, that would in theory allow a renewal to fail and be retried too.

samples/azure-service-bus-netstandard/lock-renewal/sample_transactionmode_asbs_[1,2).partial.md
```diff
@@ -0,0 +1,2 @@
+1. The transport must use the default [`SendsAtomicWithReceive`](/transports/transactions.md#transactions-transport-transaction-sends-atomic-with-receive) transaction mode for the sample to work.
+1. When using lock renewal with the [outbox feature](/nservicebus/outbox/), the transport transaction mode has to be **explicitly** set to [`SendsAtomicWithReceive`](/transports/transactions.md#transactions-transport-transaction-sends-atomic-with-receive) in the endpoint configuration code.
```
DavidBoike, line 2: An empty partial file (below :point_down:) is not required.

samples/azure-service-bus-netstandard/lock-renewal/sample.md
```diff
@@ -25,38 +25,32 @@ include: asb-connectionstring-xplat
...
 
 snippet: renewal-background-task
 
 The behavior executes the rest of the pipeline and cancels the background task when execution is completed.
 
 snippet: processing-and-cancellation
 
+### Prefetching
+
+It is adviced to disable prefetching if the majority of message in the queue result in long processing durations. This reduced unnecessary lock expiration of prefetced messages.
```
DavidBoike, line 52:
```suggestion
Disable prefetching if the majority of messages in the queue result in long processing durations. Otherwise, the lock for prefetched messages could expire before processing even starts.
```

DavidBoike (approved):
Approved with a couple small suggestions.

PR #5810, opened 2022-08-08
Fix mermaid syntax

timbussmann:
The mermaid graphs on [this page](https://docs.particular.net/transports/sqs/delayed-delivery?version=sqs_6#how-it-works-examples) are broken. `subgraph` needs to have a name, e.g. `subgraph somethingsomething` but I'm not sure why it needs a subgraph at all, so I just removed that.

### DIFF
--- a/transports/sqs/delayed-delivery.md
+++ b/transports/sqs/delayed-delivery.md
@@ -128,19 +128,13 @@ sequenceDiagram
 ```mermaid
 graph LR
 
-subgraph 
-
 Sender .-> |DelaySeconds = 845sec| Destination
-
-end
 ```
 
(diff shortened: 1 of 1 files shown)


WilliamBZA (approved):

timbussmann:
well, apparently @mauroservienti already fixed them before I could merge. 🤷

mauroservienti:
Ooooops 😬 I was reviewing docs with @soujay as part of the non-critical support process and we noticed the page was broken.

Why were there subgraphs?

timbussmann:
> Why were there subgraphs?

no idea 🤷 I looked at the history and couldn't find a good reason for subgraphs other than visual representation.

PR #5815, opened 2022-08-17
Re-org nav IA to enable better menu UX

sergioc:
This only re-organizes the global nav menu to prevent sections from linking directly to pages. Now they should just expand or collapse a section. The exception is the Getting Started section, which from some reason doesn't work yet. This probably requires a docs engine update, and main sections (e.g. transports) also seem to need a docs engine update to stop linking.

### DIFF
--- a/nservicebus/tools/migrate-to-native-delivery.md
+++ b/nservicebus/tools/migrate-to-native-delivery.md
@@ -4,6 +4,8 @@ summary: An overview of the tool supporting migrating from timeout manager to na
 reviewed: 2020-06-22
 ---
 
+NOTE: Tools are designed for single or occasional use. Only the latest version of a given tool is supported. Users should consider upgrading to the latest version of a tool before each use.
+
 The timeout migration tool is designed to help system administrators migrate existing timeouts from the legacy [timeout manager](/nservicebus/messaging/timeout-manager.md) storage to the [native delayed delivery](/nservicebus/messaging/delayed-delivery.md) infrastructure of the currently used transport.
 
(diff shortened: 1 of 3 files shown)


nservicebus/tools/migrate-to-native-delivery.md
```diff
@@ -4,6 +4,8 @@ summary: An overview of the tool supporting migrating from timeout manager to na
 reviewed: 2020-06-22
 ---
 
+NOTE: Tools are designed for single or occasional use. Only the latest version of a given tool is supported. Users should consider upgrading to the latest version of a tool before each use.
+
```
DavidBoike, line 8: Not sure this was meant to be here?

DavidBoike:
So in order for this not to "devolve" it would require the Engine not allowing a Url on levels 1 and 2, right?

nservicebus/tools/migrate-to-native-delivery.md
sergioc: I copied that note from https://docs.particular.net/nservicebus/tools/, because this page only has a single tool listed and in order to not remove the need to have that page in the navigation.

sergioc:
> So in order for this not to "devolve" it would require the Engine not allowing a Url on levels 1 and 2, right?

Yes.

sergioc:
>> So in order for this not to "devolve" it would require the Engine not allowing a Url on levels 1 and 2, right?
>
>Yes.

And require at least a linked index page on 3rd level elements that contain a sub-section of 4th level links.

nservicebus/upgrades/azure-blobstorage-data-bus.md
```diff
@@ -0,0 +1,11 @@
+---
+title: Azure BlobStorage Data Bus upgrade guides
+reviewed: 2022-08-25
+suppressRelated: true
```
DavidBoike, line 4: I can see this pattern working with the evolving systems pages but here this seems like a bad idea. If I add a new upgrade guide I’m not going to know that I need to come here and update this index too. So it won’t get updated - and that’s why the index pages generate their own navigation based on directory traversal anyway.

To add to that, these upgrade guide sections have gotten really out of control anyway. It would probably be better to change these into the proper directory structures for each thing.

PR #5827, opened 2022-08-30
Throughput tool documentation

DavidBoike:


nservicebus/upgrades/throughput-tool.md
```diff
@@ -0,0 +1,143 @@
...
+
+Once these prerequisites are comlpete, execute the tool with the resource id of the Azure Service Bus namespace, as in this example:
+
+```shell
+particular-throughput-tool azureservicebus --resourceId /subscriptions/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/resourceGroups/my-resource-group/providers/Microsoft.ServiceBus/namespaces/my-asb-namespace --outputPath throughput-report.json
+```
+
+The tool may open additional terminal windows, which are Powershell processes gathering the data from the Azure CLI.
+
+Unlike other query options, using Azure Service Bus metrics allows the tool to capture the last 30 days worth of data, which means that results are returned instantly.
```
udidahan, line 79:
```suggestion
Unlike ServiceControl, using Azure Service Bus metrics allows the tool to capture the last 30 days worth of data at once, which means that report will be generated without delay.
```

The other options of RabbitMQ and SQLTransport are not going to be relevant to this audience.

nservicebus/upgrades/throughput-tool.md
```diff
@@ -0,0 +1,143 @@
...
+
+#### Options
+
+All options are required:
+
+| Option | Description |
+|-|-|
+| <nobr>`--resourceId`</nobr> | The resource ID of the Azure Service Bus namespace, which can be found in the Azure Portal as described above. |
+| <nobr>`--outputPath`</nobr> | A local path where the JSON report file should be generated. |
```
udidahan, line 88: As mentioned above for ServiceControl.

nservicebus/upgrades/throughput-tool.md
```diff
@@ -0,0 +1,143 @@
...
+The tool will prompt for the username and password to access the RabbitMQ management interface. After that, it will take its initial reading, then sleep for 24 hours before taking its final reading and generating a report.
+
+#### Options
+
+All options are required:
+
+| Option | Description |
+|-|-|
+| <nobr>`--apiUrl`</nobr> | The URL for the RabbitMQ management site. |
+| <nobr>`--outputPath`</nobr> | A local path where the JSON report file should be generated. |
```
udidahan, line 107: As mentioned above for ServiceControl.

nservicebus/upgrades/throughput-tool.md
```diff
@@ -0,0 +1,143 @@
...
+All options are required:
+
+| Option | Description |
+|-|-|
+| <nobr>`--resourceId`</nobr> | The resource ID of the Azure Service Bus namespace, which can be found in the Azure Portal as described above. |
+| <nobr>`--outputPath`</nobr> | A local path where the JSON report file should be generated. |
+
+### RabbitMQ
+
+Once installed, execute the tool with the RabbitMQ management URL, as in this example:
```
udidahan, line 92: Could we add a link to some docs showing them how to find the RabbitMQ management URL?

nservicebus/upgrades/throughput-tool.md
```diff
@@ -0,0 +1,143 @@
...
+
+```shell
+particular-throughput-tool rabbitmq --apiUrl http://localhost:15672 --outputPath throughput-report.json
+```
+
+The tool will prompt for the username and password to access the RabbitMQ management interface. After that, it will take its initial reading, then sleep for 24 hours before taking its final reading and generating a report.
```
udidahan, line 98: It may be better for the tool to write a number of readings (values + timestamps) to temporary files in case there is some kind of crash in those 24 hours.

nservicebus/upgrades/throughput-tool.md
```diff
@@ -0,0 +1,143 @@
...
+The tool will run for slightly longer than 24 hours in order to capture a beginning and ending `RowVersion` value for each queue table. A value can only be detected when a message is waiting in the queue to be processed, and not from an empty queue, so multiple queries may be required. The tool will use a backoff mechanism to avoid putting undue pressure on the SQL Server instance.
+
+#### Options
+
+All options are required:
+
+| Option | Description |
+|-|-|
+| <nobr>`--connectionString`</nobr> | The database connection string that will provide at least read access to all queue tables. |
+| <nobr>`--outputPath`</nobr> | A local path where the JSON report file should be generated. |
```
udidahan, line 126: As mentioned above for ServiceControl.

nservicebus/upgrades/throughput-tool.md
```diff
@@ -0,0 +1,143 @@
...
+All options are required:
+
+| Option | Description |
+|-|-|
+| <nobr>`--apiUrl`</nobr> | The URL for the RabbitMQ management site. |
+| <nobr>`--outputPath`</nobr> | A local path where the JSON report file should be generated. |
+
+### SQL Transport
+
+Once installed, execute the tool with the database connection string used by SQL Server endpoints, as in this example:
```
udidahan, line 111: Could we point them to some docs as to where they could find the connection string?

Also, is there any validation that is done to see that we're looking at the "right" database, like checking for the existence of queue tables.

nservicebus/upgrades/throughput-tool.md
```diff
@@ -0,0 +1,143 @@
...
+
+```shell
+particular-throughput-tool sqlserver --connectionString "Server=SERVER;Database=DATABASE;User=USERNAME;Password=PASSWORD;" --outputPath throughput-report.json
+```
+
+The tool will run for slightly longer than 24 hours in order to capture a beginning and ending `RowVersion` value for each queue table. A value can only be detected when a message is waiting in the queue to be processed, and not from an empty queue, so multiple queries may be required. The tool will use a backoff mechanism to avoid putting undue pressure on the SQL Server instance.
```
udidahan, line 117: As above, will data collected be written into temporary files to mitigate the risk of a crash?

nservicebus/upgrades/throughput-tool.md
```diff
@@ -0,0 +1,143 @@
...
+All options are required:
+
+| Option | Description |
+|-|-|
+| <nobr>`--connectionString`</nobr> | The database connection string that will provide at least read access to all queue tables. |
+| <nobr>`--outputPath`</nobr> | A local path where the JSON report file should be generated. |
+
+## Masking private data
+
+The report that is generated will contain the names of endpoints/queues. If the queue names themselves contain confidential or proprietary information, certain strings can be masked in the report file.
```
udidahan, line 130: Maybe we could make this functionality be something that can be run on an previously generated report file? I could totally see the case where somebody forgot to put this option in up-front, and then would need to wait another 24 hours to regenerate the report just to mask the queue names.

DavidBoike:
@udidahan I moved the file into a folder to have room to put screenshots, and that screwed up comments. I'll try to address them all here:

About `--outputPath`: Would you prefer to just write to `throughput-report.json` in the current directory and be done with it—so no output path parameter at all? That might simplify things. I don't currently do any write checks until it's time to write the file, which _could_ waste 24h of work.

About writing temporary files: Yes currently a crash would lose up to 24h of data in memory. But writing temporary files would add another level of complexity and I think we should talk about what would be involved and what tradeoffs we're willing to make for that in the short and long term.

> Could we point them to some docs as to where they could find the (SQL) connection string?

I don't know what to point to. Honestly I would be a little frightened of anyone who _doesn't_ know how to find this SQL connection string actually having access to it.

> Also, is there any validation that is done to see that we're looking at the "right" database, like checking for the existence of queue tables.

Well, you'd get an empty report. But you'd get that empty report 24 hours later. No other check currently exists. So yes, it would probably be a good idea to right away list out the queues that are detected and prompt if it's OK to continue. Does that sound good? That would probably be valuable for all transport cases.

> Maybe we could make this functionality be something that can be run on an previously generated report file?

That opens up a can of worms, but if we included the previously mentioned "These are the queues" confirmation with "This is how they'll be reported" would that take care of things?

udidahan:
> About --outputPath: Would you prefer to just write to throughput-report.json in the current directory and be done with it—so no output path parameter at all? 

Yes.

> I don't currently do any write checks until it's time to write the file, which could waste 24h of work.

Would it be so difficult to check at startup?

> About writing temporary files: Yes currently a crash would lose up to 24h of data in memory. But writing temporary files would add another level of complexity and I think we should talk about what would be involved and what tradeoffs we're willing to make for that in the short and long term.

OK. Agreed. No temporary files. At least, not for now.

> So yes, it would probably be a good idea to right away list out the queues that are detected and prompt if it's OK to continue. Does that sound good? 

Yes, it does.

> That would probably be valuable for all transport cases.

Agreed.

> if we included the previously mentioned "These are the queues" confirmation with "This is how they'll be reported" would that take care of things?

I think so 👍🏼

PR #5835, opened 2022-09-06
Added an integrity test for referenced partials

TravisNickels:
This PR adds an integrity test to check if partials that are referenced in markdown files have a corresponding partial file that can be used.  Below is the error list of all the partials that were referenced, but do not have a corresponding partial file that can be used.

image

### DIFF
--- a/tests/IntegrityTests/ValidateReferencedPartials.cs
+++ b/tests/IntegrityTests/ValidateReferencedPartials.cs
@@ -1,0 +1,67 @@
+﻿using Microsoft.VisualBasic;
+using NUnit.Framework;
+using NUnit.Framework.Internal;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+
+namespace IntegrityTests
+{
+    internal class ValidateReferencedPartials
+    {
+        [Test]
+        public void ReferencedPartialsAreUsed()
+        {           
+            var allErrors = new Dictionary<string, List<string>>();
+            var allPartials = Directory.GetFiles(TestSetup.DocsRootPath, "*.partial.md", SearchOption.AllDirectories);
+            var allArticles = Directory.GetFiles(TestSetup.DocsRootPath, "*.md", SearchOption.AllDirectories).Except(allPartials);
+
+            foreach (var articlePath in allArticles)
+            {
+                //Skips the readme file because it references a partial as instructions
+                if (articlePath.Contains("README.md"))
+                {
+                    continue;
+                }
+
+                var fileName = Path.GetFileNameWithoutExtension(articlePath);
+                string text = File.ReadAllText(articlePath);
+                string searchPattern = @"(?i)partial:\s*[a-zA-Z0-9-_]*";
+                string replacePattern = @"(?i)partial:\s*";
+                var matches = Regex.Matches(text, searchPattern, RegexOptions.IgnoreCase);
+                var errors = new List<string>();
+
+                foreach (Match match in matches)
+                {
+                    var partialName = Regex.Replace(match.Value, replacePattern, "").Trim();
+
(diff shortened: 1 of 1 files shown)


tests/IntegrityTests/ValidateReferencedPartials.cs
```diff
@@ -0,0 +1,67 @@
...
+                if (errors.Count > 0)
+                {
+                    allErrors.TryAdd(articlePath, errors);
+                }
+            }
+
+            if (allErrors.Count > 0)
+            {
+                StringBuilder stringBuilder = new StringBuilder();
+                stringBuilder.Append($"There is ({allErrors.Count}) markdown file(s) that had references to partials that do not have a corresponding partial file to be used. Here are the file paths and partial references that were not used.\r\n");
```
DavidBoike, line 57: This gets you out of "is 1" vs. "are (>1)" disagreement.
```suggestion
                stringBuilder.Append($"Found ({allErrors.Count}) markdown file(s) that had references to partials that do not have a corresponding partial file to be used. Here are the file paths and partial references that were not used.\r\n");
```

tests/IntegrityTests/ValidateReferencedPartials.cs
```diff
@@ -0,0 +1,67 @@
...
+                if (articlePath.Contains("README.md"))
+                {
+                    continue;
+                }
+
+                var fileName = Path.GetFileNameWithoutExtension(articlePath);
+                string text = File.ReadAllText(articlePath);
+                string searchPattern = @"(?i)partial:\s*[a-zA-Z0-9-_]*";
+                string replacePattern = @"(?i)partial:\s*";
+                var matches = Regex.Matches(text, searchPattern, RegexOptions.IgnoreCase);
```
DavidBoike, line 35: This makes sure `partial` must occur at the beginning of a line, to guard against the (probably rare) case that `partial: ` would occur in the middle of valid text.

Also `_` can't be in a partial name since that's the filename delimiter.
```suggestion
                string searchPattern = @"^(?i)partial:\s*[a-zA-Z0-9-]*";
                string replacePattern = @"(?i)partial:\s*";
                var matches = Regex.Matches(text, searchPattern, RegexOptions.IgnoreCase | RegexOptions.Multiline);
```

tests/IntegrityTests/ValidateReferencedPartials.cs
```diff
@@ -0,0 +1,67 @@
...
+            }
+
+            if (allErrors.Count > 0)
+            {
+                StringBuilder stringBuilder = new StringBuilder();
+                stringBuilder.Append($"There is ({allErrors.Count}) markdown file(s) that had references to partials that do not have a corresponding partial file to be used. Here are the file paths and partial references that were not used.\r\n");
+                foreach (var error in allErrors)
+                {
+                    stringBuilder.Append($"\r\n  > {error.Key.ToString()}\r\n");
+                    stringBuilder.Append($"{string.Join($"\r\n", error.Value)}");
```
DavidBoike, line 61:
```suggestion
                    stringBuilder.AppendLine()
                        .AppendLine($"  > {error.Key.ToString()}")
                        .Append($"{string.Join(Environment.NewLine, error.Value)}");
```

DavidBoike (approved):

tests/IntegrityTests/ValidateReferencedPartials.cs
TravisNickels: Haha...Thanks I wasn't sure about what I had there either.  Your suggestion works way better.

tests/IntegrityTests/ValidateReferencedPartials.cs
TravisNickels: I like this suggestion, but I changed line 62 to AppendLine() as well and I think it makes the error message much easier to read.  What do you think?
image

DavidBoike: For sure, I was just sticking to your output in my suggestion. 👍

DavidBoike (approved):

PR #5842, opened 2022-09-13
Bump packages - master

HEskandari:


Snippets/Core/Core_7.7/Core_7.7.csproj
```diff
@@ -6,7 +6,7 @@
     <LangVersion>7.3</LangVersion>
   </PropertyGroup>
   <ItemGroup>
-    <PackageReference Include="NServiceBus" Version="7.7.2" />
+    <PackageReference Include="NServiceBus" Version="7.7.0" />
```
ramonsmits, line 9: This seems to be an exact version. Those can be set in brackets
```suggestion
    <PackageReference Include="NServiceBus" Version="[7.7.0]" />
```

See https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#version-ranges

samples/elsa/Core_7/Messages/Messages.csproj
```diff
@@ -6,7 +6,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="NServiceBus" Version="7.6.0" />
+    <PackageReference Include="NServiceBus" Version="[7.6.0, 8.0.0)" />
```
ramonsmits, line 9: Why isn't this 7.7.4?

ramonsmits (changes requested):

samples/elsa/Core_7/Messages/Messages.csproj
HEskandari: Think I manually undid it, but kep the range version. I'll update to `7.7.0`. We can decide to use the bracket format e.g. `[7.7.0]` in a separate PR (as we need to do it consistently)

danielmarbach: FYI we never used the bracket format in docs so far

Snippets/Core/Core_7.7/Core_7.7.csproj
danielmarbach: FYI we have done a silent release to myget using 7.8. Our goal is to release this probably mid next week

HEskandari: I rolled back the change to `7.7.2`. I think the change to how we specify exact versions should be done consistently across the board.

samples/elsa/Core_7/Messages/Messages.csproj
HEskandari: Same as above.

timbussmann:
why are you using version ranges for the docs @HEskandari ?

HEskandari:
@timbussmann It is something the newly created automated script generates.

timbussmann:
it seems that script doesn't follow the [package reference guidelines](https://github.com/Particular/docs.particular.net#nuget-package-references) and should be fixed?

PR #5846, opened 2022-09-14
Update MongoDB documentation for IMongoSynchronizedStorageSession

timbussmann:
With https://github.com/Particular/NServiceBus.Storage.MongoDB/issues/328, DI support for the shared session has been added to MongoDB. This PR adds additional snippets to show usage of the the interface via DI.

Does some namespace cleanup along the way.

### DIFF
--- a/Snippets/mongodb/MongoDB_2/SharedTransactionDI.cs
+++ b/Snippets/mongodb/MongoDB_2/SharedTransactionDI.cs
@@ -1,0 +1,33 @@
+﻿using System.Threading.Tasks;
+using NServiceBus.Storage.MongoDB;
+
+namespace MongoDB_2
+{
+    public class SharedTransactionDI
+    {
+        #region MongoDBSharedTransactionDI
+        class MyService
+        {
+            IMongoSynchronizedStorageSession sharedSession;
+
+            // Resolved from DI container
+            public MyService(IMongoSynchronizedStorageSession sharedSession)
+            {
+                this.sharedSession = sharedSession;
+            }
+
+            public Task Create()
+            {
+                return sharedSession.MongoSession.Client
+                    .GetDatabase("mydatabase")
+                    .GetCollection<MyBusinessObject>("mycollection")
+                    .InsertOneAsync(sharedSession.MongoSession, new MyBusinessObject());
+            }
+        }
+        #endregion
+
+        class MyBusinessObject
+        {
+        }
+    }
+}
--- a/Snippets/mongodb/MongoDB_2/SharedTransactionDI.cs
+++ b/Snippets/mongodb/MongoDB_2/SharedTransactionDI.cs
@@ -1,8 +1,8 @@
-﻿using System.Threading.Tasks;
-using NServiceBus.Storage.MongoDB;
-
-namespace MongoDB_2
+﻿namespace MongoDB_2
 {
+    using System.Threading.Tasks;
+    using NServiceBus.Storage.MongoDB;
+
     public class SharedTransactionDI
     {
         #region MongoDBSharedTransactionDI
--- a/Snippets/mongodb/MongoDB_3/DocumentVersion.cs
+++ b/Snippets/mongodb/MongoDB_3/DocumentVersion.cs
@@ -16,7 +16,7 @@
             //Define other update operations on the document
 
             var modifiedDocument = await collection.FindOneAndUpdateAsync<BsonDocument>(
-                filter: document => document["_id"] == documentId && document["_version"] == currentVersion,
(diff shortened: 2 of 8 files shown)


Snippets/mongodb/MongoDB_3/DocumentVersion.cs
```diff
@@ -16,7 +16,7 @@ async Task UpdateWithVersion(IMongoCollection<BsonDocument> collection, string v
             //Define other update operations on the document
 
             var modifiedDocument = await collection.FindOneAndUpdateAsync<BsonDocument>(
-                filter: document => document["_id"] == documentId && document["_version"] == currentVersion,
+                filter: document => document["_id"].AsGuid == documentId && document["_version"] == currentVersion,
```
danielmarbach, line 19: Is this working? See https://github.com/Particular/NServiceBus.Storage.MongoDB/issues/108 
I might be confusing things, though. Just double-checking

persistence/mongodb/index.md
```diff
@@ -63,14 +63,22 @@ Note that this disables the ability to use pessimistic locking for sagas which m
...
-NServiceBus supports sharing MongoDB transactions between Saga persistence and business data. The shared transaction can be used to persist document updates for both concerns atomically.
+NServiceBus supports sharing MongoDB sessions between Saga persistence, Outbox storage, and business data. The shared session can be used to persist multiple document updates atomically.
 
 To use the shared transaction in a message handler:
 
 snippet: MongoDBHandlerSharedTransaction
 
 WARN: In order to participate in the shared transaction the MongoDB session must be passed into collection API calls as demonstrated above.
 
+The shared session can also be accessed via dependency injection using the `IMongoSynchronizedStorageSession` interface:
```
danielmarbach, line 74: If I recall correctly we are usually not doing versioned partials for minors right?

Snippets/mongodb/MongoDB_3/SharedTransactionDI.cs
```diff
@@ -0,0 +1,33 @@
+﻿namespace MongoDB_3
+{
+    using System.Threading.Tasks;
+    using NServiceBus.Storage.MongoDB;
+
+    public class SharedTransactionDI
```
danielmarbach, line 6: What's the value add of that outer class?

Snippets/mongodb/MongoDB_2/SharedTransactionDI.cs
```diff
@@ -0,0 +1,33 @@
+﻿namespace MongoDB_2
+{
+    using System.Threading.Tasks;
+    using NServiceBus.Storage.MongoDB;
+
+    public class SharedTransactionDI
```
danielmarbach, line 6: What's the value add of that outer class?

danielmarbach (approved):

Snippets/mongodb/MongoDB_3/DocumentVersion.cs
timbussmann: might depend on the specific configuration and the client being used tbh. This is just a sample, alternatives would be to:
* replace `documentId` with `BsonBinaryData(guid, GuidRepresentation.X)` (where X also depends)
* change the sample to not use a GUID as the id
thoughts?

persistence/mongodb/index.md
timbussmann: not sure there is a clear guidance, I think I've seen both. One option would be to just state the minim version (e.g. "Starting from version ..."), use partials or just leave as is since it will become the only supported version anyway.

Snippets/mongodb/MongoDB_2/SharedTransactionDI.cs
timbussmann: namespacing the the types (e.g. `MyBusinessObject` which is also used in the other snippet)

SzymonPobiega (approved):

tmasternak (approved):
No other comments than these from Daniel

persistence/mongodb/index.md
danielmarbach: I have several times applied the easy route and just mentioned latest for minor changes. Otherwise the documentation pages become super complex to maintain and at the end of the day rolling forward to a new minor should be fairly straight forward for customers

Snippets/mongodb/MongoDB_3/DocumentVersion.cs
danielmarbach: I struggle to answer this because I'm lacking client knowledge. Maybe we don't have to overthink it and just leave it as is

PR #5858, opened 2022-09-20
Remove ranged version

HEskandari:


timbussmann (approved):
according to the guidelines, samples should use wildcard dependencies to make sure they are always ont he latest release. But for pre-release versions I think we've mostly targeted a specific version, we just have to remember to switch it to the wildcard dependency once v8 rtm packages are out.

danielmarbach (approved):
Not sure I follow the previous comment from Ramon. This is definitely a good change to have and brings back samples to where they should be according to the guidance

awright18:
The issue is with non-prereleae package versions. They should be using
wildcard version numbers.

You are correct that ranges shouldn't be used.

I'm working on an update to this branch to address this.

danielmarbach:
I approved this as it addresses removing the range version, as the title suggests, and as such it is self-contained and mergeable. A subsequent update could easily deal with bringing back the widest wildcard possible (doing that as well as introducing a potential test could be seen as a dedicated PR or like you suggest doing it on the same PR)

awright18:
@danielmarbach If it gets merged first that's fine.  Updating wildcards breaks some samples anyway. I'm working through that now.  Thanks

PR #5859, opened 2022-09-20
Improve wording - instead of "ghost" use "tainted" messages

udidahan:
This is for the TransactionSession and Outbox documentation.

### DIFF
--- a/nservicebus/outbox/index.md
+++ b/nservicebus/outbox/index.md
@@ -18,7 +18,7 @@ related:
 - persistence/mongodb
 ---
 
-Most message queues, and some data stores, do not support distributed transactions. This may cause problems when message handlers modify business data. Business data and messages may become inconsistent in the form of **phantom records** or **ghost messages** (see below).
+Most message queues, and some data stores, do not support distributed transactions. This may cause problems when message handlers modify business data. Business data and messages may become inconsistent in the form of **phantom records** or **tainted messages** (see below).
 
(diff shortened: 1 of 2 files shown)


udidahan:
The word "tainted" is a more [accurate](https://github.com/Particular/Strategy/blob/master/org-philosophy/organizational-philosophy.md#practice-prefer-using-specific-rather-than-general-terminology) description of what is the problem than the word "ghost".

I'd agree that the documentation should be consistent with what we have for the Outbox. I'll add a commit to this PR to update that documentation as well.

ramonsmits:
As a non native speaker I never heard of the word tainted and had to look it up:

> contaminate or pollute (something).

Maybe better to use corrupted or or unaligned message instead?

danielmarbach:
We have discussed this and came to the following conclusions:

The term "ghost" doesn't do a good job at conveying the nature of the problem.  We believe "tainted" does a much better job:
https://www.merriam-webster.com/dictionary/tainted
- contaminated by or as if by dirt
- damaged or spoiled by alteration from a correct or original state

We will also switch from "phantom records" to "orphaned records" since it is also a more accurate term describing what happens and therefore lessens the mental burden of remembering the concepts. This also applied even though the term orphaned records is more associated with relational database than NoSQL databases.

https://database.guide/what-is-an-orphaned-record/

We are currently reworking parts of the transactional session documentation that would conflict with this PR, so we are closing this one and will be following up with a dedicated PR aligning both terms in one go (including all references to it). 

Thanks for bringing it to our attention @udidahan

PR #5863, opened 2022-09-21
update docs about the ranged version use

HEskandari:


README.md
```diff
@@ -881,6 +881,10 @@ _**This is a temporary state and once a stable is released it is changed back to
 <PackageReference Include="NServiceBus.RabbitMQ" Version="4.*" />
 ```
 
+#### Version ranges
+
+Avoid using [version ranges](https://learn.microsoft.com/en-us/nuget/concepts/package-versioning#version-ranges) for package references. The integrity tests will enforce this.
```
ramonsmits, line 886: How is a range different from a wildcard? Then this should be avoid ranges and wildcards.

ramonsmits: So, we are only allowed do so a wildcard on the patch? So `1.1.*` but not `1.*` and not allowed to do `1-*` or `1.0.0-*`. 

That makes me wonder if `1.0.0-rc.*` is allowed. Assuming package restore using the highest version.

danielmarbach: @ramonsmits the wording is a bit fluffy because it depends, but this section clarifies that https://github.com/Particular/docs.particular.net#nuget-package-references

README.md
```diff
@@ -871,7 +871,7 @@ When documenting an unstable feature, those unstable packages must be explicitly
 <PackageReference Include="NServiceBus.RabbitMQ" Version="4.3.1-alpha.XYZ" />
 ```
 
-Wildcard patterns, like `Version="4.3.1-*` are not allowed for pre-releases since they tend to break the snippet and sample builds.
+Wildcard patterns, like `4.3.1-*`, or `8.0.0-alphe/beta/rc.*` are not allowed for pre-releases since they tend to break the snippet and sample builds.
```
ramonsmits, line 874: @danielmarbach It was not clear if that was allowed so I clarified by explicitly included prerelease names.

README.md
```diff
@@ -871,7 +871,7 @@ When documenting an unstable feature, those unstable packages must be explicitly
 <PackageReference Include="NServiceBus.RabbitMQ" Version="4.3.1-alpha.XYZ" />
 ```
 
-Wildcard patterns, like `4.3.1-*`, or `8.0.0-alphe/beta/rc.*` are not allowed for pre-releases since they tend to break the snippet and sample builds.
+Wildcard patterns, like `4.3.1-*`, `8.0.0-alpha.*`, `8.0.0-beta.*`, or  `8.0.0-rc.*` are not allowed for pre-releases since they tend to break the snippet and sample builds.
```
danielmarbach, line 874:
```suggestion
Wildcard patterns, like `4.3.1-*`, `8.0.0-any-pre-release-version.*` are not allowed for pre-releases since they tend to break the snippet and sample builds.
```

SemVer calls alpha, beta, rc etc. simply pre-release-version if I'm not mistaken https://semver.org/

danielmarbach (approved):

danielmarbach:
This is wonderful :heart_eyes:

PR #5866, opened 2022-09-21
Fix the folders in the TransactionalSession samples

SzymonPobiega:


samples/transactional-session/aspnetcore-webapi/SqlPersistenceTS_6/WebApplication/MyHandler.cs
```diff
@@ -19,8 +19,7 @@ public async Task Handle(MyMessage message, IMessageHandlerContext context)
     {
         log.Info("Message received at endpoint");
 
-        var entity = await dataContext.MyEntities.Where(e => e.Id == message.EntityId)
-           .FirstAsync(cancellationToken: context.CancellationToken);
+        var entity = await dataContext.MyEntities.Where(e => e.Id == message.EntityId).FirstAsync();
```
danielmarbach, line 22:
```suggestion
        var entity = await dataContext.MyEntities.Where(e => e.Id == message.EntityId)
           .FirstAsync();
```

Not hung up on this. It would be just more "consistent" with the v7 version

samples/transactional-session/aspnetcore-webapi/SqlPersistenceTS_6/WebApplication/Program.cs
```diff
@@ -5,12 +5,13 @@
...
 using Microsoft.Extensions.Hosting;
 using NServiceBus;
+using NServiceBus.Persistence;
 using NServiceBus.Persistence.Sql;
 using NServiceBus.TransactionalSession;
 
 public class Program
 {
-    const string ConnectionString = @"Server=(localdb)\MSSQLLocalDB;Integrated Security=True;";
+    private const string ConnectionString = @"Server=(localdb)\MSSQLLocalDB;Integrated Security=True;";
```
danielmarbach, line 14: So do we have different analyzers here now? Because the other uses const string

SzymonPobiega, line 14:
```suggestion
    const string ConnectionString = @"Server=(localdb)\MSSQLLocalDB;Integrated Security=True;";
```

danielmarbach (approved):

PR #5871, opened 2022-09-21
Test if the alias version matches the corresponding package version.

ramonsmits:


DavidBoike:
So, the point is to detect (for example) if there’s a Core_7 directory that references an NServiceBus version that’s not 6.something?

If so, cool. Just have to fix all the problems to merge it.

ramonsmits:
> So, the point is to detect (for example) if there’s a Core_7 directory that references an NServiceBus version that’s not 6.something?

That is not 7 something, but I think that is a typo.

Yes, most are already addressed in another PR. I'll make this PR dependant on that.

PR #5872, opened 2022-09-22
Remove DependencyVersion mode HighestMinor in nuget.cofig

ramonsmits:


bording:
@ramonsmits that nuget.config setting is effectively obsolete. It has no effect on package reference version resolution.

So, whatever you were hoping to avoid by using that setting, it's not going to work.

ramonsmits:
> @ramonsmits that nuget.config setting is effectively obsolete. It has no effect on package reference version resolution.
> 
> So, whatever you were hoping to avoid by using that setting, it's not going to work.

@bording That nuget.config is already there. If that is not responsible for `DependencyVersion` mode `HighestMinor` then what is? If we remove `nuget.config` then package restore would pull the minimum versions of packages right?

bording:
> > @ramonsmits that nuget.config setting is effectively obsolete. It has no effect on package reference version resolution.
> > So, whatever you were hoping to avoid by using that setting, it's not going to work.
> 
> @bording That nuget.config is already there. If that is not responsible for `DependencyVersion` mode `HighestMinor` then what is? If we remove `nuget.config` then package restore would pull the minimum versions of packages right?

In the world of package references, nuget.config has one real purpose: adding package sources. All of the other settings in there are ignored and have no impact on how package references work and how version resolution works. (not 100% true, but the other settings I'm thinking of are new things that have been added recently).

>. If that is not responsible for DependencyVersion mode HighestMinor then what is?

Using wildcards in the package reference is the only thing that causes it to use the latest version of something that matches the wildcard. Otherwise, NuGet will always resolve the lowest version that matches the versions specified in the package reference.

>  If we remove nuget.config then package restore would pull the minimum versions of packages right?

Since we're pretty much using package references everywhere, removing those will have no effect whatsoever. They are already being ignored.

In fact, we should probably just make the docs nuget.config match the one in [RepoStandards](https://github.com/Particular/RepoStandards/blob/master/nuget.config).

ramonsmits:
@bording Thanks for that clarification. I assumed wildcard behavior would be the same as `[6,7)` where the restore operation uses 6.0.0 but 

> When using the PackageReference format, NuGet also supports using a floating notation, *, for Major, Minor, Patch, and pre-release suffix parts of the number. When a floating version is specified, the rule is to resolve to the highest existent version that matches the version description.

Source: https://learn.microsoft.com/en-us/nuget/concepts/package-versioning


I'll update the section with your info Tomorrow and I'll add the modified nuget.config to that too.

README.md
```diff
@@ -801,6 +801,8 @@ The code used by snippets and samples is compiled on the build server. The compi
 
 ### NuGet package references
 
+Docs assumes all dependencies follow SemVer and uses [NuGet DependencyVersion mode `HighestMinor` (`~/nuget.config`)](https://learn.microsoft.com/en-us/nuget/reference/nuget-config-file#config-section)
```
ramonsmits, line 804:
```suggestion
Docs assumes all dependencies follow SemVer.
```

README.md
```diff
@@ -813,7 +815,7 @@ This applies to snippets and samples.
 
 #### Sometimes current patch
 
-In some cases, usually where a package has significant new API in a minor, it may be necessary to version snippets down to the "current patch".
+Due to DependencyVersion mode `HighestMinor` it is not needed to pin to a minor unless the referenced package is not following SemVer. It may be necessary to version snippets down to the "current patch".
```
ramonsmits, line 818:
```suggestion
In some cases, usually where a package has significant new API in a minor, it may be necessary to version snippets down to the "current patch".
```

README.md
```diff
@@ -801,6 +801,8 @@ The code used by snippets and samples is compiled on the build server. The compi
 
 ### NuGet package references
 
+Docs assumes all dependencies follow SemVer.
```
bording, line 804: I don't think this is true, or something that logically follows from the guidelines below. I'd just remove this change.

bording (changes requested):
At this point, I think the only thing that this PR should include is the change to the nuget.config.

bording (approved):

PR #5877, opened 2022-09-27
Clarify CriticalErrorContext.Stop behavior

timbussmann:
be more clear that this does never impact the host process. Also adds a snippet for the Stop method.

### DIFF
--- a/nservicebus/hosting/critical-errors.md
+++ b/nservicebus/hosting/critical-errors.md
@@ -31,11 +31,13 @@ Define a custom handler using the following code.
 
 snippet: DefiningCustomHostErrorHandlingAction
 
-### Terminating the process
+### Stopping the endpoint
 
-It is often unknown if the issue is recoverable when a critical error occurs. A sound strategy is to terminate the process when a critical error occurs and rely on the process hosting environment to restart the process as a recovery mechanism, resulting in a resilient way to deal with critical errors.
+It is often unknown if the issue is recoverable when a critical error occurs. A sound strategy is to terminate the process (e.g. via `Environment.FailFast` or `IHostApplicationLifetime.Stop`) when a critical error occurs and rely on the process hosting environment to restart the process as a recovery mechanism, resulting in a resilient way to deal with critical errors.
 
-However, this strategy only works when the endpoint instance is hosted in isolation and does not have another component. For example, when co-hosting NServiceBus with a web-service or website, terminating the process would result in these components becoming unavailable to users or other systems.
+However, this strategy only works when the endpoint instance is hosted in isolation and does not have another component. For example, when co-hosting NServiceBus with a web-service or website, terminating the process would result in these components becoming unavailable to users or other systems. When the host process should not be affected, the NServiceBus endpoint can be stopped without affecting other components or the host process:
+
+snippet: StopEndpointInCriticalError
 
 ### Host OS recoverability
 
--- a/nservicebus/hosting/critical-errors.md
+++ b/nservicebus/hosting/critical-errors.md
@@ -35,7 +35,7 @@ snippet: DefiningCustomHostErrorHandlingAction
 
 It is often unknown if the issue is recoverable when a critical error occurs. A sound strategy is to terminate the process (e.g. via `Environment.FailFast` or `IHostApplicationLifetime.Stop`) when a critical error occurs and rely on the process hosting environment to restart the process as a recovery mechanism, resulting in a resilient way to deal with critical errors.
 
-Before terminating the process, the NServiceBus endpoint can attempt a graceful shutdown:
+Before terminating the process, the NServiceBus endpoint can attempt a graceful shutdown which can be useful in non-transactional processing environments:
 
 snippet: StopEndpointInCriticalError
 
--- a/nservicebus/hosting/critical-errors.md
+++ b/nservicebus/hosting/critical-errors.md
@@ -35,7 +35,9 @@ snippet: DefiningCustomHostErrorHandlingAction
 
 It is often unknown if the issue is recoverable when a critical error occurs. A sound strategy is to terminate the process (e.g. via `Environment.FailFast` or `IHostApplicationLifetime.Stop`) when a critical error occurs and rely on the process hosting environment to restart the process as a recovery mechanism, resulting in a resilient way to deal with critical errors.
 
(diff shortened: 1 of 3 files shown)


nservicebus/hosting/critical-errors.md
```diff
@@ -31,11 +31,13 @@ Define a custom handler using the following code.
 
 snippet: DefiningCustomHostErrorHandlingAction
 
-### Terminating the process
+### Stopping the endpoint
```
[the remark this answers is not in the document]
ramonsmits: Maybe we can extend that with:

Windows Service -> FailFast
Linux Daemon -> Failfast
Dockter container -> Failfast

But all with recovery configured.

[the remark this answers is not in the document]
timbussmann: this might be something that would make more sense in dedicated documentation for the common hosting options, e.g. separate pages for windows service and linux service hosting and configuration. E.g. different cloud service might also require different behavior.

ramonsmits (changes requested):
Re-reading this I disagree with the added guidance. When passing a critical error callback you do that only for one reason and that is host termination.

Only stopping should never happen as this would disable any transport recovery mechanism and would be a very bad implementation if that is the only task done in the callback.

The only decision to then make is whether the user is allowing ungraceful termination (via `Environment.FailFast`) or are is going to allow graceful termination by signaling the host (for example via `IHostApplicationLifetime.Stop`) with a maybe a timeout to ensure termination if the host will not have its own timeout mechanism to terminate.

That is potential guidance we can add to this page but the current added content is what IMHO a user must never do.

timbussmann:
> Re-reading this I disagree with the added guidance. When passing a critical error callback you do that only for one reason and that is host termination.

I don't think that's true. There can be any range of custom logging/notifications involved to stopping the endpoint while keeping the webserver alive. There seems to have been a reason why `context.Stop` has been added (didn't exist initially).

> Only stopping should never happen as this would disable any transport recovery mechanism and would be a very bad implementation if that is the only task done in the callback.

not sure what this has to do with the transport recoverabilty mechanism? Can you further explain?

ramonsmits:
> There seems to have been a reason why `context.Stop` has been added (didn't exist initially).

To allow for a (I assume) graceful shutdown to happen as FailFast terminates. This way you can gracefully complete processing (if at all possible). That still means that users must trigger the host to stop.

I really don't see a reason why else you would stop the endpoint on critical error.

cquirosj (approved):
I agree with the fact that all this PR is doing is clarifying what criticalErrorContext currently does. Whether this method should exist or not and whether or not it makes sense I think that is a whole other conversation.

ramonsmits:
> Whether this method should exist or not and whether or not it makes sense, I think is a whole other conversation.

That is not the discussion. 

I have problems with:

1. That the title of that section got changed to "Stopping the endpoint"
2. The location of the snippet

The only reason to implement the critical error callback without stopping the message pump or not stopping the host is for alerting capabilities (*To notify support personnel when the endpoint has raised a critical error.*).

The implementation should basically never do `Stop` and not trigger the host to terminate or "rebuild".

The current location of the snippet seems to imply that just calling Stop is a wise thing to do (almost to be the default...) which is not the case. It should almost never be done when not terminating.

I'm fine with a section that explains that with v7+ we have a method that can signal the message pump to explicitly stop but with a large NOTE that stopping without signaling the host is a bad idea.

nservicebus/hosting/critical-errors.md
```diff
@@ -31,11 +31,18 @@ Define a custom handler using the following code.
...
 
-It is often unknown if the issue is recoverable when a critical error occurs. A sound strategy is to terminate the process when a critical error occurs and rely on the process hosting environment to restart the process as a recovery mechanism, resulting in a resilient way to deal with critical errors.
+It is often unknown if the issue is recoverable when a critical error occurs. A sound strategy is to terminate the process (e.g. via `Environment.FailFast` or `IHostApplicationLifetime.Stop`) when a critical error occurs and rely on the process hosting environment to restart the process as a recovery mechanism, resulting in a resilient way to deal with critical errors.
+
+Before terminating the process, the NServiceBus endpoint can attempt a graceful shutdown:
+
+snippet: StopEndpointInCriticalError
+
+Note: [The Microsoft Generic Host's](/nservicebus/hosting/extensions-hosting.md) `IHostApplicationLifetime.Stop` method also stops the NServiceBus endpoint gracefully.
```
ramonsmits, line 42:
```suggestion
Note: When using [NServiceBus.Extensions.Hosting](/nservicebus/hosting/extensions-hosting.md) the Microsoft Generic Host's `IHostApplicationLifetime.Stop` method also stops the NServiceBus endpoint gracefully.
```

ramonsmits: Also not sure if this should be here. Maybe its better to add this to https://docs.particular.net/nservicebus/hosting/extensions-hosting and link to it as something like. Please review hosting packages for behavior on critical errors and stop and/or shutdown behavior?

nservicebus/hosting/critical-errors.md
```diff
@@ -31,11 +31,18 @@ Define a custom handler using the following code.
 
 snippet: DefiningCustomHostErrorHandlingAction
 
-### Terminating the process
+### Stopping the endpoint
 
-It is often unknown if the issue is recoverable when a critical error occurs. A sound strategy is to terminate the process when a critical error occurs and rely on the process hosting environment to restart the process as a recovery mechanism, resulting in a resilient way to deal with critical errors.
+It is often unknown if the issue is recoverable when a critical error occurs. A sound strategy is to terminate the process (e.g. via `Environment.FailFast` or `IHostApplicationLifetime.Stop`) when a critical error occurs and rely on the process hosting environment to restart the process as a recovery mechanism, resulting in a resilient way to deal with critical errors.
+
+Before terminating the process, the NServiceBus endpoint can attempt a graceful shutdown:
```
ramonsmits, line 38:
```suggestion
Before terminating the process, the NServiceBus endpoint can attempt a graceful shutdown which can be useful in non-transactional processing environments:
```

ramonsmits (approved):
Some comments but this already warning the users on what side effects calling stop has and my new comment indicates when you would want to allow a graceful shutdown.

nservicebus/hosting/critical-errors.md
timbussmann: I was thinking about this but this is specifically about how to stop the endpoint/process as part of a critical error and might be a very common error to run into. Not against adding some documentation to extension hosting about how to properly shutdown the endpoint but I think that's a separate PR.

PR #5917, opened 2022-11-10
Document the impact of switching on/off full text search

mauroservienti:


servicecontrol/audit-instances/creating-config-file.md
```diff
@@ -193,6 +193,8 @@ Type: bool `true` or `false`
 
 Default: `true`.
 
+NOTE: If the audit instance uses RavenDB 5 persistence (available starting 4.26.0), changing full-text search settings causes RavenDB indexes to be redeployed and rebuilt. Depending on the amount of documents stored in the database, the operation might take a long time and documents won't be available until it's completed.
```
andreasohlund, line 196:
```suggestion
NOTE: If the audit instance uses RavenDB 5 persistence (available starting 4.26.0), changing the full-text search setting will cause indexes to be redeployed and rebuilt. Depending on the number of documents stored, this operation might take a long time and search results won't be available until completed.
```

andreasohlund (approved):

WojcikMike (approved):

PR #5932, opened 2022-11-18
Remove the showcase sample for V8

WilliamBZA:


ramonsmits (approved):

udidahan:
Some extra explanation as to why this was done may be helpful.

ramonsmits:
@udidahan The sample was already incompabitle with -alpha/beta/rc but wasn't spotted due to a Nuget version range bug/behavior. It would break at runtime. Its tracked that this sample is dropped and might be restored with a compatible version in the future.

udidahan:
OK - it would have been better to have that in the description of this PR though.

ramonsmits:
> OK - it would have been better to have that in the description of this PR though.

@udidahan Makes sense, done!

PR #5944, opened 2022-11-25
Fix wrong dependency versions in samples and snippets

ramonsmits:


ramonsmits:
@DavidBoike getting this:

> System.Exception: Could not render page: S:\docs.particular.net\transports\azure-service-bus\native-message-access.md
 ---> System.Exception: Failed to render article. PackageId:NServiceBus.Transport.AzureServiceBus. Version:[1.7.0, 1.9.0).
 ---> MissingSnippetsException: MissingSnippet.

this is after renaming the folder `ASBS_1` to `ASBS_1.7`, as the package reference is `1.7.*`. Somehow this results in this issue.

Based on the mentioned versions and folder I think it should just pass.

ramonsmits:
After diagnosing this it seems there are some assumptions on versioning. For example, take ASBS. Folder `_1` was referencing `1.7.*`. This validation would not allow this and the resolution would be to:

a. rename the folder to `_1.7`
b. set package version to `1.*`

Renaming the folder would be strange as these snippers are documentation for older API's. So to me the resolution for this problem should be:

c. rename the folder to `_1.0` and set package version to `1.0.*`

However, it seems that maybe those older minors have been removed due to supported versions?

Not sure why we would be doing that. Having snippet projects that target a minor nuget package for all our core packages seems very useful.

Wat I thought was a nice validation seems more complex......

PR #5947, opened 2022-11-30
Adjust samples using Outbox to set TransactionMode to ReceiveOnly

lailabougria:


samples/outbox/rabbit/Core_8/Sample/Program.cs
```diff
@@ -13,7 +13,10 @@ static async Task Main()
         #region ConfigureTransport
         var endpointConfiguration = new EndpointConfiguration("Samples.RabbitMQ.Outbox");
 
-        var rabbitMqTransport = new RabbitMQTransport(RoutingTopology.Conventional(QueueType.Classic), "host=localhost;username=rabbitmq;password=rabbitmq");
+        var rabbitMqTransport = new RabbitMQTransport(RoutingTopology.Conventional(QueueType.Classic), "host=localhost;username=rabbitmq;password=rabbitmq")
```
danielmarbach, line 16: I think this is the v8 pattern that we should show in the samples, so I would suggest switching to this pattern in the rest of the samples too.

danielmarbach: I agree but since you are already touching a few samples and make changes why not do it at least for those?

lailabougria: Because there's no decision on the matter yet.

danielmarbach: Well there is. We have created a new seam and reintroduced the shim APIs for smooth transition. So the new seam is the standard and the shim APIs are there to ease the transition. V8 samples should be using the new seam eventually and not show the transition APIs anymore because they will be faced out as indicated in the code with the PreObsolete attribute.

https://github.com/Particular/NServiceBus/blob/b16e590745c02a3817d02c076cf04f21946afb1b/src/NServiceBus.Core/Transports/TransportExtensions.cs#L12

danielmarbach: To add some more context around it

when you look at the existing samples they did the following

```csharp
endpointConfiguration.UseTransport(new LearningTransport());
```

which effectively translates to

```csharp
var transport = new LearningTransport();
endpointConfiguration.UseTransport(transport);
```

If you look at it from that angle, I would even argue the code was already be prepared to operate on the instance of the transport it was just inlined into the UseTransport. So in that sense, the changes here would undo that.

lailabougria: Ok, the last comment does make sense to me. What I don't want to do is go back and check all samples for the usage of the old vs. new API until we have a final decision. I'll make those changes.

danielmarbach (approved):
Approving under the assumption the code will eventually compile :tada:

PR #5953, opened 2022-12-07
Update to use .net7 to build samples

jpalac:


jpalac:
replaced by https://github.com/Particular/docs.particular.net/pull/5957

PR #5958, opened 2022-12-13
Update LangVersion from 7.3 to 8.0

danielmarbach:
This PR updates the samples to allow using the `LanguageVersion` `8.0`. This gives us at least the possibility to use slightly more modern CSharp features while striving for a good balance of supporting older versions of tooling.

This PR deliberately does not yet remove `netcoreapp3.1`


danielmarbach:
Would also unblock #5942

mikeminutillo (approved):

abparticular (approved):

PR #5960, opened 2022-12-13
Deprioritize SC, remove .NET Core 3.1

DavidBoike:


awright18 (approved):
I like the change to make service control the last option.

PR #5962, opened 2022-12-14
Clarify transactional session meta data

ramonsmits:


nservicebus/transactional-session/index.md
```diff
@@ -84,7 +84,9 @@ The default commit delay increment is set to `Timespan.FromSeconds(2)`and cannot
 
 #### Metadata
 
-It is possible to add custom headers to the control message that is used to settle the transaction outcome. Custom headers can be set by adding them to the metadata when opening the session. These headers can be accessed by a [custom behavior](/nservicebus/pipeline/manipulate-with-behaviors.md#add-a-new-step) in the `TransportReceive` part of the pipeline.
+It is possible to add meta data via custom headers to the transactional session control message that gets send. These headers can be accessed by a [custom behavior](/nservicebus/pipeline/manipulate-with-behaviors.md#add-a-new-step) when the control message is received in the `TransportReceive` part of the pipeline.
+
+Meta data passed via customer headers can useful to do XXX, XXXX
```
danielmarbach, line 89: This is a tricky one. Originally, we were even on the fence of documenting it because it is on the public API to enable downstreams to add custom headers for custom behavior like partition key extraction etc. We thought it is quite likely that no customer will ever really need this. But thinking about this, one could say it would be possible to populate tenant ID headers to the transactional session messages too. 

// cc @lailabougria @SzymonPobiega @tmasternak @timbussmann

lailabougria: Hmm, I went back into the code to refresh my memory. I do see one risk, though. This means users would be hooked into the TransportReceive stage to access these headers. We already have a behavior defined there that introduces a catch for the `ConsumeMessageException`, which we potentially throw in the `TransactionalSessionDelayControlMessageBehavior` in the physical message stage. Since we can't control the order there, is it safe to recommend this?

nservicebus/transactional-session/index.md
```diff
@@ -84,7 +84,9 @@ The default commit delay increment is set to `Timespan.FromSeconds(2)`and cannot
 
 #### Metadata
 
-It is possible to add custom headers to the control message that is used to settle the transaction outcome. Custom headers can be set by adding them to the metadata when opening the session. These headers can be accessed by a [custom behavior](/nservicebus/pipeline/manipulate-with-behaviors.md#add-a-new-step) in the `TransportReceive` part of the pipeline.
+It is possible to add meta data via custom headers to the transactional session control message that gets send. These headers can be accessed by a [custom behavior](/nservicebus/pipeline/manipulate-with-behaviors.md#add-a-new-step) when the control message is received in the `TransportReceive` part of the pipeline.
```
lailabougria, line 87:
```suggestion
It is possible to add metadata via custom headers to the transactional session control message sent. These headers can be accessed by a [custom behavior](/nservicebus/pipeline/manipulate-with-behaviors.md#add-a-new-step) when the control message is received in the `TransportReceive` part of the pipeline.
```

nservicebus/transactional-session/index.md
danielmarbach: What problem do you see related to the order? In general, if they have some infrastructure code that is in the transport receive phase, the order should not matter within the same stage. If it does, they can control the order within the stage by placing them before or after a specific behavior which could also include the already existing behavior the transactional session has added there.

lailabougria: I know you could drive the order, but IIRC we don't recommend that approach. I just wanted to ensure there's nothing we're missing on that end.

lailabougria: @ramonsmits did you get any questions surrounding this feature of the meta data? Or was this your own observation leading to this change?

ramonsmits: @lailabougria the docs are unclear. I tried to understand tranactional session behavior/features and it really needs refinement hence this PR. It still isn't clear what that API can/should be used for.

nservicebus/transactional-session/index.md
```diff
@@ -84,7 +84,9 @@ The default commit delay increment is set to `Timespan.FromSeconds(2)`and cannot
 
 #### Metadata
 
-It is possible to add custom headers to the control message that is used to settle the transaction outcome. Custom headers can be set by adding them to the metadata when opening the session. These headers can be accessed by a [custom behavior](/nservicebus/pipeline/manipulate-with-behaviors.md#add-a-new-step) in the `TransportReceive` part of the pipeline.
+It is possible to add metadata via custom headers to the transactional session control message sent. These headers can be accessed by a [custom behavior](/nservicebus/pipeline/manipulate-with-behaviors.md#add-a-new-step) when the control message is received in the `TransportReceive` part of the pipeline.
+
+Meta data passed via customer headers can useful to do XXX, XXXX
```
danielmarbach, line 89:
```suggestion
It is possible to add metadata (e.g. tenant information) via custom headers to the transactional session control message sent. These headers can be accessed by a [custom behavior](/nservicebus/pipeline/manipulate-with-behaviors.md#add-a-new-step) when the control message is received in the `TransportReceive` part of the pipeline.
```

nservicebus/transactional-session/index.md
```diff
@@ -84,7 +84,7 @@ The default commit delay increment is set to `Timespan.FromSeconds(2)`and cannot
 
 #### Metadata
 
-It is possible to add custom headers to the control message that is used to settle the transaction outcome. Custom headers can be set by adding them to the metadata when opening the session. These headers can be accessed by a [custom behavior](/nservicebus/pipeline/manipulate-with-behaviors.md#add-a-new-step) in the `TransportReceive` part of the pipeline.
+It is possible to add metadata (e.g. tenant information) via custom headers to the transactional session control message sent. These headers can be accessed by a [custom behavior](/nservicebus/pipeline/manipulate-with-behaviors.md#add-a-new-step) when the control message is received in the `TransportReceive` part of the pipeline.
```
danielmarbach, line 87: I think this is good enough. I was thinking about adding something like

"This is mostly used for extensibility of the persistence-specific implementations" but figured this is not really a hint that is required in this context. The only other alternative for me would be to remove this documentation section entirely, which, I feel, is also not a good idea.

lailabougria: Agreed. The task force has already decided not to emphasize this API. The tenant information is also the only use case I can think of right now, so I agree that this is good enough now.

lailabougria (approved):

PR #5981, opened 2023-01-23
Update Sql Persistence sample Microsoft.Data.SqlClient usage

danielmarbach:


samples/sql-persistence/injecting-services/SqlPersistence_6/Endpoint/Program.cs
```diff
@@ -13,7 +13,8 @@ static async Task Main()
         var endpointConfiguration = new EndpointConfiguration("Samples.SqlPersistence.InjectingServices");
         endpointConfiguration.EnableInstallers();
 
-        var connectionString = @"Data Source=.\SqlExpress;Initial Catalog=NsbSamplesInjectedServices;Integrated Security=True";
+        // for SqlExpress use Data Source=.\SqlExpress;Initial Catalog=NsbSamplesInjectedServices;Integrated Security=True;Encrypt=false
```
tmasternak, line 16: Why not go with `sqlexpress` as default? This would increase the chances of `F5` "just working".

tmasternak (commented):
This PR switched from SQL express to a docker-hosted SQL Server. I'm fine with that change. Should we announce it internally?

samples/sql-persistence/injecting-services/SqlPersistence_6/Endpoint/Program.cs
danielmarbach: I have no objections of flipping things around again. I just feel these days with .NET being more and more cross-platform and people embracing docker in all sorts of environments, this change would make it less windows focused. 

Also, when you look at the prerequisites we are pretty detailed about what you need to have "installed" for it to run so I wonder how much of this is really "just F5". I think you have more experience with these samples and their "whereabouts"  so I'd be happy to adjust in any direction.

tmasternak (approved):

samples/sql-persistence/sqlpersistence-prereqs.include.md
```diff
@@ -3,7 +3,7 @@
 
 ### MS SQL Server
 
- 1. Ensure an instance of SQL Server Express (Version 2016 or above for custom saga finders sample, or Version 2012 or above for other samples) is installed and accessible as `.\SqlExpress`.
+ 1. Ensure an instance of SQL Server (Version 2016 or above for custom saga finders sample, or Version 2012 or above for other samples) is installed and accessible on `localhost` and port `1433`. For example use `docker run -e "ACCEPT_EULA=Y" -e "MSSQL_SA_PASSWORD=yourStrong(!)Password" -p 1433:1433 -d mcr.microsoft.com/mssql/server:latest`.
```
mikeminutillo, line 6: We _could_ consider dropping that as a powershell, batch, or shell script into each sample. That way the instructions could just say "run the script or update the connection string to use an existing instance". I guess it depends which mechanism we think is more likely for people downloading the sample to use.

I know when I'm working on a sample I would use that. It might be too much maintenance for the benefit to the user. 

In any case, this change is better enough that I think we should merge it.

mikeminutillo (approved):

samples/sql-persistence/sqlpersistence-prereqs.include.md
danielmarbach: The funny thing about a script is that you have to assume some sort of an execution environment to run the script while when we can use a single docker command that command is universally applicable for anything that can run docker. I think the best way to move forward would be to provide a devcontainer integration. Then you can just open the sample with the devcontainer and you are ready to run.

PR #5982, opened 2023-01-24
Add a warning

SeanFeldman:


danielmarbach:
@SeanFeldman thanks for going through the effort of providing this. Honestly, I'm leaning towards not merging because a warning feels like overkill, especially given we already say there

> For exceptions where retrying does not make sense such as MyBusinessTimedOutException failed message is _not_ moved to an error queue and is discarded.

DavidBoike:
I think it could be a little more clear though, see this [discussion group post](https://discuss.particular.net/t/service-insights-doesnt-show-any-messages-that-were-discarded-due-to-custom-recoverability-policy/3379) where this came up. Maybe we could amend the text that @danielmarbach quoted rather than add a warn?

SeanFeldman:
I agree with @DavidBoike. If it would be clear, it wouldn't be confusing and lead to misunderstanding it.

SeanFeldman:
Anything that helps, works for me 🙂

danielmarbach:
@SeanFeldman I have raised an improvement to the quoted text I mentioned. The PR is available at https://github.com/Particular/docs.particular.net/pull/5991

I'll close this one for now. Thanks for the suggestion!

PR #5984, opened 2023-01-24
Update SqlTransport Microsoft.Data.SqlClient

danielmarbach:
- [x] Introduce Pool Size everywhere
- [x] Use single quote on docker command which makes it compatible with Linux too due to the `!` being escaped


mikeminutillo:
@danielmarbach does this need to update https://github.com/Particular/docs.particular.net/blob/master/samples/sql-prereq.include.md as well? I think the default database has changed for all SQL transport samples and this is included in every one of them

danielmarbach:
@mikeminutillo indeed. Will fix.

PR #5987, opened 2023-01-26
Link ServicePulse in Prometheus and Grafana sample page

sergioc:


sergioc:
Learning path banner removed temporarily in order to not drown this one.

samples/logging/prometheus-grafana/sample.md
```diff
@@ -2,11 +2,12 @@
 title: Capture and visualize metrics using Prometheus and Grafana
 summary: Illustrates using Prometheus and Grafana to capture and visualize NServiceBus metrics.
 component: Metrics
-isLearningPath: true
 reviewed: 2021-09-08
 previewImage: grafana-graph.png
 ---
 
+INFO: Monitor endpoints, detect errors, and recover from failed messages with minimal configuration. **[See how](https://particular.net/servicepulse?utm_campaign=grafana-prometheus) →**
```
DavidBoike, line 9: Is there a way to track this with only the campaign parameter?

If you're banking on the [tracker I set up with Britt to track CodeOpinion referrals](https://github.com/Particular/Website/blob/master/source/js/analytics-new.js#L69-L79), that currently only works if `utm_medium` is used with a value of `paid` or `cpc`.

DavidBoike (approved):

samples/logging/prometheus-grafana/sample.md
sergioc: You can segment by users with that campaign.

sergioc: You can segment users with that campaign and in KM you can search for views of pages containing the campaign in the URL.

PR #5990, opened 2023-01-31
Update rest of the SqlTransport and Persistence samples to Microsoft.Data.SqlClient

danielmarbach:


danielmarbach:
Pinging @tmasternak and @mikeminutillo for a review but will still merge once it is green. Anything discovered in the review can be done as an additional PR

PR #6032, opened 2023-03-30
Clarify when an audit instance is needed

andreasohlund:


fourpastmidnight:
Thank you! This is much clearer with respect to ServiceControl v4.x!

WilliamBZA (approved):

PR #6033, opened 2023-03-30
.NET 7 is required for the platform

andreasohlund:


andreasohlund:
@bording did you check of vc++2015 i still needed?

bording:
> @bording did you check if vc++2015 is still needed?

I have done some research and testing for this.

On the [RavenDB Getting Started](https://ravendb.net/docs/article-page/5.4/csharp/start/getting-started) page, it does mention it, but then says it's a requirement for .NET Core and links to Microsoft's [dependencies](https://learn.microsoft.com/en-us/dotnet/core/install/windows?tabs=net70#dependencies) page. 

That page says it's only a dependency for Windows 7/8.1 and Windows Server 2012.

However, .NET 7 is no longer supported on Windows 7/8.1, so those are no longer relevant.

I did try running SC on Windows 11 and removed the VC redistributable, and it worked fine without it.

I'm going to try and set up a Windows Server 2012 VM and see what happens there. It seems like that might be the only OS where it would still be needed.

Do we still support that version?

andreasohlund:
Since we require .net7 I guess we can’t support server2012?

andreasohlund:
Ah, missread your statement, not sure about server2012 🤷

andreasohlund:
Merging since I want to send a docs link to a customer

@bording go ahead and tweak the vc++ part as needed in a separate one

PR #6049, opened 2023-04-26
Cache the docs tool to speed up content verification

danielmarbach:
As long as there is no new docstool version pushed this improves the docs tool installation process by 24-25 times.

## Before

image

## After

image

## On cache miss

image

### DIFF
--- a/.github/workflows/validate-pull-requests.yml
+++ b/.github/workflows/validate-pull-requests.yml
@@ -6,7 +6,7 @@ env:
 jobs:
   content-verification:
     name: Content verification
-    runs-on: windows-2022
+    runs-on: ubuntu-22.04
     steps:
       - name: Checkout
         uses: actions/checkout@v3
@@ -14,13 +14,25 @@ jobs:
         uses: actions/setup-dotnet@v3.0.3
         with:
           dotnet-version: 7.0.x
+      - name: Check docstool
+        run: |
+           curl https://www.myget.org/F/particular/api/v3/flatcontainer/particular.docstool/index.json >> docstool-versions.json
+      - name: Cache docstool
+        id: cache-docstool
+        uses: actions/cache@v3.3.1
+        with:
+          path: |
+             ~/.dotnet/tools/docstool
+             ~/.dotnet/tools/.store/particular.docstool
+          key: docstool-${{ hashFiles('docstool-versions.json') }}
       - name: Install docstool
-        run: dotnet tool install -g Particular.DocsTool --add-source=https://www.myget.org/F/particular/api/v3/index.json
+        if: steps.cache-docstool.outputs.cache-hit != 'true'
+        run: dotnet tool install Particular.DocsTool --global --add-source=https://www.myget.org/F/particular/api/v3/index.json
       - name: Run docstool
-        run: docstool test
+        run: docstool test --no-version-check
   integrity-tests:
     name: Integrity tests
-    runs-on: windows-2022
+    runs-on: ubuntu-22.04
     steps:
       - name: Checkout
         uses: actions/checkout@v3
--- a/.github/workflows/verify-master.yml
+++ b/.github/workflows/verify-master.yml
@@ -9,7 +9,7 @@ env:
 jobs:
   content-verification:
     name: Content verification
-    runs-on: windows-2022
+    runs-on: ubuntu-22.04
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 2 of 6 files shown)


danielmarbach:
If this works good for us we could also attempt to move it into a composite action.

.github/workflows/validate-pull-requests.yml
```diff
@@ -14,8 +14,19 @@ jobs:
...
+      - name: Check docstool
+        run: |
+           curl https://www.myget.org/F/particular/api/v3/flatcontainer/particular.docstool/index.json >> docstool.version
+           echo "/docstool" >> $env:GITHUB_PATH
+      - name: Cache docstool
+        id: cache-docstool
+        uses: actions/cache@v3
+        with:
+          path: /docstool
+          key: ${{ runner.os }}-docstool-${{ hashFiles('docstool.version') }}
```
danielmarbach, line 26: I don't think the runner os is really necessary. Should I just use docstool-hash?

.github/workflows/validate-pull-requests.yml
```diff
@@ -14,8 +14,19 @@ jobs:
         uses: actions/setup-dotnet@v3.0.3
         with:
           dotnet-version: 7.0.x
+      - name: Check docstool
+        run: |
+           curl https://www.myget.org/F/particular/api/v3/flatcontainer/particular.docstool/index.json >> docstool.version
```
danielmarbach, line 19: This file contains all versions. Should I call it .versions?

The get request might slightly more expensive over time when we push more versions but this version is still many times faster than doing nuget list (that one took on average around 10-11 seconds)

.github/workflows/validate-pull-requests.yml
DavidBoike: Yeah as long as we only run the workflow on one platform it shouldn't matter.

Though the cache keys are a repo-wide thing that would be shared by multiple workflows.

.github/workflows/validate-pull-requests.yml
```diff
@@ -14,8 +14,19 @@ jobs:
...
+           echo "/docstool" >> $env:GITHUB_PATH
```
DavidBoike, line 20: Is this to add the tool directory to the `PATH`? Is that worth a comment?

.github/workflows/validate-pull-requests.yml
DavidBoike: How about this?
```suggestion
           curl https://www.myget.org/F/particular/api/v3/flatcontainer/particular.docstool/index.json >> docstool-versions.json
```

And at some point we should clean out some of those versions to conserve MyGet space given I can't think of a single reason to keep very old versions of DocsTool around.

.github/workflows/validate-pull-requests.yml
```diff
@@ -14,8 +14,19 @@ jobs:
...
+      - name: Cache docstool
+        id: cache-docstool
+        uses: actions/cache@v3
```
DavidBoike, line 23:
```suggestion
        uses: actions/cache@v3.3.1
```

.github/workflows/verify-master.yml
```diff
@@ -17,8 +17,19 @@ jobs:
         uses: actions/setup-dotnet@v3.0.3
         with:
           dotnet-version: 7.0.x
+      - name: Check docstool
+        run: |
+           curl https://www.myget.org/F/particular/api/v3/flatcontainer/particular.docstool/index.json >> docstool.version
+           echo "/docstool" >> $env:GITHUB_PATH
+      - name: Cache docstool
+        id: cache-docstool
+        uses: actions/cache@v3
```
DavidBoike, line 26:
```suggestion
        uses: actions/cache@v3.3.1
```

DavidBoike (approved):

.github/workflows/validate-pull-requests.yml
danielmarbach: I first tried to use $HOME/.dotnet/tools but somehow couldn't get it to work and then switched to using a dedicated path. I guess if we could use the home tools path there would be nothing need to add to the path but we might be caching more than we want. If we leave it as is adding a comment might be good yes.

.github/workflows/validate-pull-requests.yml
danielmarbach: Yeah they are but that should be a problem. I assume populating the cache is idempotent. So we even get bigger caching benefits

danielmarbach, line 26:
```suggestion
          key: docstool-${{ hashFiles('docstool-versions.json') }}
```

danielmarbach:
Concurrent cache update works like a charm

image

PR #6056, opened 2023-05-05
Use correct instance name in script

YurivanRuler:
The old audit instance must be removed from the ServiceControl (Primary) instance.


mikeminutillo (approved):

mikeminutillo:
Thanks for the fix @YurivanRuler! We appreciate it.

PR #6072, opened 2023-05-24
Do not delete ServiceControl indexes folder

ramonsmits:
Removed guidance on deleting indexes folder whch will now break SC instance that use the RavenDB 5 storage engine.

### DIFF
--- a/servicecontrol/troubleshooting.md
+++ b/servicecontrol/troubleshooting.md
@@ -186,12 +186,7 @@ To resolve these errors, the affected indexes must be rebuilt:
 - In RavenDB Management Studio, navigate to the Indexes view
 - [Reset the relevant index(es)](https://ravendb.net/docs/article-page/3.5/csharp/server/administration/index-administration)
 
-If many indexes are affected it may be easier to rebuild all indexes, although this can take a very long time if the database is large, and it will use a lot of CPU and storage IO capacity:
-
-- Stop the ServiceControl (audit or error) instance
-- Navigate to the [database folder](configure-ravendb-location.md) on disk
-- Delete the `Indexes` folder
-- Start the ServiceControl instance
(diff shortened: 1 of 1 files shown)


servicecontrol/troubleshooting.md
```diff
@@ -186,12 +186,7 @@ To resolve these errors, the affected indexes must be rebuilt:
 - In RavenDB Management Studio, navigate to the Indexes view
 - [Reset the relevant index(es)](https://ravendb.net/docs/article-page/3.5/csharp/server/administration/index-administration)
 
-If many indexes are affected it may be easier to rebuild all indexes, although this can take a very long time if the database is large, and it will use a lot of CPU and storage IO capacity:
-
-- Stop the ServiceControl (audit or error) instance
-- Navigate to the [database folder](configure-ravendb-location.md) on disk
-- Delete the `Indexes` folder
-- Start the ServiceControl instance
+WARNING: In the past a quick remedy was to purge the contents of the Indexes folder. This guidance no longer applies to since version ServiceControl 4.26.0
```
mauroservienti, line 189:
```suggestion
WARNING: In the past, a quick remedy was to purge the content of the `Indexes` folder. This guidance no longer applies since version 4.26.0 of ServiceControl.
```

mauroservienti:
Isn't it better to keep the old guidance in place and recommend it if the persister is RavenDB 3.5?

ramonsmits:
> Isn't it better to keep the old guidance in place and recommend it if the persister is RavenDB 3.5?

Not against still showing it but I want to avoid customer still accidentally using it.

ramonsmits:
Also, as mentioned I tested deleting the Indexes folder and SC audit instance happily recreated the indexes as I shown during our call so the guidance is still valid. 

Closing as I thought this would resolve some INDEX DOES NOT EXISTS errors where it seems I made the incorrect assumption that the cause was this guidance.

PR #6077, opened 2023-05-30
TransactionalSession - Use resource filter instead of middleware

danielmarbach:


samples/transactional-session/aspnetcore-webapi/SqlPersistenceTS_6/WebApplication/Program.cs
```diff
@@ -48,33 +49,56 @@ public static void Main()
...
             {
-                // Configure Entity Framework to attach to the synchronized storage session
+                // Configure Entity Framework to attach to the synchronized storage session when required
                 c.AddScoped(b =>
                 {
-                    var session = b.GetRequiredService<ISqlStorageSession>();
-                    var context = new MyDataContext(new DbContextOptionsBuilder<MyDataContext>()
-                        .UseSqlServer(session.Connection)
-                        .Options);
+                    var requiresSessionIntegration = true;
```
SzymonPobiega, line 55: Why do we need to assume `false`? Is it for the cases where `contextAccessor.HttpContext == null` ?

samples/transactional-session/aspnetcore-webapi/SqlPersistenceTS_6/WebApplication/SendMessageController.cs
```diff
@@ -18,6 +21,7 @@ public SendMessageController(ITransactionalSession messageSession, MyDataContext
 
     #region txsession-controller
     [HttpGet]
+    [RequiresTransactionalSession]
     public async Task<string> Get()
```
tmasternak, line 25:
```suggestion
    public async Task<string> Get(ITransactionalSession messageSession)
```

How about we remove the constructor injection for `ITransactionalSession` (and the class filed) and add it explicitly as a parameter to the action?

This might be a better way to show that `ITransactionalSession` is usable only in actions with proper attribute applied.

danielmarbach: @tmasternak it could be. I left it as is because with the changes, I added in place the context will automatically be switched over to the non-managed connection and the transactional session is in an unusable state but since it is not used everything will just work. I felt like leaving it like that is probably more representative of how messy certain types of controllers are in the wild in terms of what is used where. Thoughts?

samples/transactional-session/aspnetcore-webapi/SqlPersistenceTS_6/WebApplication/Program.cs
danielmarbach: When the http context is null we are in the handler pipeline for the incoming message (control message and others).

samples/transactional-session/aspnetcore-webapi/SqlPersistenceTS_6/WebApplication/SendMessageController.cs
tmasternak: > I felt like leaving it like that is probably more representative of how messy certain types of controllers are in the wild in terms of what is used where

Makes sense to me.

danielmarbach: @tmasternak I reflected about this again and went with your suggestion. Everything gets way cleaner and I think it is OK to be opinionated in the sample.

PR #6126, opened 2023-07-09
Weekly docs review

soujay:


DavidBoike:
Why do you need to add NonDurable persistence to those snippets in the first place?

DavidBoike:
To be more specific, it's failing because NonDurablePersistence was extracted to a different package in NServiceBus 8. But the snippets seem perfectly suitable without that addition.

soujay:
@DavidBoike 
> Why do you need to add NonDurable persistence to those snippets in the first place?

The other version 7.x had a snippet for InMemoryPersistence and hence I thought we could provide a similar one for 8.x.

I can remove them

PR #6127, opened 2023-07-10
Clarify wording around the use of SemVer

udidahan:
As we have a section immediately under this called "Interpretations/deviations from SemVer".

### DIFF
--- a/nservicebus/upgrades/release-policy.md
+++ b/nservicebus/upgrades/release-policy.md
@@ -14,7 +14,7 @@ While this may result in a large number of releases, only a small fraction usual
 
 ## Semantic versioning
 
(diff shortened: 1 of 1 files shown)


lailabougria (commented):
Shouldn't this also be updated to reflect the degrees of impact?
We may not follow SemVer for changes to internal seams for example (transport/persistence) based on those guidelines either.

udidahan:
Maybe we can merge this as-is and raise a different PR for the bigger changes we're discussing?

mauroservienti (approved):

lailabougria:
From @mauroservienti  and me, we're not really fond of the ambiguity this statement introduced, even though there's a subsection about deviations. Having the statement "generally" there might make users question when we do or don't, even beyond the subsections of how we deviate. But we won't oppose the change either.

PR #6174, opened 2023-08-06
add new sample for serializer converter when using System.Text.Json

soujay:
This PR is a sample that shows the workaround for the issue https://github.com/Particular/NServiceBus/issues/6804


samples/databus/blob-storage-databus-custom-serializer-converter/sample.md
```diff
@@ -0,0 +1,94 @@
+---
+title: Azure Blob Storage DataBus using converter for SystemJsonSerializer
+summary: Sending large attachments with NServiceBus via Azure blob storage using custom converter for SystemJsonSerializer
+component: ABSDataBus
+reviewed: 2023-08-05
+related:
+- nservicebus/messaging/databus
+- samples/databus/blob-storage-databus-cleanup-function
+---
+This sample shows how to send large attachments with NServiceBus via Azure blob storage using custom converter for SystemJsonSerializer
```
boblangley, line 10: The sample shows how to use a custom converter, but doesn't explain why it is necessary.

Also, the custom converter isn't unique to Azure, ASB, or Azure Blog Storage. Perhaps the sample should be in the general DataBus samples?

SeanFeldman: I'm wondering if this should be a sample at all or if it should be baked into `SystemJsonSerializer`.
Every customer using `SystemJsonSerializer` will risk encountering this issue when/if they use databus with large messages, using `DataBusProperty`.

soujay: @boblangley 

> The sample shows how to use a custom converter, but doesn't explain why it is necessary.

Although it was briefly mentioned later in the document in the section "Custom SystemJsonSerializer Converter", I see value in moving it to the beginning of the sample.


> Also, the custom converter isn't unique to Azure, ASB, or Azure Blog Storage. Perhaps the sample should be in the general DataBus samples and use Learning Transport / File DataBus?

I just happened to pick ASB to show this converter. But I agree that it is better keeping it simple and using general Learning Transport / File DataBus.

soujay: @SeanFeldman 
> I'm wondering if this should be a sample at all or if it should be baked into `SystemJsonSerializer`. Every customer using `SystemJsonSerializer` will risk encountering this issue when/if they use databus with large messages, using `DataBusProperty`.

We already have an issue around this. As a first step, I have created this sample and once merged, I will also update the https://docs.particular.net/nservicebus/serialization/system-json with a note to this effect

SeanFeldman: Introducing a workaround and replacing it isn't trivial for many places. And it ends up being a workaround that stays long after a fix has been implemented. The issue linked to has no links to any PR in progress so it's not clear wherever the bug will be worked on or not.

My client already has a custom serializer from v7 and the hope was to get rid of it when moving to NSB 8. Some places can't swap out workarounds swiftly and tend to stay with those for a long time, till the next round of updates is scheduled.

samples/databus/file-share-databus-custom-serializer-converter/sample.md
```diff
@@ -0,0 +1,89 @@
+---
+title: File Share Data Bus using converter for SystemJsonSerializer
```
boblangley, line 2:
```suggestion
title: Using databus with the SystemJsonSerializer message serializer
```

I would change the title (and menu item / summary) to something like this

boblangley (approved):

samples/databus/blob-storage-databus-custom-serializer-converter/sample.md
soujay: @SeanFeldman  - I understand that this is only a workaround until the actual bug is addressed and released. The [issue ](https://github.com/Particular/NServiceBus/issues/6804) has been queued and will be worked on as per our process. Please subscribe to the issue so that you can get notified of the latest updates.

PR #6176, opened 2023-08-08
Outbox tweaks

timbussmann:
Minor wording tweaks on the outbox, with two "major" changes:

* Removed the note about sending messages in the outbox dispatch phase.
  * This note was very prominently placed at the beginning of the document but touches an extreme edge case that seems to be related to extending the pipeline in very specific stages to do message dispatches. This note should be further down, probably in the `Important design considerations` section. However:
  * "outbox dispatch phase" is not something that is defined or documented anywhere. Afaik, there is no dispatch stage which even allows sending messages. So I'm not even sure what this tries to say.
* Added a link to the TX session in the note about using the outbox without an incoming message as this is exactly what the tx session is designed for.

### DIFF
--- a/nservicebus/outbox/index.md
+++ b/nservicebus/outbox/index.md
@@ -124,7 +124,6 @@ In phase 2, outgoing messages are sent to the messaging infrastructure and outbo
 * Because deduplication is done using `MessageId`, messages sent outside of an NServiceBus message handler (i.e. from a Web API) cannot be deduplicated unless they are sent with the same `MessageId`.
 * The outbox is _expected_ to generate duplicate messages from time to time, especially if there is unreliable communication between the endpoint and the message broker.
 * Endpoints using the outbox feature should not send messages to endpoints using DTC (see below) as the DTC-enabled endpoints treat duplicates coming from outbox-enabled endpoints as multiple messages.
-* Messages dispatched to the transport as part of the [outgoing pipeline stages](/nservicebus/pipeline/steps-stages-connectors#stages-outgoing-pipeline-stages.md) may not be batched and each message is sent in isolation.
 
 ### Transaction scope
 
--- a/nservicebus/outbox/index.md
+++ b/nservicebus/outbox/index.md
@@ -22,11 +22,9 @@ Most message queues, and some data stores, do not support distributed transactio
 
 The NServiceBus **outbox** feature ensures consistency between business data and messages. It simulates an atomic transaction, distributed across both the data store used for business data and the message queue used for messaging.
 
-Note: Messages dispatched to the transport as part of the outbox dispatch stage will not be batched and each message is sent in isolation.
-
 Note: Messages sent with [immediate dispatch](/nservicebus/messaging/send-a-message.md#dispatching-a-message-immediately) will be sent immediately and won't be handled by the outbox.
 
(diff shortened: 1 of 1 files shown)


nservicebus/outbox/index.md
```diff
@@ -22,11 +22,9 @@ Most message queues, and some data stores, do not support distributed transactio
 
 The NServiceBus **outbox** feature ensures consistency between business data and messages. It simulates an atomic transaction, distributed across both the data store used for business data and the message queue used for messaging.
 
-Note: Messages dispatched to the transport as part of the outbox dispatch stage will not be batched and each message is sent in isolation.
```
danielmarbach, line 25: I think the more accurate wording would be something like

> Note: Messages dispatched to the transport when the outbox is enabled will not be batched and each message is sent in isolation.

danielmarbach: and I agree this should be moved but I would keep the note

danielmarbach (approved):

nservicebus/outbox/index.md
timbussmann: > Messages dispatched to the transport when the outbox is enabled will not be batched and each message is sent in isolation.

how does the outbox behavior differ from regular dispatching? It uses the same pipeline and everything, so I'm not sure what information we want to give to the user here?

danielmarbach: It does differ because of this behavior https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Reliability/Outbox/ForceBatchDispatchToBeIsolatedBehavior.cs#L19 and this is what the note is about according to my understanding

danielmarbach: I raised https://github.com/Particular/NServiceBus.AmazonSQS/issues/2248 and https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/issues/857

timbussmann: > As far as I recall the Isolated has historically been interpreted differently than how you described it here.

but the scenario here would be:

1. Incoming message
2. Invoke handlers
3. handlers send message
4. message gets added to outgoing messages batch
5. Outbox persists the outgoing message batch to the DB
6. Outbox transaction committed
7. Outgoing messages (batch) is set to isolated to avoid the incoming message transaction
8. Outgoing messages are dispatched
9. Outbox records are set as dispatched

from a user perspective, the observable behavior is the same as with regular batching, as even the isolated setting does not send the messages any sooner than before (it's not the immediate dispatch path used by message handlers). So the docs message:

> Messages dispatched to the transport as part of the outbox dispatch stage will not be batched and each message is sent in isolation.

is technically true from a transport perspective but that's only due to the transport's implementation of the isolated dispatch behavior. The only thing that might differ is that a dispatch failure might roll back all outgoing messages in a batch instead of aborting in the middle but there is nothing the user can take into account in this regard as the receiver side needs to be able to deal with duplicates anyway.

Is my understanding correct @danielmarbach ?

danielmarbach: > from a user perspective, the observable behavior is the same as with regular batching, as even the isolated setting does not send the messages any sooner than before (it's not the immediate dispatch path used by message handlers)

Correct that is my understanding too see 

> Rereading that PR and your arguments here I do agree now that it is probably better to remove the note since there is no impact for the user

The most important distinction is only like you said that the sends are not enlisted in the receive transaction and I think that is already mentioned somewhere in the docs right?

PR #6190, opened 2023-08-28
Container app sample

danielmarbach:


SzymonPobiega:
This effort is being paused for now by the exploration rotation group for now. The group has invested as much resources as it could given the confidence level for KEDA and was unable to get the sample nowhere to the working state.

PR #6201, opened 2023-09-14
Fix LangVersion test for Jetbrains Rider users

dvdstelt:
JetBrains Rider does "10" instead of "10.0" when you select version in the UI

### DIFF
--- a/tests/IntegrityTests/ProjectLangVersion.cs
+++ b/tests/IntegrityTests/ProjectLangVersion.cs
@@ -36,7 +36,7 @@ namespace IntegrityTests
                     var firstLangVersionElement = xdoc.XPathSelectElement("/Project/PropertyGroup/LangVersion");
                     var value = firstLangVersionElement?.Value;
 
-                    return value == "10.0";
+                    return value is "10.0" or "10";
                 });
         }
     }


bording:
@dvdstelt We should not take this change.

`LangVersion` as [documented](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version#c-language-version-reference) should be `10.0`, so if Rider isn't doing that, it's wrong and you should fix the value.

dvdstelt:
> `LangVersion` as [documented](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version#c-language-version-reference) should be `10.0`, so if Rider isn't doing that, it's wrong and you should fix the value.

I think it'll work [just fine](https://github.com/dotnet/roslyn/blob/main/src/Compilers/CSharp/Portable/LanguageVersion.cs#L489-L492)

DavidBoike:
It's not just about whether or not it will work. This is [Postel's Law](https://en.wikipedia.org/wiki/Robustness_principle) at work. Microsoft is being liberal in what they accept. That doesn't mean we shouldn't be conservative in what we send.

The original point of this test was to keep thing's consistent so when it comes time to do a giant find/replace it's easier to do that without having to remember that oops, some of them might not have the `.0` in there.

Ultimately the fact that Rider _isn't_ following the Robustness principle isn't a good enough reason to sacrifice consistency.

I do not think this should be merged.

PR #6207, opened 2023-09-15
Update zero-downtime.md

helenktsai:


servicecontrol/upgrades/zero-downtime.md
```diff
@@ -9,6 +9,8 @@ The [ServiceControl remotes feature](/servicecontrol/servicecontrol-instances/re
 
 NOTE: PowerShell must be used to install individual instances. Example scripts are provided for each step.
 
+NOTE: For scenarios where retaining audit log data is not required (e.g. a test environment), this process is not necessary -- the audit instance can simply be destroyed and recreated with the same name.
```
soujay, line 12: It's up to the discretion of the customer whether to delete the existing instances irrespective of the environment they are in.  But usually, a customer planning to upgrade in Prod tend to follow all the recommended steps in dev/uat /test environments, so they are not presented with any unforeseen issues when they do the same in prod. I would refrain from recommending them to delete instances if its a non-prod environment.

helenktsai: @soujay Thanks for feedback 🙏 

I reworked this further in a pairing session and it should be okay for merging now.

PR #6211, opened 2023-09-20
Add external port 9094 to Kafka Trigger Sample

awright18:


dvdstelt:
It's working flawlessly. Except for 1 configuration item in the Rider run config, so I changed and pushed that. But this can be merged.

dvdstelt (approved):

PR #6212, opened 2023-09-20
Warn about tx session commit exceptions

timbussmann:
The transactional session commit operation can fail for various reasons. Users should be aware that an exception is a possible and valid outcome and that this might require additional logic in the user code to determine the appropriate reaction to such a failure. This is currently barely mentioned.

### DIFF
--- a/nservicebus/transactional-session/index.md
+++ b/nservicebus/transactional-session/index.md
@@ -60,7 +60,7 @@ snippet: committing-transactional-session
 
 Disposing the transactional session without committing will roll back any changes that were made.
 
-Warn: The `Commit` operation may fail and throw an exception for reasons outlined in the [failure scenarios section](#failure-scenarios).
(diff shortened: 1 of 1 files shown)


SzymonPobiega (approved):

danielmarbach (approved):
Wondering if Warn is really necessary and not just an Info.

timbussmann:
> Wondering if Warn is really necessary and not just an Info.

yeah no strong opinion on that, happy to tune it down to an info for now.

nservicebus/transactional-session/index.md
```diff
@@ -60,6 +60,8 @@ snippet: committing-transactional-session
 
 Disposing the transactional session without committing will roll back any changes that were made.
 
+Warn: The `Commit` operation may fail and throw an exception for reasons outlined in the [failure scenarios section](#failure-scenarios).
```
timbussmann, line 63:
```suggestion
NOTE: The `Commit` operation may fail and throw an exception for reasons outlined in the [failure scenarios section](#failure-scenarios).
```

PR #6213, opened 2023-09-20
Steer developer to use interfaces

dignite:
Having messages as interfaces have benefits, and the documentation should nudge the developer to always think of messages as interface types.

### DIFF
--- a/nservicebus/messaging/messages-events-commands.md
+++ b/nservicebus/messaging/messages-events-commands.md
@@ -50,13 +50,13 @@ Messages represent data contracts between endpoints. They should be designed acc
 
 Messages should:
 
-* be simple [POCO](https://en.wikipedia.org/wiki/Plain_old_CLR_object) types.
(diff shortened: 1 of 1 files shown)


DavidBoike:
Hi @dignite,

While messages as interfaces have some benefits, and open up interesting use cases like multiple inheritance, we don't want to recommend this strongly that all users should prefer defining their messages that way. So this wording is simply too strong.

We might, however, accept a change to that bullet list that adds a new bullet that at least teases the possibility of interface-based messages to link to that article.

dignite:
> Hi @dignite,
> 
> While messages as interfaces have some benefits, and open up interesting use cases like multiple inheritance, we don't want to recommend this strongly that all users should prefer defining their messages that way. So this wording is simply too strong.
> 
> We might, however, accept a change to that bullet list that adds a new bullet that at least teases the possibility of interface-based messages to link to that article.

Good to know, thanks! :) I have updated the docs according to your suggestion

DavidBoike (approved):

PR #6183, opened 2023-08-14
Regular diagnostics docs do not apply to azure functions

ramonsmits:
Based on a support case where a customer was not guided to the azure function pages.

PR #6214, opened 2023-09-20
Fix links startup diagnostics broken links

ramonsmits:
Fix broken links caused by:

- https://github.com/Particular/docs.particular.net/pull/6183


ramonsmits:
ok, seems already fixed by:

- https://github.com/Particular/docs.particular.net/pull/6215

PR #6233, opened 2023-10-12
Update integrity tests

bording:
With the upcoming NServiceBus 9 release, we need to take a look at how we are enforcing `LangVersion` settings. It no longer makes sense to enforce a single value across the entire repo.

Instead, we are now doing the following:
- Snippets projects (which should always be single targeted, see new test below) should not have a `LangVersion` defined. Instead, each project will use the default language version based on the TFM of the project.
- Samples should define the `LangVersion` property to match the default version of the lowest .NET TFM  used in the solution. This ensures that the samples can take advantage of language features that match the version of .NET that corresponds to the NServiceBus version used in the sample. It also means that as more TFMs are added to the sample over time (as new .NET versions come out), the `LangVersion` for that sample will stay stable and not require a newer version of tooling to use it than what it initially required.

This PR also updates the integrity tests in the following ways:

- .NET Core 3.1 should no no longer be allowed, so the test has been updated and projects still using `netcoreapp3.1` have been updated.
- There is no point to multi-targeting Snippets projects, so I have added a test to verify that they aren't.
- Re-enabled a test that had been incorrectly marked with `Ignore`. It is still generally true that we only want to have a single solution per directory, so an exception was added to the test for the single sample that violates this.
- Miscellaneous cleanup

### DIFF
--- a/Snippets/ABSDataBus/ABSDataBus_3.1/ABSDataBus_3.1.csproj
+++ b/Snippets/ABSDataBus/ABSDataBus_3.1/ABSDataBus_3.1.csproj
@@ -1,7 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFramework>net48</TargetFramework>
-    <LangVersion>10.0</LangVersion>
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="NServiceBus.DataBus.AzureBlobStorage" Version="3.1.*" />
--- a/Snippets/ABSDataBus/ABSDataBus_4/ABSDataBus_4.csproj
+++ b/Snippets/ABSDataBus/ABSDataBus_4/ABSDataBus_4.csproj
@@ -1,7 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFramework>net48</TargetFramework>
-    <LangVersion>10.0</LangVersion>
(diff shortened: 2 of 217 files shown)


DavidBoike (approved):

timbussmann:
> Samples should define the LangVersion property to match the default version of the lowest .NET TFM used in the solution. This ensures that the samples can take advantage of language features that match the version of .NET that corresponds to the NServiceBus version used in the sample. It also means that as more TFMs are added to the sample over time (as new .NET versions come out), the LangVersion for that sample will stay stable and not require a newer version of tooling to use it than what it initially required.

would there be value in having a short table on the readme that lists the corresponding language version for the TFMs that we use for easier reference?

bording:
> would there be value in having a short table on the readme that lists the corresponding language version for the TFMs that we use for easier reference?

[Microsoft already documents](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version#defaults) it, so I don't think there's value in us duplicating it.

timbussmann:
Ah that's cool, how about linking this, also mentioning the original guidance int he PR description: https://github.com/Particular/docs.particular.net/pull/6234

bording:
> Ah that's cool, how about linking this, also mentioning the original guidance int he PR description: #6234

That's fine, though when the test fails, it will tell you which `LangVersion` it expects to see, per solution, so there's no guessing/figuring it out required.

PR #6235, opened 2023-10-13
Add SQS snippets for v7

andreasohlund:
There is no upgrade guide for v7 and there is a breaking change that is documented [here](https://docs.particular.net/transports/sqs/configuration-options#v1-compatibility-mode), should we create one?


bording:
> There is no upgrade guide for v7 and there is a breaking change that is documented [here](https://docs.particular.net/transports/sqs/configuration-options#v1-compatibility-mode), should we create one?

Yeah, seems like we should.

andreasohlund:
Upgrade guide added, ready for final review

bording (approved):

PR #6264, opened 2023-10-25
Editorial for `Service fails to start due to reaching the timeout period`

helenktsai:


nservicebus/hosting/nservicebus-host/troubleshooting.md
```diff
@@ -34,17 +34,21 @@ Verify that:
...
+A timeout was reached (30000 milliseconds) while waiting for the XYZ service to connect.
+The XYZ service failed to start due to the following error:
+The service did not respond to the start or control request in a timely fashion
 ```
+
+Generally, the service will start at a later time without issue. 
+
+This problem can be mitigated by:
+
+- Configuring the service with the `depend= <dependencies>` parameter set which identifies which dependencies must start before this service starts.
```
ramonsmits, line 51:
```suggestion
- Enable service recoverability to ensure the service will auto restart in case start fails
- Configuring the service with the `depend= <dependencies>` parameter set which identifies which dependencies must start before this service starts. 

```

nservicebus/hosting/nservicebus-host/troubleshooting.md
```diff
@@ -34,17 +34,21 @@ Verify that:
...
 2. All files in the deployment (sub)folder are correct and no old files are present from previous deployments.
 
-## Service fails to start due to reaching the timeout period
+## Windows Service fails to start due to reaching the timeout period
 
-Sometimes when the server hosting a microservice that uses NServiceBus is started, the following exception is seen in windows event viewer.
-> A timeout was reached (30000 milliseconds) while waiting for the XYZ service to connect.
-> The XYZ service failed to start due to the following error:
-> The service did not respond to the start or control request in a timely fashion
+Sometimes, a Windows service that is set to start automatically will fail to start due to a dependency not yet being available before it times out. The following exception can be seen in the Windows Event Viewer:
```
ramonsmits, line 39: Now that I reviewing it again I think this whole docs should just point to https://docs.particular.net/nservicebus/hosting/windows-service which already has documentation on sc.exe

ramonsmits (changes requested):

helenktsai:
Thanks @ramonsmits, it sounds like editorial is going in the wrong direction. I don't want to take away from the intent of the original content, so I'll revisit this.

ramonsmits (approved):

PR #6267, opened 2023-10-27
Upgraded to Particular.PlatformSample 3-*

ramonsmits:


DavidBoike:
What if instead of renaming `PlatformSample_2` to `PlatformSample_3` we renamed it instead to `PlatformSample_All` since we're never going to do multi-version snippets on this package?

DavidBoike:
Regarding:

> What if instead of renaming PlatformSample_2 to PlatformSample_3 we renamed it instead to PlatformSample_All since we're never going to do multi-version snippets on this package?

I noticed that if the versioned directory is used, then the resulting documentation page has nice notices that the NuGet package with the version is being used, even if there's only ever one at a time.

DavidBoike (approved):

PR #6279, opened 2023-10-31
Update native-integration_messagetypedetection_sqs_[5.3,5.7).partial.md

HindyLambrozo:
Editors, can you please verify that my corrections are accurate?


transports/sqs/native-integration_messagetypedetection_sqs_[5.3,5.7).partial.md
```diff
@@ -3,8 +3,8 @@
...
+During message processing, the SQS transport inspects the native message attributes for an attribute with the name `MessageTypeFullName` and a value representing a full type name (i.e. `Sales.OrderAccepted`). If the attribute is present, the message is treated as a native message, and the body is deserialized into the target type represented by `MessageTypeFullName`.
 
-The native message body is loaded from the configured [S3 bucket](/transports/sqs/configuration-options.md#offload-large-messages-to-s3) when the message attribute contains an attribute with the key `S3BodyKey` and the value representing an S3 object key, including the [necessary prefix](/transports/sqs/configuration-options.md#offload-large-messages-to-s3-key-prefix) as configured by the receiving endpoint.
+The native message body is loaded from the configured [S3 bucket](/transports/sqs/configuration-options.md#offload-large-messages-to-s3) when the message attribute contains an attribute with the key `S3BodyKey` and the value representing an S3 object key, including the [necessary prefix](/transports/sqs/configuration-options.md#offload-large-messages-to-s3-key-prefix), is configured by the receiving endpoint.
```
DavidBoike, line 8: I think "as" was more correct, because it's talking about the message having a prefix. The reason it has the prefix happens to be because it was configured somewhere to do that. But I don't think that's important, and I'm also not sure it's the receiving endpoint that does that, and we're linking to more info about it anyway, so how about just drop that fragment?
```suggestion
The native message body is loaded from the configured [S3 bucket](/transports/sqs/configuration-options.md#offload-large-messages-to-s3) when the message attribute contains an attribute with the key `S3BodyKey` and the value representing an S3 object key, including the [necessary prefix](/transports/sqs/configuration-options.md#offload-large-messages-to-s3-key-prefix).
```

PR #6291, opened 2023-11-01
Add upgrade guide for Azure Table Storage v6

andreasohlund:


andreasohlund:
@lailabougria @danielmarbach @abparticular This change originated in https://github.com/Particular/NServiceBus.Persistence.AzureTable/pull/670 do you have any input on what we should tell the customer to do if they are using the compatibility mode?

danielmarbach:
@andreasohlund I have tried to gather as much context with @kentdr in https://github.com/Particular/NServiceBus.Persistence.AzureTable/issues/836

But the gist of this is that the compatibility mode has been around from v3 and has gone through all the required steps so that it can be removed. At the time, we concluded that Azure Table persistence is a bit of a niche and the customers that are using it are used to the cadence of upgrading through the versions because they depend on the storage table APIs that have been constantly evolving over the years. We didn't find it necessary to write a "migration tool" but some of these decisions also predate some of our never policies. I would suggest using the issue https://github.com/Particular/NServiceBus.Persistence.AzureTable/issues/836 to capture feedback over time

kentdr (approved):

abparticular (approved):

danielmarbach (approved):

PR #6292, opened 2023-11-01
Fix upgrade guides

bording:


components/core-dependencies/NServiceBus.AmazonSQS.txt
```diff
@@ -1,3 +1,4 @@
+7.0 : 9
```
andreasohlund, line 1: Should we consider a integrity test to check that we don't miss this in the future?

andreasohlund (approved):

kentdr (approved):

PR #6296, opened 2023-11-03
Laila's open telemetry sample with AWS X-Ray and CloudWatch

cquirosj:


udidahan:
Should this be included under:
https://docs.particular.net/samples/aws/

Also, I would have expected to see it possibly listed under "Logging and Metrics":
image

PR #6300, opened 2023-11-06
Converted Near Real-Time Transient Clients sample to .NET 6

abparticular:


samples/near-realtime-clients/Core_8/ClientHub/Program.cs
```diff
@@ -15,41 +16,36 @@ static async Task Main()
...
-            publisherEndpoint: "Samples.NearRealTimeClients.Publisher");
+        endpointConfiguration.UseTransport(new LearningTransport());
 
         endpointConfiguration.SendFailedMessagesTo("error");
 
         var conventions = endpointConfiguration.Conventions();
         conventions.DefiningEventsAs(type => type == typeof(StockTick));
 
-        endpointConfiguration.EnableInstallers();
+        var builder = WebApplication.CreateBuilder();
```
DavidBoike, line 26: Was it not possible to use NServiceBus.Extensions.Hosting with the `.UseNServiceBus(…)` extension method?

DavidBoike (approved):
I'm surprised no changes were required in the JavaScript, as I thought AspNetCore SignalR wasn't wire-compatible. Did we just get lucky because the sample is simple enough? (Thinking about ServiceControl and ServicePulse here.)

samples/near-realtime-clients/Core_8/ClientHub/Program.cs
abparticular: Thanks for the tip.  It is indeed possible and I have made the changes

abparticular:
> I'm surprised no changes were required in the JavaScript, as I thought AspNetCore SignalR wasn't wire-compatible. Did we just get lucky because the sample is simple enough? (Thinking about ServiceControl and ServicePulse here.)

I'm not sure what JavaScript you are referring to here?  The SignalR client in this sample is a console app.

PR #6305, opened 2023-11-08
Add RabbitMQ v9 samples

bording:


samples/rabbitmq/native-integration/Rabbit_9/NativeSender/Program.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            #region GenerateUniqueMessageId
+            var messageId = Guid.NewGuid().ToString();
+
+            properties.MessageId = messageId;
+
+            #endregion
+
+            #region CreateNativePayload
+            var payload = "<MyMessage><SomeProperty>Hello from native sender</SomeProperty></MyMessage>";
```
andreasohlund, line 44: Nitpick: We could change to json here to avoid using the xmlserializer

samples/rabbitmq/native-integration/Rabbit_9/Receiver/Receiver.csproj
```diff
@@ -0,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <OutputType>Exe</OutputType>
+    <LangVersion>12.0</LangVersion>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="NServiceBus" Version="9.0.0-alpha.9" />
```
andreasohlund, line 8: What is the reason to ref the core given that the transport has a ref to it?

kentdr (approved):

samples/rabbitmq/native-integration/Rabbit_9/NativeSender/Program.cs
bording: Sure, but then we'd also have to update the sample readme which talks about the serializer used, and it didn't seem like it was worth the extra effort. We'd end up needing to version the readme since there isn't a single json serializer that's available for all the versions of the sample.

Is there a reason to avoid using the XML serializer that would make it worth doing all of that?

PR #6312, opened 2023-11-14
Update azure samples to core v9

kentdr:


andreasohlund (approved):
Title should say v9 right?

bording (approved):

PR #6314, opened 2023-11-14
Add NServiceBus 9 delayed delivery sample

bording:


samples/delayed-delivery/Core_9/Client/Program.cs
```diff
@@ -0,0 +1,73 @@
...
+    static async Task Main()
+    {
+        Console.Title = "Samples.DelayedDelivery.Client";
+        var endpointConfiguration = new EndpointConfiguration("Samples.DelayedDelivery.Client");
+        endpointConfiguration.UsePersistence<LearningPersistence>();
+        endpointConfiguration.UseSerialization<SystemJsonSerializer>();
+        endpointConfiguration.UseTransport(new LearningTransport());
+
+        var endpointInstance = await Endpoint.Start(endpointConfiguration)
+            .ConfigureAwait(false);
```
andreasohlund, line 16: Wasn't there a decision to no longer use ConfigureAwait in docs?

andreasohlund (approved):

samples/delayed-delivery/Core_9/Client/Program.cs
bording: "Samples contain no calls to ConfigureAwait(bool) unless they are explicitly required."

Yes, but this was already in the other version of the sample. Makes more sense to me at this point to not worry about this for now, and then it can be cleaned up (and perhaps enforced with an integrity test) later.

PR #6362, opened 2023-12-13
Core V9 upgrade guide

andreasohlund:


andreasohlund:
@bording @kentdr this is ready for a first pass from you to confirm the proposed actions

andreasohlund:
@bording snippets have  been changed like we discussed yesterday

nservicebus/upgrades/8to9/index.md
```diff
@@ -1,18 +1,103 @@
...
+
+snippet: core-8to9-adresses-runtime
+
+### Dynamic address translation
+
+Inject the new `ITransportAddressResolver` to translate `QueueAddress`'s to a physical transport addresses at runtime.
+
+snippet: core-8to9-adresses-translation
+
+## Restricted type argument for .UsePersistence<T>
```
bording, line 78: I can't find what change you're referring to here. Looking a the approval files, I don't see any `UsePersistence` changes between `release-8.1` and `master. The links you've got in here don't really seem to do anything useful.

nservicebus/upgrades/8to9/index.md
```diff
@@ -1,18 +1,103 @@
...
+
+https://github.com/Particular/NServiceBus/compare/release-8.1...master#diff-2c08aef8335f8f17ba3dc362fe939f8a5bdddde4f411d06a067882ac204fa43eL854
+
+Proposed action: Exclude since it doesn't impact users
+
+## Exceptions no longer marked as serializable
```
bording, line 84: I think this will be okay to exclude. From my understanding you only really cared about exceptions being serializable if you needed to transport an exception across an AppDomain boundary.

nservicebus/upgrades/8to9/index.md
```diff
@@ -1,18 +1,103 @@
...
+
+https://github.com/Particular/NServiceBus/compare/release-8.1...master#diff-2c08aef8335f8f17ba3dc362fe939f8a5bdddde4f411d06a067882ac204fa43eL739
+
+Proposed action: Exclude since it doesn't impact users
+
+## Extensionpoint for non task based notifications has been deprecated
```
bording, line 90: This one also seems like would be okay to leave off, though I suppose it would be easy enough to just say that the API has been removed and link to the docs on the way to do it now?

nservicebus/upgrades/8to9/index.md
```diff
@@ -1,18 +1,103 @@
...
+
+https://github.com/Particular/NServiceBus/compare/release-8.1...master#diff-2c08aef8335f8f17ba3dc362fe939f8a5bdddde4f411d06a067882ac204fa43eR250 (TBD remove)
+
+Propose action: Don't mention this since the only use case was already deprecated in v8, see /nservicebus/upgrades/7to8/#error-notification-events and no downstream is using this API (there could be end users extending this though but that feels very unlikely?)
+
+## Audit context no longer accept time to be received
```
bording, line 96: What have we done in the past for these sort of pipeline changes? If we've documented them in the past, might be a good idea to go ahead.

nservicebus/upgrades/8to9/index.md
```diff
@@ -1,18 +1,103 @@
...
+
+snippet: core-8to9-di-features
+
+## Service collection extensions for backwards compatibility removed
+
+Service collection extensions to ease [the transition to Microsoft DI abstractions](/nservicebus/upgrades/7to8/dependency-injection.md#registercomponents-changes) have been removed. It is now required to use the registration APIs added in NServiceBus 8.
+
+snippet: core-8to9-di-shims
+
+## Making features depend on message driven subscriptions
```
bording, line 52: Do we want to document this in this way? We don't really want people to try and take that dependency at all. Otherwise, why did we make the class internal? Seems like this entry should be saying something more like "no longer possible to take this dependency" ?

andreasohlund: We decided to move it to the bottom

nservicebus/upgrades/8to9/index.md
andreasohlund: Mention that it has been removed and show an example on the new way of doing it

nservicebus/upgrades/8to9/index.md
andreasohlund: Decided to not document this since creating an audit context requires a ForkConnector to be used 

https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Pipeline/ConnectorContextExtensions.cs#L209

while a user could do that there is no way to actually invoke the fork since the `Fork` method is internal

https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Pipeline/ForkExtensions.cs#L6

and inlining the `InvokePipeline` call won't work either since its internal as well

https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Pipeline/PipelineInvocationExtensions.cs#L6

TL;DR; Not documenting this since no use outside core is possible

nservicebus/upgrades/8to9/index.md
andreasohlund: I've moved the section to the extension part and added some text with an example

bording (approved):

PR #6379, opened 2023-12-24
Clarify that moving saga-not-found messages to the error queue should be considered rare

udidahan:
Documentation update.


udidahan:
FYI @danielmarbach and @DavidBoike who introduced this in #6377.

PR #6389, opened 2024-01-08
Suggest AzureFunctions isolated worker instead of InProcess

dvdstelt:


dvdstelt:
@DavidBoike Pinged you, since you're an editor and already familiar with the previous change and the fact that we want users not to start with InProcess anymore.

nservicebus/hosting/azure-functions-service-bus/in-process/index.md
```diff
@@ -13,6 +13,8 @@ reviewed: 2023-06-28
 
 Host NServiceBus endpoints with [Azure Functions](https://docs.microsoft.com/en-us/azure/azure-functions/) and [Azure Service Bus](https://azure.microsoft.com/en-us/services/service-bus/) triggers.
 
+WARNING: Microsoft announced that .NET8 will be [the last LTF release supporting the AzureFunctions InProcess SDK](https://techcommunity.microsoft.com/t5/apps-on-azure-blog/net-on-azure-functions-august-2023-roadmap-update/ba-p/3910098). It is recommended to use [AzureFunctions Isolated Worker model](/nservicebus/hosting/azure-functions-service-bus/).
```
DavidBoike, line 16: .NET 8 will be the last LTS but that hides the fact that in-proc only ever supported LTS - in-proc never did .NET 7. So when you put that all together .NET 8 is just _the last_.

```suggestion
WARNING: Microsoft announced that .NET 8 will be [the last release supporting the in-process hosting model](https://techcommunity.microsoft.com/t5/apps-on-azure-blog/net-on-azure-functions-august-2023-roadmap-update/ba-p/3910098). Projects should be migrated to use the [isolated worker model](/nservicebus/hosting/azure-functions-service-bus/) instead. See [Migrate .NET apps from the in-process model to the isolated worker model](https://learn.microsoft.com/en-us/azure/azure-functions/migrate-dotnet-to-isolated-model?tabs=net8) in Microsoft's documentation.
```

dvdstelt: Migrating our endpoints will work completely different, as we hide the functions attribute completely and introduce our own. It's even possible that changing NuGet package does the trick. But that hasn't been tested yet. We created a small task for that and added it as mandatory to the next Azure Platform Enhancement.

PR #6393, opened 2024-01-10
Specify that unobtrusive mode allows .net standard

Asim-A:


ramonsmits:
Thanks @Asim-A

Asim-A:
Thank you as well, @ramonsmits 🤝

PR #6406, opened 2024-01-18
Correcting Transactional Session sample

### LINKED ISSUE #6239: Feedback: 'Using TransactionalSession with Entity Framework and ASP.NET Core'
Feedback for 'Using TransactionalSession with Entity Framework and ASP.NET Core' https://docs.particular.net/samples/transactional-session/aspnetcore-webapi/

Location in GitHub: https://github.com/Particular/docs.particular.net/blob/master/samples/transactional-session/aspnetcore-webapi/sample.md

The .NET 7 sample will always commit the transactional session, even when an exception is thrown in the controller. Unlike with middleware, the result needs to be explicitly checked when using filters. The change below should make it behave the same as the middleware based sample.

```
    public async Task OnResourceExecutionAsync(ResourceExecutingContext context, ResourceExecutionDelegate next)
    {
        if (context.ActionDescriptor.Parameters.Any(p => p.ParameterType == typeof(ITransactionalSession)))
        {
            var session = context.HttpContext.RequestServices.GetRequiredService<ITransactionalSession>();
            await session.Open(new SqlPersistenceOpenSessionOptions());

            var result = await next();

            if (result.Exception != null)
            {
                await session.Commit();
            }
        }
        else
        {
            await next();
        }
    }
```

awright18:
- Fixes #6239

### DIFF
--- a/samples/transactional-session/aspnetcore-webapi/SqlPersistenceTS_6/WebApplication/MessageSessionFilter.cs
+++ b/samples/transactional-session/aspnetcore-webapi/SqlPersistenceTS_6/WebApplication/MessageSessionFilter.cs
@@ -14,9 +14,12 @@ public class MessageSessionFilter : IAsyncResourceFilter
             var session = context.HttpContext.RequestServices.GetRequiredService<ITransactionalSession>();
             await session.Open(new SqlPersistenceOpenSessionOptions());
 
-            await next();
+            var result = await next();
 
-            await session.Commit();
+            if(result.Exception == null)
+            {
+                await session.Commit();
+            }            
         }
         else
         {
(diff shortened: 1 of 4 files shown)


danielmarbach (approved):

samples/transactional-session/aspnetcore-webapi/SqlPersistenceTS_6/WebApplication/MessageSessionFilter.cs
```diff
@@ -14,9 +14,12 @@ public async Task OnResourceExecutionAsync(ResourceExecutingContext context, Res
             var session = context.HttpContext.RequestServices.GetRequiredService<ITransactionalSession>();
             await session.Open(new SqlPersistenceOpenSessionOptions());
 
-            await next();
+            var result = await next();
 
-            await session.Commit();
+            if(result.Exception == null)
```
danielmarbach, line 19: Don't remember what LangVersion we use these days but if you want you could the `is null` pattern.

awright18: good idea!

PR #6413, opened 2024-01-19
Remove stream properties sample

DavidBoike:
Removing this sample from docs because it's more of a feature proof-of-concept than a sample. It's fairly complex and crosses the line into something that should be offered as a first-class feature rather than a sample someone can hack together.


DavidBoike:
Raised an issue https://github.com/Particular/NServiceBus/issues/6946 in the NServiceBus repo to include stream properties as a proper first-class NServiceBus feature at some point in the future.

bording (approved):

PR #6415, opened 2024-01-20
Clarify optimistic concurrency for saga creation.

TravisNickels:


soujay (approved):

danielmarbach:
I do understand what we are trying to achieve but I wonder if it is a good idea to do that in the generic documentation part or whether we are restricting our options too much by doing so.

nservicebus/sagas/concurrency.md
```diff
@@ -23,9 +23,9 @@ Saga state is created when a saga is started by a message. It may be updated whe
 
 ## Starting a saga
 
-When messages that start the same saga instance are received simultaneously, NServiceBus ensures that only one message starts the saga.
+When simultaneous messages initiating the same saga instance are received, NServiceBus ensures that only one message starts the saga. Optimistic concurrency is always used for saga creation.
```
TravisNickels, line 26:
```suggestion
When simultaneous messages initiating the same saga instance are received, NServiceBus ensures that only one message starts the saga. Currently, saga creation only uses optimistic concurrency.
```
> I do understand what we are trying to achieve but I wonder if it is a good idea to do that in the generic documentation part or whether we are restricting our options too much by doing so. 

I'm not 100% sure I understand what you mean by restricting our options. This was more meant to indicate the current state of saga creation.  Maybe `Optimistic concurrency is always used for saga creation` indicates that this can/will never be changed.  I could try to soften the tone?

danielmarbach (approved):

nservicebus/sagas/concurrency.md
```diff
@@ -23,9 +23,9 @@ Saga state is created when a saga is started by a message. It may be updated whe
...
 
-Using optimistic concurrency, only one message handler is allowed to succeed. The saga state is created and sent messages are dispatched. The other handlers fail, roll back, and their messages enter [recoverability](/nservicebus/recoverability/). When those messages are retried, the existing saga state is found. Handling the messages may involve changing the state or completing the saga. See below for how those scenarios are handled.
+Using optimistic concurrency, only one message handler is allowed to succeed. The saga state is created and sent messages are dispatched. The other handlers fail, roll back, and their messages enter [recoverability](/nservicebus/recoverability/). When those messages are retried, the existing saga state is found. Handling the messages may involve changing the state or completing the saga.  It's important to note that enabling pessimistic concurrency for the persister does not impact the optimistic concurrency of saga creation. See below for how those scenarios are handled.
```
TravisNickels, line 28:
```suggestion
Optimistic concurrency only allows one message handler to succeed. The saga state is created and sent messages are dispatched. The other handlers fail, roll back, and their messages enter [recoverability](/nservicebus/recoverability/). When those messages are retried, the existing saga state is found. Handling the messages may involve changing the state or completing the saga.  It's important to note that enabling pessimistic concurrency for the persister does not impact the optimistic concurrency of saga creation. See below for how those scenarios are handled.
```

PR #6417, opened 2024-01-23
Update cooperative cancellation sample

TravisNickels:


samples/cooperative-cancellation/sample.md
```diff
@@ -1,23 +1,24 @@
...
 component: Core
 ---
 
-
 Run the solution. A single console application starts up: `Server`.
 
 ## Code walk-through
 
-When the endpoint is started a `CancellationToken` is passed in from a `CancellationTokenSource`
+When the endpoint is started, a message is sent to the endpoint, which triggers a long-running message handler that enters a loop simulating a task that takes 3 seconds to execute, like writing a bunch of records to a database, by calling the async `Task.Delay()` operation.
```
DavidBoike, line 12: Not quite accurate, right? You've made it take infinite time to execute, but just log something every 3 seconds.

samples/cooperative-cancellation/Core_9/Server/LongRunningMessageHandler.cs
```diff
@@ -12,11 +13,21 @@ public async Task Handle(LongRunningMessage message, IMessageHandlerContext cont
     {
         log.Info($"Received message {message.DataId}. Entering loop.");
 
-        do
+        // The try-catch block is only for the purpose of demonstrating the sample.
+        try
+        {
+            while (true)
+            {
+                log.Info("Press any key to cancel the 3 second delay loop and stop the endpoint.");
```
DavidBoike, line 21: In my opinion, I'd take out the try-catch, and change this message to "Handler still running. Press any key to forcibly stop the endpoint."

TravisNickels: Yeah, I agree with that.  I was a little hesitant including the error handling in the handler since we explicitly state customers shouldn't do that.  I think you're right, it's better to just illustrate the proper way to implement it.  I'll make the appropriate changes.

samples/cooperative-cancellation/sample.md
```diff
@@ -1,23 +1,22 @@
...
 
-
 Run the solution. A single console application starts up: `Server`.
 
 ## Code walk-through
 
-When the endpoint is started a `CancellationToken` is passed in from a `CancellationTokenSource`
-
-snippet: StartingEndpointWithCancellationToken
+When the endpoint is started, a message is sent to the endpoint, which triggers a long-running message handler.  The handler enters an infinite loop, logging a message every two seconds, by calling the async `Task.Delay()` operation.  The `CancellationToken`, that is available in the context object where messages are being handled, is passed to `Task.Delay()` to cancel the delay operation if the `context.CancellationToken.IsCancellationRequested` property is set to `true`.
```
DavidBoike, line 12:
```suggestion
When the endpoint is started, a message is sent to the endpoint, which triggers a long-running message handler.  The handler enters an infinite loop, logging a message every two seconds, by calling the async `Task.Delay()` operation.  The `CancellationToken` provided by the message handling context object is passed to `Task.Delay()` to cancel the delay operation if the `context.CancellationToken.IsCancellationRequested` property is set to `true`.
```

DavidBoike (approved):

PR #6452, opened 2024-02-15
Use a shim to export recoverability and processing metrics to Application Insights via open telemetry

andreasohlund:
- The custom exporter is no longer needed since the one provided by MS now supports metrics as well


andreasohlund:
@DavidBoike like we discussed yesterday, do you think we need a graph as well or is it enough to mention critical time and processing time?

@lailabougria I wasn't sure if a histogram was the right abstraction to use for this?

samples/open-telemetry/application-insights/Core_8/Endpoint/Program.cs
```diff
@@ -60,6 +60,10 @@ static async Task Main(string[] args)
         endpointConfiguration.EnableOpenTelemetry();
         #endregion
 
+        #region enable-metrics-shim
+        endpointConfiguration.EnableMetricsShim();
```
lailabougria, line 64: Shouldn't this be auto-enabled if you're opting into Otel? Otherwise your endpointconfiguration becomes cluttered with the exact same intent.

andreasohlund: That was the original plan but hooking up to the retries requires access to the endpoint configuration and I couldn't find a workaround :(

andreasohlund: see https://github.com/Particular/docs.particular.net/pull/6452/files#diff-4efc07ff98883f735fa775086dce63c4935090f74a55b087de83d7eba86f9247R12

lailabougria: Let's discuss wednesday ;)

andreasohlund: @lailabougria I was able to auto-enable it, please take another look

andreasohlund:
@DavidBoike preview image added in the header + dedicated image for critical time, see if it does the job?

@lailabougria can you give this a review?

DavidBoike:
I can't review on a technical basis but it looks like from the visuals in there it will satisfy the marketing requirements.

bbrandt:
@andreasohlund This is the same shim we'd want to add to https://docs.particular.net/samples/open-telemetry/prometheus-grafana/ if we wanted nservicebus.messaging.criticaltime, nservicebus.messaging.processingtime, nservicebus.messaging.sent-to-error, and nservicebus.messaging.retries sent to Prometheus, right?

samples/open-telemetry/application-insights/Core_8/Endpoint/MetricsShim/EmitNServiceBusMetrics.cs
```diff
@@ -0,0 +1,108 @@
...
+using NServiceBus;
+using NServiceBus.Features;
+using NServiceBus.Settings;
+
+#region metrics-shim
+class EmitNServiceBusMetrics : Feature
+{
+    public EmitNServiceBusMetrics()
+    {
+        EnableByDefault();
```
bbrandt, line 16: What NSB configurations might this Feature be incompatible with?  Send-only endpoints?  Others?

samples/open-telemetry/application-insights/sample.md
```diff
@@ -58,26 +78,18 @@ snippet: enable-open-telemetry
...
-
-The custom meter exporter captures all meters that begin with `nservicebus.` of the `long` sum type, and forwards them to a `TelemetryClient`.
-
-snippet: custom-meter-exporter
-
-This exporter is installed into the meter provider builder with a custom extension method. The exporter is wrapped by a `PeriodicExportingMetricReader` instance that executes the exporter once every 10 seconds.
+#### Critical time and processing time
 
-snippet: custom-meter-exporter-installation
+[Critical time and processing time captured by the metrics package](/monitoring/metrics/definitions.md#metrics-captured) is not yet included in the built-in open telemetry support so a shim is needed to make sure that the endpoint exposes them as open telemetry metrics.
```
lailabougria, line 93:
```suggestion
[Critical time and processing time captured by the metrics package](/monitoring/metrics/definitions.md#metrics-captured) is not yet supported in OpenTelemetry's native format (using System.Diagnostics), so a shim is required to expose them as OpenTelemetry metrics.
```

samples/open-telemetry/application-insights/Core_8/Endpoint/MetricsShim/EmitNServiceBusMetrics.cs
lailabougria: I don't think there's anything incompatible; even though enabled on a send-only endpoint, it just wouldn't have any effect because no messages are consumed there.
The `OnReceivePipelineCompleted` would never be hit, and you'd never see retries, critical time, or processing time. There's also no reason to enable the shim on a SendOnly Endpoint, as it would give no added value.

samples/open-telemetry/application-insights/sample.md
```diff
@@ -37,19 +38,38 @@ Note: Although the sample uses Azure Application Insights, the solution itself d
...
-  - `nservicebus.messaging.fetches`
-
-![Graph tracking success and failed metrics in Application Insights](metrics-dashboard.png)
+Navigate to _Monitoring_ → _Metrics_ on the Azure portal dashboard for the configured Application Insight instance to start creating graphs.
 
 NOTE: It may take a few minutes for the meter data to populate to Azure. Meters will only appear on the dashboard once they have reported at least one value.
 
+#### Message processing counters
+
+To monitor the rate of messages being fetched from the queuing system, processed successfully, retried, and failed for the endpoint add the following metrics to a line chart.
```
lailabougria, line 47:
```suggestion
Add the following metrics to a line chart to monitor the rate of messages being fetched from the queuing system, processed successfully, retried, and failed for the endpoint:
```

samples/open-telemetry/application-insights/sample.md
```diff
@@ -37,19 +38,38 @@ Note: Although the sample uses Azure Application Insights, the solution itself d
...
+
+- `nservicebus.messaging.fetches`
+- `nservicebus.messaging.successes`
+- `nservicebus.messaging.failures`
+
+![Graph showing fetched, success, and failed counters in Application Insights](metrics-counters.png)
+
+#### Recoverability
+
+To monitor [recoverability](/nservicebus/recoverability/) metrics add a line graph with the following metrics.
```
lailabougria, line 57:
```suggestion
To monitor [recoverability](/nservicebus/recoverability/) metrics, add a line graph with the following metrics:
```

samples/open-telemetry/application-insights/sample.md
```diff
@@ -37,19 +38,38 @@ Note: Although the sample uses Azure Application Insights, the solution itself d
...
+
+- `nservicebus.messaging.retries`
+- `nservicebus.messaging.sent-to-error`
+
+![Graph showing recoverability metrics in Application Insights](metrics-recoverability.png)
+
+#### Critical time and processing time
+
+To monitor [critical time and processing time](/monitoring/metrics/definitions.md#metrics-captured) (in milliseconds) for messages being successfully processed by the endpoint add a line graph with the following metrics.
```
lailabougria, line 66:
```suggestion
To monitor [critical time and processing time](/monitoring/metrics/definitions.md#metrics-captured) (in milliseconds) for successfully processed messages, add a line graph with the following metrics:
```

samples/open-telemetry/application-insights/sample.md
```diff
@@ -58,26 +78,18 @@ snippet: enable-open-telemetry
 
 ### Tracing
 
-The endpoint configures an OpenTelemetry trace provider that includes the `NServiceBus.Core` source and exports traces to Azure Monitor.
+The endpoint configures an OpenTelemetry trace provider that includes the `NServiceBus.Core` source and export traces to Azure Monitor.
```
lailabougria, line 81:
```suggestion
The endpoint configures an OpenTelemetry trace provider that includes the `NServiceBus.Core` source and exports collected traces to Azure Monitor.
```

samples/open-telemetry/application-insights/sample.md
```diff
@@ -58,26 +78,18 @@ snippet: enable-open-telemetry
...
 
 snippet: enable-tracing
 
 ### Meters
 
-The endpoint also configures an OpenTelemetry meter provider that includes the `NServiceBus.Core` meter and exports data to Azure Monitor.
+The endpoint also configures an OpenTelemetry meter provider that includes the `NServiceBus.Core` meter and export data to Azure Monitor.
```
lailabougria, line 87:
```suggestion
The endpoint configures an OpenTelemetry meter provider that includes the `NServiceBus.Core` meter and exports metric data to Azure Monitor.
```

samples/open-telemetry/application-insights/sample.md
```diff
@@ -58,26 +78,18 @@ snippet: enable-open-telemetry
...
-
-snippet: custom-meter-exporter
-
-This exporter is installed into the meter provider builder with a custom extension method. The exporter is wrapped by a `PeriodicExportingMetricReader` instance that executes the exporter once every 10 seconds.
+#### Critical time and processing time
 
-snippet: custom-meter-exporter-installation
+[Critical time and processing time captured by the metrics package](/monitoring/metrics/definitions.md#metrics-captured) is not yet included in the built-in open telemetry support so a shim is needed to make sure that the endpoint exposes them as open telemetry metrics.
 
-NOTE: The shim passes `QueueName` as a custom dimension which allows filtering the graphs in Application Insights. Multi-dimensional metrics are not enabled by default. Check [the Azure Monitor documentation](https://docs.microsoft.com/en-us/azure/azure-monitor/app/get-metric#enable-multi-dimensional-metrics) for instructions on how to enable this feature.
```
lailabougria, line 83: I'm not sure this should be entirely removed. 

From MS docs:
> All metrics that you send using [OpenTelemetry](https://learn.microsoft.com/en-us/azure/azure-monitor/app/opentelemetry-add-modify), [trackMetric](https://learn.microsoft.com/en-us/azure/azure-monitor/app/api-custom-events-metrics#trackmetric) or [GetMetric and TrackValue](https://learn.microsoft.com/en-us/azure/azure-monitor/app/api-custom-events-metrics#getmetric) API calls are automatically stored in both logs and metrics stores. These metrics can be found in the customMetrics table in Application Insights and in Metrics Explorer under the Custom Metric Namespace called "azure.applicationinsights". Although the log-based version of your custom metric always retains all dimensions, the pre-aggregated version of the metric is stored by default with no dimensions. Retaining dimensions of custom metrics is a Preview feature that can be turned on from the [Usage and estimated cost](https://learn.microsoft.com/en-us/azure/azure-monitor/cost-usage#usage-and-estimated-costs) tab by selecting With dimensions under Send custom metrics to Azure Metric Store.

Maybe we should rephrase it and link to this: https://learn.microsoft.com/en-us/azure/azure-monitor/app/pre-aggregated-metrics-log-metrics#custom-metrics-dimensions-and-pre-aggregation

samples/open-telemetry/application-insights/Core_8/Endpoint/MetricsShim/EmitNServiceBusMetrics.cs
bbrandt: Thanks, Laila!  I am mostly asking because we have our own simple common NuGet packages that wrap up things like this we distribute out to all of our microservice teams and if they include the NuGet package in a SendOnly endpoint and the feature is set to `EnableByDefault()`, there's not really a decision point to say, don't turn it on in this context.  But that is all fine if it gets enabled and has no effect that is not an issue.  

More likely, we will change our copy of this code to have the feature off by default and enabled explicitly by each service.  But when a team says its not working, then the first question will be, "Did you turn it on?"  :-)

bbrandt:
@lailabougria @andreasohlund Supporting a high throughput SaaS product in production, I believe the 2 most important dimensions/labels for these metrics would be Service Name (or Cloud Role Name) and message (or handler) type.  If a service handles multiple messages or has a saga with multiple handlers, I think the only way these metrics could be useful is if they are able to reveal which of my handlers is the problem.  If we were tracking a "Time in Queue" metric,  then Message/Handler type would not affect this metric, but both processing time and critical time will vary per handler.

andreasohlund:
Thanks for the input @bbrandt, it would be nice to talk a bit more about you're requirements, my colleague @lailabougria has reached out to you over linked in to see if we can arrange a call to talk some more about your needs?

bbrandt:
> Thanks for the input @bbrandt, it would be nice to talk a bit more about you're requirements, my colleague @lailabougria has reached out to you over linked in to see if we can arrange a call to talk some more about your needs?

Thanks for meeting with me @andreasohlund and @lailabougria!  As I mentioned I either totally missed that you already have message type as a discriminator or forgot between the time I read the code and wrote my comment.  

For anyone else following along [e.TryGetMessageType(out var messageType)](https://github.com/Particular/docs.particular.net/pull/6452/files#diff-113873bb52eeddf15080df4b1906a3b08e17de1506959fdf9ade6a66d5be057bR35) is where message type is read and [nservicebus.message_type](https://github.com/Particular/docs.particular.net/pull/6452/files#diff-113873bb52eeddf15080df4b1906a3b08e17de1506959fdf9ade6a66d5be057bR119) is the tag name.  

We do not have multiple handlers that would process the same message type polymorphically, so having the `nservicebus.message_type` completely meets our need I described.  Thanks!

samples/open-telemetry/application-insights/sample.md
andreasohlund: Added the note back in

samples/open-telemetry/application-insights/Core_8/Endpoint/MetricsShim/EmitNServiceBusMetrics.cs
andreasohlund: Good catch @bbrandt I have updated the shim to handle send only endpoints now for completeness

PR #6463, opened 2024-02-20
Review ASB transaction support

danielmarbach:


danielmarbach:
Sorry I forgot we don't have strict enforcement for reviews here. I shouldn't have set it to auto-merge. If you spot anything, please make a comment and I'll update things in a separate PR.

PR #6475, opened 2024-03-01
ServicePulse in IIS Subfolder

tmasternak:


mauroservienti (approved):

poornimanayar (approved):

servicepulse/install-servicepulse-in-iis.md
```diff
@@ -66,14 +66,6 @@ NOTE: If TLS is to be applied to ServicePulse then ServiceControl also must be c
...
           <action type="Rewrite" url="/js/app.constants.js{R:1}" />
         </rule>
-        <rule name="Handle Vue.js routing paths" stopProcessing="true">
-          <match url="(.*)" />
-          <conditions logicalGrouping="MatchAll">
-            <add input="{REQUEST_FILENAME}" matchType="IsFile" negate="true" />
-            <add input="{REQUEST_FILENAME}" matchType="IsDirectory" negate="true" />
-          </conditions>
-          <action type="Rewrite" url="/" />
-        </rule>
```
mikeminutillo, line 76: Was this rule ever needed? If I understand it correctly, it is saying "if the request is not for an existing file or folder then route the request back to the root". Under what circumstances would this have been hit?

mikeminutillo (approved):

servicepulse/install-servicepulse-in-iis.md
tmasternak: I think this, was based on the assumption that ServicePulse links in Vue.js will not have `#` in the content. In such a case, on full page reload e.g. for `/dashboard` URL, we wanted to make sure that the server will serve `/` which in IIS case is `index.html`.

PR #6493, opened 2024-03-14
Scripted addition of net8.0 to all currently multi-targeted samples

DavidBoike:


ramonsmits (approved):
Seems only `csproj` are updated and `net8.0` got added. Github diff view isn't really performing for me in Chrome but the build is green and is this is the only change 👍

tamararivera (approved):

PR #6508, opened 2024-03-19
NHibernate custom saga finder: Target all supported versions of .NET, fix NHibernate configuration issue and remove double connection string init

ramonsmits:


samples/saga/nh-custom-sagafinder/NHibernate_8/Sample/SqlHelper.cs
```diff
@@ -1,4 +1,5 @@
-﻿using Microsoft.Data.SqlClient;
+﻿//using System.Data.SqlClient;
+using Microsoft.Data.SqlClient;
```
DavidBoike, line 2:
```suggestion
using Microsoft.Data.SqlClient;
```

PR #6512, opened 2024-03-19
Update realtime sample to support net6-8

DavidBoike:
* Use net6-8 throughout
* Administrative access isn't required for a localhost URL in net6+
* Copied the Core_8 sample back to Core_7 and adjusted the versions/code so we don't have to use the crufty OWIN stack and can have the sample on net6+ not net481


samples/near-realtime-clients/Core_7/ClientHub/StockTickHandler.cs
```diff
@@ -1,20 +1,28 @@
...
+    public StockTickHandler(IHubContext<StockTicksHub> hub)
+    {
+        this.hub = hub;
+    }
 
     public Task Handle(StockTick message, IMessageHandlerContext context)
     {
         log.Info($"StockTick event received for Symbol {message.Symbol} with timestamp: {message.Timestamp:O}. Press any key to exit.");
-        return hub.Clients.All.PushStockTick(message);
+        return hub.Clients.All.SendAsync("PushStockTick", message);
```
ramonsmits, line 25: Shouldn't this be in its own handler where Handle would first publish a StockUpdated event? It would show that you could have a "SignalR" channel handler but optionally also others.

samples/near-realtime-clients/Core_7/ClientHub/StockTickHandler.cs
```diff
@@ -1,20 +1,28 @@
...
+
+
 #region StockTickHandler
 
 public class StockTickHandler :
     IHandleMessages<StockTick>
 {
-    static ILog log = LogManager.GetLogger<StockTickHandler>();
-    IHubContext hub = GlobalHost.ConnectionManager.GetHubContext<StockTicksHub>();
+    static readonly ILog log = LogManager.GetLogger<StockTickHandler>();
```
ramonsmits, line 14: Ideally our samples should use ILogger

samples/near-realtime-clients/Core_7/Publisher/Program.cs
```diff
@@ -1,15 +1,14 @@
...
 static class Program
 {
     static async Task Main()
     {
-        Console.Title = "Samples.NearRealTimeClients.Publisher";
+        Console.Title = "Publisher";
         var endpointConfiguration = new EndpointConfiguration("Samples.NearRealTimeClients.Publisher");
-        endpointConfiguration.UsePersistence<InMemoryPersistence>();
-        endpointConfiguration.UseTransport<MsmqTransport>();
+        endpointConfiguration.UseTransport<LearningTransport>();
```
ramonsmits, line 11: Not using MSMQ is already a win ;-)

ramonsmits (approved):

DavidBoike:
Yeah I'm not here to make the sample perfect, only get it into net6-8. :wink:

PR #6514, opened 2024-03-19
New test for console titles too long

DavidBoike:
Now that Windows Terminal is commonplace, the long "namespaced" console title values like "Samples.SampleName.EndpointName" are now too long to be viewed within the Windows Terminal tab. All you see is 3 different windows all called "Samples.WhateverTheSampleNam…" and it gets cut off before the useful bit.

This PR adds a test to ensure that they do not contain periods (so not in that namespaced format) as as well as limits them to 26 characters.

And then it fixes everything.


ramonsmits (approved):
Nice change!

Issue #4676, opened 2020-02-05
Feedback: 'Support Policy'


boblangley:
Feedback for 'Support Policy' https://docs.particular.net/nservicebus/upgrades/support-policy

Location in GitHub: https://github.com/Particular/docs.particular.net/blob/master/nservicebus/upgrades/support-policy.md

Support policies and security advisories are not easily discoverable under upgrade guides and probably should be a top level menu item.

PR #6528, opened 2024-03-23
reorder support policy and security advisory in left nav

soujay:
- Related to  https://github.com/Particular/docs.particular.net/issues/4676

In this PR 
- the NSB support policy has been moved up to its own level and the support policy of ServiceControl has moved under the "Introduction" sub menu.
- the security advisories are not easily discoverable under upgrade guides and is moved one level higher in the left menu.


helenktsai (approved):

tamararivera (approved):
I tested this branch locally and it works perfect.
I would add somethings that I noticed and they are completely out of scope.
There is an index page for the Security Advisories that has the same name, so on the site navigation panel you see:
image
Which I found odd and confusing.
Also the site navigation is not sorted in any specific way, maybe it should be sorted by name or by relevance, but we should note that somehow.

soujay:
@tamararivera 

I renamed the Security Advisories"  landing page to "About"

helenktsai:
Reverted by https://github.com/Particular/docs.particular.net/commit/b91a2addb2b75fb557d58dbff5d191b5e6b9809c per [discussion in Slack](https://particularsoftware.slack.com/archives/C06MC3ZDK0S/p1711723078764929).

PR #6543, opened 2024-03-29
Reorder security advisories in left nav

soujay:
- Related to  https://github.com/Particular/docs.particular.net/issues/4676

In this PR 

- the security advisories are not easily discoverable under upgrade guides and is moved one level higher in the left menu.


tamararivera (approved):

helenktsai:
A couple of questions: 

1. Is renaming and putting under the nservicebus directory necessary? I thought one of David's comments was about not necessarily mapping the left-nav menu to the file system. In addition, as [Tamara comments in Slack](https://particularsoftware.slack.com/archives/C06MC3ZDK0S/p1712008312348199), not all of the security advisories belong to NserviceBus anyway.
2. Assuming we still want to move forward with putting them under the nservicebus directory, is it working on your local machine? I'm trying to run the branch locally and the links all break for me.

menu/menu.yaml
```diff
@@ -687,6 +679,23 @@
       Title: About security and NServiceBus
     - Url: nservicebus/security/generating-encryption-keys
       Title: Generating secure random strong encryption keys
+
```
helenktsai, line 682:
```suggestion
```

menu/menu.yaml
```diff
@@ -687,6 +679,23 @@
...
+    - Url: security-advisories/servicepulse-directorytraversal
+      Title: 2020-03-11
+    - Url: security-advisories/asq-alias-for-default-connectionstring
+      Title: 2018-01-19
+    - Url: security-advisories/msmq-permissions
+      Title: 2017-01-10
+    - Url: security-advisories/sqlserver-sqlinjection
+      Title: 2016-07-05
+
+
```
helenktsai, line 698:
```suggestion
```

helenktsai (approved):
I made smol suggestions to get the format consistent.

Also, the more I think about it, the more I think moving it up even one level higher makes more sense. However, hunting down an icon might be just reaching out of scope for this change, so I am just putting the comment here for posterity. 

image

PR #6573, opened 2024-04-12
Sagas, small guidance improvements

ramonsmits:


nservicebus/handlers-and-sagas.md
```diff
@@ -11,9 +11,9 @@ related:
 
 There are two standard ways of executing code when a message is processed: [handlers](handlers/) and [sagas](sagas/).
 
-Handler instances are instantiated on a per-message basis, executed, and then disposed of.
+Handler instances are instantiated on a per-message basis, executed, and then disposed of. These are often referred as stateless handlers.
```
adamralph, line 14:
```suggestion
Handler instances are instantiated on a per-message basis, executed, and then disposed of. These are sometimes referred to as “stateless handlers”.
```

nservicebus/handlers-and-sagas.md
```diff
@@ -11,9 +11,9 @@ related:
...
 
-Saga instances are also instantiated on a per-message basis, executed, and then disposed of. However they differ from handlers in that, once instantiated, they are passed an instance of a "Data" class. The "Saga Data" is persistent state that is shared between a given saga type based on a key.
+Saga instances are also instantiated on a per-message basis, executed, and then disposed of. However they differ from handlers in that, once instantiated, they are passed an instance of a "Data" class. The "Saga Data" is persistent state that is shared between a given saga type based on a key. These are often referred as statefull handlers.
```
adamralph, line 16:
```suggestion
Saga instances are also instantiated on a per-message basis, executed, and then disposed of. However they differ from handlers in that, once instantiated, they are passed an instance of a "Data" class. The "Saga Data" is persistent state that is shared between a given saga type based on a key. These are sometimes referred to as “stateful handlers”.
```

nservicebus/sagas/timeouts.md
```diff
@@ -34,13 +34,13 @@ Multiple timeouts can be requested when processing a message. The individual tim
 
 snippet: saga-multiple-timeouts
 
-NOTE: Sending a timeout with the same data multiple times will result in that timeout being processed multiple times by the saga.
+NOTE: As timeouts are just messages sending a multiple timeouts with the same state will result in that timeout being processed multiple times by the saga.
```
adamralph, line 37: Another idea:

```suggestion
## Changing timeouts

After a timeout has been requested, it cannot be changed. Requesting a timeout with the same state again, but with a different duration, will not change the original timeout. The original timeout and the subsequent timeout will both be processed by the saga.
```

nservicebus/sagas/timeouts.md
```diff
@@ -34,13 +34,13 @@ Multiple timeouts can be requested when processing a message. The individual tim
...
 
 ## Revoking timeouts
 
 A timeout that has been scheduled cannot be revoked. This means that when the timeout timestamp has elapsed then this timeout message will be queued and then processed.
 
-A timeout is a regular message and once requested, a timeout message can already be in transit or queued. Once that has happened there is no way to revoke (delete) a timeout. It is common to perform a state check in a timeout handler to see if the timeout is still applicable for processing.
+A timeout is a regular message and once requested, a timeout message can already be in transit or queued. Once that has happened there is no way to revoke (delete) a timeout. Sending a timeout will not invalidate/revoke any prior sent timeout. It is common to perform a state check (e.g. comparing a version number, a timestamp or a status field) in a timeout handler to see if the timeout is still applicable for processing.
```
adamralph, line 43: In the context of the above suggestion, I don't think anything needs to change in this section. It can remain purely about revoking timeouts, and all it needs to say is that it cannot be done.

```suggestion
A timeout is a regular message and once requested, a timeout message can already be in transit or queued. Once that has happened there is no way to revoke (delete) a timeout. It is common to perform a state check in a timeout handler to see if the timeout is still applicable for processing. 
```

nservicebus/sagas/timeouts.md
```diff
@@ -34,13 +34,11 @@ Multiple timeouts can be requested when processing a message. The individual tim
 
 snippet: saga-multiple-timeouts
 
-NOTE: Sending a timeout with the same data multiple times will result in that timeout being processed multiple times by the saga.
+## Changing or revoking timeouts
 
-## Revoking timeouts
+After a timeout has been requested, it cannot be changed (i.e. rescheduled) or revoked (i.e. deleted, cancelled). Requesting a timeout with the same state again, but with a different duration or timestamp, will not revoke or ignore the original timeout. The original timeout and the subsequent timeout will both be processed by the saga.
```
adamralph, line 39:
```suggestion
After a timeout has been requested, it cannot be changed (i.e. rescheduled) or revoked (i.e. deleted or cancelled). Requesting a timeout with the same state again, but with a different duration or timestamp, will not revoke or ignore the original timeout. The original timeout and the subsequent timeout will both be processed by the saga.
```

nservicebus/sagas/timeouts.md
```diff
@@ -34,13 +34,11 @@ Multiple timeouts can be requested when processing a message. The individual tim
...
 
-A timeout that has been scheduled cannot be revoked. This means that when the timeout timestamp has elapsed then this timeout message will be queued and then processed.
-
-A timeout is a regular message and once requested, a timeout message can already be in transit or queued. Once that has happened there is no way to revoke (delete) a timeout. It is common to perform a state check in a timeout handler to see if the timeout is still applicable for processing.
+A timeout is a delayed message and once requested, a timeout message can already be in transit or queued. Once that has happened there is no way to revoke (delete) a timeout as the message cannot be deleted. It is common to perform a state check (e.g. compare to state version or a timestamp) in a timeout handler to see if the timeout is still applicable for processing or to ignore the message.
```
adamralph, line 41: Given the changes above, this paragraph is now repeating the fact that you cannot revoke a timeout.

Maybe:

```suggestion
When a saga handles a timeout, it may choose to ignore it, depending on how the saga state has changed since the timeout was requested.
```

adamralph:
@ramonsmits @poornimanayar I went ahead and committed my suggestions and rebased the branch to resolve the conflict.

Any objections to merging?

PR #6629, opened 2024-05-17
Documentation review - Release notification

NChaganlal:


NChaganlal:
Hey @andreasohlund could you confirm the content is also accurate? I think yes, but a second pair of eyes on that wouldn't hurt.

andreasohlund (approved):

PR #6631, opened 2024-05-18
editorial review of recent Lucid->Miro changes

adamralph:
relates to https://github.com/Particular/docs.particular.net/pull/6594

### DIFF
--- a/README.md
+++ b/README.md
@@ -1092,26 +1092,15 @@ There are two css classes (`event` and `message`) that should be applied to mess
 
 #### Miro
 
-Another option is [Miro](https://miro.com/). Miro allows the creation of diagrams on a whiteboard. A board can be saved as a board backup file (.rtb) and can be used to create images.
-
-Do not use a transparent background when creating an image, as this makes the visibility of elements in the image dependent on the user's theme (e.g. light or dark). Always choose a solid background which ensure the elements of the image are visible.
+Another option is [Miro](https://miro.com/). Miro allows the creation of diagrams on a whiteboard. A board can be saved as a board backup file (.rtb) and can be used to create images. Within Miro, we create draw.io diagrams, as these can be exported as PNG images.
 
 To create an image:
 
-- Drag a rectangle around the relevant objects on the board
-- From the “three dots” (…) menu above the selection:
-  - Click “Create frame”
-  - Click “Export as image”
-- Choose the appropriate JPG size and click "Export"
-
-If you want to save it as a `.png`, copy the image to an image editing tool like "paint" and save it as a `.png`.
-
-Alternatively, if you choose to create a draw.io diagram, Miro will allow you to export to `.png`:
-
-- Double-click the draw.io diagram to edit the diagram
+- Double-click the draw.io diagram to open the diagram in the draw.io editor
 - From the "File" menu:
    - Click "Export as"
    - Click "PNG.."
+   - Do NOT select "Transparent Background", as this makes the visibility of elements in the image dependent on the user's theme (e.g. light or dark).
  - Choose the file location and click "Export"
 
 To allow images to be updated later, they must be committed to this repository along with the board backup file(s) use to create them.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1090,7 +1092,17 @@ There are two css classes (`event` and `message`) that should be applied to mess
...
-Another option is using [Miro](https://miro.com/). Miro allows the creation of diagrams in a whiteboard that can be exported and imported as a backup file (RTB) by a board owner. Diagrams created in Miro can be used to generate images that should be committed along with whiteboard backup. The idea is that the whiteboard with the diagrams can be imported in Miro by any team and evolved along with the documentation. To generate diagram images from Miro board, surround the diagram with a frame, select the frame, select the "Export as image" option and export the image as PNG.
+Another option is [Miro](https://miro.com/). Miro allows the creation of diagrams on a whiteboard. A board can be saved as a board backup file (.rtb) and can be used to create images.
+
+To create an image:
+
+- Drag a rectangle around the relevant objects on the board
+- From the “three dots” (…) menu above the selection:
+  - Click “Create frame”
+  - Click “Export as image”
+- Click the appropriate size of JPG and click "Export"
```
adamralph, line 1103: I could find a PNG option, and the [docs](https://help.miro.com/hc/en-us/articles/360017572754-How-to-export-your-board) say that PNG isn't possible.

NChaganlal: Grateful that you spotted it @adamralph !

1. For the standard case, I forgot to include **_a very important_** step which was to copy the image to pain and then save it as a `.png` . 
 
2. There is an alternative way in which Miro will allow to export to .png and that is if the diagram is a draw.io diagram (which we also have [here](https://miro.com/app/board/uXjVNknXdYo=/?moveToWidget=3458764581646656814&cot=14)). You'd have to double-click the diagram and on edit mode export to `.png` (see picture below)
image

As an additional piece of information, we should steer away from generating images with transparent backgrounds as when users have a dark theme enabled, which can be a bigger proportion for engineers, the lines in the diagram can become hard to see therefore generating a bad user experience whilst reading our documentation.

Should we add this alternative way of doing it and highlighting this note about the dark theme on the docs?

NChaganlal (approved):

README.md
```diff
@@ -1090,7 +1092,29 @@ There are two css classes (`event` and `message`) that should be applied to mess
...
+
+- Drag a rectangle around the relevant objects on the board
+- From the “three dots” (…) menu above the selection:
+  - Click “Create frame”
+  - Click “Export as image”
+- Choose the appropriate JPG size and click "Export"
+
+If you want to save it as a `.png`, copy the image to an image editing tool like "paint" and save it as a `.png`.
+
+Alternatively, if you choose to create a draw.io diagram, Miro will allow you to export to `.png`:
```
adamralph, line 1109: Again, rather than "if you choose", it feels like we should give some guidance as to whether to use draw.io within Miro or not. What are the reasons for doing so?

README.md
```diff
@@ -1102,7 +1104,16 @@ To create an image:
   - Click “Export as image”
 - Choose the appropriate JPG size and click "Export"
 
-To allow images to be updated later, they must be committed to this repo along with the board backup file(s) use to create them.
+If you want to save it as a `.png`, copy the image to an image editing tool like "paint" and save it as a `.png`.
```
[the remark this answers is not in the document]
NChaganlal: Most of the pictures already there were in `.png` so I just carried on the trend.

README.md
NChaganlal: I'd rather use draw.io as it allows the diagrams to be cleanly built and it was easier to take advantage of the shapes libraries for architecture than using Miro's libraries which are locked with our current subscription level. However, on my TF there were split opinions as some people felt it was enough to just import icons from the internet as needed. Personally, I'd rather use draw.io within Miro due to the libraries it has available and how easy it is to convert images and to not allow people to accidentally change diagrams without realizing it.

README.md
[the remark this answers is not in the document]
adamralph: Yeah, I think we should be using PNG (although we specifically don't want transparent backgrounds). I think that's enough justification to use draw.io.

I'll update the PR on that basis.

NChaganlal (approved):

PR #6633, opened 2024-05-19
Legacy docs review

lailabougria:
_part of https://github.com/Particular/GeneralPlatformExperience/issues/3135_


lailabougria:
Merging to avoid the same docs from surfacing Monday, if there are any comments, I will handle them through a follow-up PR.

PR #6656, opened 2024-05-27
Mention token credentials in ASQ

danielmarbach:
The ASQ transport has a code-only mode that is already documented. That snippets currently favors the connection string approach in the snippet. Since the clients have multiple constructors, showing all of them could be confusing. I have decided to mention that they should use whatever ctor overload that makes sense for their authentication scenarios.

### DIFF
--- a/transports/azure-storage-queues/configuration_config_asqn_[11,).partial.md
+++ b/transports/azure-storage-queues/configuration_config_asqn_[11,).partial.md
@@ -2,3 +2,5 @@
 Settings should be set when instantiating the transport:
 
 snippet: AzureStorageQueueConfigCodeOnly
+
+Use the `QueueServiceClient`, `BlobServiceClient` or `TableServiceClient` constructor overload that suits the authentication needs when creating the clients passed to the transport.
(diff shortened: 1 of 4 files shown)


transports/azure-storage-queues/configuration_config_asqn_[11,).partial.md
```diff
@@ -2,3 +2,5 @@
 Settings should be set when instantiating the transport:
 
 snippet: AzureStorageQueueConfigCodeOnly
+
+Use the [`QueueServiceClient`](https://learn.microsoft.com/en-us/dotnet/api/azure.storage.queues.queueserviceclient.-ctor?view=azure-dotnet), [`BlobServiceClient`](https://learn.microsoft.com/en-us/dotnet/api/azure.storage.blobs.blobserviceclient.-ctor?view=azure-dotnet) or [`TableServiceClient`](https://learn.microsoft.com/en-us/dotnet/api/azure.data.tables.tableserviceclient.-ctor?view=azure-dotnet) constructor overload that suits the authentication needs when creating the clients passed to the transport.
```
andreasohlund, line 6: Should we pulls this part into an .include and reuse?

danielmarbach: 10 goes out of support soon and 8 has a different version. I was thinking it is not worth it.

PhilBastian (approved):

PR #6661, opened 2024-05-27
CosmosDB: Guidance on embedding partition keys in saga timeouts state

ramonsmits:
This PR added some guidance on embedding partition keys in saga timeouts, as partition information is not automatically added to the timeout message and cannot be extracted from the incoming saga timeout message.


soujay:
@ramonsmits  - It will be good to add in a small code snippet as well.

soujay (approved):

PR #6675, opened 2024-05-29
document changes to functions allowing for identity based connections

PhilBastian:
documentation updates for changes made to better support identity based connections
- https://github.com/Particular/NServiceBus.AzureFunctions.Worker.ServiceBus/pull/514
- https://github.com/Particular/NServiceBus.AzureFunctions.InProcess.ServiceBus/pull/821


nservicebus/hosting/azure-functions-service-bus/identity-local-setting-file.include.md
```diff
@@ -0,0 +1,12 @@
+```json
+{
+  "IsEncrypted": false,
+  "Values": {
+    ...
+    "FooBar__fullyQualifiedNamespace": "<namespace>.servicebus.windows.net",
```
andreasohlund, line 6: FooBar feels a bit off, how about "MyConnectionName__" ?

nservicebus/hosting/azure-functions-service-bus/in-process/index_configuration_ASBFunctions_[4.2,].partial.md
```diff
@@ -0,0 +1,35 @@
+`ServiceBusTriggeredEndpointConfiguration` loads certain configuration values from the Azure Function host environment in the following order:
```
Comment (PhilBastian) [inline nservicebus/hosting/azure-functions-service-bus/in-process/index_configuration_ASBFunctions_[4.2,].partial.md:1]: note that this needs to be changed to next version (4.3?) once released, rather than the current version break on 4.2

andreasohlund (approved):

DavidBoike:
Looking at this PR as part of recent docs changes review, and noticed a few things that concern me.

First off, includes are global on the entire site. So the keys `connection-string-local-setting-file` and `identity-local-setting-file` don't feel specific enough for something that only relates to Azure Functions. (Snippets, on the other hand, are scoped to components)

Also, the `index_configuration_AzureFunctionsWorker` partials seem quite duplicative and could cause maintenance problems in the future. Notice that the `## Configuration`` section through the ordered list is identical, and at the end, both have `include: license-file-local-setting-file`. Seems like this could be inverted so that most of the content is on the actual page and less is stored in partials.

danielmarbach:
@DavidBoike thanks for the feedback. We can look at this next week.

danielmarbach:
> Also, the index_configuration_AzureFunctionsWorker partials seem quite duplicative and could cause maintenance problems in the future. Notice that the ## Configuration`` section through the ordered list is identical, and at the end, both have include: license-file-local-setting-file`. Seems like this could be inverted so that most of the content is on the actual page and less is stored in partials.

We discussed this and it was a deliberate decision. While there are some things that are similar, there are definitely differences, so it made more sense to keep the whole section together for us. 

> First off, includes are global on the entire site. So the keys connection-string-local-setting-file and identity-local-setting-file don't feel specific enough for something that only relates to Azure Functions. (Snippets, on the other hand, are scoped to components)

We are going to address this one. Thanks!

Issue #6585, opened 2024-04-19
Move Platform pages to a top level menu item


kbaley:
The [Service Platform](https://docs.particular.net/platform/) pages are buried under the NServiceBus menu item when they talk about the entire platform. They would be better served as a top-level menu item (possibly above NServiceBus and below Architecture.

Besides the page above and all its subpages, consider moving these pages under the same top-level heading:

- https://docs.particular.net/platform/contributing
- https://docs.particular.net/nservicebus/release-notifications

The new top level menu item will need an icon, probably the one associated with the Particular Service Platform.

soujay:
## 🚀 This is done



### ✔ What has been done



- The [Service Platform](https://docs.particular.net/platform/) pages have been moved as a top-level menu item 

### 🏺 Artifacts



- None

### 🏠 Town hall notes


- None

<sup>[📝](https://github.com/Particular/StaffSuccess/blob/master/github/saved-reply-templates.md#closing-issues) [guidelines](https://github.com/Particular/StaffSuccess/blob/master/guidelines/facilitation.md#closing-issues) version 2024-01-25</sup>

sergioc:
@soujay a couple of comments about the implementation:

- The icon you used is from the company logo. This is the actual icon for the platform: https://github.com/Particular/DeveloperEducation/blob/master/brand-style-guide/Icons/SVG/Particular_Service_Platform.svg
- There's no need to spell out `Particular Service Platform`.  Just `Service Platform` should be enough
- It shouldn't be necessary to expand a lonely "General" sub-category under the Service Platform in order to access actual content. As it is now, it doesn't make the content truly top-level.

soujay:
@sergioc - I will open another PR to apply your suggestions

PR #6680, opened 2024-05-31
move the platform pages to its own top level menu

soujay:
Address the issue: 
- https://github.com/Particular/docs.particular.net/issues/6585

### DIFF
--- a/menu/menu.yaml
+++ b/menu/menu.yaml
@@ -127,7 +127,7 @@
     - Url: platform/installation
       Title: Platform installation
     - Url: platform/platform-sample-package
-      Title: Platform Sample Installation
+      Title: Platform Sample installation
     - Url: platform/connecting
       Title: Connecting endpoints
     - Url: platform/json-schema
--- a/menu/menu.yaml
+++ b/menu/menu.yaml
@@ -118,6 +118,21 @@
       - Title: Observability
         Url: architecture/aws/observability
 
+- Name: Service Platform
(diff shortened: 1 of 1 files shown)


menu/menu.yaml
```diff
@@ -118,6 +118,26 @@
       - Title: Observability
         Url: architecture/aws/observability
 
+- Name: Particular Service Platform
+  Topics:
+  - Title: General
+    Articles: 
+    - Url: platform
+      Title: About the Service Platform
```
kbaley, line 126: On second thought, maybe `About the platform` is less awkward. It's under a menu that has the full platform name so it should be obvious what platform it refers to

menu/menu.yaml
```diff
@@ -118,6 +118,26 @@
...
+    - Url: platform/installation
+      Title: Platform Installation
```
kbaley, line 128:
```suggestion
      Title: Platform installation
```

menu/menu.yaml
```diff
@@ -118,6 +118,26 @@
...
+    - Url: platform/platform-sample-package
+      Title: Platform Sample Installation
```
kbaley, line 130:
```suggestion
      Title: Platform Sample installation
```
I'm on the fence on whether Sample needs capitalization here but installation definitely doesn't to match the rest of the menu items

kbaley (approved):

PR #6696, opened 2024-06-12
Document use of bridge to create single 360 platform view for multi-transport scenarios

andreasohlund:


NChaganlal (approved):

servicecontrol/servicecontrol-instances/remotes.md
```diff
@@ -116,6 +116,56 @@ class ServiceControl ServiceControlPrimary
 class ServiceControlAudit1,ServiceControlAudit2 ServiceControlRemote
 ```
 
+### Multi-transport deployments
+
+When using multiple transports the ServiceControl can be setup together with the [MessagingBridge](/nservicebus/bridge/) to allow the entire system to be managed via a single ServicePulse/ServiceInsight installation.
```
adamralph, line 121:
```suggestion
When a system uses multiple transports, the [Messaging Bridge](/nservicebus/bridge/) can be used to allow management of the entire system by single instances of ServicePulse and ServiceInsight.
```

adamralph, line 121: Based on https://docs.particular.net/servicecontrol/servicecontrol-instances/remotes#overview-multi-region-deployments, it looks ServiceInsight can connect to the main instance without the use of the bridge, so is this only required for ServicePulse?

servicecontrol/servicecontrol-instances/remotes.md
```diff
@@ -116,6 +116,56 @@ class ServiceControl ServiceControlPrimary
...
+classDef ServiceControlPrimary fill:#A84198,stroke:#92117E,color:#FFFFFF,stroke-width:4px
+classDef ServiceControlRemote fill:#A84198,stroke:#92117E,color:#FFFFFF
+
+class endpointsA,endpointsB,endpointsC Endpoints
+class bridgeB,bridgeC Bridge
+class ServiceInsight ServiceInsight
+class ServicePulse ServicePulse
+class primaryA,primaryB ServiceControlPrimary
+class auditA,auditB,auditC ServiceControlRemote
+```
```
adamralph, line 167: For reference, this renders as:

image

For my own understanding at least, when I retry a message using ServicePulse, how does this work? Wouldn't the bridge have to be somehow used to shuffle the message back to relevant transport and endpoint?

andreasohlund: the bridge creates proxy queues in all transports so the retry goes to a local queue in transport A that the bridge then picks up and moves to the corresponding queue in the transport its bridging. Does that make sense?

servicecontrol/servicecontrol-instances/remotes.md
andreasohlund: yes that is correct but I figured that this is the way to go for both SC and SI so that wording is kind of correct?

servicecontrol/servicecontrol-instances/remotes.md
andreasohlund: I think that it would clutter the diagram to much to get that in there?

adamralph: I get it now, thanks! I had a think about how to represent all of that in this diagram neatly but couldn't come up with anything and, in it's current form, it calls out the extra work required (on top of putting the bridge in place for the transports) to use a single ServicePulse, so I guess it makes sense as-is.

servicecontrol/servicecontrol-instances/remotes.md
adamralph: Yeah, that does make sense actually. Adding a single error instance pointing to two other error instances (as shown in that other diagram) would be an _alternate_ way do it, but just for SI, whereas as this is the way to do it for both SI and SP.

adamralph (approved):

PR #6697, opened 2024-06-13
Add documentation for MSMQ 2.0.4

mikeminutillo:
Includes instructions for how to update custom delayed delivery message store implementations.

NOTES:
- The SQL Delayed Delivery Message store included in the package is already updated. If using it, no further action is needed.
- If a custom delayed delivery message store is being used in an endpoint with installers enabled, it is already being initialized. No further action is needed.

### DIFF
--- a/transports/msmq/delayed-delivery_content_msmqtransport_[2.0,).partial.md
+++ b/transports/msmq/delayed-delivery_content_msmqtransport_[2.0,).partial.md
@@ -58,7 +58,7 @@ Defaults to `1` per sec.
 
 Create a class which implements the `IDelayedMessageStore` interface and pass an instance to the `DelayedDeliverySettings` constructor.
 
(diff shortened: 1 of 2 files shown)


transports/msmq/delayed-delivery_content_msmqtransport_[2.0,).partial.md
```diff
@@ -58,6 +58,8 @@ Defaults to `1` per sec.
 
 Create a class which implements the `IDelayedMessageStore` interface and pass an instance to the `DelayedDeliverySettings` constructor.
 
+If the custom store needs to set up some infrastructure (create tables, etc.) then it must implement `IDelayedMessageStoreWithInfrastructure`. This interface includes `IDelayedMessageStore` and adds a method for setting up the infrastructure. This method is called before `IDelayedMessageStore.Initialize()`.
```
tmasternak, line 61:
```suggestion
If the custom store needs to set up some infrastructure (create tables, etc.) then it must implement `IDelayedMessageStoreWithInfrastructure`. This interface extends `IDelayedMessageStore` adding a method for setting up the infrastructure. This new method is called before `IDelayedMessageStore.Initialize()`.
```

tmasternak (approved):

PR #6703, opened 2024-06-17
Improve platform diagram and related links for ServiceControl

NChaganlal:
Changes to the main diagram:
- Monitoring queue has been added
- "Database" inside ServiceControl component has been renamed to "Storage" for better representation and not lead customers to think that there is only one single database.
- Diagram has been ported to Miro and backup file has been added to the folder so it can be evolved.

We made the following setup links more discoverable by adding them as related articles to the ServiceControl installation page:
- https://docs.particular.net/servicecontrol/servicecontrol-instances/hardware
- https://docs.particular.net/servicecontrol/capacity-and-planning

Case that triggered this improvement:
https://discuss.particular.net/t/use-servicecontrol-to-monitor-endpoints-using-different-transports/3942


platform/architecture-overview.png
adamralph, line 1: Sorry about the nitpicking, but I can't unsee the smaller vertical spacing between the last two queues. Can we distribute them evenly? I guess it would be good to centre all four vertically while we do that, so the whole diagram is vertically symmetrical.

NChaganlal: Yes of course!

PR #6707, opened 2024-06-18
Review MSMQ sample

PhilBastian:


samples/msmq/persistence/sample.md
```diff
@@ -25,4 +25,4 @@ snippet: ConfigureMsmqPersistenceEndpoint
 
 
 > [!NOTE]
-> MSMQ Persistence only supports subscription storage so another storage is needed for Timeouts since MSMQ doesn't have native timeout support.
+> MSMQ Persistence only supports subscription storage. A different storage type is required for Timeouts since MSMQ doesn't have native timeout support.
```
mikeminutillo, line 28: I was going to suggest linking to the documentation about storage types when it is mentioned in the second sentence. I can't find one. The closest is here https://docs.particular.net/persistence/order but it's not very specific. I wonder if we're missing a page about using multiple persistences.

Alternatively, it might be worthwhile using the word Persistence in that second sentence. 

Storage Type is a specific term in NServiceBus. https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/Persistence/StorageType.cs

Timeouts _was_ a storage type (it was deprecated [here](https://github.com/Particular/NServiceBus/commit/1aa86268df1616d3faa59de011b4285898038787#diff-413a08575d33a8bcca37d63b90f5af7fc15045c7bff3a97148c57ab87f370202)).

danielmarbach (approved):
Assuming there is no need to cleanup a bit the code in the sample

samples/msmq/persistence/sample.md
danielmarbach, line 28:
```suggestion
> MSMQ Persistence [only supports subscriptions](https://docs.particular.net/persistence/msmq/). A [different persistence is required for delayed delivery](https://docs.particular.net/transports/msmq/delayed-delivery) (e.g. for [saga timeouts](https://docs.particular.net/nservicebus/sagas/timeouts) or [delayed retries](https://docs.particular.net/nservicebus/recoverability/configure-delayed-retries)) since MSMQ doesn't have native delayed delivery support.
```

danielmarbach: or leave out the inner links

```suggestion
> MSMQ Persistence [only supports subscriptions](https://docs.particular.net/persistence/msmq/). A [different persistence is required for delayed delivery](https://docs.particular.net/transports/msmq/delayed-delivery) (e.g. for saga timeouts or delayed retries) since MSMQ doesn't have native delayed delivery support.
```

mikeminutillo (approved):

PR #6717, opened 2024-06-24
Updated guidance around time-to-be-received (TTBR) and auditing

ramonsmits:


transports/postgresql/index.md
```diff
@@ -11,16 +11,17 @@ The PostgreSQL transport implements a message queuing mechanism on top of [Postg
...
+| Scale-out                                          | Competing consumer                                                                                                                                                                 |
+| Scripted Deployment                                | SQL Scripts                                                                                                                                                                        |
+| Installers                                         | Optional                                                                                                                                                                           |
+| Native integration                                 | Supported, see [SQL statements used by the transport](https://github.com/Particular/NServiceBus.SqlServer/blob/master/src/NServiceBus.Transport.PostgreSql/PostgreSqlConstants.cs) |
+| [time-to-be-received](#time-to-be-received) (TTBR) | Storage reclaimed at most 5 minutes after expiration when receiving endpoint is running                                                                                            |
```
SzymonPobiega, line 24: Why 5 minutes? The PostgreSQL transport does not have periodic task that deletes these messages. The only way to remove expired messages is by receiving them.

transports/sql/index.md
```diff
@@ -20,16 +20,17 @@ The SQL Server transport implements a message queuing mechanism on top of [Micro
...
+| Timeouts                                           | Native                                                                                                   |
+| Large message bodies                               | SqlServer can handle arbitrary message size within available resources, very large messages via data bus |
+| Scale-out                                          | Competing consumer                                                                                       |
+| Scripted Deployment                                | Sql Scripts                                                                                              |
+| Installers                                         | Optional                                                                                                 |
+| Native integration                                 | [Supported](native-integration.md)                                                                       |
+| [time-to-be-received](#time-to-be-received) (TTBR) | Storage reclaimed at most 5 minutes after expiration when receiving endpoint is running                  |
```
SzymonPobiega, line 33: Why 5 minutes? SQLT has a (disabled by default) feature to remove the expired messages on startup but it runs once (when the endpoint starts), not periodically.

transports/sql/index.md
```diff
@@ -134,3 +135,10 @@ When the SQL Server transport is used in combination with [NHibernate persistenc
 SQL Server transport supports all [transaction handling modes](/transports/transactions.md), i.e. Transaction scope, receive only, send atomic with receive, and no transactions.
 
 Refer to [transport transactions](/transports/transactions.md) for a detailed explanation of the supported transaction handling modes and available configuration options.
+
+## Time-to-be-received
+
+The SQL transport runs a periodic task that removes expired messages from the queue. The task is first executed when the endpoint starts and is subsequently scheduled to execute 5 minutes after the previous run when the task has been completed. Expired messages are not received from the queue and their disk space will be reclaimed when the periodic task executes.
```
SzymonPobiega, line 141: Not any more. It has been removed because it tended to deadloack with the receive tasks.

nservicebus/operations/auditing.md
```diff
@@ -60,34 +60,41 @@ Audit messages can be handled in a variety of ways: Save them in a database, do
...
+The [Time-to-be-received (TTBR)](/nservicebus/messaging/discard-old-messages.md) for audit messages can be overriden.
 
-> [!WARNING]
-> MSMQ forces the same TimeToBeReceived on all messages in a transaction. Therefore, OverrideTimeToBeReceived is not supported when using the [MSMQ Transport](/transports/msmq/). If OverrideTimeToBeReceived is detected when using MSMQ an exception will be thrown with the following text:
->
-> ```
-> Setting a custom OverrideTimeToBeReceived for audits is not supported on transactional MSMQ
-> ```
+> [!NOTE]
+> What happens to messages for which the TTBR expires is different. Some transports will auto deleted expired message in the backlog of the queue while no messages are dequeued while others will only delete expired messages once the expired message reaches the front of the queue due to messages being requested. More details at [discarding old messages behavior](/nservicebus/messaging/discard-old-messages.md#behavior)
```
SzymonPobiega, line 74: I would rephrase it to "most transports will not delete". To my knowledge, it is only RabbitMQ that can actually do something useful in the audit context to prevent queue from overflowing when SC is down.

SzymonPobiega (changes requested):
I am wondering if it would not be better to invert the documentation and just state "TTBR is processed at the receiver" and (maybe) add small note for RabbitMQ that in addition to that, the broker is able to remove expired messages but it looks only at the head of the queue.

transports/sql/index.md
tmasternak, line 33:
```suggestion
| [time-to-be-received](#time-to-be-received) (TTBR) | Storage reclaimed at most 5 minutes after expiration when receiving endpoint is running                  |
```

Why `5` min? Sql Server transport is purging expired messages on startup, and PostgreSQL is doing that only when messages are received.

ramonsmits: See reasoning in my other comments. So it seems behavior changed somewhere and documentation was not update to reflect that. I'll update this to ensure both old and new behavior is documented correctly.

nservicebus/operations/auditing.md
ramonsmits:
```suggestion
> What happens to messages for which the TTBR expires is different. Most transports will **not** auto deleted expired message in the backlog of the queue and purge when dequeued and are received by the client. More details at [discarding old messages behavior](/nservicebus/messaging/discard-old-messages.md#behavior)
```

transports/sql/index.md
ramonsmits: @SzymonPobiega ☝️

PR #6169, opened 2023-08-04
Improve message type detection documentation

danielmarbach:
Closes https://github.com/Particular/docs.particular.net/issues/6160

Improves the documentation slightly and catches up with the changes introduced in NServiceBus 7.4 in https://github.com/Particular/NServiceBus/pull/5696

PR #6725, opened 2024-06-27
Remove ambiguous mapping rule

timbussmann:
It was actually [discussed a while ago already](https://github.com/Particular/docs.particular.net/pull/6169#discussion_r1284302973) and there was no clear conclusion on how this option would differ from the `FullName`. I ran into this again today, as the wording could make one think that the "name" stands for just the type's name without the namespace (as the fullname option is already mentioned explicitly before), which is not the case (would be nice though...).


danielmarbach:
Hey @timbussmann 
Thanks for the contribution. I do agree in absence of more information it is probably best to remove that bullet since it is confusing as it stands.

danielmarbach (approved):

PhilBastian (approved):

mikeminutillo (approved):

PR #6740, opened 2024-07-03
remove timeout manager references in docs from package versions which no longer support it

PhilBastian:


persistence/sql/multi-tenant_timeoutsandsubscriptions_sqlpersistence_[7,).partial.md
```diff
@@ -0,0 +1,11 @@
+## Connections for subscriptions
+
+When using multi-tenant mode, [subscriptions](subscriptions.md) are stored in a single database if the message transport does not provide this feature (publish/subscribe) natively.
```
danielmarbach, line 3:
```suggestion
When using multi-tenant mode, [subscriptions](subscriptions.md) are stored in a single database if the message transport does not provide native publish/subscribe.
```

danielmarbach (approved):

PR #6744, opened 2024-07-08
include SqlServer message store methods in documentation

PhilBastian:
Since `SqlServerDelayedMessageStore` only exists on the 2.0 release branch of the MSMQ transport, and not on master, it is not easily found by people trying to implement custom stores. This change adds the logic from `SqlServerDelayedMessageStore` to the delayed messages documentation for visibility


mikeminutillo (approved):

mikeminutillo:
This might be verbose but will likely never be updated

danielmarbach (approved):

PR #6753, opened 2024-07-12
Add missing permissions on AmazonSQS Transport

NChaganlal:
Improving documentation based on feedback given on our support case

### DIFF
--- a/transports/sqs/index.md
+++ b/transports/sqs/index.md
@@ -2,7 +2,7 @@
 title: Amazon SQS Transport
 summary: A transport for Amazon Web Services Simple Queue Service.
 component: SQS
-reviewed: 2023-02-10
+reviewed: 2024-07-11
 related:
  - samples/aws/sqs-simple
 redirects:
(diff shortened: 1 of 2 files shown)


tmasternak (approved):

johnsimons:
@NChaganlal @tmasternak We do mention `sqs:sendmessage` for the subscribers, so would it make more sense to have the `sqs:publisher` for the publishers.
The docs on this particular transport don't differentiate between the different roles, publisher vs subscriber, from a security perspective, we only need `sqs:publisher` for the publisher instance, not all instances.

tmasternak:
@johnsimons I think that there are two sets of permissions here:
- permissions needed by an endpoint to run installers and operate - these are items from both of the lists combined
- permission that a queue needs to be able to subscribe to an SNS topic (shown in the json snippet)

The documentation as it is now, does not make any distinction between permissions needed during deployment and runtime, or between endpoints that only send or publish messages. 

With that in mind, I think the current state of the doco is good enough.

PR #6757, opened 2024-07-18
Mention SQS queue name prefix in Monitoring throughput queue setting

jpalac:


jpalac:
Closing as it's not valid - ie the prefix will also be applied on the send

PR #6776, opened 2024-07-26
Review of audit deployment via Powershell

abparticular:


tmasternak (approved):

bording (changes requested):
This change is not correct. The page is detailing the PowerShell Gallery version of the module, and the name of the command is `Get-ServiceControlManagementCommands` in the gallery version, regardless of version.

On the 4.x version of the page, there is a NOTE box that mentions the legacy version of the module that is installed as part of running the 4.x installer.

The PowerShell Gallery version is the preferred version to use and can be used for 4.x as well, so there's no reason to try and document the legacy version at this point.

I would suggest closing this PR, or perhaps removing the snippet change and keeping the reviewed date update.

abparticular:
@bording I've removed the versioning #ifdef in the snippet.  I still think the name needs updating since the current doc site refers to the previous name.

bording:
> @bording I've removed the versioning #ifdef in the snippet. I still think the name needs updating since the current doc site refers to the previous name.

Good call. That change was lost in the #ifdef stuff when I first took a look at the PR!

bording (approved):

Issue #6760, opened 2024-07-19
Feedback: 'Monitoring NServiceBus endpoints with Prometheus and Grafana'


SzymonPobiega:
Feedback for 'Monitoring NServiceBus endpoints with Prometheus and Grafana' https://docs.particular.net/samples/open-telemetry/prometheus-grafana/

Location in GitHub: https://github.com/Particular/docs.particular.net/blob/master/samples/open-telemetry/prometheus-grafana/sample.md

Since NServiceBus 9.1 has support for OpenTelemetry metrics, this sample should be adjusted to take advantage of the new capabilities.

PR #6777, opened 2024-07-26
Update metrics related samples to NServiceBus 9.1

mauroservienti:
Fix https://github.com/Particular/docs.particular.net/issues/6760


lailabougria:
@mauroservienti would you like me to take this off your plate?

I think what needed to be done was:
- rerun the samples so the images can be updated
- change the sample and inline parts in the docs so we can keep the shim in v9, but remove it in v9.1

Did I miss anything?

mauroservienti:
> @mauroservienti would you like me to take this off your plate?

That would be great 😊

> I think what needed to be done was:
> 
> * rerun the samples so the images can be updated

Correct.

> * change the sample and inline parts in the docs so we can keep the shim in v9, but remove it in v9.1

We need to decide if we want to keep samples for v9, too. Usually, we only keep samples for the latest minor.

> Did I miss anything?

We probably need to restructure the OpenTelemetry documentation because the various meter names, for example, are scattered around in multiple places; on many pages, we link to a centralized "Metrics" page where there is no mention of the OpenTelemetry meters. Whenever I read it, I find it puzzling.

SzymonPobiega (approved):

PR #6794, opened 2024-08-06
Disable obsolete warns in snippets

WilliamBZA:


WilliamBZA:
This will fix the current failing builds on master

PR #6812, opened 2024-08-22
Update docs on MSMQ bridge specifics

jpalac:
Update the documentation to:
- remove refence to `@machine` in the QueueAddress
- specify that the bridge needs to be installed on each MSMQ server

### DIFF
--- a/Snippets/Bridge/Bridge_2/Configuration.cs
+++ b/Snippets/Bridge/Bridge_2/Configuration.cs
@@ -196,9 +196,9 @@ public class Configuration
         #region custom-address
 
         var transport = new BridgeTransport(new MsmqTransport());
-        transport.HasEndpoint("Finance", "finance_queue_name");
+        transport.HasEndpoint("Finance", "finance@machinename");
 
-        var endpoint = new BridgeEndpoint("Sales", "sales_queue_name");
+        var endpoint = new BridgeEndpoint("Sales", "sales@machinename");
         transport.HasEndpoint(endpoint);
 
         #endregion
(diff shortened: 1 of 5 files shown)


Snippets/Bridge/Bridge_2/Configuration.cs
```diff
@@ -196,9 +196,9 @@ public void QueueName()
         #region custom-address
 
         var transport = new BridgeTransport(new MsmqTransport());
-        transport.HasEndpoint("Finance", "finance_queue_name");
+        transport.HasEndpoint("Finance", "finance@machinename");
```
andreasohlund, line 199: I wonder if we should keep both examples:

1. Where the endpoint is on a different machine => finance@machinename
2. Where the queue name is different due to https://docs.particular.net/nservicebus/endpoints/specify-endpoint-name#input-queue being used? => finance_queue_name

andreasohlund: And perhaps spell out that 1 is MSMQ specific and that 2 applies to all transports?

andreasohlund: >Where the queue name is different due to https://docs.particular.net/nservicebus/endpoints/specify-endpoint-name#input-queue being used? => finance_queue_name

I've realized that my statement here is wrong, the address override should only be used to control where messages get dispatched to and where subscription messages are sent.

To control the name of the queue that the bridge reads from the endpoint name has to be adjusted since that is what we pass into the raw endpoint that does the reading.

https://github.com/Particular/NServiceBus.MessagingBridge/blob/master/src/NServiceBus.MessagingBridge/EndpointProxyFactory.cs#L28


In short: The text in this PR is ok even though I'm not sure if "machinename" is better than "another-machine"

jpalac: I thought perhaps this PR is no longer relevant since we fixed it so that you can specify the machine name in the queue address, and we decided that just specifying a queue name isn't really supported?

andreasohlund: Got it, so we can close this PR

PR #6815, opened 2024-08-23
Add consistency section for isolated worker migration guide

TravisNickels:


nservicebus/upgrades/azure-functions-service-bus-in-process-isolated-worker.md
```diff
@@ -146,3 +146,27 @@ class HttpSender
...
 }
 ```
+
+## Ensuring consistency in the Isolated Worker model
+
+It's important to be able to maintain the consistency guarantees that were previously available in the in-process model when using `SendsAtomicWithReceive`.  To achieve this, there are a few options:
+
+1. **Implement the [Outbox pattern](/architecture/consistency.md#transactions-outbox-pattern)** to simulate atomic transactions. (Recommended)
+2. Ensure that all receiver message handlers are [idempotent](/architecture/consistency.md#idempotency).
+3. Ensure that the system infrastructure guarantees consistency between business data and messages.
```
andreasohlund, line 156: I'm not sure about this one, do U have some examples of this when running isolated workers?

nservicebus/upgrades/azure-functions-service-bus-in-process-isolated-worker.md
```diff
@@ -146,3 +146,27 @@ class HttpSender
     }
 }
 ```
+
+## Ensuring consistency in the Isolated Worker model
+
+It's important to be able to maintain the consistency guarantees that were previously available in the in-process model when using `SendsAtomicWithReceive`.  To achieve this, there are a few options:
+
+1. **Implement the [Outbox pattern](/architecture/consistency.md#transactions-outbox-pattern)** to simulate atomic transactions. (Recommended)
```
andreasohlund, line 154:
```suggestion
1. **Implement the [Outbox pattern](/architecture/consistency.md#transactions-outbox-pattern)** to enable consistent database and message queue operations. (Recommended)
```

nservicebus/upgrades/azure-functions-service-bus-in-process-isolated-worker.md
```diff
@@ -146,3 +146,27 @@ class HttpSender
...
+
+In the in-process model, `SendsAtomicWithReceive` could be enabled by setting a boolean value to true in the assembly attribute:
+
+```csharp
+[assembly: NServiceBusTriggerFunction("ASBFunctionEndpoint", SendsAtomicWithReceive = true)]
+```
+
+### Changes in the Isolated Worker model
+
+However, in the isolated worker model, the `SendsAtomicWithReceive` attribute is not supported yet.  Therefore, this setting needs to be removed from the assembly attribute.
```
andreasohlund, line 168:
```suggestion
However, in the isolated worker model, the `SendsAtomicWithReceive` attribute is not supported.  Therefore, this setting needs to be removed from the assembly attribute.
```

nservicebus/upgrades/azure-functions-service-bus-in-process-isolated-worker.md
```diff
@@ -146,3 +146,27 @@ class HttpSender
     }
 }
 ```
+
+## Ensuring consistency in the Isolated Worker model
+
+It's important to be able to maintain the consistency guarantees that were previously available in the in-process model when using `SendsAtomicWithReceive`.  To achieve this, there are a few options:
```
andreasohlund, line 152: Should we spell out that this is relevant IF `SendsAtomicWithReceive` is used in the in-process model? (Users have to opt-in to use it)

nservicebus/upgrades/azure-functions-service-bus-in-process-isolated-worker.md
danielmarbach: Maybe we should indicate that this is due to a restriction of Azure Functions and not due to lack of support on our end

nservicebus/upgrades/azure-functions-service-bus-in-process-isolated-worker.md
TravisNickels: Maybe this is my misunderstanding, but we are recommending using Outbox and in the documation for [Outbox](https://docs.particular.net/nservicebus/outbox/) it states that the outbox feature ensures consistency between business data and messages.  Even in [the consistency problem](https://docs.particular.net/nservicebus/outbox/#the-consistency-problem) it states that there are two ways to avoid zombie and ghost messages and one of the ways is to implement infrastructure which guarantees consistency, which describes the outbox feature.  I'm assuming that if the system infrastructure can guarantee consistency between business data and messages, then the outbox feature wouldn't be unnecessary?  Obviously, to me, this seems like the most difficult solution to implement from a users perspective, especially if all they had to do before in the in-process model is add the `SendsAtomicWithReceive` assembly attribute.

nservicebus/upgrades/azure-functions-service-bus-in-process-isolated-worker.md
TravisNickels: Yeah, that makes sense.  I believe I even had a previos version that expressed that, but I wasn't sure if I needed to be that explicit.

nservicebus/upgrades/azure-functions-service-bus-in-process-isolated-worker.md
andreasohlund: Yes, if the infrastructure can handle this for you (eg, using MSDTC or our SQL transport together with storing business data in SQL server), this is a good option as mentioned. 

That said in this case, using functions over Azure ServiceBus there is no infrastructure solution available (and I can't see one ever being provided by MS) so in this context I don't think it's an option for the users.

Does this make sense?

soujay (approved):

nservicebus/upgrades/azure-functions-service-bus-in-process-isolated-worker.md
```diff
@@ -146,3 +146,26 @@ class HttpSender
     }
 }
 ```
+
+## Ensuring consistency in the Isolated Worker model
+
+If `SendsAtomicWithReceive` was previously enabled in the in-process model (note that it is not enabled by default), maintaining that consistency guarantee in the isolated worker model may be important. However, achieving this consistency in the isolated worker model requires more than just adding an assembly attribute. To ensure consistency, consider the following options:
+
+- **Implement the [Outbox pattern](/architecture/consistency.md#transactions-outbox-pattern)** to enable consistent database and message queue operations. (Recommended)
+- Ensure that all receiver message handlers are [idempotent](/architecture/consistency.md#idempotency).
```
andreasohlund, line 155:
```suggestion
If `SendsAtomicWithReceive` was previously [enabled in the in-process model](/nservicebus/hosting/azure-functions-service-bus/in-process#message-consistency) (note that it is not enabled by default), maintaining that consistency guarantee in the isolated worker model is important. 

Lower [transaction modes](/transports/transactions.md#transactions) can result in the duplication of outgoing messages otherwise known as [ghost messages](/nservicebus/concepts/glossary.md#ghost-message). To ensure that [consistency](/architecture/consistency) is maintained make sure that all involved message handlers are [idempotent](https://docs.particular.net/architecture/consistency#idempotency). 
```

andreasohlund (commented):
Thinking this through I realized that its the duplicated messages that is the thing users have to deal with since that is what atomic sends with receive gives.

I've tried to reword the paragraph with this in mind, what do you think? (feel free to change the wording and fix the links as needed)

nservicebus/upgrades/azure-functions-service-bus-in-process-isolated-worker.md
andreasohlund: I was on the fence if we should repeat the options mentioned in https://docs.particular.net/architecture/consistency#idempotency but thought that just linking to them was enough?

TravisNickels: After reviewing this again I think you're right.  At first, I thought we should still mention outbox, but even outbox requires receivers to be idempotent to deal with duplicated messages.

andreasohlund (approved):

danielmarbach (approved):

PR #6824, opened 2024-09-03
Update Upgrading document w/ better, more explicit module install instructions.

OranguTech:
Specifying a minimum version will default to the latest version, which is probably what the user wants. Includes explanation of how to specify a version, and links to the official MS documentation.


DavidBoike:
Hi @OranguTech,

The problem with this PR is that this is an upgrade guide from v4 to v5, and someday when we have a 6.0, using `-MinimumVersion 5.0.0` will result in pulling version 6.

OranguTech:
You're not wrong, but the guide is specifically in the context of "upgrading from 4 to 5". I was assuming any guides for new versions would be modified appropriately. So perhaps it should just  include a `-maximumversion`, (without any of my other additions)  which has its own issue in that technically, each section in the "version" dotnet type can be rather large. (int32 I think, though I haven't tested upper bounds in powershell .) Practically, something like maxversion 5.999.999 or something could work.

Public documentation is tricky because you never know what level of knowledge the reader is coming in with, but if the docs have enough handholding details explaining how to install a module and specify a version, I assume it should have as many guardrails guiding toward the "pit of success" as possible. That is, if someone starting out just grabs the latest version of ServiceControl that that happens to be `5.1.2.3` that day, you don't want following the documentation to explicitly have them install `5.0.0`.

DavidBoike:
Thanks for your feedback. We considered it, and decided to stick with the existing text. If a user ends up on 5.0.0, that's actually OK.

PR #6828, opened 2024-09-07
Add documentation for LostLockException avoidance in azure service bus

TravisNickels:


danielmarbach (approved):
Would it make sense to also mention the option of tweaking the prefetch count and link to that documentation too?

soujay (approved):

PR #6850, opened 2024-09-25
RabbitMQ v10 upgrade guide

danielmarbach:


danielmarbach:
@bording I guess not adding it to the menu just yet is fine for now right?

PR #6854, opened 2024-09-30
add postgresql to "selecting transports" and clarify suitability of SQL brokers

PhilBastian:


transports/selecting.md
```diff
@@ -109,32 +110,58 @@ Azure Storage Queues has fewer features than Azure Service Bus but can be more c
...
+
+## PostgreSQL
+
+Similar to [SQL Server](#sql-server), the PostgreSQL transport implements queues using relational database tables. Each row of a queue table holds one message with an ID, headers, and body.
+
+### Advantages
+
+- PostgreSQL is already present in many organizations, as a relational database that doesn't have the licensing limitations of a commercial solution such as SQL Server or Oracle. This could result in lower licensing and training costs, as well as a reduction in operational risk, since the skills and knowledge required to run PostgreSQL are already present.
+- Mature tooling, such as [pgAdmin](https://www.pgadmin.org) and [Azure Data Studio](https://learn.microsoft.com/en-us/azure-data-studio/download-azure-data-studio)
+- TODO: should precious mention competing consumers, given the known issue about limited connections?
```
PhilBastian, line 148: precious = we, but this causes an exception in the docsengine because reasons

jpalac: Hmm - on one hand it's still possible, as if your system has low throughput then it will never be a problem. But then if you have low throughput then you probably also don't need to scale out.

transports/selecting.md
```diff
@@ -109,32 +110,58 @@ Azure Storage Queues has fewer features than Azure Service Bus but can be more c
...
+- TODO: Does postgres support distributed transactions?
```
jpalac, line 149: looking at the transport code, we don't seem to try and escalate to a distributed transaction, the same way we do for sqlserver

andreasohlund (approved):

jpalac (approved):

PR #6860, opened 2024-10-03
Updated quickstart tutorial to use Microsoft Extensions Logging

poornimanayar:


danielmarbach (approved):
Would it make sense to use partial classes and the source generated loggers or is this too fancy for the quickstart tutorial?

https://learn.microsoft.com/en-us/dotnet/core/extensions/logging?tabs=command-line#get-an-ilogger-from-di

poornimanayar:
I think we could add something in our docs to that effect.

PR #6900, opened 2024-11-06
Docs review andreas

andreasohlund:


monitoring/custom-checks/in-servicepulse.md
```diff
@@ -24,7 +24,7 @@ When a custom check fails, it will continue to make the main Custom Checks badge
 
 Sometimes a custom check reports an error that is easily solved. The status of the custom check will not be updated in ServicePulse until the custom check is executed again.
 
-If it is a one-off custom check, then the endpoint hosting it will need to be restarted to execute the custom check again. If it is a periodic custom check, then it will be automatically executed again after the scheduled period.
+If it is a one-off custom check, then the endpoint hosting will need to be restarted to execute the custom check again. If it is a periodic custom check, then it will be automatically executed again after the scheduled period.
```
poornimanayar, line 27:
```suggestion
If it is a one-off custom check, the endpoint hosting the check will need to be restarted to execute it again. If it is a periodic custom check, it will be automatically executed again after the scheduled period.
```

platform/platform-sample-package.md
```diff
@@ -25,7 +25,7 @@ snippet: Launch
 The Platform Sample package performs the following when the Console App is started:
 
 1. Copies platform binaries into the project output directory.
-1. Finds available ports for all the platform tools.
+1. Find available ports for all the platform tools.
```
poornimanayar, line 28: I think "Finds" is correct here gramatically.

andreasohlund: Good catch, fixed

poornimanayar (approved):

PR #6901, opened 2024-11-07
Documentation review

lailabougria:


transports/azure-service-bus/topology.md
```diff
@@ -110,18 +110,18 @@ The "single topic" design was chosen over a "topic per event type" design due to
...
 
 The need to support polymorphism was the most substantial reason to implement a "single topic" design.
 
-With simple events and without the need to support polymorphism, it is possible to map the contract type to the topic since the relationship is 1:1. Such topologies are mostly built without automatically forwarding messages in subscriptions to the destination queues. The subscription acts in such a scenario as a virtual queue. In cases when subscribers are offline or slow for some time, the quota of the topic might be reached, which can cause all publish operations to the topic to fail. Subscriptions in non-forwarding mode share the topic quota.
+For simple events or no need for polymorphism, it is possible to map the contract type to the topic since the relationship is 1:1. Such topologies are mostly built without automatically forwarding messages in subscriptions to the destination queues. The subscription acts as a virtual queue. In cases when subscribers are offline or slow for some time, the quota of the topic may be reached, which can cause all publish operations to the topic to fail. Subscriptions in non-forwarding mode share the topic quota.
```
poornimanayar, line 118:
```suggestion
For simple events or when support for polymorphism is not needed, it is possible to map the contract type to the topic since the relationship is 1:1. Such topologies are mostly built without automatically forwarding messages in subscriptions to the destination queues. The subscription acts as a virtual queue. In cases when subscribers are offline or slow for some time, the quota of the topic may be reached, which can cause all publish operations to the topic to fail. Subscriptions in non-forwarding mode share the topic quota.
```

poornimanayar (approved):

menu/menu.yaml
```diff
@@ -42,14 +42,6 @@
       Title: "Monitoring NServiceBus Demo - Message backlogs"
     - Url: tutorials/monitoring-demo/walkthrough-3
       Title: "Monitoring NServiceBus Demo - Struggling endpoints"
-  - Title: More tutorials
-    Articles:
-    - Url: tutorials/message-replay
```
andreasohlund, line 47: The folder is still there, are we linking it from other places or should it be deleted?

andreasohlund (approved):

PR #6904, opened 2024-11-14
Link to ASB metrics permission needed

andreasohlund:


johnsimons (approved):
@andreasohlund overall I don't think is a bad idea to align the permissions but I think the throughput-tool requires different permissions then the integrated throughput in ServicePulse, example I don't think the earlier needs "subscription read".
Any thoughts @DavidBoike ?

DavidBoike:
ServiceControl definitely needs more permissions. It would probably be better to just identify the Monitoring Reader role and link to it, if that's indeed all that it needs. Would that be enough permissions to enumerate the queues?

I think it's unlikely that someone would want to create a custom role (something you can't even do without Entra Premium or whatever) to run a once-a-year tool…only for provisioning permanent ServiceControl access via a principal.

andreasohlund:
Monitoring Reader has permission to read queues which is all we need so changing to link directly to that role

johnsimons (approved):

PR #6917, opened 2024-11-29
Remove outdated security advisories during docs review

andreasohlund:


andreasohlund:
As discussed in slack with @mikeminutillo 

Do we need to RFC this?

lailabougria (approved):

mikeminutillo (approved):

PR #6918, opened 2024-11-29
Add some more information for users to consider regarding prefetch and locking

lailabougria:


transports/azure-service-bus/configuration.md
```diff
@@ -60,4 +61,7 @@ To ensure smooth processing, it is recommended to configuring the `MaxAutoLockRe
 partial: lockrenewal
 
 > [!NOTE]
-> Message lock renewal is initiated by client code, not the broker. If the request to renew the lock fails after all the SDK built-in retries (.e.g due to a connection-loss), the lock won't be renewed, and the message will become unlocked and available for processing by competing consumers. Lock renewal should be treated as best-effort and not as a guaranteed operation.
+> Message lock renewal is initiated by client code, not the broker. If the request to renew the lock fails after all the SDK built-in retries (e.g., due to connection loss), the lock won't be renewed, and the message will become unlocked and available for processing by competing consumers. Lock renewal should be treated as a best effort, not as a guaranteed operation.
+
+> [!NOTE]
+> If a message lock renewal is required, it may be worth checking the duration of the handlers, and see whether these can be optimised. In addition, it may be worth checking wither the prefetch count is too high, considering that all messages are locked on peek. This may indicate that too many messages are locked for which the processing exceeds the lock duration.
```
andreasohlund, line 67:
```suggestion
> If message lock renewal is required, it may be worth checking the duration of the handlers, and see whether these can be optimised. In addition, it may be worth checking whether the prefetch count is too high, considering all messages are locked on peek. This may indicate that too many messages are locked for which the processing exceeds the lock duration.
```

andreasohlund (approved):

PR #6919, opened 2024-11-29
Use rabbitmq healtchecks instead of custom solution to check if broker is ready

andreasohlund:


andreasohlund:
All 3 versions have been deployed and tested in docker desktop

PR #6924, opened 2024-12-06
make it clear that the user needs to deal with NServiceBus messages in addition to native messages

PhilBastian:
related to https://github.com/Particular/NServiceBus.AzureStorageQueues/pull/1264


transports/azure-storage-queues/native-integration_sending_asqn_[9,12].partial.md
```diff
@@ -0,0 +1,15 @@
+### Custom envelope unwrapper
```
Comment (jpalac) [inline transports/azure-storage-queues/native-integration_sending_asqn_[9,12].partial.md:1]: This seems to be the same as the paragraph above or am I not seeing the difference?

Comment (PhilBastian) [reply transports/azure-storage-queues/native-integration_sending_asqn_[9,12].partial.md:1]: one less sentence in the warning

Comment (jpalac) [reply transports/azure-storage-queues/native-integration_sending_asqn_[9,12].partial.md:1]: I see it now - wouldn't it be the same in both though it we are backporting the change?

abparticular:
I'm not sure on the structure of the snippet/sample.  Shouldn't it be something more like (pseudocode)

`return CanDeserialize(body) ? new MessageWrapper() : null`

or

```
try
{
  return new Deserialize(body);
}
catch
{
  return null;
}
```

I'm concerned about the focus on checking for base64 (an internal implementation detail) rather than making it clear that if the message can't be handled by the custom unwrapper, then null should be returned.

abparticular (approved):

abparticular (approved):

abparticular (changes requested):
Sorry I didn't mean to approve this PR

abparticular (approved):

jasontaylordev (approved):

Issue #6923, opened 2024-12-05
Feedback: 'ServiceControl Hardware Considerations'


jbogard:
Feedback for 'ServiceControl Hardware Considerations' https://docs.particular.net/servicecontrol/servicecontrol-instances/hardware

Location in GitHub: https://github.com/Particular/docs.particular.net/blob/master/servicecontrol/servicecontrol-instances/hardware.md

This section isn't correct now with container hosting options:

https://docs.particular.net/servicecontrol/servicecontrol-instances/hardware#general-recommendations-hosting-in-the-cloud

PR #6935, opened 2024-12-16
ServiceControl hosting options

ramonsmits:
- Resolves https://github.com/Particular/docs.particular.net/issues/6923


ramonsmits:
@DavidBoike @bording @boblangley is RavenDB Cloud a working/supported option?

SzymonPobiega (approved):

boblangley:
Some of this info is duplicated in https://docs.particular.net/platform/installation#installation-servicecontrol

Perhaps we should adjust both documents to use an include?

mauroservienti (approved):

boblangley:
Should we play the game of calling out specific container hosting applications?

I don't think we should.

Our images are OCI compliant, we shouldn't have much of an opinion about what app is used.

We are also calling out Docker Desktop like it is a production environment, which it isn't.

servicecontrol/hosting.md
```diff
@@ -0,0 +1,68 @@
...
+- [Monitoring instances](/servicecontrol/monitoring-instances/deployment/)
+- [Audit instances](/servicecontrol/audit-instances/deployment/)
+- [Error instances](/servicecontrol/servicecontrol-instances/deployment/)
+
+## Windows Server
+
+Instances can be installed on Windows Server (virtual) machines using:
+
+- Management UI: ServiceControl Management Utility
+- Commandline: ServiceControl Powershell Module
```
boblangley, line 25: We should be downplaying SCMU, as it isn't very devops friendly and promote powershell for Windows more.

ramonsmits: I've never seen a customer use Powershell. I'm not disagreeing with you and change the order but most of our customers especially those that need support are the point and clicky type users.

ramonsmits:
> Should we play the game of calling out specific container hosting applications?
> 
> I don't think we should.
> 
> Our images are OCI compliant, we shouldn't have much of an opinion about what app is used.
> 
> We are also calling out Docker Desktop like it is a production environment, which it isn't.

Not really disagreeing here but its not how users seek information. The majority wants to know does it work for my environment. They don't know if Azure/AWS K8s can run our OCI complient image.

But I'll included that our image is OCI complient

tmasternak (approved):

servicecontrol/hosting.md
```diff
@@ -0,0 +1,47 @@
...
+component: ServiceControl
+reviewed: 2025-03-20
+---
+
+ServiceControl instances can be hosted on:
+
+- Windows Server (virtual machines)
+- Linux containers
+
+
```
mauroservienti, line 13:
```suggestion

```

servicecontrol/hosting.md
```diff
@@ -0,0 +1,47 @@
...
+- [Monitoring instances](/servicecontrol/monitoring-instances/deployment/)
+- [Audit instances](/servicecontrol/audit-instances/deployment/)
+- [Error instances](/servicecontrol/servicecontrol-instances/deployment/)
+
+## Windows Server
+
+Instances can be installed on Windows Server (virtual) machines using:
+
+- Commandline: ServiceControl Powershell Module
+- Management UI: ServiceControl Management Utility
```
mauroservienti, line 25: Can we add links here? The cumbersome thing is that we would have to add different links for each instance type.

mauroservienti (approved):

boblangley:
> Are we providing mission critical support for all containerized environments?

What does that mean?

SzymonPobiega (approved):

PR #6938, opened 2024-12-17
Update layout of thrid party dependencies

johnsimons:


DavidBoike:
Made some adjustments, perhaps there could be more improvements but this is good enough to fix the page for the short term.

PR #6947, opened 2025-01-03
Update for .NET 9 

cquirosj:
Update samples in Docs to add the new .NET version so that a version is available to download for that .NET version on the Docs site.

### DIFF
--- a/.github/workflows/build-samples-and-snippets.yml
+++ b/.github/workflows/build-samples-and-snippets.yml
@@ -16,7 +16,9 @@ jobs:
       - name: Setup .NET SDK
         uses: actions/setup-dotnet@v4.1.0
         with:
-          dotnet-version: 8.0.x
+          dotnet-version: |
+            9.0.x
+            8.0.x
       - name: Add MSBuild to PATH
         uses: microsoft/setup-msbuild@v2
       - name: Restore NuGet Cache
--- a/.github/workflows/validate-pull-requests.yml
+++ b/.github/workflows/validate-pull-requests.yml
@@ -13,7 +13,9 @@ jobs:
       - name: Setup .NET SDK
         uses: actions/setup-dotnet@v4.1.0
         with:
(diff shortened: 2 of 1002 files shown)


hazel-bohon:
The error occurs when building samples/hosting/aspire/Core_9/AspireDemo.AppHost/AspireDemo.AppHost.csproj it seems that there is a dependency on aspire which needs to be installed:

```
  Error: C:\Program Files\dotnet\sdk\9.0.101\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.ImportWorkloads.targets(38,5): error NETSDK1147: To build this project, the following workloads must be installed: aspire [D:\a\docs.particular.net\docs.particular.net\samples\hosting\aspire\Core_9\AspireDemo.AppHost\AspireDemo.AppHost.csproj]
  Error: C:\Program Files\dotnet\sdk\9.0.101\Sdks\Microsoft.NET.Sdk\targets\Microsoft.NET.Sdk.ImportWorkloads.targets(38,5): error NETSDK1147: To install these workloads, run the following command: dotnet workload restore [D:\a\docs.particular.net\docs.particular.net\samples\hosting\aspire\Core_9\AspireDemo.AppHost\AspireDemo.AppHost.csproj]
```

I downloaded it and ran it in my local environment and got the same error. In local env, the solution is to just run the suggested command and it runs just fine. I am not 100% sure how we ought to resolve this in the CI though. I think we probably need to add the command `dotnet workload restore` right before [line 110 in the build-samples-and-snippets.ps1 file](https://github.com/Particular/docs.particular.net/blob/support-net9/tools/build-samples-and-snippets.ps1#L110).

We could add a logic test of some sort to only run the command for those solutions which need it, but unless it leads to unacceptably slow performance, I think that it would be best practice to just run it all the time for every solution. So, that this doesnt accidentally catch some future dev by surprise when they add a sample that has a build dependency not normally anticipated by the script.

PR #6979, opened 2025-02-07
Update AWS configuration to use EC2Launch v2 agent

abparticular:


abparticular:
Still trying to determine whether the first screenshot needs to be updated.

jpalac (approved):

PR #6984, opened 2025-02-11
Topology migration sample for ASB

SzymonPobiega:
The sample shows that V4 and V5 of Azure Service Bus transport can communicate with one another using V5's migration topology.


SzymonPobiega:
@danielmarbach I added the sample markdown file so this is now ready for review

PR #6988, opened 2025-02-13
Rework the message conventions slightly

danielmarbach:


danielmarbach:
@DavidBoike I started the first draft here. I think I might also want to somehow mention the message interfaces assembly and I wonder whether expanding the design guidance section a bit more would make sense

nservicebus/messaging/messages-events-commands.md
```diff
@@ -81,14 +109,16 @@ The simplest way to identify messages is to use interfaces.
...
-```cs
+```csharp
 public class MyCommand : ICommand { }
 
 public class MyEvent : IEvent { }
 
 public class MyMessage : IMessage { }
 ```
 
+Those interfaces are available in [NServiceBus.MessageInterfaces](https://www.nuget.org/packages/NServiceBus.MessageInterfaces). The project targets `netstandard2.0` has a stable version number which is highly unlikely to change. Using these well-defined interfaces should be prefered over conventions since NServiceBus.MessageInterfaces package can be used to create a shared message assembly that can be used by multiple major versions of NServiceBus, and in projects using different target frameworks, while still relying on the `ICommand` and `IEvent` marker interfaces.
```
danielmarbach, line 120:
```suggestion
Those interfaces are available in [NServiceBus.MessageInterfaces](https://www.nuget.org/packages/NServiceBus.MessageInterfaces). The project targets `netstandard2.0` has a stable version number which is highly unlikely to change. Using these well-defined interfaces should be prefered over conventions since `NServiceBus.MessageInterfaces` package can be used to create a shared message assembly that can be used by multiple major versions of NServiceBus, and in projects using different target frameworks, while still relying on the `ICommand` and `IEvent` marker interfaces.
```

danielmarbach:
@DavidBoike merging this now. We can always do follow ups or edits later. This has been hanging around long enough 👯‍♂️

PR #6989, opened 2025-02-14
Remove hierarchy migration sample

danielmarbach:
With the introduction of v5 and the topic-per-event topology we don't want to encourage people to migrate to the hierarchy anymore. Removing the sample. It hurts my heart because I really think this was a great sample showing how to migrate things while keeping the system running but we can always resurrect ideas again from this sample. 

Related issue https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/issues/718


danielmarbach:
Moving forward since this doesn't really need a review given @SzymonPobiega and I have already agreed on deleting this sample

PR #6996, opened 2025-02-17
Improve structure of the ASB command-line tool docs

SzymonPobiega:
By introducing another set of partials

### DIFF
--- a/transports/azure-service-bus/operational-scripting_migration-endpoint-command_asbs_[5,).partial.md
+++ b/transports/azure-service-bus/operational-scripting_migration-endpoint-command_asbs_[5,).partial.md
@@ -1,9 +1,19 @@
-### asb-transport endpoint create
+## Migration-specific commands
+
+These commands are dedicated to endpoints that are configured with the migration topology for the purpose of migrating from the single-topic publishing approach (forwarding topology) to topic-per-event approach.
+
+- `migration endpoint create`
+- `migration endpoint subscribe`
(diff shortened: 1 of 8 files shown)


transports/azure-service-bus/operational-scripting_migration-endpoint-command_asbs_[5,).partial.md
```diff
@@ -0,0 +1,120 @@
...
+
+`-c` | `--connection-string` : Overrides the environment variable 'AzureServiceBus_ConnectionString'
+
+`-n` | `--namespace` : Sets the fully qualified namespace to connect with cached credentials, e.g., credentials from Azure PowerShell or CLI. This setting cannot be used in conjunction with the connection string setting.
+
+`-t` | `--topic`: Topic name to subscribe on (defaults to 'bundle-1')
+
+`-b` | `--subscription`: Subscription name (defaults to endpoint name)
+
+`-r` | `--rule-name`: Rule name (defaults to event type)
```
danielmarbach, line 63: Aren't we having a similar problem here to what we discussed on the tool PR. I think "defaults to" is ambiguous here. As I would read, it is that both the subscription name and rule name are optional... This comment also applies for the other "defaults to"

danielmarbach: The rest looks good to me

PR #6997, opened 2025-02-17
Azure Service Bus Topology Description

danielmarbach:


transports/azure-service-bus/topology_description_asbs_[5,).partial.md
```diff
@@ -0,0 +1,167 @@
...
+- Each topic is dedicated to one event type, so message consumption is isolated.
+- If any single topic hits its 5 GB quota, only that event type is affected, reducing the blast radius of a partial outage.
+- The maximum limit of 1,000 topics per messaging unit can comfortably support hundreds of event types, especially when factoring that not all event types are high-volume
+
+> [!NOTE]
+> If the system has numerous event types beyond these limits, an architectural review is recommended. Additional messaging units or other partitioning strategies may be required.
+
+#### Subscription rule matching
+
+In this topology, no SQL or Correlation filtering is strictly required on the topic itself because messages in a topic are all of the same event type. Subscriptions can use a default “match-all” rule (`1=1`) or the default catch-all rule on each topic subscription.
```
SzymonPobiega, line 51:
```suggestion
In this topology, no SQL or Correlation filtering is required on the topic itself because messages in a topic are all of the same event type. Subscriptions can use a default “match-all” rule (`1=1`) or the default catch-all rule on each topic subscription.
```

transports/azure-service-bus/topology_description_asbs_[5,).partial.md
```diff
@@ -0,0 +1,167 @@
...
+
+- Publish both versions of the event on the publisher side to individual topics and setting up the subscribers where necessary to receive both _or_
+- Multiplex all versions of the event to the same topic and filter the versions on the subscriber side within specialized filter rules
+
+When publishing both versions of the event the subscribers need to opt-into receiving those events by adding an explicit mapping:
```
SzymonPobiega, line 133:
```suggestion
When publishing both versions of the event the subscribers need to opt-in to receiving those events by adding an explicit mapping:
```

SzymonPobiega (approved):
I added some minor comments.

I think one thing we might want to explicitly describe here is a case of funny type names e.g. nested types.

danielmarbach:
Do it as a follow up?

> I think one thing we might want to explicitly describe here is a case of funny type names e.g. nested types.

PR #7025, opened 2025-03-06
Near real-time clients review

helenktsai:


awright18:
I think using the occasionally-connected terminology is way better term than transient, and having transient as an adjective for client is really bizarre, because it's the client's connection that's transient not the client.

helenktsai:
> I think using the occasionally-connected terminology is way better term than transient, and having transient as an adjective for client is really bizarre, because it's the client's connection that's transient not the client.

Good point, I agree with the misuse of "transient" as the adjective. Do you think we should omit it entirely though? I think it's technically a different concept with slight differences from "occasionally-connected" and people might identify with one but not the other, so I hesitate to remove it entirely.

helenktsai:
I made some updates to possibly address both concerns.

samples/near-realtime-clients/sample.md
```diff
@@ -10,36 +10,36 @@ related:
...
 
-One of the basic features of message queuing is the ability for the receiving endpoints to maintain service even when offline.  In this scenario, the long lasting, durable nature of message queuing can result in a backlog of messages that are no longer relevant which, if disconnected long enough, can result in queue quotas being exceeded. This can ultimately result in exceptions on the message sender possibly impacting other parts of the system.
+One of the key features of message queuing is the ability for the receiving endpoints to maintain service even when offline. Messages remain in the queue for processing when the system comes back online. However, while this feature is critical for some systems, the long lasting, durable nature of message queuing can result in a backlog of messages that are no longer relevant for occasionally-connected ones. If the client is disconnected long enough, it can result in queue quotas being exceeded. This can ultimately result in exceptions on the message sender and possibly impact other parts of the system.
```
tamararivera, line 17:
```suggestion
One of the key features of message queuing is the ability for the receiving endpoints to maintain service even when offline. Messages remain in the queue for processing when the system comes back online. However, while this feature is critical for some systems, the long lasting, durable nature of message queuing can result in a backlog of messages that are no longer relevant for occasionally-connected ones. If the client is disconnected long enough, it can result in queue quotas being exceeded, which in turn can generate exceptions on the message sender and possibly impact other parts of the system.
```

samples/near-realtime-clients/sample.md
```diff
@@ -10,36 +10,36 @@ related:
 ---
 
 > [!NOTE]
-> SignalR can be used in a variety of ways. Browse the official [SignalR tutorials](https://learn.microsoft.com/en-us/aspnet/core/tutorials/signalr?tabs=visual-studio&view=aspnetcore-6.0) and [SignalR samples](https://github.com/aspnet/SignalR-samples) for information on how to use SignalR. This document focuses on how to relay NServiceBus events to SignalR clients.
+> SignalR can be used in many different ways. For general guidance, check out the official [SignalR tutorials](https://learn.microsoft.com/en-us/aspnet/core/tutorials/signalr?tabs=visual-studio&view=aspnetcore-6.0) and [SignalR samples](https://github.com/aspnet/SignalR-samples). This guide specifically focuses on how to relay NServiceBus events to SignalR clients.
```
tamararivera, line 13: **Suggestion**: I think this note will make more sense if we put it before 'Project structure'

samples/near-realtime-clients/sample.md
```diff
@@ -10,36 +10,36 @@ related:
...
 
 ## Project structure
 
-Before running the sample, review the solution structure, the projects, and the classes. The `Publisher` and `ClientHub` projects are command-line applications that host an instance of NServiceBus.
+Before running the sample, review the solution structure, the projects, and the classes.
 
-The `ClientHub` project also hosts a SignalR server. The `Client` project is a command-line application that hosts a SignalR client.
+- `Publisher` project: command-line application that hosts an instance of NServiceBus.
+- `ClientHub` project: command-line application that hosts an instance of NServiceBus and a SignalR server.
+- `Client` project: command-line application that hosts a SignalR client.
```
tamararivera, line 31: **Suggestion:** Maybe this is better? I'm not convinced 
```suggestion
Projects in the solution:
- `Publisher`: command-line application that hosts an instance of NServiceBus.
- `ClientHub`: command-line application that hosts an instance of NServiceBus and a SignalR server.
- `Client`: command-line application that hosts a SignalR client.
```

samples/near-realtime-clients/sample.md
```diff
@@ -10,36 +10,36 @@ related:
...
+One way to prevent this is to [unsubscribe](/nservicebus/messaging/publish-subscribe/controlling-what-is-subscribed.md#manually-subscribing-to-a-message) when shutting down. However, this approach is unreliable since a crash could prevent the unsubscribe from happening.
 
-Another solution is to avoid implementing each client as an NServiceBus endpoint, but instead use a push technology, such as [SignalR](https://signalr.net/), to update clients only while they are connected.
+A better solution is to avoid implementing each client as an NServiceBus endpoint. Instead, use a push-based technology like [SignalR](https://signalr.net/) to only send updates when clients are connected.
 
-This sample demonstrates how to use a SignalR server, which also acts as an NServiceBus endpoint, to push subscribed NServiceBus events to any connected SignalR clients.
+This sample demonstrates how to use a SignalR server that also acts as an NServiceBus endpoint, to push subscribed NServiceBus events to any connected SignalR clients.
 
 ## Project structure
```
tamararivera, line 25: **Suggestion:** I think this should be Solution structure
```suggestion
## Solution structure
```

samples/near-realtime-clients/sample.md
```diff
@@ -10,36 +10,36 @@ related:
...
 
-Before running the sample, review the solution structure, the projects, and the classes. The `Publisher` and `ClientHub` projects are command-line applications that host an instance of NServiceBus.
+Before running the sample, review the solution structure, the projects, and the classes.
 
-The `ClientHub` project also hosts a SignalR server. The `Client` project is a command-line application that hosts a SignalR client.
+- `Publisher` project: command-line application that hosts an instance of NServiceBus.
+- `ClientHub` project: command-line application that hosts an instance of NServiceBus and a SignalR server.
+- `Client` project: command-line application that hosts a SignalR client.
```
tamararivera, line 32: **Suggestion:** According to the section below, the list is missing the `StockEvents` project.

samples/near-realtime-clients/sample.md
```diff
@@ -10,36 +10,36 @@ related:
...
 ## Sharing message types with SignalR
 
-The `StockEvents` project contains the definition of a message class that is shared with both NServiceBus endpoints, the SignalR hub, and the SignalR client. Open "StockTick" to see the message that will be published by this sample. Note that this event is defined using [unobtrusive mode message conventions](/nservicebus/messaging/unobtrusive-mode.md), allowing the `Client` project, which only uses SignalR, to reference the message type without requiring a reference to NServiceBus.
+The `StockEvents` project contains the definition of a message class that is shared with both NServiceBus endpoints, the SignalR hub, and the SignalR client. Open `StockTick` to see the message that will be published by this sample. Note that this event is defined using [unobtrusive mode message conventions](/nservicebus/messaging/unobtrusive-mode.md), allowing the `Client` project, which only uses SignalR, to reference the message type without requiring a reference to NServiceBus.
 
 snippet: MessageConventionsForNonNSB
 
 ## Hosting SignalR with NServiceBus
 
 The sample shows how to host a self-hosted SignalR server side-by-side with an NServiceBus endpoint.
```
tamararivera, line 42: **Suggestion:** I think we can delete this section, as it is very similar to line 23
```suggestion
```

samples/near-realtime-clients/sample.md
helenktsai: I personally like it feels like metadata about the document as a whole?

tamararivera: My problem is that we start talking about SignalR on the note, and then we don't mention it until the last paragraph of the intro, so it feels disjointed

helenktsai: Oops, seeing your note after the merge. I see what you mean, moved: https://github.com/Particular/docs.particular.net/commit/df710245b4138dce5b298b1ac2d5b1f9b591984e

PR #7028, opened 2025-03-07
Update performance-tuning.md

tamararivera:
This PR deletes the references to ServicePointManager from the Performance tuning page of the AWS SQS transport. 
I also updated the table to have the same order as the formula above.


transports/sqs/performance-tuning.md
```diff
@@ -14,59 +14,40 @@ The Amazon SQS transport uses HTTP/S connections to send and receive messages fr
...
+The transport will automatically increase the degree of parallelism by applying the following formula.
 
 ```
 Degree of parallelism = Math.Ceiling(MaxConcurrency / NumberOfMessagesToFetch)
 ```
 
 The following examples illustrate how the formula is applied when the concurrency is greater or equal to 10.
 
-|`MaxConcurrency` | `DegreeOfReceiveParallelism` | `NumberOfMessagesToFetch` |
+|DegreeOfReceiveParallelism | MaxConcurrency | NumberOfMessagesToFetch |
```
danielmarbach, line 25: FYI the order of these columns has been originally chosen like this because the thought process was that someone coming from NServiceBus might be most likely familiar with max concurrency (either the default value or the value they have explicitly set). We were assuming someone would then come from the angle of "I set the concurrency to LimitProcessingConcurrencyTo(100) and which row is that in the table..."

I would still argue the previous way makes more sense for most users. // cc @mauroservienti

tamararivera: Interesting! then can we flip the formula? As a reader, I expect the table to match the order of the formula

danielmarbach: What if you use the following order?

|MaxConcurrency | NumberOfMessagesToFetch | DegreeOfReceiveParallelism |

then it leas naturally from left to right following the formula and the two first columns are the input formula and the right most is the result.

bording:
The ServicePointManager stuff would still be relevant for earlier versions of NServiceBus that can still run on .NET Framework, so it seems like it would be better to introduce partials that could keep that information around for those. It would be good to then point out that they are only relevant for .NET Framework, though.

PR #7037, opened 2025-03-10
Update AttributeMessageConvention.cs

mozerov:
Fix mismatch in IsCommandType and IsEventType methods


awright18:
@mozerov, thank you very much for this contribution!

PR #7071, opened 2025-04-01
Add handler span customization to the v9 OpenTelemetry customization sample

andreasohlund:
I did some testing using the sample and ended up doing some minor cleanup as well.

- [x] Is it worth keeping the processor part that adds the machine name?
- [x] Apply changes to the v8 sample as well


samples/open-telemetry/customizing/Core_9/Sample/NetHostProcessor.cs
```diff
@@ -6,12 +6,7 @@
 
 class NetHostProcessor : BaseProcessor<Activity>
```
andreasohlund, line 7: @lailabougria is this worth keeping? It feels that we are showing an OTEL feature that has nothing to do with the NServiceBus? (The machine name is already visible in the traces via the `nservicebus.originating_machine` header)

samples/open-telemetry/customizing/Core_9/Sample/Program.cs
```diff
@@ -3,64 +3,55 @@
...
 using OpenTelemetry.Trace;
 using System;
-using System.Threading.Tasks;
 
-public static class Program
-{
-    public static async Task Main()
-    {
-        Console.Title = "CustomTelemetry";
+Console.Title = "MyEndpoint";
```
andreasohlund, line 7: Changed the name to make it more clear in the traces that this is an NServiceBus endpoint
image

samples/open-telemetry/customizing/Core_9/Sample/Program.cs
```diff
@@ -3,64 +3,55 @@
...
+var endpointConfiguration = new EndpointConfiguration("MyEndpoint");
+endpointConfiguration.EnableOpenTelemetry();
 
-            endpointConfiguration.Pipeline.Register(
-                new TraceOutgoingMessageSizeBehavior(),
-                "Captures body size of outgoing messages as OpenTelemetry tags"
-            );
+#endregion
 
-            var endpoint = await Endpoint.Start(endpointConfiguration);
```
andreasohlund, line 40: I thought about using extensions hosting but I don't think it would have made the code more terse

```
// Initialize OpenTelemetry
builder.Services.AddOpenTelemetry()
    .ConfigureResource(resource => /* Resource configuration goes here */)
    .WithLogging(logging => /* Logging configuration goes here */)
    .WithMetrics(metrics => /* Metrics configuration goes here */)
    .WithTracing(tracing => /* Tracing configuration goes here */));
```

samples/open-telemetry/customizing/Core_9/Sample/NetHostProcessor.cs
andreasohlund: Discuessed with @lailabougria and we decided to remove

andreasohlund:
@lailabougria I've made the changes we discussed. You were right about the wildcards

andreasohlund:
@lailabougria added an example of adding tags to the handler span as well

PR #7083, opened 2025-04-05
Quickstart fix logging snippet

awright18:
Make sure structured logging is illustrated in the docs and used in the quick start tutorial


awright18:
This was handled in https://github.com/Particular/docs.particular.net/pull/7086

PR #7101, opened 2025-04-17
Update scmu.md

PhilBastian:


servicecontrol/monitoring-instances/deployment/scmu.md
```diff
@@ -11,8 +11,6 @@ The ServiceControl installation package includes a utility to manage the install
 
 Use a Monitoring instance to vizualize [endpoint performance metrics](/monitoring/#endpoint-performance) in ServicePulse.
 
-Monitoring instances are not as [resource intensive as the Error and Audit instances](/servicecontrol/servicecontrol-instances/hardware.md) and can be run on the same machine as the Error instance.
```
PhilBastian, line 14: I couldn't see any guidance around installing audit instances on a separate machine to Error/Monitoring instances, so I also removed this. Are either of you @SzymonPobiega @saratry aware of guidance around this that isn't in https://docs.particular.net/servicecontrol/servicecontrol-instances/hardware?

servicecontrol/monitoring-instances/deployment/scmu.md
```diff
@@ -44,6 +39,9 @@ Clicking the upgrade link will:
 
 Click on the `+ NEW` link at the top of the screen and select "Monitoring instance" to launch the "New instance form". Complete the form to register a new monitoring instance.
 
+> [!WARNING]
```
PhilBastian, line 42: this fits better here than where it was, but even here it doesn't explain _why_ it shouldn't be used in production?

servicecontrol/monitoring-instances/deployment/scmu.md
saratry: @PhilBastian I would leave it. Even if it states the obvious, it doesn't hurt.

saratry: I agree that you can deduce it. Could we leave the first part of the sentece and remove the latest?

saratry:
```suggestion
Monitoring instances are not as [resource intensive as the Error and Audit instances](/servicecontrol/servicecontrol-instances/hardware.md).
```

SzymonPobiega: I think the assumption behind it is that the Monitoring instance has no embedded database. So: when running with out-of-the-box configuration with embedded RavenDB, it makes sense to co-host the Monitoring instance because, compared to RavenDB, it consumes little resources. 

When hosting in K8s when DB is a separate container/service, the Monitoring instance should be treated more-or-less the same as any other instance.

servicecontrol/monitoring-instances/deployment/scmu.md
SzymonPobiega: I don't know where this warning comes from. I wasn't aware of it and I see nothing wrong in having 2 monitoring instances if you have a system that uses two different transports.

PR #7102, opened 2025-04-17
[DynamoDB] 2.1 updates

danielmarbach:


persistence/dynamodb/transactions_options_dynamodb_[2.1,).partial.md
```diff
@@ -0,0 +1,3 @@
+Starting with Version 2.1 of the persistence it is possible to override or extend the default JsonSerializationOptions used by the mapper. This allows making the mapper more aware of the object model attributes used by the `DynamoDBContext` if desired.
```
mauroservienti, line 1:
```suggestion
Starting with Version 2.1 of the persistence it is possible to override or extend the default `JsonSerializationOptions` used by the mapper. This allows the mapper to be more aware of the object model attributes used by the `DynamoDBContext` if desired.
```

persistence/dynamodb/sagas_options_dynamodb_[2.1,).partial.md
```diff
@@ -0,0 +1,3 @@
+Sagas being persisted use the built-in mapper as described above. With Version 2.1 of the persistence it is possible to override the saga persistence JsonSerializationOptions to support scenarios like [source generated serialization contexts](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/source-generation).
```
mauroservienti, line 1:
```suggestion
Sagas being persisted use the built-in mapper as described above. With Version 2.1 of the persistence it is possible to override the saga persistence `JsonSerializationOptions` to support scenarios such as [source-generated serialization contexts](https://learn.microsoft.com/en-us/dotnet/standard/serialization/system-text-json/source-generation).
```

Snippets/DynamoDB/DynamoDB_2.1/prerelease.txt
mauroservienti, line 1: why do we need this?

persistence/dynamodb/sagas_consistency_dynamodb_[2.1,).partial.md
```diff
@@ -0,0 +1,9 @@
+Amazon DynamoDB supports two read models: strongly consistent and eventually consistent. By default, the AWS SDK uses eventual consistency, which returns the most recently committed version of a record—but without guaranteeing that the latest write is visible at the time of the read. While this model is sufficient for many applications, it can introduce correctness issues when used with concurrency control mechanisms like optimistic locking.
```
mauroservienti, line 1:
```suggestion
Amazon DynamoDB supports two read models: strongly consistent and eventually consistent. By default, the AWS SDK uses eventual consistency, which returns the most recently committed version of a record but does not guarantee that the latest write is visible at the time of the read. While this model is sufficient for many applications, it can introduce correctness issues when used in conjunction with concurrency control mechanisms, such as optimistic locking.
```

Snippets/DynamoDB/DynamoDB_2.1/prerelease.txt
danielmarbach: Currently only because I reference the alpha version to make the engine happy. Should be removed once we bring in the final version and the nuget package version should be changed

mauroservienti (approved):

Snippets/DynamoDB/DynamoDB_2.1/DynamoDB_2.1.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="NServiceBus.Persistence.DynamoDB" Version="2.1.0-alpha.2" />
```
lailabougria:
```suggestion
    <PackageReference Include="NServiceBus.Persistence.DynamoDB" Version="2.1.0" />
```

PR #7115, opened 2025-05-02
Documentation review

lailabougria:


lailabougria:
I am merging this so it doesn't come up again for next week. We can discuss the question posted on Slack and raise a follow-up PR if needed later:
> The documentation I'm reviewing contains a bunch of information around migrating from non-native to native delayed delivery. The versions from which users would be migrating are long out of support, so from that angle I could remove it from the docs, however, I'm on the fence.

PR #7122, opened 2025-05-09
Removing "None" transaction level as it's not supported

afprtclr:


ramonsmits:
@afprtclr interesting, so what happens if you pass None?

afprtclr:
@ramonsmits 

image

ramonsmits (approved):

PR #7131, opened 2025-05-13
Describe how lack of handling failed messages may lead to zombie messages

afprtclr:


nservicebus/outbox/index.md
```diff
@@ -140,6 +140,15 @@ The performance of the outbox feature depends on the scope of the transactions u
 * Transactions scoped to multiple databases on a single server, also known as _cross-database transactions_, are supported by some persisters, such as those which use SQL Server. These transactions may have reasonably good performance but they may introduce other concerns. For example, [cross-database transactions are not supported by all types of tables in SQL Server](https://docs.microsoft.com/en-us/sql/relational-databases/in-memory-oltp/cross-database-queries).
 * Transactions scoped to multiple databases on multiple servers, also known as _distributed transactions_, are supported by persisters which use SQL Server, but they require the use of MSDTC and should generally be avoided for the same reasons as those listed in the [comparison with MSDTC](#comparison-with-msdtc) below.
 
+### Errors when dispatching the outbox messages
+
+The dispatch of the outbox messages is triggered in two different ways:
```
lailabougria, line 145:
```suggestion
The dispatch of the pending messages is triggered in two different ways:
```

nservicebus/outbox/index.md
```diff
@@ -140,6 +140,15 @@ The performance of the outbox feature depends on the scope of the transactions u
...
+
+* In the incoming scenario, by the flow processing the message. This happens after the user transaction has been committed but before the incoming messsage is acknowledged.
+* In the non-incoming scenario, by the control message. This happens when processing the control message.
```
lailabougria, line 148: I wouldn't this mention here, this is mixing in a transactional session concern that is not part of the outbox.

nservicebus/outbox/index.md
```diff
@@ -140,6 +140,15 @@ The performance of the outbox feature depends on the scope of the transactions u
 * Transactions scoped to multiple databases on a single server, also known as _cross-database transactions_, are supported by some persisters, such as those which use SQL Server. These transactions may have reasonably good performance but they may introduce other concerns. For example, [cross-database transactions are not supported by all types of tables in SQL Server](https://docs.microsoft.com/en-us/sql/relational-databases/in-memory-oltp/cross-database-queries).
 * Transactions scoped to multiple databases on multiple servers, also known as _distributed transactions_, are supported by persisters which use SQL Server, but they require the use of MSDTC and should generally be avoided for the same reasons as those listed in the [comparison with MSDTC](#comparison-with-msdtc) below.
 
+### Errors when dispatching the outbox messages
+
+The dispatch of the outbox messages is triggered in two different ways:
+
+* In the incoming scenario, by the flow processing the message. This happens after the user transaction has been committed but before the incoming messsage is acknowledged.
```
lailabougria, line 147:
```suggestion

The dispatch of the outgoing messages occurs after the user transaction has been committed but before the incoming message is acknowledged.
```

nservicebus/outbox/index.md
```diff
@@ -140,6 +140,15 @@ The performance of the outbox feature depends on the scope of the transactions u
...
+* In the non-incoming scenario, by the control message. This happens when processing the control message.
+
+If the message triggering the dispatch fails to process, the message processing is retried as usual. However, if the message fails to process all the retry attempts, the message will be moved to the error queue, but the outbox messages will not be sent. The message triggering the dispatch should be then processed manually. If the message triggering the dispatch is not processed, it may result in zombie records.
```
lailabougria, line 150:
```suggestion
If the message triggering the dispatch fails to process but has been able to commit the database transaction, the message processing is retried as usual. However, if the message fails to process through all the retry attempts, the message will be moved to the error queue, and the outbox messages will not be dispatched. Once the error is resolved, the failed message should be retried to ensure the outgoing messages are dispatched. If this doesn't happen, this may result in zombie records.
```

afprtclr: I don't want to say explicitly that "the failed message should be retried". It may be impossible to dispatch the messages (for instance if the queue doesn't exist or something). I want to be very generic here and only suggest that "something" needs to be done.

nservicebus/outbox/index.md
lailabougria: Yes, but then we should mention that on the TransactionalSession docs, not here. Most users of the outbox don't have transactionalsession enabled, so it mixes in a concern that is not applicable to most readers of this document.

nservicebus/outbox/index.md
lailabougria: I updated the suggestion, can you take a look?

nservicebus/outbox/index.md
afprtclr: Okay, I removed this part.

SzymonPobiega (approved):

lailabougria (approved):

PR #7135, opened 2025-05-13
Reviewing SQL transport benchmark

afprtclr:


lailabougria (commented):
Not sure about changing from msg to messages. I think it’s a well understood abbreviation and seeing it now I wanted to also fully qualify seconds

afprtclr:
> Not sure about changing from msg to messages. I think it’s a well understood abbreviation and seeing it now I wanted to also fully qualify seconds

I replaced "s" with "second" now.

lailabougria (approved):

PR #7138, opened 2025-05-15
 Describe how lack of handling failed control messages may lead to zombie records

afprtclr:


nservicebus/transactional-session/index.md
```diff
@@ -178,6 +178,8 @@ Sending the control message first ensures that eventually, the transaction will
 
 If dispatching the control message fails, the transactional session changes will roll back, and an error will be raised to the user committing the session.
 
+If the transaction finishes and the control message fails to process through all the retry attempts, the control message will be moved to the error queue, and the outgoing messages will not be dispatched. Once the error is resolved, the control message should be retried to ensure the outgoing messages are dispatched. If this doesn't happen, this may result in zombie records. If that's undesirable, the system should be returned to a consistent state through another action.
```
ramonsmits, line 181:
```suggestion
If the transaction completes and the control message fails to process through all the retry attempts, the control message will be moved to the error queue, and the outgoing messages will not be dispatched. Once the error is resolved, the control message must be manually retried in ServicePulse to ensure the outgoing messages are dispatched. If this doesn't happen, the stored outgoing messages will never delivered. If that's undesirable, the system should be returned to a consistent state through another action.
```

ramonsmits: That last sentence is a bit strange:

> If that's undesirable, the system should be returned to a consistent state through another action.

What is meant? What is the action we ask from users?

afprtclr: We don't want to ask for any specific action. We want to be very generic to suggest that "something" should be done but we can't tell what exactly (as it depends on the business scenario).

tmasternak (approved):

PR #7153, opened 2025-05-23
Unity Sample in NSB 9

poornimanayar:


danielmarbach:
@poornimanayar I do not understand why we need to add this sample. Can you clarify?

In case we want to keep it, didn't we recently do an effort to move many samples towards the generic host? Should we also stick to that or does it make sense for the DI samples to be purely about the container integration only?

PR #7154, opened 2025-05-25
Update default value of immediate retries

poornimanayar:
There seems to be a mismatch in the docs with the amount of default immediate retries. This fixes it.


poornimanayar:
Crap! I should have verified it via core. So that means it's the docs about the retries itself that needs changing to 5

poornimanayar:
Looks like immediate retry default is 5 and delayed is 3.

PR #7181, opened 2025-06-05
Improve alternative sc 4 to 5 guide

ramonsmits:


servicecontrol/upgrades/4to5/alternative.md
```diff
@@ -1,118 +1,118 @@
...
+3. Add:
+   ```xml
+   <add key="ServiceControl/IngestErrorMessages" value="False" />
+   ```
+4. Save the file.
+
+### Change ServiceControl Instance Queue
+
+> [!NOTE]  
+> This step is only needed if **heartbeats** are used.
```
tmasternak: Is it? The instance input queue has to be changed to properly handle future message retry commands as well.

tmasternak (approved):
One minor question. Apart from that LGTM

servicecontrol/upgrades/4to5/alternative.md
ramonsmits: @tmasternak Renaming, this should be framed differently. How about:

```suggestion
> [!NOTE]  
> The swapping of the instance names is only needed if **heartbeats** are used to avoid requiring to update the configuration of all endpoints.
```

servicecontrol/upgrades/4to5/alternative.md
```diff
@@ -1,118 +1,118 @@
 ---
-title: Alternative ServiceControl v4 to newer migration
-reviewed: 2025-05-21
-summary: Advanced alternative ServiceControl v4 migration for environment that cannot retry all messages immediately
+title: Alternative ServiceControl v4 to newer migration  
+reviewed: 2025-05-21  
+summary: Advanced alternative ServiceControl v4 migration for environments that cannot retry all messages immediately.
```
ramonsmits, line 4:
```suggestion
summary: Advanced alternative ServiceControl v4 migration for environments that cannot retry all messages immediately
```

PR #7198, opened 2025-06-11
Bridge hosting

danielmarbach:
This PR improves the guidance on how to host the bridge. I did not yet include the licensing part since that is covered in https://particular.net/pricing/faq.

### DIFF
--- a/nservicebus/bridge/index.md
+++ b/nservicebus/bridge/index.md
@@ -26,6 +26,12 @@ The messaging bridge enables several scenarios:
 
 More details on these scenarios are provided in the [messaging bridge scenarios](scenarios.md) article.
 
+## Hosting
+
+While it's technically possible to co-host the bridge within the same process as a regular NServiceBus endpoint, it's generally recommended to run the bridge in a dedicated hosting process. The bridge's purpose is purely infrastructural—it connects logical endpoints. 
+
+To fulfill this role as efficiently as possible, it may be necessary to scale the bridge either [vertically](performance.md#performance-tuning) or [horizontally](performance.md#scaling-out). Because of its infrastructural nature, the bridge also exhibits significantly different CPU, memory, and monitoring characteristics compared to a typical NServiceBus endpoint that handles business logic.
+
 ## Bridge configuration
 
 partial: configuration


nservicebus/bridge/index.md
```diff
@@ -26,6 +26,12 @@ The messaging bridge enables several scenarios:
 
 More details on these scenarios are provided in the [messaging bridge scenarios](scenarios.md) article.
 
+## Hosting
+
+While it's technically possible to co-host the bridge within the same process as a regular NServiceBus endpoint, it's generally recommended to run the bridge in a dedicated hosting process. The bridge's purpose is purely infrastructural—it connects logical endpoints. 
+
+To fulfill this role as efficiently as possible, it may be necessary to scale the bridge either [vertically](performance.md#performance-tuning) or [horizontally](performance.md#scaling-out). Because of its infrastructural nature, the bridge also exhibits significantly different CPU, memory, and monitoring characteristics compared to a typical NServiceBus endpoint that handles business logic.
+
```
poornimanayar, line 34: If the bridge is hosted in the same process as endpoint, would it show an increase in messages processed? Would that be another reason to not host it alongside an endpoint?

danielmarbach: It really depends. A process might have CPU or memory limits. A process might also have "connection" limits. So if you share the same process the underlying connections are increased. 

>  would it show an increase in messages processed? Would that be another reason to not host it alongside an endpoint?

Not sure what that means.

poornimanayar: Would hosting the bridge in the same process as the endpoint show an increase in the number of messages processed by that endpoint? And have an impact on the usage report?

poornimanayar: It could be a dumb question totally!

danielmarbach: There are no dumb questions @poornimanayar. All good questions. I'll check and report back

danielmarbach: As far as I understand the bridge has no input queue and only shuffles from the source to the destination and therefore has no impact on the endpoint usage report

PR #7201, opened 2025-06-12
Document benefits of using a dedicated transactional session processor endpoint

andreasohlund:
TODO:

- Add links as needed

### DIFF
--- a/nservicebus/transactional-session/index.md
+++ b/nservicebus/transactional-session/index.md
@@ -82,8 +82,8 @@ The transactional session feature requires a supported persistence package to st
 It's recommended to not mix the processing of control messages with business messages in order to get:
 
 - Predictable control message dispatch: Processing of control messages will be more reliable since there is no risk of getting delayed behind slow business messages
-- More accurate metrics: Metrics like critical time and queue lenght will accurately represent the performance of the control message processing and not be skewed by business messages
-- Simplified management: Knowning that the endpoint only processes control message makes it possible to always retry all failed message related to the endpoint via tools like ServicePulse
+- More accurate metrics: Metrics like critical time and queue length will accurately represent the performance of the control message processing and not be skewed by business messages
+- Simplified management: Knowing that the endpoint only processes control messages makes it possible to always retry all failed messages related to the endpoint via tools like ServicePulse
 
 ## Transaction consistency
 
--- a/nservicebus/transactional-session/index.md
+++ b/nservicebus/transactional-session/index.md
@@ -77,6 +77,14 @@ The transactional session feature requires a supported persistence package to st
 * [MongoDB](/persistence/mongodb)
 * [DynamoDB](/persistence/dynamodb/)
 
+## Design considerations
+
+It's recommended to not mix the processing of control messages with business messages in order to get:
+
+- Predictable control message dispatch: Processing of control messages will be more reliable since there is no risk of getting delayed behind slow business messages
+- More accurate metrics: Metrics like critical time and queue lenght will accurately represent the performance of the control message processing and not be skewed by business messages
+- Simplified management: Knowning that the endpoint only processes control message makes it possible to always retry all failed message related to the endpoint via tools like ServicePulse
+
 ## Transaction consistency
 
 To guarantee atomic consistency across database and message operations, the transactional session requires the [outbox](/nservicebus/outbox) to be enabled. This combination of features provides the strongest consistency guarantees and is, therefore, the recommended, safe-by-default configuration.
--- a/nservicebus/transactional-session/index_remote-processor_transactionalsession_[3.3,).partial.md
+++ b/nservicebus/transactional-session/index_remote-processor_transactionalsession_[3.3,).partial.md
@@ -17,7 +17,7 @@ The processor endpoint [must have both the outbox and the transactional session 
 
 Using the transactional session in send-only endpoints has the following benefits:
 
-- More predictable dispatch:
+- Predictable control message dispatch:
   - Processing of control messages will be more reliable since the processor is dedicated to dispatching outbox records
   - Downtime is limited since redeployment of the processor is only necessary when new versions of the transactional session are released
 - Simplified management: For short-lived endpoints, there is no longer a need to make sure that all control messages have been processed before decommissioning them 
--- a/nservicebus/transactional-session/index_remote-processor_transactionalsession_[3.3,).partial.md
+++ b/nservicebus/transactional-session/index_remote-processor_transactionalsession_[3.3,).partial.md
@@ -15,10 +15,10 @@ The processor endpoint [must have both the outbox and the transactional session 
 
 ### Benefits
 
-In addition to unlopcking the use of send-only endpoint using a dedicated processor endpoint has the following benefits since it would only process transactional session control messages:
+In addition to unlocking the use of a send-only endpoint, using a dedicated processor endpoint has the following benefits since it would only process transactional session control messages:
 
-- More predicable dispatch: Since the endpoint is dedicated to dispatching outbox records processing times will be more reliable. And redeployment of the endpoint is only needed when new versions of the transactional session is released.
-- Targeted scaling: Scaling can be tailored to the load of control messages
+- More predictable dispatch: Since the endpoint is dedicated to dispatching outbox records, processing times will be more reliable. Redeployment of the endpoint is only necessary when new versions of the transactional session are released.
+- Targeted scaling: Scaling can be tailored with only the load of control messages in mind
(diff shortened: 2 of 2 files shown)


andreasohlund:
@timbussmann FYI

nservicebus/transactional-session/index_remote-processor_transactionalsession_[3.3,).partial.md
```diff
@@ -13,6 +13,14 @@ The processor endpoint [must have both the outbox and the transactional session
 > [!NOTE]
 > If migrating to this mode from an endpoint couldn't be made send-only due to the previous limitations of the transactional session, see the migration guidance below.
 
+### Benefits
+
+In addition to unlocking the use of a send-only endpoint, using a dedicated processor endpoint has the following benefits since it would only process transactional session control messages:
+
+- More predictable dispatch: Since the endpoint is dedicated to dispatching outbox records, processing times will be more reliable. Redeployment of the endpoint is only necessary when new versions of the transactional session are released.
```
timbussmann: Assuming, that only the control message dispatching prevented you from using a send-only endpoint so far, that endpoint would only handle control messages anyway, therefore that benefit would be neglectable or only really be interesting for non-send-only endpoints (but the remote processor seems to be primarily advertised for send-only endpoints in the section above)

timbussmann: one important aspect is for use cases where the endpoint is short lived (e.g. Kubernetes job, spinning up, dispatching a few messages and the shutting down again) - stopping the endpoint before it consumed the control messages would delay the messages for potentially very long time and this currently can only be avoided by adding artifical delays before the shutdown of the process - that issue is solved with the remote processor.

nservicebus/transactional-session/index_remote-processor_transactionalsession_[3.3,).partial.md
```diff
@@ -13,6 +13,14 @@ The processor endpoint [must have both the outbox and the transactional session
...
+- Targeted scaling: Scaling can be tailored with only the load of control messages in mind
+- Auditing can be turned off: There is no need to audit the control messages since the resulting business messages will be audited individually when processed.
```
timbussmann: why wouldn't that be a built-in auditing feature if there is no value in auditing them? ;)

nservicebus/transactional-session/index_remote-processor_transactionalsession_[3.3,).partial.md
andreasohlund: Good points, I've made some tweaks, see if it makes sense

nservicebus/transactional-session/index_remote-processor_transactionalsession_[3.3,).partial.md
andreasohlund: Very good point. I'll raise an issue or PR to fix this

@danielmarbach do you see any flaws in this thinking?

danielmarbach: Honestly I was surprised that we audit control messages because it makes little sense. That being said ServiceControl is already aware of that fact and treats messages with the control message header as system messages (similar to scheduled tasks). That information is then exposed over the metadata or the indexes. 

I see no flaw in the thinking as long as the filtering capability is at the level of audit allowing certain messages with certain headers to be skipped. At the time of writing the transactional session we used the control message header because we knew it will cause core to skip certain parts of the pipeline. To achieve what you want you probably need a way to mark control message of the transactional session belonging to that package so that you can filter out those only. In your case it might be enough to the have processor endpoint header and the control message but I wonder for non-sendonly scenarios whether that would work too.

andreasohlund: I was contemplating to add a audit behavior to skip in the transactional session, similar to this https://docs.particular.net/samples/pipeline/audit-filtering/#code-walk-through-auditfilterbehavior

danielmarbach: yep that works well. Glad we wrote that extension seam 🤜 🤛

nservicebus/transactional-session/index_remote-processor_transactionalsession_[3.3,).partial.md
```diff
@@ -13,6 +13,16 @@ The processor endpoint [must have both the outbox and the transactional session
 > [!NOTE]
 > If migrating to this mode from an endpoint couldn't be made send-only due to the previous limitations of the transactional session, see the migration guidance below.
 
+### Benefits
+
+Using the transactional session in send-only endpoints has the following benefits:
+
+- Predictable control message dispatch:
```
timbussmann: one could argue this is still a pure marketing statement as the described benefits aren't really new/valuable:
- "reliable" processing of control messages works the same way when using a non-send-only endpoint that is only consuming messages from the transactional session (what you would do as of now, before switching to the new send-only mode) -> no benefits gained other than saving a queue. You'd probably even worsen the behavior when many active send-only endpoints share a processor ;)
- limited downtime: assuming you'd have to update your "send-only" endpoint anyway, there already is downtime and you're not saving on that downtime given you're relying on the same packages?

so all in all - those points are not incorrect but also not really an actual benefit I could see (at least in our use cases)

what about less infra (queues) required and potentially involved costs on cloud environment for that? (probably also a fairly weak argument but hey ;) )

timbussmann (approved):

nservicebus/transactional-session/index_remote-processor_transactionalsession_[3.3,).partial.md
andreasohlund: Turns out that the usage pattern of the tx-session prevents auditing from happening, see https://github.com/Particular/NServiceBus.TransactionalSession/pull/441

nservicebus/transactional-session/index_remote-processor_transactionalsession_[3.3,).partial.md
andreasohlund: Yea, you are right in that these benefits might have been there earlier if your endpoint did not process any other types of messages. I'll try to reword

andreasohlund:
@timbussmann made another pass on this based on your feedback

@danielmarbach any angles you see missing?

nservicebus/transactional-session/index.md
```diff
@@ -77,6 +77,14 @@ The transactional session feature requires a supported persistence package to st
 * [MongoDB](/persistence/mongodb)
 * [DynamoDB](/persistence/dynamodb/)
 
+## Design considerations
+
+It's recommended to not mix the processing of control messages with business messages in order to get:
+
+- Predictable control message dispatch: Processing of control messages will be more reliable since there is no risk of getting delayed behind slow business messages
+- More accurate metrics: Metrics like critical time and queue lenght will accurately represent the performance of the control message processing and not be skewed by business messages
```
poornimanayar:
```suggestion
- More accurate metrics: Metrics like critical time and queue length will accurately represent the performance of the control message processing and not be skewed by business messages
```

nservicebus/transactional-session/index.md
```diff
@@ -77,6 +77,14 @@ The transactional session feature requires a supported persistence package to st
...
+- Simplified management: Knowning that the endpoint only processes control message makes it possible to always retry all failed message related to the endpoint via tools like ServicePulse
```
poornimanayar:
```suggestion
- Simplified management: Knowing that the endpoint only processes control messages makes it possible to always retry all failed messages related to the endpoint via tools like ServicePulse
```

nservicebus/transactional-session/index.md
```diff
@@ -77,6 +77,14 @@ The transactional session feature requires a supported persistence package to st
 * [MongoDB](/persistence/mongodb)
 * [DynamoDB](/persistence/dynamodb/)
 
+## Design considerations
+
+It's recommended to not mix the processing of control messages with business messages in order to get:
+
+- Predictable control message dispatch: Processing of control messages will be more reliable since there is no risk of getting delayed behind slow business messages
```
danielmarbach, line 84: In general, I think this is a good enough first step. I'm starting to wonder though if the term control message is "too technical" and we should use another term that is more aligned with the transactional session behavior. I know we use this term in the entire documentation, so please consider this out of scope for this PR. But maybe something like "an internal message to settle the outbox outcome is dispatched" or something similar might make more sense.

andreasohlund: Agreed, I’ll handle this in a separate pr

Issue #5801, opened 2020-12-04
Improve the documentation for how to migrate data between ServiceControl instances


timbussmann:
## Relevant information

When migrating a ServiceControl instance to a new server or system, it's not obvious how to transfer the data of the existing SC instance to the new SC instance.
It seems that the only approach at the moment seems to just copy the whole database (or create a backup in maintenance mode) (see https://docs.particular.net/servicecontrol/backup-sc-database#important-notes-and-restrictions) which seems to be a brittle approach, as according to our documentation, the instances need to have the same service name (which might not be desired by the user):

> When restoring a ServiceControl database, the name of the ServiceControl Windows Service must be the same on the target machine as it was on the source machine.

mikeminutillo:
We have guidannce for performing a [zero-downtime](https://docs.particular.net/servicecontrol/upgrades/zero-downtime) upgrade which could be extracted and repurposed.

PR #7207, opened 2025-06-13
ServiceControl documentation for how to retain data between instances

ngallegos:
This adds documentation specific to retaining data between ServiceControl deployments as a result of issue:

- https://github.com/Particular/docs.particular.net/issues/5801

There is documentation for replacing [audit](https://docs.particular.net/servicecontrol/migrations/replacing-audit-instances/) and [error](https://docs.particular.net/servicecontrol/migrations/replacing-error-instances/) instances, as well as documentation for [backing up and restoring a ServiceControl database](https://docs.particular.net/servicecontrol/backup-sc-database), but nothing that specifies these as strategies for retaining data across installations.

### DIFF
--- a/servicecontrol/migrations/retain-data-between-instances.md
+++ b/servicecontrol/migrations/retain-data-between-instances.md
@@ -35,6 +35,18 @@ ServiceControl data consists of audit and error message data, managed by [audit 
 > [!NOTE]
 > It is worth assessing whether audit and error message data retention is required. For scenarios where retaining audit and error message data is not required (e.g. transient or test environment data that does not merit effort to retain), these strategies are not necessary.
 
+Each of the strategies presented here have trade-offs:
+
+- Instance replacement
+  - Requires little to no downtime
+  - Does not require ServiceControl version, instance name, and operating system to be the same
+  - Requires that old instances remain in service for the duration of the audit message retention period
+
+- Database backup and restore
+  - Requires downtime from when the database is backed up to when it is restored
+  - Requires ServiceControl version, instance name, and operating system to be the same
+  - Allows old instance decommission immediately after the new instance is running with restored data
+
 ### Instance replacement
 
 This strategy is an incremental approach to data retention. Data is retained by replacing audit and error instances separately and allowing new instances to access old data through remote configuration and stopping ingestion. The following instance replacement guides should be followed for this strategy.


servicecontrol/migrations/migrate-data-between-instances.md
```diff
@@ -0,0 +1,52 @@
...
+    - servicecontrol/data-migration
+---
+
+ServiceControl, which exists to serve the management of distributed systems, is itself a distributed system. As a result, migrating all ServiceControl data between deployments means each piece of the system that accesses and writes data is migrated separately.
+
+This document describes in general terms how to migrate ServiceControl data, and links to more specific information on how to accomplish data migration for each piece of the system.
+
+## Overview
+
+ServiceControl data consists of audit and error message data, managed by [audit instances](/servicecontrol/audit-instances/) and [error instances](/servicecontrol/servicecontrol-instances/) respectively. To migrate all ServiceControl data from one deployment to another means replacing both audit and error instances.
```
danielmarbach: Could this be confusing because it is not clear what deployment to deployment means? Reading this someone might wonder whether that procedure is necessary for each patch or minor deployment or whether this is only necessary when doing major to major...

Or it might be necessary to indicate when switching from one instance to another is required, desirable or even encouraged to give some more hints

servicecontrol/migrations/migrate-data-between-instances.md
```diff
@@ -0,0 +1,52 @@
+---
+title: Migrate ServiceControl data between instances
```
DavidBoike: Is this a misleading title? Because … you can't actually migrate any data.

menu/menu.yaml
```diff
@@ -1598,6 +1598,8 @@
     Articles:
     - Title: Data expiration
       Url: servicecontrol/how-purge-expired-data
+    - Title: Data migration
+      Url: servicecontrol/migrations/migrate-data-between-instances
```
DavidBoike: I think it's weird to have this in two spots in the navigation. We try to minimize that as much as possible, as both sections of the nav will be expanded if it exists in 2 places. I think it belongs under Migrations, so if links can be added to any relevant pages in this section, so be it, but I don't think it should be in the nav here.

menu/menu.yaml
```diff
@@ -1646,6 +1648,8 @@
         Url: servicecontrol/migrations/replacing-audit-instances/containers
     - Title: New persistence format
       Url: servicecontrol/migrations/new-persistence
+    - Title: Migrating data between instances
+      Url: servicecontrol/migrations/migrate-data-between-instances
```
DavidBoike: If the point of this PR is to make kind of a "cover page" to the [replacing an audit instance](https://docs.particular.net/servicecontrol/migrations/replacing-audit-instances/) and [replacing an error instance](https://docs.particular.net/servicecontrol/migrations/replacing-error-instances/) pages, then this article should probably appear in the nav above those sections, not after new persistence format.

servicecontrol/migrations/migrate-data-between-instances.md
ngallegos: How about this?

```suggestion
title: Retain ServiceControl data for new deployments
```

If it sounds good I'll update the doc to use `retain` over `migrate` where appropriate.

servicecontrol/migrations/migrate-data-between-instances.md
ngallegos: I think I found a happy medium by rewording and adding examples of when data retention would be preferable in [this commit](https://github.com/Particular/docs.particular.net/pull/7207/commits/6a1afbd92a73696b0c3b2a37d4816263836f3aa1)

servicecontrol/migrations/retain-data-between-instances.md
```diff
@@ -0,0 +1,61 @@
...
+3. Creating a new error instance configured to use the same audit instance(s)
+
+Follow the [detailed error instance replacement process](/servicecontrol/migrations/replacing-error-instances/) to achieve retention of this data.
+
+## Alternative retention strategies
+
+Data retention can also be achieved by [backing up the ServiceControl data](/servicecontrol/backup-sc-database.md) and restoring it to the new instance.
+
+> [!WARNING]
+> The [restrictions](/servicecontrol/backup-sc-database.md#important-notes-and-restrictions) should be strongly considered before moving forward with this approach.
```
danielmarbach:
```suggestion
> The [restrictions](/servicecontrol/backup-sc-database.md#important-notes-and-restrictions) must considered before moving forward with this approach.
```

danielmarbach (approved):

servicecontrol/migrations/retain-data-between-instances.md
```diff
@@ -0,0 +1,61 @@
+---
+title: Retain ServiceControl data between instances
```
ramonsmits, line 2: Maybe better to call it migrate installations?

servicecontrol/migrations/retain-data-between-instances.md
```diff
@@ -0,0 +1,61 @@
...
+
+- ServiceControl instances are deployed on aging hardware and need to be moved
+- ServiceControl instances are being migrated to a different kind of deployment, such as moving from on-premises to cloud-hosted servers or from ServiceControl Management Utility installation to containers
+- ServiceControl instance upgrades are not possible
+
+This document describes in general terms how to retain ServiceControl data, and links to more specific information on how to accomplish data retention for each piece of the system.
+
+## Overview
+
+ServiceControl data consists of audit and error message data, managed by [audit instances](/servicecontrol/audit-instances/) and [error instances](/servicecontrol/servicecontrol-instances/) respectively. To retain all ServiceControl data means replacing both audit and error instances.
```
ramonsmits: I think the procedure would be:

1. Get version on machine A
2. Stop it
3. Install version on machine B with same config and name
4. Copy all data
5. Start it

That applies to both error and audit instances.

servicecontrol/migrations/retain-data-between-instances.md
```diff
@@ -0,0 +1,61 @@
...
+
+[Monitoring instances](/servicecontrol/monitoring-instances/) do not use a persistent data store, and therefore have no data to retain.
+
+> [!NOTE]
+> It is worth assessing whether audit and error message data retention is required. For scenarios where retaining audit and error message data is not required (e.g. transient or test environment data that does not merit effort to retain), this process is not necessary.
+
+### Audit data
+
+Retaining audit data involves:
```
ramonsmits: Currently not written as migrating an installation.

servicecontrol/migrations/retain-data-between-instances.md
```diff
@@ -0,0 +1,61 @@
...
+
+1. Adding the new audit instance as a remote
+2. Disabling audit queue ingestion on the old audit instance
+3. Decommission of the old audit instance _**after**_ all audit information has expired
```
ramonsmits: That is not great advice when retention is set to a very long duration.

servicecontrol/migrations/retain-data-between-instances.md
ngallegos: I rewrote a bit of the document to make it more clearly focused on data retention strategies.  The intent of this is to fill a gap for users who may be looking for a way to retain data between ServiceControl instances but aren't immediately aware that instance replacement or database backup and restore can achieve that.

servicecontrol/migrations/retain-data-between-instances.md
```diff
@@ -0,0 +1,50 @@
...
+
+ServiceControl data consists of audit and error message data, managed by [audit instances](/servicecontrol/audit-instances/) and [error instances](/servicecontrol/servicecontrol-instances/) respectively.
+
+[Monitoring instances](/servicecontrol/monitoring-instances/) do not use a persistent data store, and therefore have no data to retain.
+
+## Strategies
+
+> [!NOTE]
+> It is worth assessing whether audit and error message data retention is required. For scenarios where retaining audit and error message data is not required (e.g. transient or test environment data that does not merit effort to retain), these strategies are not necessary.
+
```
ramonsmits, line 37: I think it helps to state when to go for Backup/Restore versus Instance replacement.

Just guessing some aspects:

- Downtime
- Lift and shift
- Version upgrades
- Very long audit retention periods
- Virtualization difference (VM to container)
- OS difference (Windows 2019 to 2022)

Issue #7193, opened 2025-06-10
Feedback: 'Message Headers' referring to non-existing helper class DateTimeExtensions


peterhutgasunie:
Feedback for 'Message Headers' https://docs.particular.net/nservicebus/messaging/headers

Location in GitHub: https://github.com/Particular/docs.particular.net/blob/master/nservicebus/messaging/headers.md

I could not find "The helper class DateTimeExtensions" with "ToWireFormattedString() and ToUtcDateTime() methods". It looks like it should probably be the DateTimeOffsetHelper class with similar methods: https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/DateTimeOffsetHelper.cs

mauroservienti:
I raised https://github.com/Particular/docs.particular.net/pull/7210 to address @peterhutgasunie's findings.

PR #7218, opened 2025-06-19
Fix Timestamp format block in headers based on version

DharmishthaVPatel:
- Issue - https://github.com/Particular/docs.particular.net/issues/7193

Added a partial class to support version-specific rendering of the Timestamp format block. This resolves incorrect references to the non-existent DateTimeExtensions helper.

Reference : https://github.com/Particular/NServiceBus/blob/master/src/NServiceBus.Core/DateTimeOffsetHelper.cs

### DIFF
--- a/nservicebus/messaging/headers.md
+++ b/nservicebus/messaging/headers.md
@@ -17,7 +17,13 @@ The headers of a message are similar to HTTP headers and contain metadata about 
 
 ## Timestamp format
 
+#if-version [8,)
 For all timestamp message headers, the format is `yyyy-MM-dd HH:mm:ss:ffffff Z`, where the time is UTC. The helper class `DateTimeOffsetHelper` supports converting from UTC to wire format and vice versa by using the `ToWireFormattedString(DateTimeOffset)` and `ToDateTimeOffset(string)` methods.
+#end-if
+
+#if-version [,8)
+For all timestamp message headers, the format is `yyyy-MM-dd HH:mm:ss:ffffff Z` where the time is UTC. The helper class `DateTimeExtensions` supports converting from UTC to wire format and vice versa by using the `ToWireFormattedString(DateTime)` and `ToUtcDateTime(string)` methods.
+#end-if
 
 ### ISO 8601 format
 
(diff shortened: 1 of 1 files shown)


nservicebus/messaging/headers.md
```diff
@@ -17,7 +17,13 @@ The headers of a message are similar to HTTP headers and contain metadata about
 
 ## Timestamp format
 
+#if-version [8,)
```
ramonsmits, line 20: nice addition!

nservicebus/messaging/headers.md
```diff
@@ -42,17 +48,30 @@ Use the following code to generate or read the NServiceBus custom timestamp form
 ```cs
 const string Format = "yyyy-MM-dd HH:mm:ss:ffffff Z"; // Not ISO 8601
 
+#if-version [8,)
```
ramonsmits: The existing code does not have anything to do with either `DateTimeOffsetHelper` or `DateTimeExtensions`. There is no need for versioned snippets as the usage of `DateTime` or `DateTimeOffset` is unrelated to NServiceBus.

As listed at the start of this section this is about **When doing native intergration**, so in the context there is no NServiceBus package.

DharmishthaVPatel: I agree, but to stay consistent with our codebase and Using DateTimeOffset helps align with the type expected by NServiceBus. Users can use this snippet to convert to the required format.

ramonsmits: This snippet should not be used where there is a reference to NServiceBus. What type is or isn't expected by NServiceBus is not very relevant in this context.

However, to avoid confusion the page could be extended with something like

```
#if-version [8,)
> [!NOTE]
> Use `DateTimeOffsetHelper` when you have a reference to the NServiceBus package.
#end-if

#if-version [,8)
> [!NOTE]
> Use `DateTimeExtensions` when you have a reference to the NServiceBus package.
#end-if
```

DharmishthaVPatel: Removed code snippet since DateTimeOffsetHelper usage is already explained at the beginning.

soujay (approved):

soujay (approved):

PR #7219, opened 2025-06-19
Transport case-sensitivity

ramonsmits:


danielmarbach:
Let's add similar notes to SQS (double check) and RabbitMQ

PR #7232, opened 2025-06-25
Use the name "dispatch message" rather than control message to describe the tx session behavior of sending a special message

andreasohlund:


andreasohlund:
@danielmarbach this is based on the remark you did on the previous PR, thoughts about the name?

andreasohlund:
I thought about that name but discarded it since it wouldn't be correct based on my mental model since its not involved in committing the TX session but rather "making sure that the outgoing messages are being sent to the transport".

@danielmarbach thoughts?

poornimanayar (approved):

PR #7427, opened 2025-07-07
Refactor logger calls in samples to use structured logging

DharmishthaVPatel:
Part of 
- https://github.com/Particular/GeneralPlatformExperience/issues/3482

- Updated all logger.LogInformation and similar calls in the samples to use structured logging message templates and named parameters.
- No functional changes


DharmishthaVPatel:
Created multiple small PRS

PR #7439, opened 2025-07-07
Update tests to allow .NET 10.0

tamararivera:
This PR adds the .NET 10.0 framework to the integrity tests and attempts to future proof us for when C# 14 is not preview anymore.

### DIFF
--- a/tests/IntegrityTests/ProjectLangVersion.cs
+++ b/tests/IntegrityTests/ProjectLangVersion.cs
@@ -83,7 +83,7 @@ namespace IntegrityTests
 
                     solutionLangVersion ??= fallbackLangVersion;
 
-                    var solutionLangVersionString = solutionLangVersion > latestStableLangVersion ? "preview" : solutionLangVersion.Value.ToString("N1");
+                    var solutionLangVersionString = solutionLangVersion > latestReleasedLangVersion ? "preview" : solutionLangVersion.Value.ToString("N1");
 
                     bool valid = true;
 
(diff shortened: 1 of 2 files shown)


tests/IntegrityTests/ProjectLangVersion.cs
```diff
@@ -162,6 +168,12 @@ public void ShouldHaveLangVersionMappingForEachAllowedTargetFramework()
             { "net8.0-windows", 12 },
             { "net9.0", 13 },
             { "net9.0-windows", 13 },
+            {"net10.0", 14}
+        };
+
+        private static readonly Dictionary<string, string> supportPhase = new()
```
tamararivera: I called it supportPhase, because that is the wording on the [Microsoft site](https://dotnet.microsoft.com/en-us/download/dotnet), but I'm taking suggestions for better names

tests/IntegrityTests/ProjectLangVersion.cs
```diff
@@ -162,6 +163,11 @@ public void ShouldHaveLangVersionMappingForEachAllowedTargetFramework()
             { "net8.0-windows", 12 },
             { "net9.0", 13 },
             { "net9.0-windows", 13 },
+            { "net10.0", 14 },
+            { "net10.0-windows", 14 }
         };
+
+        // This allows us to use the preview version of C# before is officially released
+        private const int latestStableVersion = 14;
```
bording:
```suggestion
        private const int latestStableVersion = 13;
```

Also, let's go with `latestStableLangVersion`

bording (approved):

Issue #7260, opened 2025-07-02
Update Sample: azure/azure-table/simple


RyanThomasForge:
Add an NServiceBus 10 version of this sample: [azure/azure-table/simple](https://github.com/Particular/docs.particular.net/blob/master/samples/azure/azure-table/simple/)

> Path: samples/azure/azure-table/simple/

afprtclr:
PR merged. Closing.

PR #7446, opened 2025-07-11
Updating Azure Table Simple sample to NET 10

afprtclr:
For https://github.com/Particular/docs.particular.net/issues/7260

### DIFF
--- a/samples/azure/azure-table/simple/ASTP_7/Client/Client.csproj
+++ b/samples/azure/azure-table/simple/ASTP_7/Client/Client.csproj
@@ -1,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFramework>net10.0</TargetFramework>
+    <OutputType>Exe</OutputType>
+    <LangVersion>preview</LangVersion>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="NServiceBus.Extensions.Hosting" Version="4.0.0-alpha.1" />
+  </ItemGroup>
+  <ItemGroup>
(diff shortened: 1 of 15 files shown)


samples/azure/azure-table/simple/ASTP_7/Client/Client.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <PropertyGroup>
+    <TargetFrameworks>net10.0</TargetFrameworks>
+    <OutputType>Exe</OutputType>
+    <LangVersion>preview</LangVersion>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="NServiceBus.Extensions.Hosting" Version="3.0.1" />
```
RyanThomasForge: Ensure you are using the latest versions of NServiceBus/Particular packages. You can use feedz.io to determine the latest version available for each package. 

NServiceBus.Extensions.Hosting, for example, should be 4.0.0-alpha.1

samples/azure/azure-table/simple/ASTP_7/Client/InputLoopService.cs
```diff
@@ -0,0 +1,38 @@
+﻿
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Hosting;
+using NServiceBus;
+public class InputLoopService(IMessageSession messageSession) : BackgroundService
```
tmasternak: I think that we want to get rid of `InputLoopSerivce`

afprtclr: Fixed.

Issue #7327, opened 2025-07-02
Update Sample: logging/notifications


RyanThomasForge:
Add an NServiceBus 10 version of this sample: [logging/notifications](https://github.com/Particular/docs.particular.net/blob/master/samples/logging/notifications/)

> Path: samples/logging/notifications/

PR #7462, opened 2025-07-15
Updating logging-notification sample to NET 10

irinascurtu:
For https://github.com/Particular/docs.particular.net/issues/7327

### DIFF
--- a/samples/logging/notifications/Core_10/Sample/Program.cs
+++ b/samples/logging/notifications/Core_10/Sample/Program.cs
@@ -1,0 +1,49 @@
+using System;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using NServiceBus;
+
+
+var host = Host.CreateDefaultBuilder(args)
+    .UseNServiceBus(x =>
+    {
+        #region endpointConfig
+
+        var endpointConfiguration = new EndpointConfiguration("Samples.Notifications");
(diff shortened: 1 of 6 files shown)


samples/logging/notifications/Core_10/Sample/Program.cs
```diff
@@ -0,0 +1,49 @@
...
+var messageSession = host.Services.GetRequiredService<IMessageSession>();
+
+var message = new MyMessage
+{
+    Property = "PropertyValue"
+};
+
+await messageSession.SendLocal(message);
+
+Console.WriteLine("Press any key to exit");
```
tmasternak: I think we need support for `Ctrl+C` in the `Console.WriteLine`. Similar to what we have in [the EventsSubscription sample](https://github.com/Particular/docs.particular.net/pull/7478/files#diff-78c88b00f3c8881847c9c71553207e27499f04ed813d86bc4cb8198d134aa2daR59)

tmasternak (changes requested):

samples/logging/notifications/Core_10/Sample/Program.cs
irinascurtu: Pressing any key, stops the execution...

tmasternak: I think, it's only `Ctrl+C` that is problematic

tmasternak (approved):

PR #7485, opened 2025-07-16
update header propogation sample for NSB 10

### LINKED ISSUE #7358: Update Sample: pipeline/header-propagation
Add an NServiceBus 10 version of this sample: [pipeline/header-propagation](https://github.com/Particular/docs.particular.net/blob/master/samples/pipeline/header-propagation/)

> Path: samples/pipeline/header-propagation/

PhilBastian:
fixes #7358

### DIFF
--- a/samples/pipeline/header-propagation/Core_10/Receiver/MyMessageHandler.cs
+++ b/samples/pipeline/header-propagation/Core_10/Receiver/MyMessageHandler.cs
@@ -1,0 +1,24 @@
+﻿using System;
+using System.Threading.Tasks;
+using NServiceBus;
+
+class MyMessageHandler : IHandleMessages<ProcessOrder>, IHandleMessages<ShipOrder>
+{
+    #region handlers
+    public Task Handle(ProcessOrder message, IMessageHandlerContext context)
+    {
+        context.MessageHeaders.TryGetValue("CustomerId", out var customerId);
+        Console.WriteLine($"Recieved ProcessOrder for customer {customerId}. Shipping");
+
+        return context.SendLocal(new ShipOrder());
+    }
+
+    public Task Handle(ShipOrder message, IMessageHandlerContext context)
+    {
+        context.MessageHeaders.TryGetValue("CustomerId", out var customerId);
+        Console.WriteLine($"Shipping Order for customer {customerId}.");
+
+        return Task.CompletedTask;
+    }
+    #endregion
+}
--- a/samples/pipeline/header-propagation/Core_10/Receiver/Program.cs
+++ b/samples/pipeline/header-propagation/Core_10/Receiver/Program.cs
@@ -1,0 +1,34 @@
+using System;
+using System.Threading.Tasks;
+using NServiceBus;
+
+public static class Program
+{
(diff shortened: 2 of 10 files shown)


jasontaylordev (approved):

samples/pipeline/header-propagation/Core_10/Receiver/MyMessageHandler.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Threading.Tasks;
+using NServiceBus;
+
+class MyMessageHandler : IHandleMessages<ProcessOrder>, IHandleMessages<ShipOrder>
+{
+    #region handlers
+    public Task Handle(ProcessOrder message, IMessageHandlerContext context)
+    {
+        context.MessageHeaders.TryGetValue("CustomerId", out var customerId);
+        Console.WriteLine($"Recieved ProcessOrder for customer {customerId}. Shipping");
```
warwickschroeder, line 11: Consider using ILogger with structured logging

samples/pipeline/header-propagation/Core_10/Receiver/Program.cs
```diff
@@ -0,0 +1,34 @@
...
+        endpointConfiguration.UseTransport<LearningTransport>();
+
+        #region register-behavior
+        endpointConfiguration.Pipeline.Register(
+            new PropagateCustomerIdHeaderBehavior(),
+            "Copies CustomerId header from incoming to outgoing messages"
+        );
+        #endregion
+
+        var endpoint = await Endpoint.Start(endpointConfiguration);
```
warwickschroeder, line 22: Consider using NServiceBus.Extensions.Hosting

samples/pipeline/header-propagation/Core_10/Receiver/Program.cs
```diff
@@ -0,0 +1,34 @@
...
+
+        Console.WriteLine("Press [ESC] to quit.");
+
+        while (Console.ReadKey(true).Key != ConsoleKey.Escape)
+        {
+        }
```
warwickschroeder, line 28: If using *.Extensions.Hosting, you could use RunAsync and pass a cancellation token to it. Quiting would be CTRL+C. An empty while loop looks wrong.

samples/pipeline/header-propagation/Core_10/Receiver/MyMessageHandler.cs
PhilBastian: would need a DI container for that

samples/pipeline/header-propagation/Core_10/Receiver/Program.cs
PhilBastian: i feel this leads to a far messier solution when all the work is done in program.cs

samples/pipeline/header-propagation/Core_10/Receiver/Program.cs
PhilBastian: maybe... but then the endpoint would need to be retrieved inline rather than using di, which is messier imo

PR #7523, opened 2025-07-18
Assembly sharing to .NET 10

irinascurtu:


irinascurtu:
Uses Nsb..old version

Issue #7322, opened 2025-07-02
Update Sample: logging/extensions-logging


RyanThomasForge:
Add an NServiceBus 10 version of this sample: [logging/extensions-logging](https://github.com/Particular/docs.particular.net/blob/master/samples/logging/extensions-logging/)

> Path: samples/logging/extensions-logging/

PR #7585, opened 2025-07-31
Add Extensions Logging sample for NServiceBus 10

hsvotwa:
Fixes https://github.com/Particular/docs.particular.net/issues/7322.

### DIFF
--- a/samples/logging/extensions-logging/Extensions.Logging_4/Sample/MyHandler.cs
+++ b/samples/logging/extensions-logging/Extensions.Logging_4/Sample/MyHandler.cs
@@ -9,7 +9,7 @@ public class MyHandler :
 
     public Task Handle(MyMessage message, IMessageHandlerContext context)
     {
-        log.Warn("Hello from MyHandler");
+        log.Info("Hello from MyHandler");
         return Task.CompletedTask;
     }
 }
--- a/samples/logging/custom-factory/Core_9/Sample/MyMessage.cs
+++ b/samples/logging/custom-factory/Core_9/Sample/MyMessage.cs
@@ -1,6 +1,3 @@
 ﻿using NServiceBus;
 
-public class MyMessage :
-    IMessage
-{
-}
+public class MyMessage : IMessage;


samples/logging/extensions-logging/Extensions.Logging_4/Sample/MyHandler.cs
```diff
@@ -0,0 +1,15 @@
...
+using NServiceBus.Logging;
+
+public class MyHandler :
+    IHandleMessages<MyMessage>
+{
+    static ILog log = LogManager.GetLogger<MyHandler>();
+
+    public Task Handle(MyMessage message, IMessageHandlerContext context)
+    {
+        log.Warn("Hello from MyHandler");
```
mauroservienti: Is this a warning because it was a warning in the previous version sample? I guess it is to get colored output in the console.

samples/logging/extensions-logging/Extensions.Logging_4/Sample/MyMessage.cs
```diff
@@ -0,0 +1,6 @@
+﻿using NServiceBus;
+
+public class MyMessage :
+    IMessage
+{
+}
```
mauroservienti: This should work with latest C#, or something similar to that.

```suggestion
public class MyMessage : IMessage;
```

samples/logging/extensions-logging/Extensions.Logging_4/Sample/Program.cs
```diff
@@ -0,0 +1,57 @@
...
+        return endpointConfiguration;
+    })
+    .Build();
+
+await host.StartAsync();
+
+var endpointInstance = host.Services.GetRequiredService<IMessageSession>();
+
+var myMessage = new MyMessage();
+await endpointInstance.SendLocal(myMessage);
```
mauroservienti:
```suggestion
await endpointInstance.SendLocal(new MyMessage());
```

WilliamBZA (approved):
The tweaks Mauro suggested are great :+1:

samples/logging/extensions-logging/Extensions.Logging_4/Sample/MyHandler.cs
hsvotwa: @mauroservienti yes, it's because it was a warning in the previous version. I've updated this to Info.

Issue #7320, opened 2025-07-02
Update Sample: logging/datadog


RyanThomasForge:
Add an NServiceBus 10 version of this sample: [logging/datadog](https://github.com/Particular/docs.particular.net/blob/master/samples/logging/datadog/)

> Path: samples/logging/datadog/

PR #7590, opened 2025-07-31
Add Datadog Logging sample for NServiceBus 10

hsvotwa:
Fixes https://github.com/Particular/docs.particular.net/issues/7320.

### DIFF
--- a/samples/logging/datadog/Core_10/Endpoint/LoadSimulator.cs
+++ b/samples/logging/datadog/Core_10/Endpoint/LoadSimulator.cs
@@ -1,0 +1,56 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+using NServiceBus;
+
+// Simulates busy (almost no delay) / quiet time in a sine wave
+class LoadSimulator(IEndpointInstance endpointInstance, TimeSpan minimumDelay, TimeSpan idleDuration)
+{
+    CancellationTokenSource tokenSource = new();
+    TimeSpan idleDuration = TimeSpan.FromTicks(idleDuration.Ticks / 2);
+    Task fork;
+
+    public Task Start()
+    {
+        fork = Loop();
+        return Task.CompletedTask;
+    }
+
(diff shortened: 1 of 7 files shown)


samples/logging/datadog/Core_10/Endpoint/SomeCommand.cs
```diff
@@ -0,0 +1,5 @@
+﻿using NServiceBus;
+
+class SomeCommand : ICommand
+{
+}
```
mauroservienti:
```suggestion
class SomeCommand : ICommand;
```

samples/logging/datadog/Core_10/Endpoint/Program.cs
mauroservienti, line 1: Is there a reason this is not using the Host?

samples/logging/datadog/Core_10/Endpoint/LoadSimulator.cs
```diff
@@ -0,0 +1,56 @@
...
+        var delay = TimeSpan.FromMilliseconds(idleDuration.TotalMilliseconds * Math.Sin(angleInRadians));
+        delay += idleDuration;
+        delay += minimumDelay;
+        return delay;
+    }
+
+    Task Work()
+    {
+        return endpointInstance.SendLocal(new SomeCommand());
+    }
```
mauroservienti:
```suggestion
    Task Work() => endpointInstance.SendLocal(new SomeCommand());
```

samples/logging/datadog/Core_10/Endpoint/Program.cs
WilliamBZA: Probably because the v9 sample doesn't. I think it's a good idea to change the v10 sample to use the Host so it's the same as the rest of the samples. :+1:

hsvotwa: This has been updated to use Host.

PR #7721, opened 2025-08-12
Added the 2 new options for endpoint type

johnsimons:
These are introduced in https://github.com/Particular/ServicePulse/pull/2580

### DIFF
--- a/servicepulse/usage.md
+++ b/servicepulse/usage.md
@@ -35,7 +35,7 @@ The report file must be provided to Particular upon request, it is **not** autom
 
 ## Viewing usage summary
 
-At any time the system usage can be viewed on the Usage page.
+At any time, the system usage can be viewed on the Usage page.
 This page has two tabs:
 
 - [Detected endpoints](#viewing-usage-summary-detected-endpoints)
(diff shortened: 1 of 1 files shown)


servicepulse/usage.md
```diff
@@ -89,6 +89,14 @@ An endpoint that is expected to no longer be used within the next 30 days. These
 
 Not an NServiceBus endpoint. These are broker queues or other messaging infrastructure that should not be included in NServiceBus licensing calculations.
 
+#### Gateway or Bridging Endpoint
+
+This is an infrastructure endpoint used by the Particular Platform<
```
mikeminutillo: Should these `<` be `.`?

mikeminutillo: Given that these have the same descriptions, will customers know which one to pick?

johnsimons: This is based on https://github.com/Particular/CustomerSuccess/issues/4834, which suggests to add a single entry.
Maybe that is the right way to go, thoughts @mikeminutillo ?

My concern is that "Infrastructure Endpoint" is too generic and users can confuse it with their own endpoints that deal with infrastructure.
That said, maybe we can be more explicit and name it "Particular Platform Endpoint"?

mikeminutillo: I don't mind having them separate, but their descriptions should maybe provide more detail as to what fits in each category.

The Particular Platform Endpoints might explicitly mention Error, Audit, and Monitoring queues, as well as the ServiceControl instance input queues. 

The Gateway or Bridging Endpoints might link off to the documentation for Gateway and Bridge (if links are possible)

johnsimons: ok I'll update these

johnsimons:
@mikeminutillo I have updated this, please review again.
I have also updated https://github.com/Particular/ServicePulse/pull/2580

PR #7839, opened 2025-09-25
Update cosmosdb persistence details

warwickschroeder:
Update a result of https://github.com/Particular/NServiceBus.Persistence.CosmosDB/pull/896

Update how the container info and partition key is set and retrieved.

_Not to be merged until the above PR is released_


persistence/cosmosdb/defaultcontainer.include.md
```diff
@@ -1,11 +1,16 @@
-Set the default container using the following configuration API:
+The container that is used by default for all incoming messages is specified via the `DefaultContainer(..)` configuration API:
 
 snippet: CosmosDBDefaultContainer
 
-The container that is used by default for all incoming messages is specified via `DefaultContainer(..)`. When installers are enabled, this (default) container will be created if it doesn't exist.
+#if-version [3, )
+If the container information is being extracted during runtime from a message instance (header or message body), the default container specified will be overwritten by the last extractor in the pipeline. For example, if a [Header Extractor](/persistence/cosmosdb/transactions.md#specifying-the-container-to-use-for-the-transaction-using-message-header-values) (physical stage) and a [Message Extractor](/persistence/cosmosdb/transactions.md#specifying-the-container-to-use-for-the-transaction-using-the-message-contents) (logical stage) are both configured, container information within the Message Extractor would be used.
```
PhilBastian:
```suggestion
If the container information is being extracted during runtime from a message instance (header or message body), the default container specified will be overwritten by the last extractor in the pipeline. For example, if a [Header Extractor](/persistence/cosmosdb/transactions.md#specifying-the-container-to-use-for-the-transaction-using-message-header-values) (physical stage) and a [Message Extractor](/persistence/cosmosdb/transactions.md#specifying-the-container-to-use-for-the-transaction-using-the-message-contents) (logical stage) are both configured, then the container information within the Message Extractor would be used.
```

persistence/cosmosdb/index_outboxstorageformat_CosmosDB_[3,).partial.md
```diff
@@ -0,0 +1,7 @@
+A default [synthetic partition key](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/synthetic-partition-keys) will be used, in the format `{endpointName}-{messageId}`, if not explicitly [overwritten](/persistence/cosmosdb/transactions.md#specifying-the-partitionkey-to-use-for-the-transaction) at runtime.
+
+> [!NOTE]
+> To support backward compatibility of control messages during migration, the persistence includes a fallback mechanism. When enabled (default), if a record is not found using the synthetic key format, the system falls back to the legacy `{messageId}` format. The fallback can be disabled using `DisableReadFallback()` once all legacy records have expired. This is recommended as the fallback mechanism involves an additional read operation to the Outbox container.
```
PhilBastian:
```suggestion
> To support backward compatibility of control messages during migration, the persistence includes a fallback mechanism. When enabled (default), and if a record is not found using the synthetic key format, the system falls back to the legacy `{messageId}` format. The fallback can be disabled using `DisableReadFallback()` once all legacy records have expired, which is recommended since the fallback mechanism involves an additional read operation on the Outbox container.
```

PhilBastian (approved):

persistence/cosmosdb/index_outboxstorageformat_CosmosDB_[3,).partial.md
```diff
@@ -0,0 +1,7 @@
+A default [synthetic partition key](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/synthetic-partition-keys) will be used, in the format `{endpointName}-{messageId}`, if not explicitly [overwritten](/persistence/cosmosdb/transactions.md#specifying-the-partitionkey-to-use-for-the-transaction) at runtime.
```
jpalac: This needs to be split into v3 up to 3.1.2 and then new text for v3.1.3 onwards - that is the only one that will have the synthetic partition key.

persistence/cosmosdb/index_outboxstorageformat_CosmosDB_[3,).partial.md
```diff
@@ -0,0 +1,7 @@
...
+
+> [!NOTE]
+> To support backward compatibility of control messages during migration, the persistence includes a fallback mechanism. When enabled (default), and if a record is not found using the synthetic key format, the system falls back to the legacy `{messageId}` format. The fallback can be disabled using `DisableReadFallback()` once all legacy records have expired, which is recommended since the fallback mechanism involves an additional read operation on the Outbox container.
```
jpalac: Same as above - only applies to 3.1.3+

persistence/cosmosdb/index_outboxstorageformat_CosmosDB_[3,).partial.md
```diff
@@ -0,0 +1,7 @@
...
+
+> [!WARNING]
+> When the default synthetic partition key is not used, but rather overwritten at runtime, Outbox rows are not automatically separated by endpoint name. As a result, multiple logical endpoints cannot share the same database and container since [message identities are not unique across endpoints from a processing perspective](/nservicebus/outbox/#message-identity). To avoid conflicts, either separate different endpoints into different containers, use the default synthetic partition key, or [override the default synthetic partition key](transactions.md).
```
jpalac: Same as above - only applies from 3.1.3+

persistence/cosmosdb/index_outboxstorageformat_CosmosDB_[3,).partial.md
PhilBastian: docs doesn't support patch level (or even minor iirc) versioning. This would need to be specified in a note or other text within the document

persistence/cosmosdb/index_outboxstorageformat_CosmosDB_[3.2,).partial.md
```diff
@@ -0,0 +1,9 @@
+A default [synthetic partition key](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/synthetic-partition-keys) will be used for all incoming messages, in the format `{endpointName}-{messageId}`, if not explicitly [overwritten](/persistence/cosmosdb/transactions.md#specifying-the-partitionkey-to-use-for-the-transaction) at runtime.
+
+> [!NOTE]
+> The [default partition key must be overwritten](/persistence/cosmosdb/transactions.md#specifying-the-partitionkey-to-use-for-the-transaction) whenever the message handler creates or updates business documents in CosmosDB. To guarantee atomicity, explicitly set the Outbox partition key to match the partition key of your business document (e.g., an `OrderId` or `CustomerId`). This is a requirement for including both the business document and the Outbox record in the same [Cosmos DB transactional batch](https://learn.microsoft.com/en-us/azure/cosmos-db/partitioning-overview). Conversely, for simplicity, you can use the default partition key when a handler's logic does not involve persisting business data.
```
jpalac: I wouldn't say that it "must" be overwritten - perhaps "should"? Also - giving the example of a key such as "OrderId" would cause the outbox issue that we have just solved by changing the default key to be synthetic. 
So perhaps best not to give an example? Unless we can come up with one that doesn't cause a problem.

persistence/cosmosdb/index_outboxstorageformat_CosmosDB_[3.2,).partial.md
```diff
@@ -0,0 +1,9 @@
...
+
+To support backward compatibility of control messages during migration, the persistence includes a fallback mechanism. When enabled (default), and if a record is not found using the synthetic key format, the system falls back to the legacy `{messageId}` format. The fallback can be disabled using `DisableReadFallback()` once all legacy records have expired, which is recommended since the fallback mechanism involves an additional read operation on the Outbox container.
+
+> [!WARNING]
+> When the default synthetic partition key is not used, but rather overwritten at runtime, Outbox records are not automatically separated by endpoint name. As a result, multiple logical endpoints cannot share the same database and container since [message identities are not unique across endpoints from a processing perspective](/nservicebus/outbox/#message-identity). To avoid conflicts, either separate different endpoints into different containers, use the default synthetic partition key, or [override the default synthetic partition key](transactions.md) in a way that ensures message identities are unique to each processing endpoint.
```
jpalac: Wouldn't this be dependant on the key used? Ie if they use a key that includes the endpoint name then it's not the case. So perhaps something like: 

```suggestion
>  Since [message identities are not unique across endpoints from a processing perspective](/nservicebus/outbox/#message-identity), when overwriting the default synthetic key, either separate different endpoints into different containers or [override the default synthetic partition key](transactions.md) in a way that ensures message identities are unique to each processing endpoint.
```

persistence/upgrades/cosmosdb-3.1to3.2.md
```diff
@@ -0,0 +1,27 @@
...
+
+As a result of the above issue, customers may be unintentionally relying on the use of the default container over the configured message container extractor.
+
+### Solution
+
+Affected customers should perform one of the below options prior to updating to version 3.2:
+
+1. Customers can remove the configured message container extractor and only rely on the default container.
+2. Customers can update their message container extractor to use the same container specified as the default container.
+3. Customers can [migrate records](https://learn.microsoft.com/en-us/azure/cosmos-db/container-copy?tabs=online-copy&pivots=api-nosql) from the default container to the container specified in the message container extractor. This option may require [changing the container partition key](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/change-partition-key).
```
jpalac: It may not be all records, but only the ones they originally were trying to separate out with the message extractor.

```suggestion
3. Customers can [migrate relevant records](https://learn.microsoft.com/en-us/azure/cosmos-db/container-copy?tabs=online-copy&pivots=api-nosql) from the default container to the container specified in the message container extractor. This option may require [changing the container partition key](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/change-partition-key).
```

jpalac (commented):
Looks good - just a couple of comments.

persistence/cosmosdb/index_outboxstorageformat_CosmosDB_(,3.1].partial.md
```diff
@@ -0,0 +1,2 @@
+> [!WARNING]
+> For control messages, a default partition key in the format `{messageId}` will be used, however these Outbox records are not separated by endpoint name. As a result, multiple logical endpoints cannot share the same database and container since [message identities are not unique across endpoints from a processing perspective](/nservicebus/outbox/#message-identity). To avoid conflicts, either separate different endpoints into different containers or [override the partition key](transactions.md).
```
Comment (jpalac) [inline persistence/cosmosdb/index_outboxstorageformat_CosmosDB_(,3.1].partial.md]: Should we add to this note that this issue has been solved in the 3,2 so we recommend customers use that version instead?

persistence/cosmosdb/index_outboxstorageformat_CosmosDB_[3.2,).partial.md
```diff
@@ -0,0 +1,9 @@
...
+> The [default partition key must be overwritten](/persistence/cosmosdb/transactions.md#specifying-the-partitionkey-to-use-for-the-transaction) whenever the message handler creates or updates business documents in CosmosDB. To guarantee atomicity, explicitly set the Outbox partition key to match the partition key of your business document (e.g., an `OrderId` or `CustomerId`). This is a requirement for including both the business document and the Outbox record in the same [Cosmos DB transactional batch](https://learn.microsoft.com/en-us/azure/cosmos-db/partitioning-overview). Conversely, for simplicity, you can use the default partition key when a handler's logic does not involve persisting business data.
+
+To support backward compatibility of control messages during migration, the persistence includes a fallback mechanism. When enabled (default), and if a record is not found using the synthetic key format, the system falls back to the legacy `{messageId}` format. The fallback can be disabled using `DisableReadFallback()` once all legacy records have expired, which is recommended since the fallback mechanism involves an additional read operation on the Outbox container.
```
jpalac: Perhaps put the  `DisableReadFallback()` into a code block that includes the object it is to be set on? So the sentence could become something like:

"Since the fallback mechanism involves an additional read operation on the Outbox container, it is recommended to turn it off once all legacy records have expired."

`config.DisableReadFallback()`

persistence/cosmosdb/defaultcontainer.include.md
```diff
@@ -1,11 +1,16 @@
...
+> [!NOTE]
+> **Added in version 3.2.0:** The exception to the above is if the extractor fails to pull container information from the message (logical), then the container information in the header (physical) will be used. If this also fails, the configured default container will be used as a fallback. If there is no default container configured, an exception will be thrown.
+#end-if
+
+When installers are enabled, this (default) container will be created if it doesn't exist. To opt-out of creating the default container, either disable the installers or use `DisableContainerCreation()`:
 
 snippet: CosmosDBDisableContainerCreation
 
-Any other containers that are resolved by extracting partition information from incoming messages need to be [manually created in Azure](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/how-to-create-container).
+Any other containers that are resolved by extracing partition information from incoming messages need to be [manually created in Azure](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/how-to-create-container).
```
PhilBastian:
```suggestion
Any other containers that are resolved by extracting partition information from incoming messages need to be [manually created in Azure](https://learn.microsoft.com/en-us/azure/cosmos-db/nosql/how-to-create-container).
```

persistence/cosmosdb/defaultcontainer.include.md
```diff
@@ -1,11 +1,16 @@
-The default container can be set by using the following configuration API:
+The container that is used by default for all incoming messages is specified via the `DefaultContainer(..)` configuration API:
 
 snippet: CosmosDBDefaultContainer
 
-The container that is used by default for all incoming messages is specified via `DefaultContainer(..)`. When installers are enabled, this (default) container will be created if it doesn't exist.
+#if-version [3, )
+**Added in version 3.2.0:** If the container information is being extracted during runtime from a message instance (header or message body), the default container specified will be overwritten by the last extractor in the pipeline. For example, if a [Header Extractor](/persistence/cosmosdb/transactions.md#specifying-the-container-to-use-for-the-transaction-using-message-header-values) (physical stage) and a [Message Extractor](/persistence/cosmosdb/transactions.md#specifying-the-container-to-use-for-the-transaction-using-the-message-contents) (logical stage) are both configured, then the container information within the Message Extractor would be used.
```
PhilBastian, line 6: why make reference to the version it was added in, if the user has already selected that version (or above) in the docs version selector? Can this docs snippet not be made conditional on 3.2 and above?

warwickschroeder: We cant select minor versions in the docs version selector. So if the user is on version 3.1 of the package, this wont be available to them. It seems this is the recommended way (now) to handle minor versions in docs, based off the failed checks when I added a minor versioned snippet.

PR #7868, opened 2025-10-13
Update native integration documentation to specify that ContentType is copied to NServiceBus.ContentType

mauroservienti:
Documentation for:

- https://github.com/Particular/NServiceBus.Transport.AzureServiceBus/pull/1256

### DIFF
--- a/transports/azure-service-bus/native-integration.md
+++ b/transports/azure-service-bus/native-integration.md
@@ -4,7 +4,7 @@ summary: How to integrate NServiceBus endpoints with non-NServiceBus endpoints o
 component: ASBS
 related:
  - samples/azure-service-bus-netstandard/native-integration
-reviewed: 2024-05-17
+reviewed: 2025-10-14
 ---
 
 This document guides integrating NServiceBus endpoints with non-NServiceBus endpoints by sharing an Azure Service Bus (ASB) namespace as a mutual communication channel.


transports/azure-service-bus/native-integration.md
```diff
@@ -14,12 +14,23 @@ This document guides integrating NServiceBus endpoints with non-NServiceBus endp
...
-1. By default, the transport creates its own entities when they don't exist in the namespace. However non-NServiceBus endpoints may require manual creation of entities. Refer to the [Azure Service Bus documentation](https://docs.microsoft.com/en-us/azure/service-bus-messaging/) for more information on available ASB SDKs and tools to perform these tasks.
+#if-version [5.1.1, )
 
-1. The native message must allow NServiceBus to [detect the message type either via the headers or the message payload](/nservicebus/messaging/message-type-detection.md).
+> [!NOTE]
+> If the incoming Service Bus message specifies a value for the `ContentType` property, the value is used to populate the `NServiceBus.ContentType` header. 
+
+#end-if
+
+#if-version [4.2.5, 5)
```
afprtclr: What about version 3.2.7?

mauroservienti: I missed it.

@DavidBoike, is there any way to do a sort of a OR in those IFs? Ideally, it should be:

```
#if-version [5.1.1, ) || [4.2.5, 5) || [3.2.1, 4)
```

Because the text is always the same for all those versions.

DavidBoike: Partials and inline ifs are never going to care about versions that don't show up in the version dropdown. And that means versions that have their own snippets. And we're really trying to stop having snippet projects for minors, so really this should just mean majors going forward.

The user is going to support major 1, 2, 3, 4, or 5 from the dropdown, they aren't typing in "I have 4.2.5 so show me the docs for that."

If you need any sort of note that is more nuanced than a major, you should include that in the text itself instead of trying to branch off of it.

PR #7874, opened 2025-10-16
Delete partial for PlatformSample

tamararivera:
This partial is used on samples that use the PlatformSample, that used to be Windows only, but not anymore (I run a few on Mac). It also mentions that the tools are not necessary to run the sample, which is not accurate, as all the samples that reference it are samples that make use of ServiceControl or ServicePulse


PhilBastian:
I believe what it was trying to say was that you don't need to install ServiceControl/ServicePulse separately, since the sample bundles a portable version of them that will get launched with the sample

PR #7893, opened 2025-11-04
Fix Sample Receiver Endpoint startup on newer frameworks

mattmercurio:
In Core_8 and previous versions of this sample, `Endpoint.Start()` was used in the Receiver endpoint. `Console.ReadKey()` was provided as a way to exit the console.

In our newer samples, the Receiver endpoint uses `RunAsync()` on the host; `Console.ReadKey()` was preventing the startup of the endpoint and has been removed here.


hazel-bohon (approved):
lgtm

PR #7917, opened 2025-11-28
fix: asb versioning typo

janv8000:
I believe the intention was to publish both message types to the same topicname.

Cleaned up the redundant duplicate region


danielmarbach (approved):

danielmarbach:
Thanks!

PR #7932, opened 2025-12-05
SqlSaga class is deprecated

DavidBoike:
Edits related to the deprecation of `SqlSaga<T>` in version 8.3 (NServiceBus 9) ahead of the deprecation with error in version 9.0 (NServiceBus 10).

Also removes snippets for SQL Persistence versions 4 and 5 which are unsupported.

### DIFF
--- a/persistence/sql/sqlsaga.md
+++ b/persistence/sql/sqlsaga.md
@@ -16,6 +16,4 @@ reviewed: 2025-12-05
 >
 > The generated [table name](/persistence/sql/saga.md#table-structure-table-name) can also be set using the `[SqlSaga]` attribute.
 
-partial: required-in-some-versions
-
 partial: main-content


persistence/sql/sqlsaga_main-content_sqlpersistence_[,8).partial.md
```diff
@@ -0,0 +1,35 @@
+## SqlSaga definition
+
+A saga can be implemented using the `SqlSaga` base class as follows:
```
danielmarbach, line 5: Should we add a note here about it no longer being the preferred thing?

DavidBoike: Yes … you gave me an idea, I'm going to reorganize.

PR #7947, opened 2025-12-11
Document ServiceControl / ServicePulse support for authentication and SSL/TLS

jasontaylordev:
Add security documentation for the following PRs:
- https://github.com/Particular/ServiceControl/pull/5197
- https://github.com/Particular/ServicePulse/pull/2729

### DIFF
--- a/servicecontrol/security/configuration/tls-troubleshooting.include.md
+++ b/servicecontrol/security/configuration/tls-troubleshooting.include.md
@@ -40,7 +40,7 @@
 
 **Symptom**: Browser displays "Your connection is not private" or similar certificate warnings.
 
-**Cause**: The certificate is self-signed, expired, or doesn't match the hostname.
+**Cause**: The client (browser or another container) doesn't trust the CA that signed the server's certificate.
 
 **Solutions**:
 
@@ -51,3 +51,40 @@
   - If running in a Windows Service as `Local System`, the certificate should be in the `Local Computer` store.
   - If running as yourself, the certificate should be in the `Current User` store.
 - For internal/development use, add the self-signed certificate to the trusted root store
+- For containers: Add the CA certificate to the CA bundle and set `SSL_CERT_FILE`
+
+### The remote certificate is invalid according to the validation procedure
+
+**Symptom**: Service starts but fails when calling other services (e.g., ServiceControl can't reach ServiceControl-Audit) or when validating tokens with Azure AD.
+
+**Cause**: The CA bundle doesn't contain the CA certificate that signed the remote server's certificate.
+
+**Solutions**:
+
+- Ensure the CA bundle includes your local CA certificate (e.g., mkcert's rootCA.pem)
+- For Azure AD authentication, append the Mozilla CA bundle: curl https://curl.se/ca/cacert.pem >> ca-bundle.crt
+- Verify `SSL_CERT_FILE` environment variable points to the correct path inside the container
+- Check the CA bundle is mounted correctly
+
+### The certificate doesn't match the hostname
+
+**Symptom**: Browser or client rejects the certificate with a hostname mismatch error
+
+**Cause**: The certificate's Subject Alternative Names (SANs) don't include the hostname being used to connect.
+
+**Solutions**:
+
+- Regenerate the certificate with all required hostnames
+- Include both the Docker service name (for container-to-container) and localhost (for host access)
+
+### Health checks fail with certificate errors
+
+**Symptom**: Container health check reports unhealthy, logs show SSL errors from the health check command.
+
+**Cause**: The health check binary inside the container doesn't trust the certificate.
+
+**Solutions**:
+
+- Ensure `SSL_CERT_FILE` is set so the health check process can find the CA bundle
+- Check the health check URL uses the correct protocol: `https://` not `http://`
+- Verify the certificate includes localhost in its SANs (health checks run inside the container)
--- a/servicecontrol/security/configuration/tls.md
+++ b/servicecontrol/security/configuration/tls.md
@@ -50,6 +50,55 @@ When ServiceControl handles TLS directly using a PFX certificate:
 <add key="ServiceControl/Https.CertificatePassword" value="{certificate-password}" />
 ```
 
+**Container**:
+
+> [!NOTE]
+> The following is a docker compose snippets. For full examples, see the [Platform Container Examples repository](https://github.com/Particular/PlatformContainerExamples).
+
+```bash
+servicecontrol:
+    image: particular/servicecontrol:latest
+    env_file: .env
+    ports:
+      - "33333:33333"
+    environment:
+      RAVENDB_CONNECTIONSTRING: http://servicecontrol-db:8080
+      REMOTEINSTANCES: '[{"api_uri":"https://servicecontrol-audit:44444/api"}]'
+      SERVICECONTROL_HTTPS_ENABLED: "true"
+      SERVICECONTROL_HTTPS_CERTIFICATEPATH: "/usr/share/ParticularSoftware/certificate.pfx"
+      SERVICECONTROL_HTTPS_CERTIFICATEPASSWORD: "..."
+      SSL_CERT_FILE: "/etc/ssl/certs/ca-bundle.crt"
+      SERVICECONTROL_AUTHENTICATION_ENABLED: "true"
+      SERVICECONTROL_AUTHENTICATION_AUTHORITY: "https://login.microsoftonline.com/{tenantId}"
+      SERVICECONTROL_AUTHENTICATION_AUDIENCE: "api://{api-id}"
+      SERVICECONTROL_AUTHENTICATION_SERVICEPULSE_CLIENTID: "{servicepulse-clientid}"
+      SERVICECONTROL_AUTHENTICATION_SERVICEPULSE_AUTHORITY: "https://login.microsoftonline.com/{tenantId}/v2.0"
+      SERVICECONTROL_AUTHENTICATION_SERVICEPULSE_APISCOPES: '["api://{api-id}/access_as_user"]'
+    command: --setup-and-run
+    restart: unless-stopped
+    volumes:
+      - C:\path\to\servicecontrol\cert.pfx:/usr/share/ParticularSoftware/certificate.pfx
+      - C:\path\to\servicecontrol\ca-bundle.crt:/etc/ssl/certs/ca-bundle.crt:ro
+    healthcheck:
+      test: ["CMD", "/healthcheck/healthcheck", "https://localhost:33333/api"]
+      interval: 30s
+      timeout: 10s
+      start_period: 60s
+      retries: 3
+    depends_on:
+      servicecontrol-db:
+        condition: service_healthy
+      rabbitmq:
+        condition: service_healthy
+```
+
+**SSL_CERT_FILE (CRT)**: This tells SSL/TLS libraries (including .NET's HttpClient, OpenSSL, and libcurl) the path to a CA (Certificate Authority) certificate bundle file. This bundle contains the public certificates of trusted CAs used to verify the authenticity of SSL/TLS certificates presented by remote servers. This certificate is used to verify HTTPS (trust others).
+
+**SERVICEPULSE_HTTPS_CERTIFICATEPATH (PFX)**: This is a binary archive format that bundles together the private key, public certificate, and certificate chain. This certificate is used to serve HTTPS (prove identity).
+
+> [!NOTE]
+> When containers communicate with each other over HTTPS, they use the Docker service names (like `servicecontrol`, `servicecontrol-audit`) as hostnames, and TLS validation will fail if the certificate doesn't include these names in its Subject Alternative Names (SANs).
+
 ### Direct HTTPS with HSTS
 
 When ServiceControl handles TLS directly and you want to enable HSTS:
--- a/servicepulse/security/configuration/tls.md
+++ b/servicepulse/security/configuration/tls.md
@@ -69,14 +69,41 @@ When ServicePulse handles TLS directly using a PFX certificate:
 
 **Container:**
 
+> [!NOTE]
+> The following is a docker compose snippets. For full examples, see the [Platform Container Examples repository](https://github.com/Particular/PlatformContainerExamples).
+
 ```bash
-docker run -e SERVICEPULSE_HTTPS_ENABLED=true \
-           -e SERVICEPULSE_HTTPS_CERTIFICATEPATH=/certs/servicepulse.pfx \
-           -e SERVICEPULSE_HTTPS_CERTIFICATEPASSWORD={certificate-password} \
-           -v /path/to/certs:/certs \
-           ...
-           particular/servicepulse:latest
+servicepulse:
+    image: particular/servicepulse:latest
+    ports:
+      - "9090:9090"
+    environment:
+      SERVICECONTROL_URL: https://servicecontrol:33333
+      MONITORING_URL: https://servicecontrol-monitoring:33633
+      SERVICEPULSE_HTTPS_ENABLED: "true"
+      SERVICEPULSE_HTTPS_CERTIFICATEPATH: "/usr/share/ParticularSoftware/certificate.pfx"
+      SERVICEPULSE_HTTPS_CERTIFICATEPASSWORD: "..."
+      SSL_CERT_FILE: "/etc/ssl/certs/ca-bundle.crt"
+      ASPNETCORE_URLS: "https://+:9090"
+    restart: unless-stopped
+    volumes:
+      - C:\path\to\servicecontrol\cert.pfx:/usr/share/ParticularSoftware/certificate.pfx
+      - C:\path\to\servicecontrol\ca-bundle.crt:/etc/ssl/certs/ca-bundle.crt:ro
+    depends_on:
+      servicecontrol:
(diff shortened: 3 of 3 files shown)


servicecontrol/security/configuration/authentication.md
```diff
@@ -0,0 +1,144 @@
...
+
+### Core Settings
+
+| Environment Variable                | App.config                          | Default | Description                                                                               |
+|-------------------------------------|-------------------------------------|---------|-------------------------------------------------------------------------------------------|
+| `{PREFIX}_AUTHENTICATION_ENABLED`   | `{PREFIX}/Authentication.Enabled`   | `false` | Enable JWT authentication                                                                 |
+| `{PREFIX}_AUTHENTICATION_AUTHORITY` | `{PREFIX}/Authentication.Authority` | (none)  | OpenID Connect authority URL (e.g., `https://login.microsoftonline.com/{tenant-id}/v2.0`) |
+| `{PREFIX}_AUTHENTICATION_AUDIENCE`  | `{PREFIX}/Authentication.Audience`  | (none)  | The audience identifier (typically your API identifier or client ID, e.g., `api://servicecontrol`)                      |
+
+### Token Validation Settings
```
abparticular: This table submarines under the right hand menu for me, it's maybe too wide? Although not sure what to do about it.  Maybe take out the default column and say all values default to true in the header text?

abparticular: Or maybe it should follow the format in http://docs.particular.net/servicecontrol/audit-instances/configuration#host-settings-servicecontrol-auditinstancename.  The app.config name wraps because of the '.', but the env var names are very long and the Descriptions end up being squished.  

These are already documented in the configuration.md files under each instance, are they really needed here again?  Can we link to the sections for each instance instead?

servicecontrol/security/configuration/forward-headers.md
```diff
@@ -0,0 +1,115 @@
+---
+title: Forward Headers Configuration
+summary: How to enable and configure forward headers for ServiceControl instances
+reviewed: 2026-01-12
+component: ServiceControl
+---
+
+When ServiceControl instances are deployed behind a reverse proxy (like NGINX, Traefik, or a cloud load balancer) that terminates SSL/TLS, you need to configure forwarded headers so ServiceControl correctly understands the original client request.
```
abparticular:
```suggestion
When ServiceControl instances are deployed behind a reverse proxy that terminates SSL/TLS (like NGINX, Traefik, or a cloud load balancer), you need to configure forwarded headers so ServiceControl correctly understands the original client request.
```

servicecontrol/security/configuration/index.md
```diff
@@ -0,0 +1,37 @@
...
+### [CORS](cors.md)
+
+Configure Cross-Origin Resource Sharing to control which web applications can access the ServiceControl API.
+
+## Configuration methods
+
+ServiceControl instances can be configured via environment variables or App.config. Each instance type uses a different prefix:
+
+include: servicecontrol-instance-prefix
+
```
abparticular:
```suggestion
```
This is mentioned in plenty of places, not sure it's needed here?

servicecontrol/security/configuration/tls.md
```diff
@@ -0,0 +1,105 @@
...
+| `{PREFIX}_HTTPS_CERTIFICATEPASSWORD`   | `{PREFIX}/Https.CertificatePassword`         | (none)     | Password for the certificate file                              |
+| `{PREFIX}_HTTPS_REDIRECTHTTPTOHTTPS`   | `ServiceControl/Https.RedirectHttpToHttps`   | `false`    | Redirect HTTP requests to HTTPS                                |
+| `{PREFIX}_HTTPS_PORT`                  | `ServiceControl/Https.Port`                  | (none)     | HTTPS port for redirect (required for reverse proxy scenarios) |
+| `{PREFIX}_HTTPS_ENABLEHSTS`            | `{PREFIX}/Https.EnableHsts`                  | `false`    | Enable HTTP Strict Transport Security                          |
+| `{PREFIX}_HTTPS_HSTSMAXAGESECONDS`     | `ServiceControl/Https.HstsMaxAgeSeconds`     | `31536000` | HSTS max-age in seconds (default: 1 year)                      |
+| `{PREFIX}_HTTPS_HSTSINCLUDESUBDOMAINS` | `ServiceControl/Https.HstsIncludeSubDomains` | `false`    | Include subdomains in HSTS policy                              |
+
+## Security Considerations
+
+### Certificate Management
```
abparticular: This section seems more like dev notes than public documentation?  

Maybe 
> Restrict read access to the certificate file to only the ServiceControl service account
can be expanded upon and belong here?

servicecontrol/security/configuration/tls.md
```diff
@@ -0,0 +1,105 @@
...
+- Use environment variables rather than App.config (environment variables are not persisted to disk)
+
+### HSTS Considerations
+
+- HSTS should not be tested on localhost because browsers cache the policy, which could break other local development
+- HSTS is disabled in Development environment (ASP.NET Core excludes localhost by default)
+- HSTS can be configured at either the reverse proxy level or in ServiceControl (but not both)
+- HSTS is cached by browsers, so test carefully before enabling in production
+- Start with a short max-age during initial deployment
+- Consider the impact on subdomains before enabling `includeSubDomains`
```
abparticular: > - HSTS should not be tested on localhost because browsers cache the policy, which could break other local development
> - HSTS is disabled in Development environment (ASP.NET Core excludes localhost by default)

dev notes?

> - HSTS can be configured at either the reverse proxy level or in ServiceControl (but not both)
> - HSTS is cached by browsers, so test carefully before enabling in production
> - Start with a short max-age during initial deployment

These should probably be documented and expanded upon

> - Consider the impact on subdomains before enabling `includeSubDomains`

I'm not sure what this means, maybe the considerations should be spelled out more?

servicecontrol/security/configuration/authentication.md
```diff
@@ -0,0 +1,144 @@
...
+<!-- ServicePulse settings -->
+<add key="ServiceControl/Authentication.ServicePulse.ClientId" value="{spa-client-id}" />
+<add key="ServiceControl/Authentication.ServicePulse.Authority" value="https://{keycloak-host}/realms/{realm}" />
+<add key="ServiceControl/Authentication.ServicePulse.ApiScopes" value="[&quot;{scope}&quot;]" />
+```
+
+## Limitations
+
+### Anonymous endpoints
+
```
abparticular, line 106: I don't think `Anonymous endpoints`, `Scatter-gather endpoints` and `Token Forwarding Security Considerations` sections belong on this configuration page.  Maybe move these sections to the index.md page which links here?

abparticular: Also maybe get rid of limitations header and promote the remaining subsection?

servicecontrol/security/index.md
```diff
@@ -0,0 +1,103 @@
...
+    Note over Primary,Audit: Scatter-gather (token forwarded)
+    Primary->>Audit: Forward request + Bearer {token}
+    Audit-->>Primary: Response data
+    Primary-->>ServicePulse: Aggregated API response
+    ServicePulse-->>User: Display data
+```
+
+Certain endpoints remain accessible without authentication to support API discovery, client bootstrapping, and scatter-gather communication between ServiceControl instances. See [Authentication Configuration](configuration/authentication.md#limitations) for the full list of anonymous endpoints and security considerations.
+
+See [Authentication Configuration](configuration/authentication.md) for settings, or [Microsoft Entra ID](entra-id-authentication.md) for a complete setup guide.
```
abparticular: Lines 59 and 61 both have links to `Authentication Configuration` but they are different URLs.  I'm not sure the "limitations" URL should even have the list of anonymous endpoints.

servicecontrol/security/configuration/authentication.md
```diff
@@ -0,0 +1,144 @@
...
+| Endpoint                     | Purpose                                                   |
+|------------------------------|-----------------------------------------------------------|
+| `/api`                       | API root/discovery - returns available endpoints          |
+| `/api/instance-info`         | Returns instance configuration information                |
+| `/api/configuration`         | Returns instance configuration information (alias)        |
+| `/api/configuration/remotes` | Returns remote instance configurations for scatter-gather |
+
+These endpoints allow the Primary instance to discover remote instances.
+
+### Same Authentication Configuration Required
```
abparticular:
```suggestion
### Matching authentication configuration required
```

servicecontrol/security/configuration/authentication.md
warwickschroeder: I think thats a good idea. No need to duplicate this here.

servicecontrol/security/configuration/index.md
warwickschroeder: Agreed. Will remove.

servicecontrol/security/configuration/tls.md
warwickschroeder: I've cleaned up that list a bit, but have left a couple that the customer should know about.
>- ServiceControl supports PFX (PKCS#12) certificate files
>- Store certificate files securely with appropriate file permissions

servicecontrol/security/configuration/tls.md
warwickschroeder: I've taken a different approach and have removed the points altogether. I now have a short description of what HSTS is, and then a link that goes into more detail about the options.

johnsimons:
[This](https://docs.particular.net/servicecontrol/securing-servicecontrol) page has not been updated at all, I was expecting to see some changes there

servicecontrol/security/configuration/tls-troubleshooting.include.md
```diff
@@ -0,0 +1,50 @@
...
+**Symptom**: The application fails to start with certificate-related errors when using a PFX certificate file.
+
+**Cause**: The certificate file path is incorrect, the file is not readable, or the password is wrong.
+
+**Solutions**:
+
+- Verify the certificate path is correct and the file exists
+- Check that the certificate file has appropriate permissions
+- Confirm the certificate password is correct
+- For containers, ensure the volume mount is correct (e.g., `-v /path/to/certs:/certs`)
```
johnsimons: Is this a good example? I thought you needed to volume map the file itself, not the folder?

servicecontrol/security/configuration/tls.md
```diff
@@ -0,0 +1,86 @@
...
+related:
+- servicecontrol/security/hosting-guide
+- servicepulse/security/configuration/tls
+---
+
+ServiceControl instances can be configured to use HTTPS directly, enabling encrypted connections without relying on a reverse proxy for SSL termination.
+
+## Configuration
+
+ServiceControl instances can be configured via environment variables or App.config. Each instance type uses a different prefix. See the [Hosting Guide](../hosting-guide.md) for example usage of these configuration settings in conjustion with [Authentication](authentication.md) and [Forward Headers](forward-headers.md) configuration settings in a scenario based format.
```
johnsimons:
```suggestion
ServiceControl instances can be configured via environment variables or App.config. Each instance type uses a different prefix. See the [Hosting Guide](../hosting-guide.md) for example usage of these configuration settings, along with [Authentication](authentication.md) and [Forward Headers](forward-headers.md), in a scenario-based format.
```

servicepulse/security/configuration/forward-headers.md
```diff
@@ -0,0 +1,205 @@
...
+---
+
+> [!NOTE]
+> This page is **not** relevant if the [ServicePulse static files have been extracted](/servicepulse/install-servicepulse-in-iis.md), and is being hosted in anything other than the [Container](/servicepulse/containerization/) or [Windows Service](/servicepulse/installation.md) hosting options provided. Forward header configuration is only required if using a reverse proxy.
+
+When ServicePulse is deployed behind a reverse proxy that terminates SSL/TLS (like nginx, Traefik, or a cloud load balancer), you need to configure forwarded headers so ServicePulse correctly understands the original client request.
+
+## Configuration
+
+There are two hosting options for ServiceControl, [Container](/servicepulse/containerization/) and [Windows Service](/servicepulse/installation.md). The container is configured via environment variables and the windows service is configured using command-line arguments. See the [Hosting Guide](../hosting-guide.md) for example usage of these configuration settings in conjustion with [Authentication](authentication.md) and [TLS](tls.md) configuration settings in a scenario based format.
```
johnsimons:
```suggestion
There are two hosting options for ServiceControl, [Container](/servicepulse/containerization/) and [Windows Service](/servicepulse/installation.md). The container is configured via environment variables, while the Windows Service is configured using command-line arguments. See the [Hosting Guide](../hosting-guide.md) for example usage of these configuration settings, along with [Authentication](authentication.md) and [TLS](tls.md), in a scenario-based format.
```

servicepulse/security/configuration/tls.md
```diff
@@ -0,0 +1,168 @@
...
+---
+
+> [!NOTE]
+> This page is **not** relevant if the [ServicePulse static files have been extracted](/servicepulse/install-servicepulse-in-iis.md), and is being hosted in anything other than the [Container](/servicepulse/containerization/) or [Windows Service](/servicepulse/installation.md) hosting options provided. If using [authentication](/servicepulse/security/configuration/authentication.md), it is recommended to use TLS encryption.
+
+ServicePulse can be configured to use HTTPS directly, enabling encrypted connections without relying on a reverse proxy for SSL termination.
+
+## Configuration
+
+There are two hosting options for ServiceControl, [Container](/servicepulse/containerization/) and [Windows Service](/servicepulse/installation.md). The container is configured via environment variables and the windows service is configured using command-line arguments. See the [Hosting Guide](../hosting-guide.md) for example usage of these configuration settings in conjustion with [Authentication](authentication.md) and [Forward Headers](forward-headers.md) configuration settings in a scenario based format.
```
johnsimons:
```suggestion
There are two hosting options for ServiceControl, [Container](/servicepulse/containerization/) and [Windows Service](/servicepulse/installation.md). The container is configured via environment variables, while the Windows Service is configured using command-line arguments. See the [Hosting Guide](../hosting-guide.md) for example usage of these configuration settings, along with [Authentication](authentication.md) and [Forward Headers](forward-headers.md), in a scenario-based format.
```

servicecontrol/security/configuration/cors.md
```diff
@@ -0,0 +1,98 @@
+---
+title: ServiceControl CORS Configuration
+summary: How to configure Cross-Origin Resource Sharing for ServiceControl instances
+reviewed: 2026-01-14
+component: ServiceControl
+related:
+- servicecontrol/security/hosting-guide
+---
+
+Cross-Origin Resource Sharing (CORS) controls which web applications can make requests to ServiceControl. This is important when ServicePulse is hosted on a different domain than ServiceControl.
```
PhilBastian, line 10: isn't it also important if ServicePulse is hosted on the same domain but a different port? Essentially it's required unless SP is hosted out of SC, which hasn't been released yet

servicecontrol/security/configuration/cors.md
```diff
@@ -0,0 +1,98 @@
...
+</appSettings>
+```
+
+## Troubleshooting
+
+### ServicePulse cannot connect to ServiceControl
+
+**Symptom**: ServicePulse displays connection errors or fails to load data, and browser developer tools show CORS errors like "Access-Control-Allow-Origin" or "blocked by CORS policy".
+
+**Cause**: The ServicePulse origin is not in the allowed origins list, or the origin URL doesn't match exactly (including protocol and port).
```
PhilBastian, line 53: there is no reference made to port anywhere above

servicecontrol/servicecontrol-instances/configuration.md
```diff
@@ -175,6 +175,444 @@ Run [ServiceControl error instance in maintenance mode](/servicecontrol/ravendb/
...
+
+| Type | Default value |
+| --- | --- |
+| bool | `true` |
+
+### ServiceControl/ForwardedHeaders.TrustAllProxies
+
+_Added in version 6.11.0_
+
+Trusts forwarded headers from any source. Set to `false` when using `KnownProxies` or `KnownNetworks`.
```
PhilBastian, line 530: > Set to `false` when using `KnownProxies` or `KnownNetworks`.

this is ambiguous in its meaning. Should it be:
- Set this to `false` when using `KnownProxies` or `KnownNetworks`; or
- This is set to `false` when using `KnownProxies` or `KnownNetworks`?

the warning below suggests the former

servicecontrol/security/configuration/cors.md
warwickschroeder: The port does matter, so we should probably add this. Its also worth noting that by default, CORS allows all origins. So if the customer doesn't change this, the request from the same host but different port will succeed.

servicecontrol/security/configuration/cors.md
warwickschroeder: We should add info about the port and protocol where you flagged this higher up in the doc

servicecontrol/servicecontrol-instances/configuration.md
warwickschroeder: So the warning is just making it clear that its recommended that this setting should be set to false, however which way that is done. But the user can either set it explicitly, or if the users sets `KnownProxies` or `KnownNetworks`, then this setting is automatically set to false. So it should be latter.

PR #7969, opened 2025-12-19
Upgrade Guide: Features, Installers, IWant

DavidBoike:
* Features will not be assembly scanned in the future and should be added explicitly, plus method declarations
* Installers will not be assembly scanned in the future and should be added explicitly
* `IWantToRunBeforeConfigurationIsFinalized` deprecated

### DIFF
--- a/nservicebus/upgrades/9to10/index.md
+++ b/nservicebus/upgrades/9to10/index.md
@@ -103,7 +103,7 @@ endpointConfiguration.EnableFeature<MyFeature>();
 endpointConfiguration.DisableFeature<MyFeature>();
 ```
 
-Lastly, any `Feature` classes must now have a public paramaterless constructor.
+Lastly, any `Feature` classes must now have a paramaterless constructor.
 
 ### Installers
 


nservicebus/upgrades/9to10/index.md
```diff
@@ -69,6 +74,55 @@ This section describes changes to advanced extensibility APIs.
...
+```csharp
+// Instead of:
+endpointConfiguration.EnableFeature(typeof(MyFeature));
+endpointConfiguration.DisableFeature(typeof(MyFeature));
+// Use instead:
+endpointConfiguration.EnableFeature<MyFeature>();
+endpointConfiguration.DisableFeature<MyFeature>();
+```
+
+Lastly, any `Feature` classes must now have a public paramaterless constructor.
```
danielmarbach: The visibility is not required. You can make it internal as long as the code that enables it sees the ctor

```suggestion
Lastly, any `Feature` classes must now have a paramaterless constructor.
```

PR #7976, opened 2026-01-05
Saga not found handler updates

danielmarbach:
This PR documents the changes introduced in https://github.com/Particular/NServiceBus/pull/7516 for NServiceBus v10

Instead of using the interface name all over the place, we are now using the concept, which avoids having to duplicate countless pages.

### DIFF
--- a/nservicebus/sagas/analyzers_not-found-handler_core_[10,).partial.md
+++ b/nservicebus/sagas/analyzers_not-found-handler_core_[10,).partial.md
@@ -2,4 +2,4 @@
 
 A [saga not found handler](/nservicebus/sagas/saga-not-found.md) provides a way to deal with messages that are not allowed to start a saga but cannot find existing saga data.
 
-Saga not found handlers may require specific dependencies to be injected into the class and may implement non-trivial handling logic. To separate those concerns from the saga an independent class should implement `ISagaNotFoundHandler` and be mapped accordingly.
+Saga not found handlers may require specific dependencies to be injected into the class and may implement non-trivial handling logic. To separate those concerns from the saga an independent class should implement `ISagaNotFoundHandler` and be mapped with `mapper.ConfigureNotFoundHandler<MyNotFoundHandler>()`.


Snippets/Core/Core_10/Sagas/SagaNotFoundHandler.cs
```diff
@@ -16,11 +16,43 @@ public Task Handle(object message, IMessageProcessingContext context)
...
 
-public class SagaDisappearedMessage
+public class SagaDisappearedMessage;
+
+class SagaUsingNotFoundHandler : Saga<NotFoundSagaData>, IAmStartedByMessages<StartSagaMessage>, IHandleMessages<AnotherSagaMessage>
+{
+    protected override void ConfigureHowToFindSaga(SagaPropertyMapper<NotFoundSagaData> mapper)
+    {
+        mapper.ConfigureNotFoundHandler<SagaNotFoundHandler>();
+        #endregion
```
danielmarbach, line 26: I simply made this v10 snippet larger, also containing the registration. To not overwhelm the page with too much noise, I moved the region inside the ConfigureHowToFindSaga method. I figured doing a partial in the page for the snippet to add another sentence seems to be overkill.

nservicebus/sagas/analyzers_not-found-handler_core_[10,).partial.md
```diff
@@ -0,0 +1,5 @@
+* **Example message**: A saga should not implement `ISagaNotFoundHandler`, as this gives access to the uninitialized saga data property. Implement `ISagaNotFoundHandler` on a separate class instead.
+
+A [saga not found handler](/nservicebus/sagas/saga-not-found.md) provides a way to deal with messages that are not allowed to start a saga but cannot find existing saga data.
+
+Saga not found handlers may require specific dependencies to be injected into the class and may implement non-trivial handling logic. To separate those concerns from the saga an independent class should implement `ISagaNotFoundHandler` and be mapped accordingly.
```
DavidBoike: Can we be more specific on what "and be mapped accordingly" means here?

nservicebus/sagas/saga-not-found.md
```diff
@@ -7,29 +7,21 @@ related:
 - samples/saga
 ---
 
-The messages that are handled by sagas can either start a new saga (if handled by `IAmStartedByMessages<T>`) or update an existing saga (if handled by `IHandleMessages<T>`). If the incoming message is meant to be handled by a saga but is not expected to start a new one, then NServiceBus uses [correlation rules](/nservicebus/sagas/#correlating-messages-to-a-saga) to find an existing saga. If no existing saga can be found, all implementations of `IHandleSagaNotFound` are executed. If no implementation can be found, the message is discarded without additional notification.
+The messages that are handled by sagas can either start a new saga (if handled by `IAmStartedByMessages<T>`) or update an existing saga (if handled by `IHandleMessages<T>`). If the incoming message is meant to be handled by a saga but is not expected to start a new one, then NServiceBus uses [correlation rules](/nservicebus/sagas/#correlating-messages-to-a-saga) to find an existing saga. If no existing saga can be found, all saga not found handlers are executed. If no handler is configured, the message is discarded without additional notification.
```
DavidBoike: Isn't this wrong since they're all tied to specific sagas now? Are we trying to say that all sagas that use this message type will execute their own not found handlers if they have them?

Or do we save that description for the multiple sagas section below?

But that's only for v10 - does this belong in a partial?

Snippets/Core/Core_10/Sagas/SagaNotFoundHandler.cs
DavidBoike: I hate split snippets but I get it. I wish we had some sort of `#region snippet-ignore` that turned multiple lines into `...`

Snippets/Core/Core_10/Sagas/SagaNotFoundHandler.cs
```diff
@@ -16,11 +16,43 @@ public Task Handle(object message, IMessageProcessingContext context)
...
+        mapper.MapSaga(saga => saga.CorrelationId)
+            .ToMessage<StartSagaMessage>(msg => msg.CorrelationId);
+    }
```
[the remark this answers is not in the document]
danielmarbach: That was deliberate

nservicebus/sagas/analyzers_not-found-handler_core_[10,).partial.md
danielmarbach:
```suggestion
Saga not found handlers may require specific dependencies to be injected into the class and may implement non-trivial handling logic. To separate those concerns from the saga an independent class should implement `ISagaNotFoundHandler` and be mapped with `mapper.ConfigureNotFoundHandler<MyNotFoundHandler>()`.
```

nservicebus/sagas/saga-not-found.md
danielmarbach: Good catch. I'll rework this

danielmarbach:
```suggestion
The messages that are handled by sagas can either start a new saga (if handled by `IAmStartedByMessages<T>`) or update an existing saga (if handled by `IHandleMessages<T>`). If the incoming message is meant to be handled by a saga but is not expected to start a new one, then NServiceBus uses [correlation rules](/nservicebus/sagas/#correlating-messages-to-a-saga) to find an existing saga. If no existing saga can be found, saga not found handlers are executed. If no handler is configured, the message is discarded without additional notification.
```

or

```suggestion
The messages that are handled by sagas can either start a new saga (if handled by `IAmStartedByMessages<T>`) or update an existing saga (if handled by `IHandleMessages<T>`). If the incoming message is meant to be handled by a saga but is not expected to start a new one, then NServiceBus uses [correlation rules](/nservicebus/sagas/#correlating-messages-to-a-saga) to find an existing saga. If no existing saga can be found, configured saga not found handlers are executed. If no handler is configured, the message is discarded without additional notification.
```

danielmarbach: Would it be enough to simply remove the "all"?

If we want to be more specific we could remove the all and after that create a new paragraph in the partial that explains the different in pre v10 vs after v10.

Thoughts?

DavidBoike: I think either option works pretty well.

PR #7989, opened 2026-01-09
Update snippet and sample alpha versions

DavidBoike:


DavidBoike:
Merging with failing tests because the AWSSDK vulnerability is breaking those on master as well.

PR #7991, opened 2026-01-12
Update critical errors documentation with revisions - Weekly legacy docs review

poornimanayar:


nservicebus/hosting/critical-errors.md
```diff
@@ -69,20 +67,21 @@ snippet: CustomHostErrorHandlingAction
 
 ### Implementation concerns
 
-partial: override
+> [!NOTE]
```
WilliamBZA, line 70: What are the benefits of not using the partial for this?

poornimanayar: Better readability. IMO it made no sense to have the two partials with the text alone separated away in the partial given we only support 8+ now. I believe the partial was used when NServiceBus 6 was still supported, looking at the name of the markdown files. With NSB6 out of support we could remove that and provide better readability to the whole page.

PR #7996, opened 2026-01-12
Update the SI support policy to clarify how long support is available

ngallegos:
ServiceInsight has been sunset, but it isn't clear what that means for support.  This PR clarifies that until a migration path to ServicePulse is defined and ServiceInsight is moved to a Deprecated status, support will still be offered.

### DIFF
--- a/serviceinsight/support-policy.md
+++ b/serviceinsight/support-policy.md
@@ -14,9 +14,9 @@ related:
 ---
 
 > [!WARNING]
-> ServiceInsight has been sunset and will receive no further updates.  
+> ServiceInsight has been sunset and will receive no further improvements or features.  
 >
-> The latest version of ServiceControl is supported until migration instructions to ServicePulse are available.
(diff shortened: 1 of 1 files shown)


soujay (approved):

serviceinsight/support-policy.md
```diff
@@ -14,9 +14,11 @@ related:
 ---
 
 > [!WARNING]
-> ServiceInsight has been sunset and will receive no further updates.  
+> ServiceInsight has been sunset and will receive no further improvements or features.  
+>
+> The latest version of ServiceControl is supported (including bug fixes) until migration instructions to ServicePulse are available.
```
udidahan, line 19: Is this supposed to say "ServiceControl" or should it be "ServiceInsight"?

Fixing in #8026

ngallegos: Yes, thank you for catching that.

PR #8174, opened 2026-04-14
Move docstool cache management file one directory up

DavidBoike:


DavidBoike:
Isn't going to work, hashFiles doesn't support using files outside the GITHUB_WORKSPACE

PR #8176, opened 2026-04-14
CI improvements

tamararivera:
This PR implements a few quality of life improvements:
- Rename `.github/workflows/verify-master.yml`to `.github/workflows/validate-master.yml` to align with naming of the workflow for pull requests
- ~~Move the content of `.github/workflows/build-samples-and-snippets.yml` to `.github/workflows/validate-master.yml` (including its cron schedule)~~
- Add workflow_dispatch to `.github/workflows/validate-master.yml` to easily trigger the workflow
- Set a default shell on both
- Change the key for the docstool version cache to just be the latest version found


.github/workflows/validate-master.yml
```diff
@@ -1,11 +1,17 @@
 name: VerifyMaster
 on:
+  schedule:
+    - cron: '0 2/4 * * 1-5' # Every 4 hours, Monday-Friday
```
DavidBoike: Why does it need to run on a schedule if it runs on every push to master?

.github/workflows/validate-master.yml
```diff
@@ -19,31 +25,31 @@ jobs:
...
+           echo "DOCSTOOL_VERSION=$(curl --connect-timeout 5 --max-time 15 --silent --show-error https://f.feedz.io/particular-software/packages/nuget/v3/packages/particular.docstool/index.json)" >> $GITHUB_ENV
       - name: Cache docstool
         id: cache-docstool
         uses: actions/cache@v5.0.5
         with:
           path: |
              ~/.dotnet/tools/docstool
              ~/.dotnet/tools/.store/particular.docstool
-          key: docstool-${{ hashFiles('docstool-versions.json') }}
+          key: docstool-${{ FromJson(env.DOCSTOOL_VERSION).version }}
```
DavidBoike, line 36: I don't get it. The result of https://f.feedz.io/particular-software/packages/nuget/v3/packages/particular.docstool/index.json is:

```json
{"versions":["6.19.0","6.19.1","6.19.2","7.0.0","7.1.0","7.1.1","7.1.2"]}
```
So `.version` off that is going to be undefined, but even `.versions` wouldn't be the single version you want.

Did you copy this one from elsewhere in the PR and is that other place using a different URL?

.github/workflows/validate-master.yml
```diff
@@ -82,4 +87,50 @@ jobs:
           $result = Invoke-RestMethod -Method POST -Uri https://slack.com/api/chat.postMessage -ContentType "application/json; charset=utf-8" -Headers $headers -Body $body
           Write-Output $result
           exit $(If ($result.ok) { 0 } Else { 1 })
-
+  build-samples-and-snippets:
```
DavidBoike: Oh, the schedule is because you're trying to bring this job into the same workflow. But why? And the result would be this would run against ALL samples and snippets on every push to master, which would not be good.

.github/workflows/validate-pull-requests.yml
```diff
@@ -17,15 +20,15 @@ jobs:
         # Caching is done to limit data transfer on each PR, not to optimize CI speed
       - name: Check docstool version for cache
         run: |
-           curl --connect-timeout 5 --max-time 15 --silent --show-error https://s3.amazonaws.com/particular.downloads/Particular.DocsTool/version.json >> docstool-version.json
+           echo "DOCSTOOL_VERSION=$(curl --connect-timeout 5 --max-time 15 --silent --show-error https://s3.amazonaws.com/particular.downloads/Particular.DocsTool/version.json)" >> $GITHUB_ENV
```
DavidBoike, line 23: Aha! Here it is the exact version. But I can't think of a reason this URL couldn't be used up above. It does represent the current release of DocsTool that you'd get from a dotnet tool install.

DavidBoike (changes requested):

.github/workflows/validate-master.yml
tamararivera: Yes, I just assumed both URL will return the same content

.github/workflows/validate-pull-requests.yml
tamararivera: Yeap, I didn't want to change it, in case it was there for some reason, but maybe both should use this URL

DavidBoike (approved):

PR #8182, opened 2026-04-16
NServiceBus.SqlTransport version 9 to 10 upgrade guide

tmasternak:


danielmarbach:
@tmasternak why not merge this for a prerelease so that it becomes visible and isn't lost?

danielmarbach:
Ping @tmasternak

PR #8188, opened 2026-04-17
Remove Samples for Version 8

hazel-bohon:


mauroservienti:
Looks like this `samples/unit-of-work/sample.md` sample needs to be dropped since there are no more versions for it. Worth checking why, though.

mauroservienti (approved):

kylierozegnal (approved):

mauroservienti:
@hazel-bohon, I pushed some commits to start addressing the broken links due to the samples removal

kylierozegnal (approved):

willie-levy:
I dont have th context, so not approving but removing myself

PR #8203, opened 2026-04-24
Change version check back to excluding 11+

rbev:
This flips the display of these unsupported types to 

* appear in the unsupported list in versions `< 10` 
* show a note saying they are supported from version 10.2
* remove them from the list from version 11 (as all versions of v11 will support them)


DavidBoike:
I'm so sorry, I missed that it was a list of UNSUPPORTED types.

I tried to change the italicized note to hopefully help other people like me who are reading it inverted.

PR #8204, opened 2026-04-24
Note about using the Azure Service Bus emulator with ServiceControl

mauroservienti:


afprtclr (approved):

servicecontrol/transports.md
```diff
@@ -85,9 +84,11 @@ For example, given a hierarchy namespace of `my-hierarchy` and error, audit, and
 - The [audit queue name](/servicecontrol/audit-instances/configuration.md#transport-servicebusauditqueue) would need to be set to `my-hierarchy/audit`.
 - The [monitoring queue name](/servicecontrol/monitoring-instances/configuration.md#transport-monitoringerrorqueue) would need to be set to `my-hierarchy/monitoring`.
 
-
 ### Example connection string
 
+> [!NOTE]
+> when using the [Azure Service Bus emulator](https://learn.microsoft.com/en-us/azure/service-bus-messaging/overview-emulator), the default connection string points to `localhost` for the Endpoint value. That results in an error being looged during ServiceControl instances startup. The error can be safely ignored as it doesn't afftect ServiceControl operations, instead it's related to the throughput and licensing portion of Servicecontrol, which will be anyway not available when using the Azure Service Bus emulator. It's possible to prevent the error to show up in the logs by creating an alias for `localhost` containing a `.`, such as for example, `servicebus-emulator.local` or by specifying an IP address, such as `127.0.0.1` for the Endpoint value.
```
danielmarbach:
```suggestion
> when using the [Azure Service Bus emulator](https://learn.microsoft.com/en-us/azure/service-bus-messaging/overview-emulator), the default connection string points to `localhost` for the Endpoint value. That results in an error being looged during ServiceControl instances startup. The error can be safely ignored as it doesn't affect ServiceControl operations, instead it's related to the throughput and licensing portion of Servicecontrol, which will be anyway not available when using the Azure Service Bus emulator. It's possible to prevent the error to show up in the logs by creating an alias for `localhost` containing a `.`, such as for example, `servicebus-emulator.local` or by specifying an IP address, such as `127.0.0.1` for the Endpoint value.
```

danielmarbach:
Quick question: I think @stevedowling and @PhilBastian looked into the restrictions of the emulator and concluded some of the limits are too low to realistically run the "platform on it". Has that turned out to be not a problem, or is it still, and should this be documented too?

mauroservienti:
We only validated that ServiceControl doesn't blow up. Considering the non-production nature of the emulator, I don't think we should put too much effort into validating if it can _reliably_ run the Platform. Unless there are more critical things.

afprtclr:
I also verified that SC receives error messages and shows them properly.

stevedowling:
The biggest restriction on using the emulator with Service Platform is Microsoft's hard limit of 10 connections. Each logical connection ends up being 2 physical because we open separate connections for send and receive.

mauroservienti:
> The biggest restriction on using the emulator with Service Platform is Microsoft's hard limit of 10 connections. Each logical connection ends up being 2 physical because we open separate connections for send and receive.

So, beyonfd samples, it's basically unusable 🤦

stevedowling:
I did hack the ASB transport locally to share a connection for send and receive. Using that, I was able to get a completely local (but slightly cut down) Azure specific version of the AWS Loan Broker going using Service Platform, docker compose and the emulator, but I am not sure of the value of modifying the transport to extend the usefulness of the emulator slightly.

It's a shame that Microsoft have taken such a hard stance on the emulator - based on [this](https://github.com/Azure/azure-service-bus-emulator-installer/issues/58#issuecomment-2984760245) comment. It could be a really useful tool for local dev, prototyping, and integration tests if it wasn't so restricted. Local dev in docker containers is a bit broken when you have to have each developer on their own ASB namespace as well. Too bad if you wanted to work on a plane.

servicecontrol/transports.md
```diff
@@ -85,9 +84,11 @@ For example, given a hierarchy namespace of `my-hierarchy` and error, audit, and
 - The [audit queue name](/servicecontrol/audit-instances/configuration.md#transport-servicebusauditqueue) would need to be set to `my-hierarchy/audit`.
 - The [monitoring queue name](/servicecontrol/monitoring-instances/configuration.md#transport-monitoringerrorqueue) would need to be set to `my-hierarchy/monitoring`.
 
-
 ### Example connection string
 
+> [!NOTE]
+> when using the [Azure Service Bus emulator](https://learn.microsoft.com/en-us/azure/service-bus-messaging/overview-emulator), the default connection string points to `localhost` for the Endpoint value. That results in an error being looged during ServiceControl instances startup. The error can be safely ignored as it doesn't affect ServiceControl operations, instead it's related to the throughput and licensing portion of Servicecontrol, which will be anyway not available when using the Azure Service Bus emulator. It's possible to prevent the error to show up in the logs by creating an alias for `localhost` containing a `.`, such as for example, `servicebus-emulator.local` or by specifying an IP address, such as `127.0.0.1` for the Endpoint value.
```
mauroservienti:
```suggestion
> When using the [Azure Service Bus emulator](https://learn.microsoft.com/en-us/azure/service-bus-messaging/overview-emulator), the default connection string points to `localhost` for the Endpoint value. That results in an error being logged during the startup of ServiceControl instances. The error can be safely ignored as it doesn't affect ServiceControl operations, instead it's related to the throughput and licensing portion of Servicecontrol, which will be anyway not available when using the Azure Service Bus emulator. It's possible to prevent the error from showing up in the logs by creating an alias for `localhost` containing a `.`, such as for example, `servicebus-emulator.local` or by specifying an IP address, such as `127.0.0.1` for the Endpoint value.
```

mauroservienti:
> I did hack the ASB transport locally to share a connection for send and receive. Using that, I was able to get a completely local (but slightly cut down) Azure specific version of the AWS Loan Broker going using Service Platform, docker compose and the emulator, but I am not sure of the value of modifying the transport to extend the usefulness of the emulator slightly.
> 
> It's a shame that Microsoft have taken such a hard stance on the emulator - based on [this](https://github.com/Azure/azure-service-bus-emulator-installer/issues/58#issuecomment-2984760245) comment. It could be a really useful tool for local dev, prototyping, and integration tests if it wasn't so restricted. Local dev in docker containers is a bit broken when you have to have each developer on their own ASB namespace as well. Too bad if you wanted to work on a plane.

Thanks, @stevedowling. This looks like a note that should go on the transport page when we release 6.3.

cquirosj (approved):

PR #8206, opened 2026-04-24
Core-hosting with AddNServiceBusEndpoint

mattmercurio:


nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,102 @@
+---
+title: Recommended hosting model
+summary: Use AddNServiceBusEndpoint with Microsoft.Extensions.Hosting for the recommended single-endpoint and multi-endpoint hosting model.
+component: Core
```
DavidBoike, line 4:
```suggestion
component: Core
versions: '[10,)'
```

Snippets/Core/Core_10/Hosting/AddNServiceBusEndpoint.cs
```diff
@@ -0,0 +1,60 @@
...
+
+    void MultipleEndpoints(HostApplicationBuilder builder)
+    {
+        #region AddNServiceBusEndpointMulti
+
+        var salesConfig = new EndpointConfiguration("Sales");
+        var billingConfig = new EndpointConfiguration("Billing");
+
+        builder.Services.AddNServiceBusEndpoint(salesConfig, "Sales");
+        builder.Services.AddNServiceBusEndpoint(billingConfig, "Billing");
```
andreasohlund: Use the endpoint name to avoid duplicating the strings , eg`salesConfig.EndpointName`

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,76 @@
...
+
+NServiceBus also supports hosting multiple logical endpoints in one process. This is typically useful when a single-endpoint host is not the right fit. Common scenarios:
+
+- Multi-tenant systems where each tenant requires an isolated endpoint.
+- Partitioned throughput where each partition is an endpoint sharing a host.
+- Co-located infrastructure endpoints that do not justify a separate process.
+
+Compared to a single-endpoint host, each additional endpoint adds registration, startup overhead, and coordination within the shared process. Keep the configuration and endpoint identities explicit so the host remains easy to reason about.
+
+Each endpoint is registered with its own `EndpointConfiguration`. Pass an identifier string as the second argument to distinguish them in dependency injection:
```
andreasohlund:
```suggestion
Each endpoint is registered with its own `EndpointConfiguration`. EndpointName is used to distinguish them in dependency injection:
```

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,76 @@
...
+
+snippet: AddNServiceBusEndpointMulti
+
+The endpoint name is the recommended identifier. A distinct identifier is only required when the same endpoint definition is hosted more than once with different per-instance configuration — for example, a per-tenant deployment where the endpoint name is composed at runtime.
+
+### Endpoint-scoped dependencies
```
andreasohlund, line 56: Do we need to somehow mention "global" dependencies as well?

danielmarbach: @andreasohlund It is

> Services that do not vary per endpoint are registered normally on `IServiceCollection`

The question is: Do you want it to be more prominent?

nservicebus/hosting/core-hosting.md
danielmarbach: I actually prefer the current wording because from the perspective of the API it is correct how it is worded right now

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,76 @@
...
+
+When a shared service needs its own per-endpoint instance, register it as a [keyed service](https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection#keyed-services) using the endpoint name as the key:
```
danielmarbach:
```suggestion
When a shared service needs its own per-endpoint instance, register it as a [keyed service](https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection#keyed-services) using the previously chosen (e.g. endpoint name) identifier as a key:
```

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,76 @@
...
+
+snippet: AddNServiceBusEndpointKeyedServices
+
+Each endpoint resolves its own instance from the keyed slot. Services that do not vary per endpoint are registered normally on `IServiceCollection` and every endpoint resolves the same instance.
```
danielmarbach:
```suggestion
Each endpoint resolves its own instance as keyed services using the previously chosen identifier as a key. Services that do not vary per endpoint are registered normally (non-keyed) on `IServiceCollection` and every endpoint resolves the same instance.
```

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,76 @@
...
+
+For the migration path from `EndpointConfiguration.RegisterComponents`, see the [NServiceBus 10 to 11 upgrade guide](/nservicebus/upgrades/10to11/).
```
danielmarbach:
```suggestion
It is still possible to use `EndpointConfiguration.RegisterComponents` and the API will internally automatically use the right approach regardless whether a single endpoint or multiple endpoints are used. The API is obsoleted with a warning and it is recommended to migrate to explicit registrations shown above. For more migration guidance see the [NServiceBus 10 to 11 upgrade guide](/nservicebus/upgrades/10to11/).
```

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,76 @@
...
+## Endpoint identity and the DI identifier
+
+- **Endpoint name** — the NServiceBus identity set on `EndpointConfiguration`. Drives the input queue, routing, diagnostics, and log context. Unique across the system.
+- **Endpoint identifier** — the dependency-injection slot used to resolve this endpoint's `IMessageSession` and keyed dependencies. Scoped to a single process. Defaults to the endpoint name.
+
+In a single-endpoint host, the two are the same value and the distinction does not matter. In a multi-endpoint host, each endpoint retains its own input queue, `IMessageSession`, and per-endpoint configuration.
+
+## Logging
+
+Core uses [`Microsoft.Extensions.Logging`](https://learn.microsoft.com/en-us/dotnet/core/extensions/logging/) as its built-in logging infrastructure. Log events flow through the host's configured `ILoggerFactory` with endpoint context attached. The `NServiceBus.Extensions.Logging` package is no longer required.
```
danielmarbach:
```suggestion
NServiceBus uses [`Microsoft.Extensions.Logging`](https://learn.microsoft.com/en-us/dotnet/core/extensions/logging/) as its built-in logging infrastructure. Log events flow through the host's configured `ILoggerFactory` with endpoint context attached. The `NServiceBus.Extensions.Logging` package is no longer required.
```

Snippets/Core/Core_10/Hosting/AddNServiceBusEndpoint.cs
danielmarbach: Would it be helpful to add a more "complex scenario" for example hosting the same endpoint per tenant twice with different names to show the "deviation" pattern explained in the documentation?

danielmarbach:
This is already pretty good!

nservicebus/hosting/core-hosting.md
andreasohlund: ah, missed that, I think that is fine

Snippets/Core/Core_10/Hosting/AddNServiceBusEndpoint.cs
mattmercurio: applied both suggestions

nservicebus/hosting/core-hosting.md
mattmercurio: The new per-tenant snippet below this one passes a tenant (not EndpointName) as the identifier, so "Pass an identifier string" stays accurate

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,80 @@
+---
+title: Recommended hosting model
```
DavidBoike: I'm struggling with what we call this. This can only be the "recommended hosting model" as long as we don't replace it, then what would it be called? But "Microsoft Generic Host" is a dead term, right?

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,80 @@
...
+summary: Use AddNServiceBusEndpoint with Microsoft.Extensions.Hosting for the recommended single-endpoint and multi-endpoint hosting model.
+component: Core
+versions: '[10,)'
+reviewed: 2026-04-27
+related:
+  - nservicebus/upgrades/10to11
+  - samples/hosting/generic-host
+---
+
+Starting in NServiceBus version 10.2, the recommended way to host NServiceBus is through `Microsoft.Extensions.Hosting` by registering endpoints with `IServiceCollection` using `AddNServiceBusEndpoint`. This aligns endpoint startup, dependency injection, and logging with the standard .NET hosting model, and it supports both the common single-endpoint case and advanced scenarios that host multiple endpoints in one process.
```
DavidBoike: Wouldn't it be more accurate to say something like 10.2 takes a dependency on Microsoft.Extensions.Hosting and offers hosting capabilities without requiring any other NServiceBus packages?

BUT

Remember that 6 months after 10.2 is released, 10.1 is no longer supported, and this page is V10 and up only. We should be thinking about this documentation through the lens of what it will look like to someone building an endpoint after NServiceBus 11 is released, and less like an upgrade guide.

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Choosing an integration path
```
DavidBoike: Because of the previous comment, this should not be presented as "yet another choice you have to make" to use NServiceBus. We need to consider how people reading this when V11 is out and presenting it as "This is the way".

If we want to put in a partial that shows a note box about the old way in V10 only, that's possible though.

Snippets/Core/Core_10/Hosting/AddNServiceBusEndpoint.cs
```diff
@@ -0,0 +1,75 @@
...
+using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using NServiceBus;
+
+class AddNServiceBusEndpointHosting
+{
+    async Task SingleEndpoint()
+    {
+        #region AddNServiceBusEndpointSingle
```
danielmarbach, line 13: Would it be helpful for each scenario to show an example on how to resolve things explicitely? For example we could show something like this for a single endpoint

Using the provider directly

```
serviceProvider.GetRequiredService<IMessageSession>();
````

injected into a class

```
MyComponent(IMessageSession session)
````

and for the other samples accordingly

```
serviceProvider.GetRequiredKeyedService<IMessageSession>("ChosenIdentifier");
````

and

```
MyComponent([FromKeyedServices("ChosenIdentifier")] IMessageSession session)
````

and maybe even something like

```
// otherThing is global
MyComponent(IMyOtherThing otherThing, [FromKeyedServices("ChosenIdentifier")] IMessageSession session)
```

and so on until it becomes clear?

File: nservicebus/hosting/core-hosting.md [hunk 13]
```diff
@@ -0,0 +1,80 @@
...
+NServiceBus provides two integration paths with `Microsoft.Extensions.Hosting`:
+
+- **Built-in integration** (recommended): use `AddNServiceBusEndpoint` on `IServiceCollection`. Available starting in NServiceBus 10.2.
+- **Package-based integration** (legacy): the [`NServiceBus.Extensions.Hosting`](https://www.nuget.org/packages/NServiceBus.Extensions.Hosting) package and its `UseNServiceBus` method are no longer the recommended path. Existing applications should plan to migrate; new applications should use the built-in integration.
+
+For details on the `UseNServiceBus` approach, see [NServiceBus.Extensions.Hosting](/nservicebus/hosting/extensions-hosting.md).
+
+## Hosting a single endpoint
+
+Register the endpoint on the host's service collection. For new development, this is the recommended alternative to the `UseNServiceBus` pattern in `NServiceBus.Extensions.Hosting`.
```
DavidBoike: Don't compare to UseNServiceBus just tell how to do it

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,80 @@
...
+
+snippet: AddNServiceBusEndpointSingle
+
+The endpoint starts and stops with the host's lifecycle. `IMessageSession` is available in dependency injection for components that need to send or publish from outside a message handler.
```
DavidBoike: Could we use slightly more precise language about what happens with the IMessageSession?

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,80 @@
...
+
+For most applications, a single endpoint per process is the simplest place to start.
+
+## Hosting multiple endpoints
+
+NServiceBus also supports hosting multiple logical endpoints in one process. This is typically useful when a single-endpoint host is not the right fit. Common scenarios:
```
DavidBoike: After saying "It supports more than one" it's kind of silly to say "this is useful for when 1 is bad". Kind of sounds like the commercials that say not to take the drug if you're allergic to the drug.

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,80 @@
...
+
+- Multi-tenant systems where each tenant requires an isolated endpoint.
+- Partitioned throughput where each partition is an endpoint sharing a host.
+- Co-located infrastructure endpoints that do not justify a separate process.
```
DavidBoike: Would it make sense to add something hand-wavy about modular monoliths?

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,80 @@
...
+
+Compared to a single-endpoint host, each additional endpoint adds registration, startup overhead, and coordination within the shared process. Keep the configuration and endpoint identities explicit so the host remains easy to reason about.
```
DavidBoike: > Keep the configuration and endpoint identities explicit so the host remains easy to reason about.

I think a reader will ask "What does that mean?"

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,80 @@
...
+
+Each endpoint is registered with its own `EndpointConfiguration`. Pass an identifier string as the second argument to distinguish them in dependency injection:
```
DavidBoike: I think this part should lead the section, with all the "scenarios" and stuff coming after.

I would also prefer if we were explicit about the endpointIdentifier being used as a ServiceKey (or whatever DI-docs-recognizable noun) and also linking to MS docs about keyed services.

Snippets/Core/Core_10/Hosting/AddNServiceBusEndpoint.cs
```diff
@@ -0,0 +1,75 @@
...
+        #endregion
+    }
+
+    void PerTenantEndpoints(HostApplicationBuilder builder, string[] tenants)
+    {
+        #region AddNServiceBusEndpointPerTenant
+
+        foreach (var tenant in tenants)
+        {
+            var endpointConfig = new EndpointConfiguration($"Sales-{tenant}");
```
DavidBoike: Isn't this a weird example though, because the endpoint name would then already be different? Shouldn't all the Sales endpoints have the same endpoint name, but different queue names? Could we show that?

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,80 @@
...
+
+The endpoint name is the recommended identifier. A distinct identifier is only required when the same endpoint definition is hosted more than once with different per-instance configuration — for example, a per-tenant deployment where the endpoint name is composed at runtime:
+
+snippet: AddNServiceBusEndpointPerTenant
+
+In this case the endpoint name distinguishes each runtime instance for routing, while the tenant key serves as the DI identifier so callers can resolve a specific tenant's `IMessageSession` and keyed services.
+
+### Endpoint-scoped dependencies
+
+When a shared service needs its own per-endpoint instance, register it as a [keyed service](https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection#keyed-services) using the previously chosen (e.g. endpoint name) identifier as a key:
```
DavidBoike: Is that wording awkward at the beginning? Would "When each endpoint requires a different implementation of a shared service" be clearer?

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,80 @@
...
+
+snippet: AddNServiceBusEndpointKeyedServices
+
+Each endpoint resolves its own instance as keyed services using the previously chosen identifier as a key. Services that do not vary per endpoint are registered normally (non-keyed) on `IServiceCollection` and every endpoint resolves the same instance.
```
DavidBoike:
```suggestion
Each endpoint resolves its own `DatabaseService` instance as keyed services using the previously chosen identifier as a key. Services that do not vary per endpoint are registered normally (non-keyed) on `IServiceCollection` and every endpoint resolves the same instance.
```

DavidBoike: But also, "resolves the same instance" isn't correct unless it's a singleton, right? You could still have scoped or transient service descriptors. How do we make that more accurate?

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,80 @@
...
+
+It is still possible to use `EndpointConfiguration.RegisterComponents` and the API will internally automatically use the right approach regardless whether a single endpoint or multiple endpoints are used. The API is obsoleted with a warning and it is recommended to migrate to explicit registrations shown above. For more migration guidance see the [NServiceBus 10 to 11 upgrade guide](/nservicebus/upgrades/10to11/).
```
DavidBoike: This is upgrade guide material.
```suggestion
```

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,80 @@
...
+
+## Endpoint identity and the DI identifier
```
DavidBoike: I don't like abbreviating DI in the title, and I really don't like just dropping into a bullet list. Maybe the heading should be "Understanding endpoint identity" and there should be an intro sentence about how it's important to understand the difference between the endpoint name and the identifier used to group/correlate/specify services for endpoints.

And does queue name deserve a mention?

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,80 @@
...
+
+- **Endpoint name** — the NServiceBus identity set on `EndpointConfiguration`. Drives the input queue, routing, diagnostics, and log context. Unique across the system.
```
DavidBoike: > Unique across the system

I don't think that's true?

Also, either we can write complete sentences, or if we're really trying to be terse, we can put all of this in a comparison table.

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,80 @@
...
+- **Endpoint identifier** — the dependency-injection slot used to resolve this endpoint's `IMessageSession` and keyed dependencies. Scoped to a single process. Defaults to the endpoint name.
+
+In a single-endpoint host, the two are the same value and the distinction does not matter. In a multi-endpoint host, each endpoint retains its own input queue, `IMessageSession`, and per-endpoint configuration.
+
+## Logging
```
DavidBoike, line 91: I could see a sample of how log output looks with multiple endpoints, but most of the stuff below this line is upgrade guide material.

nservicebus/hosting/critical-errors.md
```diff
@@ -27,7 +27,7 @@ Critical errors can sometimes be persistent. When a critical error persists, it
 
 ### Stop the endpoint
 
-[Microsoft Generic Host's](/nservicebus/hosting/extensions-hosting.md) `IHostApplicationLifetime.Stop` method stops the NServiceBus endpoint gracefully.
+When hosted with [`Microsoft.Extensions.Hosting`](https://learn.microsoft.com/en-us/dotnet/core/extensions/generic-host), calling `IHostApplicationLifetime.StopApplication()` stops the host gracefully, which in turn stops the NServiceBus endpoint.
```
DavidBoike: I don't know if "when hosting with package I never added to the project myself" is going to land right?

nservicebus/hosting/index.md
```diff
@@ -14,9 +14,12 @@ NServiceBus is a library at its core so that it can be hosted in any .NET proces
...
 There are several approaches to hosting.
 
-## Microsoft Generic Host
+## Microsoft.Extensions.Hosting
 
-The [Microsoft Generic Host](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/generic-host) is the most common way to host NServiceBus on .NET Core. NServiceBus can be integrated with the generic host using the [NServiceBus.Extensions.Hosting package](/nservicebus/hosting/extensions-hosting.md).
+The [.NET Generic Host](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/generic-host) is the most common way to host NServiceBus in modern .NET applications. There are two integration paths:
+
+- [Recommended hosting model](/nservicebus/hosting/core-hosting.md) using the built-in `AddNServiceBusEndpoint` integration.
+- [NServiceBus.Extensions.Hosting](/nservicebus/hosting/extensions-hosting.md) for the existing package-based `UseNServiceBus` integration.
```
DavidBoike: This needs to be turned into partials, because in V9 and below there is only option and in V10 and above (discounting 10.0 and 10.1 which ultimately go away) there is only one (albeit different) choice.

nservicebus/hosting/web-application.md
```diff
@@ -17,9 +17,9 @@ As most web technologies operate in a scale-out manner, NServiceBus can be hoste
 
 ## Dependency injection integration
 
-### Using the Generic Host
+### Using Microsoft.Extensions.Hosting
 
-NServiceBus can be integrated into any web host that supports the [Microsoft Generic Host](https://docs.microsoft.com/en-us/dotnet/core/extensions/generic-host) using the [NServiceBus Generic Host integration](/nservicebus/hosting/extensions-hosting.md). The Generic Host integration automatically registers `IMessageSession` with the dependency injection container.
+NServiceBus integrates with any web host that supports the [.NET Generic Host](https://docs.microsoft.com/en-us/dotnet/core/extensions/generic-host). For new development, use the [recommended hosting model](/nservicebus/hosting/core-hosting.md). Existing applications can continue with [NServiceBus.Extensions.Hosting](/nservicebus/hosting/extensions-hosting.md). In both paths, `IMessageSession` is available through dependency injection.
```
DavidBoike: Same here about partials.

Snippets/Core/Core_10/Hosting/AddNServiceBusEndpoint.cs
danielmarbach: I was thinking about this as well when I looked at it but then went down mentally the path of thinking, "But then we need OverrideLocalAddress and potentially other things" and concluded this gets the point across enough to show in this sample the tenant identifier is the endpoint identifier and the endpoint name is different. But maybe I'm already too much in the weeds with my thought process because I assume it is clear

nservicebus/hosting/core-hosting.md
danielmarbach: Let's avoid going down the route of scopes and try to find a language that describes "endpoint-specific services" vs "services shared between endpoints" which we sometimes call "global". But maybe we need a crisper distinction.

nservicebus/hosting/core-hosting.md
danielmarbach: Is it though? When I added this suggestion, I came from the angle of "this is deprecated with warning for v10" which means you can, for the lifetime of v10 decide to ignore this or use it "because it is more convenient or you already had this in place". So I saw this section as "Hey, look, you can, but you should really consider the upgrade guide over there"

I can see both approaches having their pro and cons and some of your comments indicate some form of principle we should always follow when writing major version documentation and that major version is embarking in a different direction. Maybe we can extract these principles and populate them into the docs review skill?

nservicebus/hosting/core-hosting.md
mattmercurio: @DavidBoike Yea, I've gone back and forth on this title a few times. What about "Hosting with Microsoft.Extensions.Hosting" ? Technically right but not "hey you should favor this"

Snippets/Core/Core_10/Hosting/AddNServiceBusEndpoint.cs
mattmercurio: Yes, good stuff. I weaved that in: image 
image

nservicebus/hosting/core-hosting.md
mattmercurio: For the menu, I think this fits:
image

The page title as "Hosting with Microsoft.Extensions.Hosting" 
image

nservicebus/hosting/core-hosting.md
mattmercurio: I reframed this and dropped the legacy note

nservicebus/hosting/core-hosting.md
mattmercurio: Line now reads simply "Register the endpoint on the host's service collection:"

nservicebus/hosting/core-hosting.md
mattmercurio: tightened up with snippets

nservicebus/hosting/core-hosting.md
mattmercurio: thoughts? image

nservicebus/hosting/critical-errors.md
mattmercurio: This is reverted and back to original wording

nservicebus/hosting/index.md
mattmercurio: Turned into partials

nservicebus/hosting/core-hosting.md
mattmercurio: I cleaned up upgrade guide material in general. I did not show a log output sample as you suggested yet.

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,88 @@
...
+summary: Host NServiceBus endpoints with Microsoft.Extensions.Hosting using AddNServiceBusEndpoint, for both single-endpoint and multi-endpoint scenarios.
+component: Core
+versions: '[10,)'
+reviewed: 2026-04-27
+related:
+  - nservicebus/upgrades/10to11
+  - samples/hosting/generic-host
+---
+
+NServiceBus endpoints are hosted with [`Microsoft.Extensions.Hosting`](https://learn.microsoft.com/en-us/dotnet/core/extensions/generic-host) by registering them on `IServiceCollection` using `AddNServiceBusEndpoint`. Endpoint startup, dependency injection, and logging align with the standard .NET hosting model, and the same registration approach supports both single-endpoint and multi-endpoint hosts.
```
DavidBoike:
```suggestion
NServiceBus endpoints are hosted with [`Microsoft.Extensions.Hosting`](https://learn.microsoft.com/en-us/dotnet/core/extensions/generic-host) by registering them on `IServiceCollection` using the `AddNServiceBusEndpoint` method. Endpoint startup, dependency injection, and logging align with the standard .NET hosting model, and the same registration approach supports both single-endpoint and multi-endpoint hosts.
```

Snippets/Core/Core_10/Hosting/AddNServiceBusEndpoint.cs
```diff
@@ -0,0 +1,134 @@
...
+
+        var builder = Host.CreateApplicationBuilder();
+
+        var endpointConfiguration = new EndpointConfiguration("Sales");
+        // configure transport, persistence, etc.
+
+        builder.Services.AddNServiceBusEndpoint(endpointConfiguration);
+
+        var host = builder.Build();
+        await host.RunAsync();
```
DavidBoike: Since this is the first, simplest option shown, I was thinking keep it simple as possible:
```suggestion
        await builder.Build().RunAsync();
```

Snippets/Core/Core_10/Hosting/AddNServiceBusEndpoint.cs
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    void ResolveFromProvider(IServiceProvider serviceProvider)
+    {
+        #region AddNServiceBusEndpointGetSession
+
+        var session = serviceProvider.GetRequiredService<IMessageSession>();
+
+        #endregion
+    }
```
DavidBoike: I think I'd like to see this in context? 
```suggestion
    void ResolveFromProvider()
    {
        var builder = Host.CreateApplicationBuilder();

        #region AddNServiceBusEndpointGetSession

        var host = builder.Build();
        var messageSession = host.Services.GetRequiredService<IMessageSession>();

        #endregion
    }
```

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,88 @@
...
+NServiceBus also supports hosting multiple logical endpoints in one process. Common scenarios:
+
+- Multi-tenant systems where each tenant requires an isolated endpoint.
+- Modular monoliths where each module owns its own endpoint within a shared host.
+- Partitioned throughput where each partition is an endpoint sharing a host.
+- Co-located infrastructure endpoints that do not justify a separate process.
+
+Compared to a single-endpoint host, each additional endpoint adds registration, startup overhead, and coordination within the shared process.
+
+Each endpoint is registered with its own `EndpointConfiguration`. The second argument to `AddNServiceBusEndpoint` is the DI identifier — a [service key](https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection#keyed-services) used to distinguish that endpoint's `IMessageSession` and per-endpoint keyed services:
```
DavidBoike: Still not sure I like inventing a term "DI identifier" that is completely absent in the source code.

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,88 @@
...
+
+snippet: AddNServiceBusEndpointMulti
+
+A distinct DI identifier is only needed when the same endpoint is registered more than once in a process — for example, a per-tenant deployment where each tenant has its own input queue:
```
DavidBoike:
```suggestion
Most of the time, the endpoint name is the ideal identifier to separate endpoint-specific services in the service collection. A distinct DI identifier is only needed when the same endpoint is registered more than once in a process — for example, a per-tenant deployment where each tenant has the same point but with its own input queue:
```

Snippets/Core/Core_10/Hosting/AddNServiceBusEndpoint.cs
```diff
@@ -0,0 +1,134 @@
...
+    {
+        #region AddNServiceBusEndpointPerTenant
+
+        foreach (var tenant in tenants)
+        {
+            var endpointConfig = new EndpointConfiguration("Sales");
+            endpointConfig.OverrideLocalAddress($"Sales-{tenant}");
+            // additional per-tenant configuration
+
+            builder.Services.AddNServiceBusEndpoint(endpointConfig, tenant);
```
DavidBoike: If you hosted per-tenant Billing services in the same host as well this would not work unless:
```suggestion
            builder.Services.AddNServiceBusEndpoint(endpointConfig, "Sales-{tenant});
```

Snippets/Core/Core_10/Hosting/AddNServiceBusEndpoint.cs
```diff
@@ -0,0 +1,134 @@
...
+{
+    public Task Submit(Order order) => session.Send(order);
+}
+
+#endregion
+
+#region AddNServiceBusEndpointInjectMixed
+
+// MyService is a global (non-keyed) service; IMessageSession is keyed for the "Sales" endpoint.
+class SalesOrderRouter(MyService service, [FromKeyedServices("Sales")] IMessageSession session)
```
DavidBoike: Would be easier here if you just renamed MyService to MyGlobalService

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,88 @@
...
+| Set via        | `new EndpointConfiguration(name)`                   | Second argument to `AddNServiceBusEndpoint(config, identifier)`            |
+| Required       | Always                                              | When more than one endpoint is registered on the same `IServiceCollection` |
+| Unique across  | Logical endpoints in the system                     | Endpoint registrations in the process                                      |
+
+When a DI identifier is needed, the endpoint name is the recommended value. A different value is only needed when the same endpoint definition is hosted more than once in one process — for example, a per-tenant deployment. See [Hosting multiple endpoints](#hosting-multiple-endpoints) for the registration patterns.
+
+## Logging
+
+NServiceBus uses [`Microsoft.Extensions.Logging`](https://learn.microsoft.com/en-us/dotnet/core/extensions/logging/) as its built-in logging infrastructure. Log events flow through the host's configured `ILoggerFactory` with endpoint context attached. The `NServiceBus.Extensions.Logging` package is no longer required.
```
DavidBoike:
```suggestion
NServiceBus uses [`Microsoft.Extensions.Logging`](https://learn.microsoft.com/en-us/dotnet/core/extensions/logging/) as its built-in logging infrastructure. Log events flow through the host's configured `ILoggerFactory` with endpoint context attached.
```

nservicebus/hosting/extensions-hosting.md
```diff
@@ -1,13 +1,16 @@
...
 component: Extensions.Hosting
-reviewed: 2025-08-19
+reviewed: 2026-04-27
 related:
  - samples/hosting/generic-host
  - samples/dependency-injection/aspnetcore
 ---
 
+> [!NOTE]
+> For new development, prefer [hosting with Microsoft.Extensions.Hosting](/nservicebus/hosting/core-hosting.md) using the built-in `AddNServiceBusEndpoint` integration on `IServiceCollection`. This page describes the existing package-based `NServiceBus.Extensions.Hosting` and `UseNServiceBus` integration path.
```
DavidBoike: Thinking add "which is deprecated starting in NServiceBus version 10.2.0" and also a link to the upgrade guide.

nservicebus/hosting/index.md
```diff
@@ -14,9 +14,7 @@ NServiceBus is a library at its core so that it can be hosted in any .NET proces
 
 There are several approaches to hosting.
 
-## Microsoft Generic Host
-
-The [Microsoft Generic Host](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/host/generic-host) is the most common way to host NServiceBus on .NET Core. NServiceBus can be integrated with the generic host using the [NServiceBus.Extensions.Hosting package](/nservicebus/hosting/extensions-hosting.md).
+partial: generic-host
 
 ## Self-hosting
```
DavidBoike, line 19: When I preview this, I still get Endpoint.Create() snippets. Andreas marked those as ignore the obsolete, so we need to track down where else they're being used to see what we need to do here.

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,88 @@
...
+
+The endpoint starts and stops with the host's lifecycle. `IMessageSession` is registered on the host's `IServiceCollection` and can be resolved from the service provider:
+
+snippet: AddNServiceBusEndpointGetSession
+
+Or injected into controllers, background services, or any other component that needs to send or publish from outside a message handler or other NServiceBus extension point:
+
+snippet: AddNServiceBusEndpointInjectSession
+
+For most applications, a single endpoint per process is the simplest place to start.
```
danielmarbach:
```suggestion
For most applications, a single endpoint per process should be the default.
```

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,88 @@
...
+## Hosting multiple endpoints
+
+NServiceBus also supports hosting multiple logical endpoints in one process. Common scenarios:
+
+- Multi-tenant systems where each tenant requires an isolated endpoint.
+- Modular monoliths where each module owns its own endpoint within a shared host.
+- Partitioned throughput where each partition is an endpoint sharing a host.
+- Co-located infrastructure endpoints that do not justify a separate process.
+
+Compared to a single-endpoint host, each additional endpoint adds registration, startup overhead, and coordination within the shared process.
```
danielmarbach:
```suggestion
Compared to a single-endpoint host, each additional endpoint adds registration, startup, and coordination overhead within the shared process.
```

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,88 @@
...
+
+Each endpoint is registered with its own `EndpointConfiguration`. The second argument to `AddNServiceBusEndpoint` is the endpoint identifier — a [service key](https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection#keyed-services) used to distinguish that endpoint's `IMessageSession` and per-endpoint keyed services:
+
+snippet: AddNServiceBusEndpointMulti
+
+Most of the time, the endpoint name is the ideal identifier to separate endpoint-specific services in the service collection. A distinct endpoint identifier is only needed when the same endpoint is registered more than once in a process — for example, a per-tenant deployment where each tenant has the same point but with its own input queue:
```
danielmarbach, line 50:
```suggestion
Most of the time, the endpoint name is the ideal identifier to separate endpoint-specific services in the service collection. A distinct endpoint identifier is only needed when the same endpoint is registered more than once in a process — for example, a per-tenant deployment where each tenant has the same endpoint name but with its own input queue:
```

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,88 @@
...
+
+snippet: AddNServiceBusEndpointPerTenant
+
+All tenants share the `Sales` endpoint name; each gets its own [input queue](/nservicebus/endpoints/specify-endpoint-name.md#input-queue). A tenant-scoped identifier (`Sales-<tenant>`) lets callers resolve a specific tenant's `IMessageSession` and keyed services. Including the endpoint name in the identifier keeps it unique when other endpoints (for example, `Billing`) are also hosted per tenant in the same process.
```
danielmarbach, line 54:
```suggestion
All tenants share the `Sales` endpoint name; each gets its own [input queue](/nservicebus/endpoints/specify-endpoint-name.md#input-queue). A tenant-scoped identifier (`Sales-<tenant>`) lets callers resolve a specific tenant's `IMessageSession` and [keyed services](https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection#keyed-services). Including the endpoint name in the identifier keeps it unique when other endpoints (for example, `Billing`) are also hosted per tenant in the same process.
```

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,88 @@
...
+
+### Endpoint-scoped dependencies
+
+When each endpoint requires a different implementation of a shared service, register it as a [keyed service](https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection#keyed-services) using the chosen endpoint identifier (typically the endpoint name) as the key:
+
+snippet: AddNServiceBusEndpointKeyedServices
+
+Each endpoint resolves its own `DatabaseService` instance as keyed services using the previously chosen identifier as a key. Services that do not vary per endpoint are registered normally (non-keyed) on `IServiceCollection` and every endpoint resolves the same instance.
```
danielmarbach:
```suggestion
In the snippet above each endpoint resolves its own `DatabaseService` instance as keyed services using the previously chosen identifier as a key. Services that do not vary per endpoint are registered normally (non-keyed) on `IServiceCollection` and every endpoint resolves the same instance.
```

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,88 @@
...
+
+For most applications, a single endpoint per process is the simplest place to start.
+
+## Hosting multiple endpoints
+
+NServiceBus also supports hosting multiple logical endpoints in one process. Common scenarios:
+
+- Multi-tenant systems where each tenant requires an isolated endpoint.
+- Modular monoliths where each module owns its own endpoint within a shared host.
+- Partitioned throughput where each partition is an endpoint sharing a host.
```
danielmarbach:
```suggestion
- Partitioned throughput where each partition is an endpoint sharing a host.
- Competing consumers co-located in the same process that require a shared in-memory synchronization primitive
```

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,88 @@
...
+- Co-located infrastructure endpoints that do not justify a separate process.
```
danielmarbach:
```suggestion
- Co-located infrastructure endpoints that do not justify a separate process (for example to save the memory overhead of multiple processes to achieve a more dense hosting).
```

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,88 @@
...
+
+Inside NServiceBus extension points — message handlers, sagas, features, and installers — per-endpoint services resolve automatically from the right scope. `[FromKeyedServices]` is not needed in those contexts.
+
+Outside extension points (for example, controllers, background services, or any host-level component that needs to send or publish), use `GetRequiredKeyedService<T>(identifier)` from the service provider or `[FromKeyedServices(identifier)]` for constructor injection:
+
+snippet: AddNServiceBusEndpointGetKeyedSession
+
+snippet: AddNServiceBusEndpointInjectKeyedSession
+
+Global (non-keyed) services and per-endpoint keyed services can be mixed in the same constructor:
```
danielmarbach:
```suggestion
Non-keyed (global) services and per-endpoint keyed services can be mixed in the same constructor:
```

Snippets/Core/Core_10/Hosting/AddNServiceBusEndpoint.cs
```diff
@@ -0,0 +1,143 @@
...
+class SalesOrderService([FromKeyedServices("Sales")] IMessageSession session)
+{
+    public Task Submit(Order order) => session.Send(order);
+}
+
+#endregion
+
+#region AddNServiceBusEndpointInjectMixed
+
+// MyGlobalService is a global (non-keyed) service; IMessageSession is keyed for the "Sales" endpoint.
```
danielmarbach:
```suggestion
// MyGlobalService is a non-keyed (global) service; IMessageSession is keyed for the "Sales" endpoint.
```

Snippets/Core/Core_10/Hosting/AddNServiceBusEndpoint.cs
```diff
@@ -0,0 +1,143 @@
...
+        #region AddNServiceBusEndpointMulti
+
+        var salesConfig = new EndpointConfiguration("Sales");
+        var billingConfig = new EndpointConfiguration("Billing");
+
+        builder.Services.AddNServiceBusEndpoint(salesConfig, salesConfig.EndpointName);
+        builder.Services.AddNServiceBusEndpoint(billingConfig, billingConfig.EndpointName);
+
+        #endregion
+    }
```
danielmarbach, line 38:
```suggestion
    }
    
    void MultipleEndpointsLean(HostApplicationBuilder builder)
    {
        #region AddNServiceBusEndpointMultiLean

        builder.Services.AddSalesEndpoint();
        builder.Services.AddBillingEndpoint();

        #endregion
    }```

File: nservicebus/hosting/core-hosting.md [hunk 49]
```diff
@@ -0,0 +1,88 @@
...
+- Multi-tenant systems where each tenant requires an isolated endpoint.
+- Modular monoliths where each module owns its own endpoint within a shared host.
+- Partitioned throughput where each partition is an endpoint sharing a host.
+- Co-located infrastructure endpoints that do not justify a separate process.
+
+Compared to a single-endpoint host, each additional endpoint adds registration, startup overhead, and coordination within the shared process.
+
+Each endpoint is registered with its own `EndpointConfiguration`. The second argument to `AddNServiceBusEndpoint` is the endpoint identifier — a [service key](https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection#keyed-services) used to distinguish that endpoint's `IMessageSession` and per-endpoint keyed services:
+
+snippet: AddNServiceBusEndpointMulti
```
danielmarbach, line 44:
```suggestion
snippet: AddNServiceBusEndpointMulti

It is recommended to encapsulate endpoint specific configuration including the call to `builder.AddNServiceBusEndpoint`inside an extension method to keep the composition root lean and simple to understand.

snippet: AddNServiceBusEndpointMultiLean
```

Snippets/Core/Core_10/Hosting/AddNServiceBusEndpoint.cs
danielmarbach: @mattmercurio You probably need to create those fake extension methods somewhere outside the snippet region to make the snippet compile

danielmarbach (approved):

Snippets/Core/Core_10/Hosting/AddNServiceBusEndpoint.cs
danielmarbach: This one not worth taking?

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,93 @@
...
+
+## Hosting multiple endpoints
+
+NServiceBus also supports hosting multiple logical endpoints in one process. Common scenarios:
+
+- Multi-tenant systems where each tenant requires an isolated endpoint.
+- Modular monoliths where each module owns its own endpoint within a shared host.
+- Partitioned throughput where each partition is an endpoint sharing a host.
+- Competing consumers co-located in the same process that require a shared in-memory synchronization primitive
+- Co-located infrastructure endpoints that do not justify a separate process (for example to save the memory overhead of multiple processes to achieve a more dense hosting).
```
DavidBoike: Aligning periods since none of these are complete sentences
```suggestion
- Multi-tenant systems where each tenant requires an isolated endpoint
- Modular monoliths where each module owns its own endpoint within a shared host
- Partitioned throughput where each partition is an endpoint sharing a host
- Competing consumers co-located in the same process that require a shared in-memory synchronization primitive
- Co-located infrastructure endpoints that do not justify a separate process (for example to save the memory overhead of multiple processes to achieve a more dense hosting)
```

nservicebus/hosting/core-hosting.md
```diff
@@ -0,0 +1,93 @@
...
+
+Compared to a single-endpoint host, each additional endpoint adds registration, startup, and coordination overhead within the shared process.
+
+Each endpoint is registered with its own `EndpointConfiguration`. The second argument to `AddNServiceBusEndpoint` is the endpoint identifier — a [service key](https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection#keyed-services) used to distinguish that endpoint's `IMessageSession` and per-endpoint keyed services:
```
DavidBoike:
```suggestion
Each endpoint is registered with its own `EndpointConfiguration`. The second argument to `AddNServiceBusEndpoint` is the endpoint identifier — a [service key](https://learn.microsoft.com/en-us/dotnet/core/extensions/dependency-injection#keyed-services) used to distinguish that endpoint's `IMessageSession` and per-endpoint keyed services within the dependency injection container:
```

nservicebus/hosting/extensions-hosting.md
```diff
@@ -1,13 +1,16 @@
...
 component: Extensions.Hosting
-reviewed: 2025-08-19
+reviewed: 2026-05-04
 related:
  - samples/hosting/generic-host
  - samples/dependency-injection/aspnetcore
 ---
 
+> [!NOTE]
+> For new development, prefer [hosting with Microsoft.Extensions.Hosting](/nservicebus/hosting/core-hosting.md) using the built-in `AddNServiceBusEndpoint` integration on `IServiceCollection`. The package-based `NServiceBus.Extensions.Hosting` and `UseNServiceBus` integration described on this page is deprecated starting with NServiceBus 10.2.0. See the [upgrade guide](/nservicebus/upgrades/10to11/) for migration details.
```
DavidBoike:
```suggestion
> For new development, prefer [hosting with Microsoft.Extensions.Hosting](/nservicebus/hosting/core-hosting.md) using the built-in `AddNServiceBusEndpoint` integration on `IServiceCollection`. The **NServiceBus.Extensions.Hosting** package's `UseNServiceBus` hosting method described on this page is deprecated starting with NServiceBus 10.2.0. See the [upgrade guide](/nservicebus/upgrades/10to11/) for migration details.
```

PR #8227, opened 2026-04-30
Add a -FullBuild option to samples and snippets build script

andreasohlund:
This updates tools/build-samples-and-snippets.ps1 to support explicit full builds and to run safely outside GitHub Actions.

  What changed

   - Added a -FullBuild switch to force building all .sln files found in the repo, regardless of branch or git diff state.
   - Kept existing default behavior intact: on non-master branches without -FullBuild, the script still builds only affected solutions.
   - Added a Write-GitHubStepSummary helper so failure details are only written when GITHUB_STEP_SUMMARY is available.
   - Prevented local PowerShell failures caused by attempting to write to $Env:GITHUB_STEP_SUMMARY when the script is run outside GitHub Actions.

### DIFF
--- a/tools/build-samples-and-snippets.ps1
+++ b/tools/build-samples-and-snippets.ps1
@@ -1,3 +1,8 @@
+[CmdletBinding()]
+param(
+    [switch] $FullBuild
+)
+
 # Assumes working directory is Docs repository root folder
 function CombinePaths()
 {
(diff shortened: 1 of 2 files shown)


tools/build-samples-and-snippets.ps1
```diff
@@ -104,10 +144,10 @@ foreach($solution in $solutions) {
         if( -not $? ) {
             $exitCode = 1
             Write-Output ("::error::Build failed: {0}" -f $solution.FullName)
-            Write-Output ("🔴 Build failed: {0}" -f $solution.FullName) | Out-File -FilePath $Env:GITHUB_STEP_SUMMARY -Encoding utf-8 -Append
-            Write-Output ('```txt') | Out-File -FilePath $Env:GITHUB_STEP_SUMMARY -Encoding utf-8 -Append
-            Write-Output $out | Out-File -FilePath $Env:GITHUB_STEP_SUMMARY -Encoding utf-8 -Append
-            Write-Output ('```') | Out-File -FilePath $Env:GITHUB_STEP_SUMMARY -Encoding utf-8 -Append
+            Write-GitHubStepSummary ("🔴 Build failed: {0}" -f $solution.FullName)
+            Write-GitHubStepSummary '```txt'
+            Write-GitHubStepSummary $out
```
DavidBoike: Is this even useful when running locally? Should there just be a condition that only does this block if $Env:GITHUB_STEP_SUMMARY does not exist?

andreasohlund: good catch, updated

DavidBoike (approved):

PR #8243, opened 2026-05-06
Update hosting index page

DavidBoike:


nservicebus/hosting/index.md
```diff
@@ -47,14 +37,16 @@ Related:
  * [Hosting endpoints in Docker Linux containers](/samples/hosting/docker/)
  * [Generic host](/samples/hosting/generic-host)
 
-### Send-only hosting
+## Send-only hosting
```
andreasohlund, line 40: More of a sidenote but send only being documented under hosting has always felt off, its more of a ”mode of operation” for an endpoint and not a hosting option. I wonder if we should move it?

danielmarbach (approved):

nservicebus/hosting/index.md
danielmarbach: Feels to me like a dedicated PR

andreasohlund (approved):

mattmercurio (approved):

PR #8265, opened 2026-05-20
Update aspire hosting samples

rbev:
* add new nuget to packages list
* simplify `samples/hosting/aspire` using Aspire hosting package
* new samples
  * `samples/aspire/platform`
  * `samples/aspire/platform-asb`
* new section under platform `platform/aspire`


components/nugetAlias.txt
```diff
@@ -7,6 +7,7 @@ ASQ: NServiceBus.Azure.Transports.WindowsAzureStorageQueues
 ASQN: NServiceBus.Transport.AzureStorageQueues
 ASP: NServiceBus.Persistence.AzureStorage
 ASTP: NServiceBus.Persistence.AzureTable
+Aspire: Particular.Aspire.Hosting.ServicePlatform
```
jpalac: When we have a client package down the track (which seems to be a valid assumption to make), should we name this so that's more explicit and allows for other packages? So perhaps:

```suggestion
AspirePlatformHost: Particular.Aspire.Hosting.ServicePlatform
```

platform/aspire/index.md
```diff
@@ -0,0 +1,557 @@
...
+| [IBM MQ](/transports/ibmmq/)                              | Not yet supported           |
+| [MSMQ](/transports/msmq/)                                 | Not available in containers |
+
+### Persistence
+
+| Persistence                          | Status    |
+| ------------------------------------ | --------- |
+| [RavenDB](/servicecontrol/ravendb/)  | Supported |
+
+### Throughput reporting
```
jpalac: I don't think we need this section as it's kind of part of the transport, and it's not a feature the user looks for - it's for our benefit only.

platform/aspire/index.md
```diff
@@ -0,0 +1,557 @@
...
+//     .WithThroughputReporting(new ThroughputReportingAzureServiceBus(
+//         tenantId: tenantId,
+//         subscriptionId: subscriptionId,
+//         clientId: clientId,
+//         clientSecret: clientSecret));
+```
+
+See [Usage reporting when using the Azure Service Bus transport](/servicecontrol/servicecontrol-instances/configuration.md#usage-reporting-when-using-the-azure-service-bus-transport) for what each value is used for in ServiceControl.
+
+#### Options
```
jpalac: Is it worth listing this here if these are the same as linked to above? Otherwise if those defaults\requirements change then we'll have to repeat the same here.

platform/aspire/index.md
```diff
@@ -0,0 +1,557 @@
...
+//
+// var error = platform.AddServiceControlErrorInstance("servicecontrol", persistence)
+//     .WithThroughputReporting(new ThroughputReportingRabbitMq(apiUrl, username, password));
+```
+
+See [Usage reporting when using the RabbitMQ transport](/servicecontrol/servicecontrol-instances/configuration.md#usage-reporting-when-using-the-rabbitmq-transport) for what each value is used for in ServiceControl.
+
+#### Options
+
+| Option | Default |
```
jpalac: Is it worth listing this here if these are the same as linked to above? Otherwise if those defaults\requirements change then we'll have to repeat the same here.

platform/aspire/index.md
```diff
@@ -0,0 +1,557 @@
...
+// TODO: replace with a real snippet from snippets/aspire
+// platform.AddServicePulse("servicepulse", error, monitoring)
+//     .WithEndpoint("servicepulse", e =>
+//     {
+//         e.Port = 9091;
+//         e.TargetPort = 9090;
+//     });
+```
+
+Refer to the following table for the default endpoint names and ports exposed by each component:
```
jpalac: Again, if we mention this info already elsewhere, should we just point to it instead of repeating it?

components/nugetAlias.txt
rbev: I'm all for this 
I think I made it `Aspire.Hosting`in my first pass

platform/aspire/index.md
```diff
@@ -0,0 +1,457 @@
...
+
+Aspire supports multiple deployment targets, including [Docker Compose](https://aspire.dev/deployment/docker-compose/), Kubernetes, and Azure Container Apps. See [Aspire pipelines](https://aspire.dev/deployment/pipelines/) for the deployment model and the supported targets.
+
+Two integration-specific points to keep in mind for any publish target:
+
+- The license parameter default is materialized into the manifest at publish time. Supply the license through Aspire parameter inputs in production so the value does not live in the published output. See [Supplying the license in production](#production-considerations-supplying-the-license-in-production).
+- The Learning transport is blocked in publish mode unless explicitly opted in. See [Learning transport in publish mode](#production-considerations-learning-transport-in-publish-mode).
```
rbev: This refers to a section that no longer exists
```suggestion
- The Learning transport is blocked in publish mode.
```

platform/aspire/index.md
warwickschroeder: the `endpoint name` info isn't documented elsewhere. I think you need this info for `WithEndpoint(endpointName, callback)`, and those names are `internal const` in the package

platform/aspire/index.md
warwickschroeder: I think its worth keeping. These document the API specifically. If these changed on the transport, we'd need to change it in the package API anyway. The table also gives a bit more info on the type of param for the API.

platform/aspire/index.md
warwickschroeder: It does still exist - https://github.com/Particular/docs.particular.net/pull/8265/changes#diff-13046e09c17257a6933217945f8862a99e21a6ed8259c3d9e08773a99a02146eR395

platform/aspire/index.md
warwickschroeder: In fact, maybe we need to make those consts public?

platform/aspire/index.md
```diff
@@ -0,0 +1,446 @@
...
+- **ServiceControl Error**: the error instance API
+- **ServiceControl Audit**: the audit instance API
+- **ServiceControl Monitoring**: the monitoring instance API
+- **Management Studio**: the RavenDB management UI (when using the managed RavenDB persistence instance)
+
+Externally supplied resources are not nested under the platform. An Azure Service Bus or RabbitMQ broker passed in via `AddConnectionString("...")` (or any other `Add*` call), and an existing RavenDB instance attached via `WithPersistenceRavenDb(existingRaven)`, appear as separate top-level resources in the dashboard. The platform does not own their lifecycle, so it does not group them under itself.
+
+The platform node tracks readiness as its children come up. It starts in `Starting`, transitions to `Running` once every child reports healthy, and moves to `RuntimeUnhealthy` if any child stops.
+
+<!-- TODO: add a screenshot of the running platform in the Aspire dashboard. -->
```
jpalac:
```suggestion
```

platform/aspire/index.md
```diff
@@ -0,0 +1,446 @@
...
+snippet: aspire-components-error
+
+#### Options
+
+| Option | Default |
+| --- | --- |
+| `name` (`string`) parameter on `AddServiceControlErrorInstance` | Required |
+| `persistence` (`IResourceBuilder<IResource>`) parameter on `AddServiceControlErrorInstance` | Required |
+| `queueName` (`string`) parameter on `WithErrorQueueName` | `error` |
+| `queueName` (`string`) parameter on `WithThroughputQueue` | Not set (throughput reporting off) |
```
jpalac: This is for receiving throughput from the monitoring instance. If no queue is specified, the default of "ServiceControl.ThroughputData" is used.
```suggestion
| `queueName` (`string`) parameter on `WithThroughputQueue` | `ServiceControl.ThroughputData` |
```

platform/aspire/index.md
```diff
@@ -0,0 +1,446 @@
...
+snippet: aspire-components-monitoring
+
+#### Options
+
+| Option | Default |
+| --- | --- |
+| `name` (`string`) parameter on `AddServiceControlMonitoringInstance` | Required |
+| `queueName` (`string`) parameter on `WithMonitoringQueueName` | `Particular.Monitoring` |
+| `errorInstance` (`IResourceBuilder<ServiceControlErrorInstanceResource>`) parameter on `WithThroughputQueueFrom` | Required (when called) |
+| `queueName` (`string`) parameter on `WithThroughputQueue` | Not set |
```
jpalac: Default value would be used if it's not set. 
```suggestion
| `queueName` (`string`) parameter on `WithThroughputQueue` | `ServiceControl.ThroughputData` |
```

platform/aspire/index.md
```diff
@@ -0,0 +1,446 @@
...
+#### Options
+
+| Option | Default |
+| --- | --- |
+| `tenantId` (`IExpressionValue`) constructor parameter | Required |
+| `subscriptionId` (`IExpressionValue`) constructor parameter | Required |
+| `clientId` (`IExpressionValue`) constructor parameter | Required |
+| `clientSecret` (`IExpressionValue`) constructor parameter | Required, should be a secret parameter |
+| `serviceBusName` (`IExpressionValue?`) constructor parameter | Inferred from the transport configuration when omitted |
+| `managementUrl` (`IExpressionValue?`) constructor parameter | Public Azure management URL |
```
jpalac: if omitted, the default is used based on some logic:https://github.com/Particular/ServiceControl/blob/master/src/ServiceControl.Transports.ASBS/AzureQuery.cs#L112

This is why I think repeating the logic here is not ideal :-)

platform/aspire/index.md
jpalac: It is documented. This is the default setting for SC error: https://docs.particular.net/servicecontrol/servicecontrol-instances/configuration#transport-servicebuserrorqueue

We have the same info for all the instances, eg audit queue: https://docs.particular.net/servicecontrol/audit-instances/configuration#transport-servicebusauditqueue

Not sure what the `http` is for Managed RavenDB persistence?

warwickschroeder: I dont think these are the same things. These aren't the default queue names, they are the HTTP endpoint names given to aspire instance resource. the ravens "PrimaryEndpointName" is "http", so the user would need to know this if they want to set their own host port. This isnt documented anywhere (other than here).

jpalac: Ah - that's confusing then - perhaps because of the use of "Endpoint name" which to me implied NSB endpoint.

warwickschroeder: You've worked for Particular too long ;-)

platform/aspire/index.md
rbev: oh, I thought I removed it - which we absolutely should do.

The learning transport is not supported in production and we should not be encouraging users to publish it into real infrastructure. The ability to disable that behaviour was only included so that the tests could verify it since they leverage the publish pipeline.

warwickschroeder: Yep, lets remove it

platform/aspire/index.md
warwickschroeder:
```suggestion
Refer to the following table for the default HTTP endpoint names and ports exposed by each component:
```

jpalac:
```suggestion
Refer to the following table for the Aspire instance resource names and ports exposed by each component:
```

platform/aspire/index.md
```diff
@@ -0,0 +1,444 @@
...
+
+Refer to the following table for the default endpoint names and ports exposed by each component:
+
+| Component                   | Endpoint name  | Internal Port  |
+| --------------------------- | -------------- | -------------- |
+| ServicePulse                | `servicepulse` | 9090           |
+| ServiceControl Error        | `error`        | 33333          |
+| ServiceControl Audit        | `audit`        | 44444          |
+| ServiceControl Monitoring   | `monitoring`   | 33633          |
+| Managed RavenDB persistence | `http`         | 8080           |
```
warwickschroeder: Remove

platform/aspire/index.md
warwickschroeder: Remove this line

platform/aspire/index.md
warwickschroeder:
```suggestion
```

platform/aspire/index.md
```diff
@@ -0,0 +1,444 @@
...
+
+Each managed container exposes an endpoint allowing Aspire to manage the exposed endpoint to avoid conflicts with ports already in use by other processes. As these ports are selected randomly by Aspire at startup you can refer to the dashboard for the url to access each component.
+
+In some cases it may be desirable to have a fixed port exposed by these components to provide a stable url (for example `http://localhost:9090` for ServicePulse) without having to inspect the Aspire dashboard.
+
+Override a port with the standard Aspire `WithEndpoint(endpointName, callback)` API:
+
+snippet: aspire-host-ports
+
+Refer to the following table for the default endpoint names and ports exposed by each component:
```
warwickschroeder:
```suggestion
```

platform/aspire/index.md
```diff
@@ -0,0 +1,436 @@
...
+- **ServiceControl Error**: the error instance API
+- **ServiceControl Audit**: the audit instance API
+- **ServiceControl Monitoring**: the monitoring instance API
+- **Management Studio**: the RavenDB management UI (when using the managed RavenDB persistence instance)
+
+Externally supplied resources are not nested under the platform. An Azure Service Bus or RabbitMQ broker passed in via `AddConnectionString("...")` (or any other `Add*` call), and an existing RavenDB instance attached via `WithPersistenceRavenDb(existingRaven)`, appear as separate top-level resources in the dashboard. The platform does not own their lifecycle, so it does not group them under itself.
+
+The platform node tracks readiness as its children come up. It starts in `Starting`, transitions to `Running` once every child reports healthy, and moves to `RuntimeUnhealthy` if any child stops.
+
+
```
warwickschroeder:
```suggestion
```

platform/aspire/index.md
```diff
@@ -0,0 +1,436 @@
...
+| Option | Default |
+| --- | --- |
+| `tenantId` (`IExpressionValue`) constructor parameter | Required |
+| `subscriptionId` (`IExpressionValue`) constructor parameter | Required |
+| `clientId` (`IExpressionValue`) constructor parameter | Required |
+| `clientSecret` (`IExpressionValue`) constructor parameter | Required, should be a secret parameter |
+| `serviceBusName` (`IExpressionValue?`) constructor parameter | Inferred from the transport configuration when omitted |
+| `managementUrl` (`IExpressionValue?`) constructor parameter | Public Azure management URL |
+
```
warwickschroeder:
```suggestion
```

platform/aspire/index.md
```diff
@@ -0,0 +1,436 @@
...
+See [Usage reporting when using the RabbitMQ transport](/servicecontrol/servicecontrol-instances/configuration.md#usage-reporting-when-using-the-rabbitmq-transport) for what each value is used for in ServiceControl.
+
+#### Options
+
+| Option | Default |
+| --- | --- |
+| `apiUrl` (`IExpressionValue?`) constructor parameter | Falls back to the RabbitMQ transport connection string |
+| `userName` (`IExpressionValue?`) constructor parameter | Falls back to the RabbitMQ transport connection string |
+| `password` (`IExpressionValue?`) constructor parameter | Falls back to the RabbitMQ transport connection string, should be a secret parameter when supplied |
+
```
warwickschroeder:
```suggestion
```

platform/aspire/index.md
```diff
@@ -0,0 +1,436 @@
...
+Each managed container exposes an endpoint allowing Aspire to manage the exposed endpoint to avoid conflicts with ports already in use by other processes. As these ports are selected randomly by Aspire at startup you can refer to the dashboard for the url to access each component.
+
+In some cases it may be desirable to have a fixed port exposed by these components to provide a stable url (for example `http://localhost:9090` for ServicePulse) without having to inspect the Aspire dashboard.
+
+Override a port with the standard Aspire `WithEndpoint(endpointName, callback)` API:
+
+snippet: aspire-host-ports
+
+
```
warwickschroeder:
```suggestion
```

Snippets/Aspire.Hosting/Aspire.Hosting_All/Production.cs
```diff
@@ -0,0 +1,44 @@
...
+        var monitoring = platform.AddServiceControlMonitoringInstance("monitoring");
+
+        #region aspire-host-ports
+
+        platform.AddServicePulse("servicepulse", error, monitoring)
+            .WithEndpoint("servicepulse", e =>
+            {
+                e.Port = 9091;
+                e.TargetPort = 9090;
+            });
```
warwickschroeder:
```suggestion
            .WithEndpoint("http", e =>
            {
                e.Port = 9091;
            });
```

samples/aspire/platform-sqs/sample.md
```diff
@@ -0,0 +1,86 @@
...
+
+`AddParticularPlatform` registers the Particular Platform as a resource named `particular`. The `WithTransportAmazonSqs` extension points the platform at the SQS configuration defined earlier, so that the ServiceControl instances connect to the same AWS region as the endpoints.
+
+`AddDefaultComponents` registers the remaining platform components using their default configuration — the ServiceControl audit and monitoring instances and ServicePulse. The error instance is added explicitly above so that usage reporting can be configured on it.
+
+snippet: platform-config
+
+#### Endpoints
+
+Each NServiceBus endpoint is added as an Aspire project and linked to the platform with `WithParticularPlatform`. This wires the endpoint to the platform's transport connection string. The `ClientUI` endpoint additionally uses `WaitFor(sales)` so that the `Sales` endpoint exists before it starts sending messages to it.
```
jpalac: There is no transport connection string. Perhaps mention the env variables that we set that allow the endpoints to use the same id, secret and region as the platform.

samples/aspire/platform-sqs/sample.md
```diff
@@ -0,0 +1,86 @@
...
+
+### AspireDemo.ServiceDefaults
+
+The [Aspire service defaults](https://aspire.dev/get-started/csharp-service-defaults/) project provides extension methods to configure application hosts and NServiceBus endpoints in a standardized way. This project is referenced by all of the NServiceBus endpoint projects.
+
+The OpenTelemetry configuration has been updated to include NServiceBus metrics and traces.
+
+snippet: add-nsb-otel
+
+Each endpoint project retrieves the connection string for the Azure ServiceBus broker and configures NServiceBus to use it as a transport:
```
jpalac: No connection  string and not ASB :-)

samples/aspire/platform-sqs/sample.md
```diff
@@ -0,0 +1,87 @@
...
+- Two projects, each of which is an NServiceBus endpoint. All of these projects reference the `platform` resource.
+  - `clientui`
+  - `sales`
+- ServiceControl error, audit and monitoring instances
+- ServicePulse
+
+#### Platform configuration
+
+- `region`
+- `queueNamePrefix` - change this to avoid queue conflicts in your region
```
jpalac: Should this be more clear that it's not required? maybe: 
```suggestion
- `queueNamePrefix` - optional - use this to avoid queue conflicts in your region
```

And then in the code actually fill it in with the sample name to demonstrate it?

platform/aspire/index.md
```diff
@@ -0,0 +1,452 @@
...
+| --------------------------------------------------------- | ---------------------------- |
+| [Learning](/transports/learning/)                         | Supported (Development only) |
+| [Azure Service Bus](/transports/azure-service-bus/)       | Supported                    |
+| [RabbitMQ](/transports/rabbitmq/)                         | Supported                    |
+| [Amazon SQS](/transports/sqs/)                            | Supported                    |
+| [Azure Storage Queues](/transports/azure-storage-queues/) | Not yet supported            |
+| [Microsoft SQL Server](/transports/sql/)                  | Not yet supported            |
+| [PostgreSQL](/transports/postgresql/)                     | Not yet supported            |
+| [IBM MQ](/transports/ibmmq/)                              | Not yet supported            |
+| [MSMQ](/transports/msmq/)                                 | Not available in containers  |
```
warwickschroeder: Does this status make sense? Should we just say "Not supported?
```suggestion
| [MSMQ](/transports/msmq/)                                 | Not yet supported  |
```

rbev: should it even be included in the table at all?

jpalac:
```suggestion
```

platform/aspire/index.md
```diff
@@ -450,12 +450,15 @@ If `AddParticularPlatform` is called but no platform components are added, the i
...
-### "Platform [name] has container images not using the 'latest' tag" warning at startup
+### "Platform [name] has mismatched ServiceControl container image versions" in Aspire host console
 
-This warning appears when one or more ServiceControl or ServicePulse container images have been pinned to a specific version tag instead of using latest. For example:
+This warning appears when the ServiceControl container images (error, audit, and monitoring instances) are configured with different version tags. For example:
 
 snippet: aspire-components-version
 
-All platform container images default to the latest tag, which ensures the components are always compatible with each other and with this version of the hosting package. Pinning to a specific version risks version skew between components and may cause compatibility issues.
+ All ServiceControl components (error, audit, and monitoring) must run the same version to ensure compatibility. The warning lists each component with its configured image and tag so the mismatch is easy to identify.
```
warwickschroeder: Should we link "/servicecontrol/upgrades/support-policy" somewhere ?

platform/aspire/index.md
```diff
@@ -0,0 +1,464 @@
...
+
+### "Platform [name] has mismatched ServiceControl container image versions" in Aspire host console
+
+This warning appears when the ServiceControl container images (error, audit, and monitoring instances) are configured with different version tags. For example:
+
+snippet: aspire-components-version
+
+ All ServiceControl components (error, audit, and monitoring) must run the same version to ensure compatibility. The warning lists each component with its configured image and tag so the mismatch is easy to identify.
+
+To resolve the warning, ensure all ServiceControl instances use the same image tag, or remove explicit `.WithImage()` calls to use the default latest tag, which always refers to a compatible set of images.
```
warwickschroeder:
```suggestion
To resolve the warning, ensure all ServiceControl instances use the same image tag, or remove explicit `.WithImage()` calls to use the default `latest` tag, which always refers to a compatible set of images.
```

warwickschroeder (approved):

platform/aspire/index.md
jpalac: Maybe "must" is too strong?

```suggestion
All ServiceControl components (error, audit, and monitoring) should run the same [supported version](https://docs.particular.net/servicecontrol/upgrades/supported-versions) to ensure compatibility. The warning lists each component with its configured image and tag so the mismatch is easy to identify.
```

nservicebus/platform-nuget-packages.md
```diff
@@ -37,6 +37,7 @@ This page contains a curated list of NuGet packages commonly used to build a dis
 - [NServiceBus.AzureFunctions.InProcess.ServiceBus](https://www.nuget.org/packages/NServiceBus.AzureFunctions.InProcess.ServiceBus)
 - [NServiceBus.AzureFunctions.Worker.ServiceBus](https://www.nuget.org/packages/NServiceBus.AzureFunctions.Worker.ServiceBus)
 - [NServiceBus.AwsLambda.Sqs](https://www.nuget.org/packages/NServiceBus.AwsLambda.Sqs)
+- [Particular.Aspire.Hosting.ServicePlatform](https://www.nuget.org/packages/Particular.Aspire.Hosting.ServicePlatform)
```
mikeminutillo: I wonder if this deserves it's own Orchestration section. It is not the same thing as the other hosting packages.

platform/aspire/index.md
```diff
@@ -0,0 +1,464 @@
+---
+title: Aspire
+summary: Describes how to orchestrate the Particular Platform via the Aspire AppHost
```
mikeminutillo:
```suggestion
summary: Describes how to orchestrate the Particular Platform via an Aspire AppHost
```

platform/aspire/index.md
```diff
@@ -0,0 +1,464 @@
...
+reviewed: 2026-05-21
+component: Aspire.Hosting
+related:
+- samples/hosting/aspire
+- samples/aspire/platform
+- samples/aspire/platform-asb
+---
+
+The `Particular.Aspire.Hosting.ServicePlatform` package is an [Aspire](https://aspire.dev/) hosting integration that runs the Particular Platform — the ServiceControl instances, ServicePulse, persistence, and message transport — as part of an Aspire AppHost. It is intended for developers and technical leads who run the platform locally during development and want the same AppHost to carry through to publish-mode deployments, without maintaining a separate set of infrastructure scripts.
```
mikeminutillo:
```suggestion
The `Particular.Aspire.Hosting.ServicePlatform` package is an [Aspire](https://aspire.dev/) hosting integration that runs the Particular Service Platform — the ServiceControl instances, ServicePulse, persistence, and message transport — as part of an Aspire AppHost. It is intended for developers and technical leads who run the platform locally during development and want the same AppHost to carry through to publish-mode deployments, without maintaining a separate set of infrastructure scripts.
```

platform/aspire/index.md
```diff
@@ -0,0 +1,464 @@
...
+## Overview
+
+The `Particular.Aspire.Hosting.ServicePlatform` package extends an [Aspire](https://aspire.dev/) AppHost so it can run a complete Particular Platform alongside the rest of a distributed application.
+
+A single `AddParticularPlatform(...)` call registers a platform resource that owns:
+
+- [ServiceControl](/servicecontrol) error, audit, and monitoring instances
+- [ServicePulse](/servicepulse)
+- A managed RavenDB persistence instance, or one supplied by the AppHost
+- The configured message transport (Learning, Azure Service Bus, or RabbitMQ)
```
mikeminutillo: This list is missing SQS. Maybe remove the list and link to the section below.

platform/aspire/index.md
```diff
@@ -0,0 +1,464 @@
...
+
+snippet: aspire-apphost
+
+## Supported components
+
+The integration is being built out across the Particular Platform in stages. The tables below show what is currently wired through the AppHost.
+
+If you are using the Particular Service Platform with Aspire today and would like to see more Aspire support, [reach out](https://github.com/Particular/NServiceBus/issues/6941).
+
+### Transport
```
mikeminutillo, line 36: Each transport has a quick summary table. e.g. [AzureServiceBus](https://github.com/Particular/docs.particular.net/blob/master/transports/azure-service-bus/index.md?plain=1#L18-L30). I wonder if we should include Aspire compatibility in there and link back to this section (for Yes or No).

platform/aspire/index.md
```diff
@@ -0,0 +1,464 @@
...
+### Persistence
+
+| Persistence                          | Status    |
+| ------------------------------------ | --------- |
+| [RavenDB](/servicecontrol/ravendb/)  | Supported |
+
+## Prerequisites
+
+- The [Aspire CLI](https://aspire.dev/get-started/install-cli/) and the .NET 10 SDK, used to build and run the AppHost project.
+- A container runtime such as [Docker Desktop](https://www.docker.com/products/docker-desktop/) or [Podman](https://podman.io/). The platform components are pulled from Docker Hub as `particular/servicecontrol`, `particular/servicecontrol-audit`, `particular/servicecontrol-monitoring`, `particular/servicecontrol-ravendb`, and `particular/servicepulse`.
```
mikeminutillo: Consider linking this to [here](https://docs.particular.net/servicecontrol/hosting#containers) instead as it contains a list of supported container runtimes.

platform/aspire/index.md
```diff
@@ -0,0 +1,464 @@
...
+dotnet add package Particular.Aspire.Hosting.ServicePlatform
+```
+
+Only the AppHost project needs this reference. NServiceBus endpoint projects pick up the platform license and transport connection string from environment variables injected by `WithParticularPlatform(...)`, so they do not need their own reference to this package.
+
+## Quick start
+
+`AddDefaultComponents()` wires up the Learning transport, a managed RavenDB persistence instance, the ServiceControl error, audit, and monitoring instances, and ServicePulse with sensible defaults:
+
+snippet: aspire-quick-start-1
```
mikeminutillo: Consider giving these snippets a name rather than a number. If someone wants to add one, or shuffle them around in the future, it will be confusing.

platform/aspire/index.md
```diff
@@ -0,0 +1,464 @@
...
+
+To attach an [NServiceBus endpoint](/nservicebus/endpoints/) so it picks up the platform's license and transport connection string, chain `WithParticularPlatform(platform)` on the endpoint's project resource:
+
+snippet: aspire-quick-start-2
+
+These defaults are intended for local development. For deployment scenarios, configure an explicit transport, persistence, and license as described in [Configuring the transport](#configuring-the-transport), [Configuring persistence](#configuring-persistence), and [Configuring the license](#configuring-the-license).
```
mikeminutillo, line 83: It may be worth providing a sample snippet here with a full configuration just to show it.

platform/aspire/index.md
```diff
@@ -0,0 +1,464 @@
...
+
+The platform node tracks readiness as its children come up. It starts in `Starting`, transitions to `Running` once every child reports healthy, and moves to `RuntimeUnhealthy` if any child stops.
+
+## Configuring the transport
+
+The platform uses whichever transport is configured via a `WithTransport*` extension on the platform resource. The same transport connection string is then propagated to every ServiceControl instance and to any NServiceBus endpoint attached via `WithParticularPlatform(...)`. See [Supported components](#supported-components) for the transports currently wired through the integration.
+
+### Learning transport
+
+The [Learning transport](/transports/learning/) stores messages on the local file system. By default they are persisted to `.learningtransport` under the AppHost project directory. Pass a custom path to override.
```
mikeminutillo: Ideally this would be a `.learningtransport` next to the solution file. Consider linking to https://docs.particular.net/transports/learning/#usage-storage-directory

Your AppHost is likely to be part of the same solution as your endpoints. If they default to the same location then the endpoints do not need any configuration of a specific folder.

platform/aspire/index.md
```diff
@@ -0,0 +1,464 @@
...
+
+snippet: aspire-transport-asb
+
+#### Options
+
+| Option | Default |
+| --- | --- |
+| `azureServiceBus` (`IResourceBuilder<IResourceWithConnectionString>`) parameter on `WithTransportAzureServiceBus` | Required |
+
+Azure Service Bus tuning such as topic name, partitioning, hierarchy namespace, and websocket mode is configured on the connection string itself. See [Azure Service Bus transport configuration](/servicecontrol/transports.md#azure-service-bus) for the full list of connection-string options.
```
mikeminutillo, line 134: I'm not sure how this would work if I was working with a resource itself and not a connection string. That might be something to think about for a future version of the package.

platform/aspire/index.md
```diff
@@ -0,0 +1,464 @@
...
+| `QueueNamePrefix`          (`string?`) property on `AmazonSQSTransportSettings`                        | Optional |
+| `TopicNamePrefix`          (`string?`) property on `AmazonSQSTransportSettings`                        | Optional |
+| `S3BucketForLargeMessages` (`IExpressionValue?`) property on `AmazonSQSTransportSettings`              | Optional |
+| `QueueNamePrefix`          (`string?`) property on `AmazonSQSTransportSettings`                        | Optional |
+
+## Configuring persistence
+
+ServiceControl uses a persistence backend to store error and audit data, retry state, and saga history. See [Supported components](#supported-components) for the persisters currently wired through the integration.
+
+Configure persistence on the platform resource, then pass the resulting persistence builder into the ServiceControl Error and Audit instances that need it. The Monitoring instance does not require persistence.
```
mikeminutillo: Should we have some statement about error and audit being able to share a persistence (which it does for the default configuation)? Should we be making (or preferrably linking off to) some guidance around this?

platform/aspire/index.md
```diff
@@ -0,0 +1,464 @@
...
+
+Aspire supports multiple deployment targets, including [Docker Compose](https://aspire.dev/deployment/docker-compose/), Kubernetes, and Azure Container Apps. See [Aspire pipelines](https://aspire.dev/deployment/pipelines/) for the deployment model and the supported targets.
+
+Two integration-specific points to keep in mind for any publish target:
+
+- The license parameter default is materialized into the manifest at publish time. Supply the license through Aspire parameter inputs in production so the value does not live in the published output. See [Supplying the license in production](#production-considerations-supplying-the-license-in-production).
+- The Learning transport is blocked in publish mode.
```
mikeminutillo, line 439: Both of these points are covered already above. It probably doesn't hurt to repeat them, just pointing it out in case it was not intentional.

mikeminutillo (approved):

nservicebus/platform-nuget-packages.md
rbev: There were three options I looked at from the existing list when I put it here:

* Hosting
* Tooling - The other items were more cli/oneshot/maintenance tools
* Other - didn't feel like it fit in very well with that group

Hosting felt the best out of those, but you are correct another category might be more appropriate.

platform/aspire/index.md
rbev: I think it's probably worthwhile to change our default to be solution-relative, it's still worth having the example here as we allow the path to be passed in though the extension method 

changing the default will make it more likely to "just work" if someone discovers our package without the samples

platform/aspire/index.md
rbev: Do we want to break out the `aspire-transport-asb` into two snippets rather than the inline comment: one with transport and one with connection string?

samples/aspire/platform-sqs/Core_10/samplepublish/docker-compose.yaml
rbev, line 1: I think that this publish folder may have been accidentally included. I think it should probably be deleted

platform/aspire/index.md
jpalac:
```suggestion
- A [container runtime](/servicecontrol/hosting#containers). The platform components are pulled from Docker Hub as `particular/servicecontrol`, `particular/servicecontrol-audit`, `particular/servicecontrol-monitoring`, `particular/servicecontrol-ravendb`, and `particular/servicepulse`.
```

platform/aspire/index.md
jpalac: will make the change to the default location

platform/aspire/index.md
jpalac: keeping it as is for initial release

platform/aspire/index.md
jpalac:
```suggestion
Configure persistence on the platform resource, then pass the resulting persistence builder into the ServiceControl Error and Audit instances that need it. The Monitoring instance does not require persistence. Follow the [managing persistence guidance](/servicecontrol/ravendb/containers) when setting up for production use.
```

jpalac (approved):

PR #8269, opened 2026-05-21
More Core 10.2 docs updates

DavidBoike:


Snippets/CosmosDB/CosmosDB_4/CustomProvider.cs
```diff
@@ -19,11 +20,11 @@ public CustomCosmosClientProvider(CosmosClient cosmosClient)
 
 class CosmosDBCustomClientProviderRegistration
 {
-    public CosmosDBCustomClientProviderRegistration(EndpointConfiguration endpointConfiguration)
+    public CosmosDBCustomClientProviderRegistration(IHostApplicationBuilder builder)
     {
         #region CosmosDBCustomClientProviderRegistration
 
-        endpointConfiguration.RegisterComponents(c => c.AddTransient<IProvideCosmosClient, CustomCosmosClientProvider>());
+        builder.Services.AddTransient<IProvideCosmosClient, CustomCosmosClientProvider>();
```
andreasohlund, line 27: @danielmarbach shouldn't this be a singleton?

Snippets/DynamoDB/DynamoDB_4/CustomProvider.cs
```diff
@@ -21,11 +23,11 @@ public CustomDynamoClientProvider(IAmazonDynamoDB dynamoClient)
 
 class DynamoDBCustomClientProviderRegistration
 {
-    public DynamoDBCustomClientProviderRegistration(EndpointConfiguration endpointConfiguration)
+    public DynamoDBCustomClientProviderRegistration(IHostApplicationBuilder builder)
     {
         #region DynamoDBCustomClientProviderRegistration
 
-        endpointConfiguration.RegisterComponents(c => c.AddTransient<IDynamoClientProvider, CustomDynamoClientProvider>());
+        builder.Services.AddTransient<IDynamoClientProvider, CustomDynamoClientProvider>();
```
andreasohlund, line 30: Same here

Snippets/CosmosDB/CosmosDB_4/CustomProvider.cs
danielmarbach: It would be better but ultimately doesn't matter because the persister uses it once by injecting into the seams as far as I recall so effectively it becomes a singleton

danielmarbach: But doesn't hurt to double check or change it

PR #8278, opened 2026-05-23
Move send-only endpoint documentation to /endpoints#send-only

andreasohlund:
Hosting was never the right place for it

### DIFF
--- a/nservicebus/hosting/index.md
+++ b/nservicebus/hosting/index.md
@@ -2,11 +2,10 @@
 title: Hosting
 summary: Describes the various approaches to endpoint hosting
 component: Core
-reviewed: 2026-05-05
+reviewed: 2026-05-23
 redirects:
 - nservicebus/hosting/self-hosting
 - nservicebus/hosting/self-hosting-v4.x
-- nservicebus/hosting/send-only-endpoints
 - nservicebus/hosting/service-fabric-hosting
 ---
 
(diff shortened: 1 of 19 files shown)


nservicebus/hosting/index.md
```diff
@@ -66,3 +57,7 @@ NServiceBus endpoints have certain requirements for the hosting environment:
 * The endpoint process needs write access to write log files. See the [logging documentation](/nservicebus/logging) for more details about the default log file location and how to configure logging.
 * The endpoint process needs write access to write the startup diagnostics file. See the [startup diagnostics documentation](/nservicebus/hosting/startup-diagnostics.md) for more details about the diagnostic file.
 * Since NServiceBus makes assumptions on aspects like assembly names, ILMerging any of the NServiceBus* assemblies is not supported.
+
+## Send-only hosting
+
+Send-only is documented as an [endpoint concept](/nservicebus/endpoints/#send-only), rather than a hosting approach.
```
danielmarbach: This is a little weird because it documents a previous state we moved away from as documentation. I'm on the fence. What's the benefit of this addition?

If you want to leave it I'd rephrase it to. For send only endpoints consult XYZ documentation

danielmarbach (approved):

nservicebus/hosting/index.md
andreasohlund: I left it there so as not to break external links since our redirects don't support sub-sections but I'm on the fence as well if we need that.

@DavidBoike ?

PR #8303, opened 2026-06-03
Review of RavenDB connection documentation

tamararivera:
Review for clarity and consolidate some partials


persistence/ravendb/connection.md
```diff
@@ -7,33 +7,51 @@ redirects:
...
 ---
 
-include: dtc-warning
+## Database creation
 
-include: cluster-configuration-info
+When using RavenDB as persistence, the RavenDB client won't create the database it if it doesn't already exist.
+This is why the database needs to be previously created or created when initializing the endpoint.
 
-The following sections outline various ways to connect to the RavenDB server. Specifying an external shared store (providing a fully configured RavenDB `DocumentStore` instance) is preferred so that the configuration of the DocumentStore is not obscured.
```
tamararivera, line 17: I marked the external shared store for all persisters as the recommended approach, but I couldn't get the reasoning why it is preferred. It only says is not obscured, by what? how?

persistence/ravendb/connection.md
```diff
@@ -7,33 +7,51 @@ redirects:
...
+The following sections outline various ways to connect to the RavenDB server.
 
-include: raven-dispose-warning
+### Default
 
-partial: externalatinitialization
+By default, a `DocumentStore` is created that connects to `http://localhost:8080` and uses the endpoint name as its database name. This default connection is used for all the persisters.
```
tamararivera: I moved this to the top, as it makes sense to me to have the default behavior first

persistence/ravendb/connection.md
```diff
@@ -7,33 +7,51 @@ redirects:
...
 
-partial: shared
```
tamararivera, line 23: I deleted these partials, as the minimum version was lower than the minimum we support.

persistence/ravendb/connection.md
```diff
@@ -7,33 +7,51 @@ redirects:
...
+### External shared store for all persisters
 
+This is the recommended method to connect to the database.
+The RavenDB `DocumentStore` instance can be provided directly to NServiceBus to use for all persisters. This enables sharing the same application database for NServiceBus data as well.
```
tamararivera, line 20: What is a persister in this case? on the snippet, it shows subscription, is that the same?

persistence/ravendb/connection.md
```diff
@@ -7,33 +7,51 @@ redirects:
...
 
-See also [How to specify endpoint name](/nservicebus/endpoints/specify-endpoint-name.md).
+To use an external `DocumentStore`, but defer its creation until NServiceBus initializes, a custom factory delegate can be provided which will allow the `DocumentStore` to be created with access to the settings and the dependency injection container. This gives the ability to configure the `DocumentStore` based on conventions derived from endpoint data present in the settings object. For example, the `DocumentStore` instance can be configured to use the [Endpoint Name](/nservicebus/endpoints/specify-endpoint-name.md) as its database name by accessing `readOnlySettings.EndpointName()`.
 
-## Database creation
+snippet: ravendb-persistence-create-store-by-func
+
+include: raven-dispose-warning
 
-Prior to RavenDB version 4.0, RavenDB automatically creates the database if it doesn't already exist. This is no longer the case in RavenDB versions 4.0 and above.
```
tamararivera, line 39: This is at the top of the document now, but I reworded to not talk about versions we don't support anymore.

persistence/ravendb/connection.md
```diff
@@ -7,33 +7,51 @@ redirects:
...
 
-include: dtc-warning
+## Database creation
 
-include: cluster-configuration-info
+When using RavenDB as persistence, the RavenDB client won't create the database it if it doesn't already exist.
+This is why the database needs to be previously created or created when initializing the endpoint.
 
-The following sections outline various ways to connect to the RavenDB server. Specifying an external shared store (providing a fully configured RavenDB `DocumentStore` instance) is preferred so that the configuration of the DocumentStore is not obscured.
+The following sections outline various ways to connect to the RavenDB server.
```
ngallegos: I'm not sure we even need a dedicated section to tell readers that the database needs to be created in lines 13-16. It could just be a short note before the rest of the article.

```suggestion

> [!NOTE]
> The database must be created before using RavenDB persistence.

```

persistence/ravendb/connection.md
```diff
@@ -7,33 +7,51 @@ redirects:
...
  - nservicebus/ravendb/connecting
  - nservicebus/ravendb/connection
-reviewed: 2024-09-26
+reviewed: 2026-06-03
 ---
 
-include: dtc-warning
+## Database creation
 
-include: cluster-configuration-info
```
ngallegos, line 15: If we [make it a note](https://github.com/Particular/docs.particular.net/pull/8303/changes#r3356105420) we don't a dedicated section.

persistence/ravendb/connection.md
ngallegos: Do we want the default first, or the preferred approach first?

I agree, it seems like something named "Default" should be first, but if we want to nudge the user in the right direction by having the recommended option (external store) to be first, maybe we could rename this section to "Local store" and keep it where it was?

bartekwasielak: As a general rule, the default behaviour should come first. In this case it sounds like the default one is a fail-safe, that should always work, because the recommended one requires more conscious configuration. It could be phrased like "we recommend doing this; if you don't set it, it will fall back to this".

persistence/ravendb/connection.md
bartekwasielak: Agree with this. Especially that this article is about the ravendb connection options. Even this note on cluster configuration info is also present in https://docs.particular.net/persistence/ravendb/ - no need to repeat it here.

persistence/ravendb/connection.md
```diff
@@ -7,33 +7,51 @@ redirects:
...
+When using RavenDB as persistence, the RavenDB client won't create the database it if it doesn't already exist.
+This is why the database needs to be previously created or created when initializing the endpoint.
```
bartekwasielak:
```suggestion
```

ngallegos (approved):

PR #8305, opened 2026-06-04
Review callback sample

ngallegos:
This simplifies the callback sample by removing the `WebSender` project. It duplicated callback examples in a more complicated environment and didn't add any benefit.

The disclaimer was also moved from an `include` file to the sample directly because that was the only place it was being referenced.


bartekwasielak:
The web project does exactly the same as the console application, right? I feel that the intention here might have been to use it to show the NSB's usage not only in console apps, but also in the web, where callbacks might be more popular.

Do you think we could do the opposite, remove the console sender and leave the web one? I think that this is a more rare configuration, so it might be valuable on its own.

ngallegos:
@bartekwasielak:
 
> The web project does exactly the same as the console application, right?

Correct, it does the exact same thing except uses web pages and requests to trigger the messages. 

> I feel that the intention here might have been to use it to show the NSB's usage not only in console apps, but also in the web, where callbacks might be more popular.

> Do you think we could do the opposite, remove the console sender and leave the web one? I think that this is a more rare configuration, so it might be valuable on its own.Do you think we could do the opposite, remove the console sender and leave the web one? I think that this is a more rare configuration, so it might be valuable on its own.

I agree that was probably the intention, but my thought in removing it was that the sample is supposed to show the use of callbacks, and the web project is a more complicated way to show that.  It could imply that a web project is needed to use callbacks.  

The other reason I wanted to keep the console sender instead of the web project is that there is already a dedicated sample to show how to [use NServiceBus in a web application already](https://docs.particular.net/samples/web/asp-web-application/).

tamararivera (approved):

bartekwasielak (approved):

PR #8307, opened 2026-06-04
Update Aspire sample to 13.4

bording:


bording:
Looks like there is a bug in 13.4 related to multi-targeting: https://github.com/microsoft/aspire/issues/17912

PR #8324, opened 2026-06-10
Legacy docs review: connecting endpoints

bartekwasielak:


platform/connecting.md
```diff
@@ -20,17 +20,17 @@ The ServicePlatform Connector plugin provides a unified API to connect an NServi
...
 
-The connection details can be parsed from JSON-compliant text with [a specific configuration schema](json-schema.md).
+The connection details can be parsed from JSON with [a specific configuration schema](json-schema.md).
 
 snippet: PlatformConnector-FromJson
 
 The JSON file looks like this:
 
-snippet: PlatformConnector-Json
+snippet: PlatformConnector-json
```
ngallegos: I don't see a corresponding change to the snippet code. Is it case-insensitive or does the snippet identifier need to be updated?

bartekwasielak: I thought it was supposed to be lower case. If it's not necessary to change, I'll revert this.

ngallegos (approved):

PR #8332, opened 2026-06-12
Legacy docs review: Multi-tenant RavenDB sample

tamararivera:
Moved some things around to help with clarity.

### DIFF
--- a/samples/multi-tenant/ravendb/sample.md
+++ b/samples/multi-tenant/ravendb/sample.md
@@ -25,12 +25,9 @@ The databases created by this sample are:
  * `MultiTenantSamples-A`
  * `MultiTenantSamples-B`
 
-include: persistence-running
-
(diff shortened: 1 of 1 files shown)


bartekwasielak (approved):
I guess we don't have a defined order for "running the project" and "code walk-through".

tamararivera:
I swapped them, because on the running the project section it talks about a SQL DB:
> Open SQL Server Management Studio and go to the tenant databases. Verify that there are rows in saga state table (dbo.OrderLifecycleSagaData) and in the orders table (dbo.Orders) for each message sent.

I thought it was a mistake, but a SQL DB was later described on the code walk-through section:
> The Receiver mimics a back-end system. It is configured to use SQL persistence in multi-tenant mode.

So I think it will be easier to understand this way

PR #8346, opened 2026-06-17
Use major versions for pub/sub deprecation guides

mnadareski-particular:
Instead of targeting the next minor version release for pub/sub hybrid mode deprecation, they should target the next major version instead.

This updates the name, description, and link to each page in question.

### DIFF
--- a/menu/menu.yaml
+++ b/menu/menu.yaml
@@ -1060,7 +1060,7 @@
       Articles:
       - Url: persistence/upgrades/nhibernate-10to11
         Title: Version 10 to 11
-      - Url: persistence/upgrades/nhibernate-8too9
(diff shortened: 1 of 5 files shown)


mauroservienti (approved):

menu/menu.yaml
```diff
@@ -1060,7 +1060,7 @@
       Articles:
       - Url: persistence/upgrades/nhibernate-10to11
         Title: Version 10 to 11
-      - Url: persistence/upgrades/nhibernate-8to9
+      - Url: persistence/upgrades/nhibernate-8too9
```
mauroservienti: This change can be reverted, right?

mnadareski-particular: I'm not entirely sure how that happened, but yes.

PR #8356, opened 2026-06-19
Switch aspire samples over to Aspire component

rbev:
* Fix nuget alias
* Switch samples to depend on hosting package components
* Remove unnecessary partials


components/nugetAlias.txt
```diff
@@ -7,7 +7,7 @@ ASQ: NServiceBus.Azure.Transports.WindowsAzureStorageQueues
 ASQN: NServiceBus.Transport.AzureStorageQueues
 ASP: NServiceBus.Persistence.AzureStorage
 ASTP: NServiceBus.Persistence.AzureTable
-AspirePlatformHost: Particular.Aspire.Hosting.ServicePlatform
+Aspire.Hosting: Particular.Aspire.Hosting.ServicePlatform
```
jpalac: So what happens when we have other host packages - shouldn't this name be more explicit about what it is?

rbev: I just aligned it with the name that was in the other metadata file because I couldn't get the samples to run.
Happy to change it to be more specific if we want to - will have to update the snippets too

jpalac: Aspire.Hosting.Platform or Aspire.Hosting.ServicePlatform if we wanted to align more with the package name

rbev: I'm actually not sure why an alias is something desired at all, seems like using nuget package names would have been better. 

Should we just make our alias the *same* as our package name?

danielmarbach: The alias goes into the file name which has length restrictions. That's probably why the alias exists

warwickschroeder: It seems the path it ends up being is the problem (windows max 260 chars), and the reason why its short. `AHP` would do, but so would `AspireHostPlatform` tbh. The longest alias in the file is `SqlTransportLegacySystemClient` at 30 chars

jpalac (approved):

mikeminutillo (approved):

warwickschroeder (approved):

PR #8360, opened 2026-06-22
Legacy documentation review: Persistence - CosmosDB - Transactions

ramonsmits:
Legacy documentation review of `persistence/cosmosdb/transactions.md`.

### Technical correction
- The page referenced a non-existent API `IMessageHandlerContext.SynchronizedStorageSession.GetSharedTransactionalBatch()`. The shared `TransactionalBatch` is actually obtained via the `CosmosPersistenceSession()` extension method (then `.Batch`), which is exactly what the `CosmosDBHandlerSharedTransaction` snippet shown earlier on the page uses. Corrected the prose to match.

### DIFF
--- a/Snippets/CosmosDB/CosmosDB_1/SharedTransactions.cs
+++ b/Snippets/CosmosDB/CosmosDB_1/SharedTransactions.cs
@@ -38,12 +38,9 @@ class UsageHandler : IHandleMessages<MyMessage>
 
 #region CosmosDB-TransactionalBatchRegisteredWithDependencyInjectionResolvedInHandler
 
-class MyHandler : IHandleMessages<MyMessage>
+class MyHandler(ICosmosStorageSession storageSession) : IHandleMessages<MyMessage>
 {
-    public MyHandler(ICosmosStorageSession storageSession)
-    {
-        transactionalBatch = storageSession.Batch;
-    }
+    readonly TransactionalBatch transactionalBatch = storageSession.Batch;
 
     public Task Handle(MyMessage message, IMessageHandlerContext context)
     {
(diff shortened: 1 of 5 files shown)


ramonsmits:
**Verification of the API correction** (checked against the `NServiceBus.Persistence.CosmosDB` package history, since `GetSharedTransactionalBatch()` could plausibly have been an earlier major's API):

- `GetSharedTransactionalBatch()` existed **only during pre-release development (Sept–Oct 2020)** and was renamed to `CosmosPersistenceSession()` before the first release (commit `b2764df`).
- **No released tag** (`2.0.5` through `4.0.1`) ever contained `GetSharedTransactionalBatch()`; **every** released version exposes `CosmosPersistenceSession()` (returning `ICosmosStorageSession`, with a `.Batch` property).
- Neither CosmosDB upgrade guide (`cosmosdb-1to2`, `cosmosdb-3to4`) references it.

So this was never a public/shipped API — the page documented a name that never shipped. The correction matches the `CosmosDBHandlerSharedTransaction` snippet on the same page, so no versioned/upgrade note is needed.

ramonsmits:
Refined the "With dependency injection" wording per review feedback. The snippet after that sentence injects `ICosmosStorageSession` and reads its `.Batch`, so naming the type is clearer:

> Alternatively to using the `context.SynchronizedStorageSession.CosmosPersistenceSession()` extension method, the `ICosmosStorageSession` can be injected directly into the handler:

This reflects the actual model: both the `CosmosPersistenceSession()` extension method and dependency injection surface an `ICosmosStorageSession` (the method's return type is `ICosmosStorageSession`), and `.Batch` on it is the shared `TransactionalBatch`.

ramonsmits:
Reworked the "With dependency injection" section and modernized the snippets (commit fd667d9):

- **Reordered** the section: it now shows `ICosmosStorageSession` injected **directly into the handler** first, then presents the custom-dependency variant framed as enabling more complex DI graphs (the previous order led with the more complex case).
- **Primary constructors** for `MyHandler`, `MyCustomDependency`, and `MyHandlerWithCustomDependency` — all four snippet projects target net8.0+/net10.0 with no `LangVersion` override, so C# 12+ is the default.
- **Dropped the redundant `private`** field modifier (it's the default), and removed the explicit backing field where the primary-constructor parameter can be captured directly.

Verified: all four projects (`CosmosDB_1`–`CosmosDB_4`, NSB 7→10) build with **0 warnings / 0 errors**, and `docstool test` passes.

PR #8405, opened 2026-07-15
Document ASB QueueLengthQueryDelayInterval as a monitoring-only setting

ramonsmits:
The `QueueLengthQueryDelayInterval` connection string option was buried in the ASB topic-per-event topology section. It only affects the Monitoring instance, so it now has its own "Queue length monitoring" section noting the value can be increased if rate throttling is observed.

### DIFF
--- a/servicecontrol/transports.md
+++ b/servicecontrol/transports.md
@@ -1,7 +1,7 @@
 ---
 title: Transport configuration
 summary: ServiceControl can be configured to use one of the supported message transports, which are configured for each instance type
-reviewed: 2026-07-03
+reviewed: 2026-07-15
 component: ServiceControl
 ---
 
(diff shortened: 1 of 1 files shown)


servicecontrol/transports.md
```diff
@@ -84,6 +85,15 @@ For example, given a hierarchy namespace of `my-hierarchy` and error, audit, and
 - The [audit queue name](/servicecontrol/audit-instances/configuration.md#transport-servicebusauditqueue) would need to be set to `my-hierarchy/audit`.
 - The [monitoring queue name](/servicecontrol/monitoring-instances/configuration.md#transport-monitoringerrorqueue) would need to be set to `my-hierarchy/monitoring`.
 
+### Queue length monitoring
+
+> [!NOTE]
+> This setting applies only to the [Monitoring instance](/servicecontrol/monitoring-instances/), which is the only instance type that tracks queue lengths.
+
+The Monitoring instance periodically queries the Azure Service Bus namespace to refresh the queue length data behind the throughput graphs:
```
johnsimons: Maybe consider adding a link to the graphs you referring to

servicecontrol/transports.md
```diff
@@ -84,6 +85,15 @@ For example, given a hierarchy namespace of `my-hierarchy` and error, audit, and
...
+
+* `QueueLengthQueryDelayInterval=<value_in_milliseconds>` - The delay between successive queue length refresh queries. The default value is 500 ms. Increase this value if Azure Service Bus rate throttling is observed.
```
johnsimons, line 95: Is the unit type milliseconds or TimeSpan? I thought when we looked at the source it was TimeSpan.

johnsimons (approved):
It LGTM but I asked a couple of questions to check

servicecontrol/transports.md
ramonsmits: No, its milliseconds:

https://github.com/Particular/ServiceControl/blob/6.18.2/src/ServiceControl.Transports.ASBS/ConnectionStringParser.cs#L39

servicecontrol/transports.md
ramonsmits: Added it in 4329d0d3f43115b2a13c83fba0a762c602d99f17

PR #8416, opened 2026-07-24
update masking information to include scope

PhilBastian:


nservicebus/throughput-tool/index.md
```diff
@@ -432,7 +432,7 @@ The following requests will be sent to the monitoring instance:
 
 ## Masking private data
 
-The generated report will contain the names of endpoints/queues. Certain strings can be masked in the report file if the queue names contain confidential or proprietary information.
+The generated report will contain the names of endpoints/queues and, for some transport types, a scope (e.g. the table schema in SQL Transport). Certain strings can be masked in the report file if the queue names contain confidential or proprietary information. Masking is not case sensitive.
```
mikeminutillo: I wouldn't use the term scope here as it is an internal term and unlikely to mean anything to the reader. Perhaps we should turn this into a list with more specific examples.

Something like this:

```suggestion
The generated report may contain terms which are considered confidential or proprietary information, including:
- Queue/endpoint names
- Catalog/schema names (SQL Server or PostgreSQL transport)
- Virtual host names (RabbitMQ transport)

The tool can be configured to mask selected terms. Masking is not case sensitive.
```

mikeminutillo (approved):

Issue #681, opened 2015-06-21
Configure.SetEndpointSLA needs snippet/version varients


SimonCropp:
http://docs.particular.net/nservicebus/operations/monitoring-endpoints#nservicebus-performance-counters-sla-violation-countdown

SimonCropp:
snippets added here http://docs.particular.net/nservicebus/operations/performance-counters

PR #8423, opened 2026-07-27
Docs: Clarify AWS Lambda SQS endpoint limitation

johnsimons:
The documentation for hosting NServiceBus in AWS Lambda with SQS has been updated.

It now clarifies that a single AWS Lambda application currently supports only one NServiceBus endpoint and encourages feedback on issue #681 for multi-endpoint support.

### DIFF
--- a/nservicebus/hosting/aws-lambda-simple-queue-service/index.md
+++ b/nservicebus/hosting/aws-lambda-simple-queue-service/index.md
@@ -6,7 +6,7 @@ related:
  - samples/aws/lambda-sqs
  - samples/aws/lambda-sqs-annotations
  - samples/showcase/loan-broker-showcase
-reviewed: 2026-05-26
+reviewed: 2026-07-27
 redirects:
  - previews/aws-lambda-simple-queue-service
 ---
@@ -15,7 +15,7 @@ Host NServiceBus endpoints with [AWS Lambda](https://aws.amazon.com/lambda/) usi
 
 ## Basic usage
 
-An NServiceBus endpoint is hosted in AWS Lambda by creating an `AwsLambdaSQSEndpoint` instance and calling the `Process` method from within an AWS Lambda definition.
+An NServiceBus endpoint is hosted in AWS Lambda by creating an `AwsLambdaSQSEndpoint` instance and calling the `Process` method from within an AWS Lambda definition. An AWS Lambda application currently supports one NServiceBus endpoint. Support for multiple endpoints in a single Lambda application is being considered. If multi-endpoint support would be valuable, [add a thumbs-up or share the use case on issue #681](https://github.com/Particular/NServiceBus.AwsLambda.Sqs/issues/681).
 
 ### `AwsLambdaSQSEndpoint` creation
 
--- a/nservicebus/hosting/aws-lambda-simple-queue-service/index.md
+++ b/nservicebus/hosting/aws-lambda-simple-queue-service/index.md
@@ -15,7 +15,9 @@ Host NServiceBus endpoints with [AWS Lambda](https://aws.amazon.com/lambda/) usi
 
 ## Basic usage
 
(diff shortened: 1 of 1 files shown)


warwickschroeder:
I sort a had in mind that there would be a "Limitations" heading of some sort with a nav item in the menu for components. This way its really easy to find and the GitHub links aren't hidden on a page somewhere. Thoughts?

nservicebus/hosting/aws-lambda-simple-queue-service/index.md
```diff
@@ -15,7 +15,7 @@ Host NServiceBus endpoints with [AWS Lambda](https://aws.amazon.com/lambda/) usi
 
 ## Basic usage
 
-An NServiceBus endpoint is hosted in AWS Lambda by creating an `AwsLambdaSQSEndpoint` instance and calling the `Process` method from within an AWS Lambda definition.
+An NServiceBus endpoint is hosted in AWS Lambda by creating an `AwsLambdaSQSEndpoint` instance and calling the `Process` method from within an AWS Lambda definition. An AWS Lambda application currently supports one NServiceBus endpoint. Support for multiple endpoints in a single Lambda application is being considered. If multi-endpoint support would be valuable, [add a thumbs-up or share the use case on issue #681](https://github.com/Particular/NServiceBus.AwsLambda.Sqs/issues/681).
```
jpalac: I think this should be its own paragraph at least so that it is more visible.

```suggestion
An NServiceBus endpoint is hosted in AWS Lambda by creating an `AwsLambdaSQSEndpoint` instance and calling the `Process` method from within an AWS Lambda definition. An AWS Lambda application currently supports one NServiceBus endpoint.

Support for multiple endpoints in a single Lambda application is being considered. If multi-endpoint support would be valuable, [add a thumbs-up or share the use case on issue #681](https://github.com/Particular/NServiceBus.AwsLambda.Sqs/issues/681).
```

johnsimons:
To me that is what sets apart the documentation in Microsoft, the relevance of the callout is contextualized to the document the user is reading and hence is a natural area to draw their attention, this is exactly what I explained in yesterdays meeting.
The Microsoft documentation actually has these callouts sprinkled everywhere not just under "Limitations", have a look at https://github.com/search?q=repo%3Adotnet%2FEntityFramework.Docs%20%22https%3A%2F%2Fgithub.com%2Fdotnet%2Fefcore%2Fissues%2F%22&type=code

My current approach to this is the following:
1. add contextual guidance first, in other words, inline it with current paragraphs when it makes sense
2. use a Limitations section on the page (not standalone) as the fallback when no natural context exists
3. avoid NOTE and WARNING callouts unless the behavior creates an immediate operational risk, this is so that doesn't feel we are begging for it!

johnsimons:
Microsoft also uses the 👍 emoji, which draws attention, is this something we should do?
image

warwickschroeder:
I think having links inline is good, but I also think having somewhere that immediately jumps out at you would be beneficial as well. i.e. an easy to find all "call to action" items for a component

warwickschroeder:
> Microsoft also uses the 👍 emoji, which draws attention, is this something we should do?

I dont think we use emojis anywhere in docs at the moment. Im ok with this, but just wondering what our writing guide says about them

nservicebus/hosting/aws-lambda-simple-queue-service/index.md
johnsimons: I don't mind ☝️

johnsimons: @warwickschroeder is this a good comprimise?

johnsimons:
> I dont think we use emojis anywhere in docs at the moment.

We can always introduce it.
The docs engine supports them.

jasontaylordev (approved):

PR #8425, opened 2026-07-27
Add documentation for component limitations and unsupported features

johnsimons:
This change introduces callouts and dedicated sections to clarify current limitations and unsupported features across different NServiceBus components:
*   Cosmos DB Persistence: custom saga finding and transactional batching.
*   Azure Service Bus Transport: message sessions and native message ID correlation.
*   Amazon SQS Transport: FIFO input queues for endpoint input queues.

Each clarification includes a link to a relevant GitHub issue where users can provide feedback and share their use cases.

### DIFF
--- a/nservicebus/sagas/saga-finding.md
+++ b/nservicebus/sagas/saga-finding.md
@@ -1,7 +1,7 @@
 ---
 title: Complex saga finding logic
 summary: Use IFindSaga to write custom code that resolves sagas.
-reviewed: 2026-06-29
+reviewed: 2026-07-27
 component: Core
 related:
 - samples/saga/nh-custom-sagafinder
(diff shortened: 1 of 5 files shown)


nservicebus/sagas/saga-finding.md
```diff
@@ -18,6 +18,9 @@ snippet: saga-finder
 
 partial: mapping
 
+> [!NOTE]
+> Cosmos DB persistence does not support custom saga finders. Use [message-property correlation](/nservicebus/sagas/message-correlation.md#message-property-expression) with a partition-aware mapping. For more information, see [Cosmos DB saga finding limitations](/persistence/cosmosdb/#usage-saga-finding-limitations).
+
```
jpalac, line 23: Feels a bit weird to be calling out a specific technology on this page. Should it be more generic?


```suggestion
> [!NOTE]
> Not every persistence supports custom saga finders. Check the individual persistence page and where not supported use [message-property correlation](/nservicebus/sagas/message-correlation.md#message-property-expression) with a partition-aware mapping. 

```

johnsimons: The way I look at it, is that we should be upfront and not make the user have to go somewhere else to find that information.

warwickschroeder (approved):

nservicebus/sagas/saga-finding.md
jpalac: If we have more than just cosmosdb not supporting it, then this will become quite a list. It's easier to maintain if the information is in one place (ie on the individual pages)

johnsimons: well, we can cross that bridge when it happens

PR #8446, opened 2026-08-05
Azure Service Bus: Better migration guidance and small fixes

danielmarbach:
The 4-to-5 guide had one path: the migration topology, event by event, until you reach topic-per-event. That's the right default for most systems. It provisions a topic per migrated event, though, and the fallback topic that would let you avoid that only becomes available after the migration is done.

For namespaces with a lot of event types near the Premium 1,000-entity cap, that ordering hurts. The migration spends the headroom it was meant to preserve, and the thing that would let you skip it shows up only once you no longer need it.

Transport 6.4 changes this. You can skip the migration topology and point `TopicTopology.Default` at the existing single topic. Unmapped events keep flowing through the shared topic, and you carve out dedicated topics only where volume or isolation actually matters. No per-event `EventToMigrate`/`MigratedX` mapping, and topic count grows with the events you carve out instead of with your total event-type count. The blend lands on `TopicTopology.Default`, so it's a supported permanent end-state, and it moves people off the migration topology that's on its way out.

I kept the detailed tradeoffs deeper in the guide and surfaced the choice at the top, so the alternative is visible from the entry point without bloating the migration steps.

## What's in here

- A **Choosing a migration path** section near the top that lays out both paths with links.
- A **Blending topic-per-event with a fallback topic** section with the inverse-migration path, when to pick it, and the tradeoffs.
- `SqlLikeFilter` as the fallback mode so new subscriptions stay compatible with existing single-topic endpoints, with the reasoning for why `CorrelationFilter` is a later optimization rather than the starting point.
- Tradeoffs that call out rule count, not filter type, as the dominant cost on a shared topic, and that the shared topic belongs on the low-volume tail.
- A note that this needs transport 6.4 or later, which for some systems means going from 4 straight to 6 instead of stopping at 5.

## Small fixes

- `version 6.1` -> `version 6.4` in the fallback topic note (stale version).
- `TopicRoutingMode.SqlFilter` -> `TopicRoutingMode.SqlLikeFilter` to match the real enum.

### DIFF
--- a/transports/azure-service-bus/topology_description_asbs_[5,).partial.md
+++ b/transports/azure-service-bus/topology_description_asbs_[5,).partial.md
@@ -175,7 +175,7 @@ snippet: asb-versioning-publisher-customization
 which would allow adding either a Correlation filter (preferred) or a SQL filter based on the promoted full name.
 
 > [!NOTE]
-> Starting with version 6.1 of the transport, the [FallbackTopic](#subscription-rule-matching-fallback-topic) feature provides a simpler alternative for managing message contract evolution by routing all unmapped events to a shared topic with built-in filtering support.
+> Starting with version 6.4 of the transport, the [FallbackTopic](#subscription-rule-matching-fallback-topic) feature provides a simpler alternative for managing message contract evolution by routing all unmapped events to a shared topic with built-in filtering support.
 
 ##### Advanced multiplexing strategies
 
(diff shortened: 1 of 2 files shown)


danielmarbach:
Pinging @SzymonPobiega more as an FYI

afprtclr (approved):

PR #8456, opened 2026-08-12
Add a section about license signature type to docs

jpalac:


nservicebus/licensing/index.md
```diff
@@ -46,6 +46,22 @@ partial: license-management
 
 Sometimes the license must be embedded in a single line of text, for example, in a command line when deploying an endpoint with Docker. For these scenarios, the license can be minified, removing all spaces and line breaks, by adding `minify=true` to the query string of the URL used to download a license file from the Particular Software website.
 
+### License signature type
+
+By default, licenses are signed using `Rsa1024Sha1` (1024-bit RSA with SHA-1). For environments that require stronger cryptographic signing, licenses can be signed using a 4096-bit RSA key pair with SHA-256 instead.
+
+To download a license with the stronger signature, add `signaturetype=Rsa4096Sha256` to the query string of the license download URL.
```
rbev: Should we hold off on putting the query string manipulation into public documentation until we settle on the customer site changes?

rbev (approved):

andreasohlund (approved):

nservicebus/licensing/index.md
jpalac:
```suggestion
To download a license with the stronger signature, add `signature-version=V2` to the query string of the license download URL.
```

rbev (approved):

andreasohlund (approved):

johnsimons:
This new section feels a bit out of place to me.

If we ask customers whether they want more security, the answer is going to be "yes, please" every single time. So framing this as something you opt into for "environments that require stronger cryptographic signing" reads a little oddly.

I understand it is not really about security, or at least not from that angle. It is about customers running locked-down environments where the weaker algorithms are blocked by policy.

So my question is: why do we need two options at all? Why isn't Rsa4096Sha256 simply the default from now on?

I assume the constraint is supported versions; for example, older versions can't validate the stronger signature. If that is the case, I think that constraint belongs in the email template that sends customers the download link, where it can be surfaced at the moment it matters, rather than as a query string parameter in the public docs that customers have to know exists.

rbev:
> I assume the constraint is supported versions; for example, older versions can't validate the stronger signature. If that is the case, I think that constraint belongs in the email template that sends customers the download link, where it can be surfaced at the moment it matters, rather than as a query string parameter in the public docs that customers have to know exists.

Yes, the primary reason is the version constraint. To make this easier we intend to add this in to the new license download screen in the customer portal to remove the need for the customer or support to modify the URL manually, this manual method is just the first step that we can take at this point.

We are also working under the assumption that at some point the default signature type will change. Given how recent this support is the number of customers on versions that don't support this signature type will be quite high so to minimise support issues at license renewal time we opted to keep the more compatible license as the default for now.

PR #8461, opened 2026-08-17
Cosmosdb sample with encryption

jpalac:
Addresses https://github.com/Particular/NServiceBus.Persistence.CosmosDB/issues/451


persistence/cosmosdb/index.md
```diff
@@ -207,6 +208,27 @@ They may also be set when using a `CosmosClientBuilder` via the [`WithThrottling
...
+
+Azure Cosmos DB supports [client-side encryption](https://learn.microsoft.com/en-us/azure/cosmos-db/how-to-always-encrypted) (also known as Always Encrypted), which encrypts sensitive data in the application before it is stored in Cosmos DB. This ensures that plaintext data is never visible to the Cosmos DB service.
+
+Client-side encryption can be used with NServiceBus saga data. Individual saga data properties can be encrypted by configuring an encryption policy on the Cosmos DB container. This is useful when saga data contains sensitive information such as personal identifiers, financial data, or other regulated content.
+
+### Limitations
+
+The following properties cannot be encrypted:
+
+- **`id`** and **partition key** properties — Cosmos DB requires these in plaintext for point reads and routing.
```
stevedowling: Is there a source for Cosmos requiring this or is it NServiceBus that requires it? It looks like [Cosmos should be able to do it](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/3241) if the encryption is deterministic because it encrypts the lookup values again before doing a point read.

Filtering on non-deterministically encrypted fields won't work or would fail silently because the values would be different each time it was encrypted. Not sure if we need to include that but a link to [this](https://learn.microsoft.com/en-gb/azure/cosmos-db/how-to-always-encrypted?tabs=dotnet#randomized-vs-deterministic-encryption) couldn't hurt.

jpalac: I have made a change to the sample to use v2 which allows both.

jasontaylordev (approved):

stevedowling (approved):

persistence/cosmosdb/index.md
```diff
@@ -207,6 +208,39 @@ They may also be set when using a `CosmosClientBuilder` via the [`WithThrottling
 
 snippet: CosmosDBConfigureThrottlingWithBuilder
 
+## Client-side encryption
+
+Azure Cosmos DB supports [client-side encryption](https://learn.microsoft.com/en-us/azure/cosmos-db/how-to-always-encrypted) (also known as Always Encrypted), which encrypts sensitive data in the application before it is stored in Cosmos DB. This ensures that plaintext data is never visible to the Cosmos DB service.
+
+Client-side encryption can be used with NServiceBus saga data. Individual saga data properties can be encrypted by configuring an encryption policy on the Cosmos DB container. This is useful when saga data contains sensitive information such as personal identifiers, financial data, or other regulated content.
```
warwickschroeder, line 215: > Client-side encryption can be used with NServiceBus saga data

With the outbox and business data too?

persistence/cosmosdb/index.md
```diff
@@ -207,6 +208,39 @@ They may also be set when using a `CosmosClientBuilder` via the [`WithThrottling
...
+Cosmos DB Always Encrypted supports two [encryption types](https://learn.microsoft.com/en-us/azure/cosmos-db/how-to-always-encrypted#randomized-vs-deterministic-encryption):
+
+- **Randomized** — more secure, but prevents queries from filtering on encrypted properties. Suitable for saga data properties that are only read via point reads.
+- **Deterministic** — always produces the same ciphertext for a given plaintext value, allowing equality filters on encrypted properties.
+
+NServiceBus derives the saga ID from the correlation property and performs a point read, so saga data properties can use randomized encryption.
+
+### Encrypting id and partition key
+
+With [encryption policy format version 1](https://learn.microsoft.com/en-us/azure/cosmos-db/how-to-always-encrypted#choosing-a-client-encryption-policy), the `id` and partition key properties cannot be encrypted. [Policy format version 2](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/3241) lifts this restriction, allowing `id` and partition key to be encrypted using deterministic encryption. The [encryption sample](/samples/cosmosdb/encryption/) demonstrates this using policy format version 2.
```
warwickschroeder: Is there not rather some Microsoft documentation we can link to for the v2 policy format? I'm also not 100% sure we can do this i.e. "any property that is NOT used with a direct equality operator cannot be encrypted". Do we use `id` in this way? If not we probably should say its possible.

persistence/cosmosdb/index.md
```diff
@@ -207,6 +208,39 @@ They may also be set when using a `CosmosClientBuilder` via the [`WithThrottling
...
+
+### Limitations
+
+The following property cannot be encrypted:
+
+- **`_NServiceBus-Persistence-Metadata`** — The persistence metadata property cannot be encrypted because [pessimistic locking](/persistence/cosmosdb/saga-concurrency.md#sagas-concurrency-control-pessimistic-locking-internals) patches a nested path within this object. Only top-level properties can be included in an encryption policy, so encrypting the metadata property would encrypt the entire subtree and prevent the lock from functioning.
```
warwickschroeder, line 234: I know in the issue this was given as an example, but is this the only one?

persistence/cosmosdb/index.md
```diff
@@ -207,6 +208,39 @@ They may also be set when using a `CosmosClientBuilder` via the [`WithThrottling
...
+
+Outbox records are not supported with client-side encryption.
```
warwickschroeder: Why?

persistence/cosmosdb/index.md
```diff
@@ -207,6 +208,39 @@ They may also be set when using a `CosmosClientBuilder` via the [`WithThrottling
...
+
+[`EnableInstallers`](/nservicebus/operations/installers.md) cannot be used with client-side encryption. The NServiceBus installer creates a standard container without an encryption policy, and an encryption policy can only be set at container creation time — it cannot be added or modified afterward. The container must be created manually with the required client encryption key and encryption policy before the endpoint starts.
```
warwickschroeder, line 238: Technically it can, but if it is you need to use `persistence.DisableContainerCreation();`

samples/cosmosdb/encryption/sample.md
```diff
@@ -0,0 +1,75 @@
...
+* `OrderSaga` publishes an `OrderCompleted` event when the `CompleteOrder` timeout fires.
+
+## Running the sample
+
+1. Start the `Server` project and wait for the endpoint to report that it has started.
+1. Start the `Client` project.
+1. Press <kbd>S</kbd> in the client window to start an order.
+1. After the saga timeout expires, verify that the server reports the completed saga and the client receives the `OrderCompleted` event.
+
+## Implementation highlights
```
warwickschroeder, line 42: Might be worth having a prerequisite step to manually create the container first

samples/cosmosdb/encryption/CosmosDB_3/Server/Program.cs
```diff
@@ -0,0 +1,87 @@
...
+             endpointConfiguration.UseSerialization<SystemJsonSerializer>();
+             endpointConfiguration.EnableInstallers();
+                      
+             return endpointConfiguration;
+         });
+
+    static async Task<CosmosClient> CreateEncryptingClientAsync()
+    {
+        var connection = Environment.GetEnvironmentVariable("COSMOS_CONNECTION_STRING")
+            ?? @"AccountEndpoint=https://localhost:8081/;AccountKey=C2y6yDjf5/R+ob0N8A7Cgv30VRDJIWEHLM+4QDU5DE2nQ9nDuVTqobD4b8mGGyPMbIZnqyMsEcaGQy67XIw/Jw==";
```
warwickschroeder, line 50: Assuming the hostname and port is the default for the cosmos emulator? What about the account key? Replace with a tag? Same in CosmosDB_4

warwickschroeder (approved):
Some minor things for your review. The simplicity of using a demo key resolver and the emulator is great, but just wondering if a more real life sample for Azure would also be beneficial. e.g. using Azure Key Vault with all of its security setup.

persistence/cosmosdb/index.md
jpalac:
```suggestion
Outbox records are not supported with client-side encryption. The outbox internally uses stream-based Cosmos DB APIs that bypass the encryption client, so outbox records would be written and read without encryption regardless of the container's encryption policy.
```

persistence/cosmosdb/index.md
jpalac:
```suggestion
[`EnableInstallers`](/nservicebus/operations/installers.md) cannot be used with client-side encryption. The NServiceBus installer creates a standard container without an encryption policy, and an encryption policy can only be set at container creation time — it cannot be added or modified afterward. The container must be created manually with the required client encryption key and encryption policy before the endpoint starts.

If you need the installers for other reasons, you can instead disable container creation by calling `persistence.DisableContainerCreation();`
```

persistence/cosmosdb/index.md
jpalac: The id is the key - it is used with a direct equality operator.
We've updated the link.

```suggestion
With [encryption policy format version 1](https://learn.microsoft.com/en-us/azure/cosmos-db/how-to-always-encrypted#choosing-a-client-encryption-policy), the `id` and partition key properties cannot be encrypted. [Policy format version 2](https://learn.microsoft.com/en-us/dotnet/api/microsoft.azure.cosmos.clientencryptionpolicy?view=azure-dotnet) [lifts this restriction](https://github.com/Azure/azure-cosmos-dotnet-v3/pull/3241), allowing `id` and partition key to be encrypted using deterministic encryption. The [encryption sample](/samples/cosmosdb/encryption/) demonstrates this using policy format version 2.
```

samples/cosmosdb/encryption/CosmosDB_3/Server/Program.cs
jpalac: It is the default for the emulator and we use this in other samples as well.

persistence/cosmosdb/index.md
jpalac: Outbox not - stated below. Business data yes - assuming the user sets up their cosmosdb that way - it's outside the scope of nsb encryption.

samples/cosmosdb/encryption/sample.md
jpalac: The sample does the [container creation](https://github.com/Particular/docs.particular.net/pull/8461/changes#diff-2f4e5c7d6da5d2ab138f8b3ac16f3060d740df99f420806d6500ae49d42f10deR68-R71)

persistence/cosmosdb/index.md
jpalac: As far as we can find - yes

Issue #8226, opened 2023-05-15
Better guidance for deploying endpoints without installers enabled


danielmarbach:
#### Is your improvement related to a problem? Please describe.
Currently users deploying an endpoint without installers enabled need to use the AWS CLI or any other way to deploy the tables and there is no guidance available of what is required. Having some guidance or helpful samples around what is required would be nice

jasontaylordev:
DynamoDB is now covered by #8463, which adds `persistence/dynamodb/table-creation.md`.

PR #8463, opened 2026-08-19
Add DynamoDB table creation page

jasontaylordev:
This pull request addresses https://github.com/Particular/docs.particular.net/issues/8226 by adding documentation for manually creating the DynamoDB table required by NServiceBus persistence, and updates related navigation and references to improve clarity for users deploying to production environments.

### DIFF
--- a/persistence/dynamodb/table-creation.md
+++ b/persistence/dynamodb/table-creation.md
@@ -1,0 +1,71 @@
+---
+title: Table creation (DynamoDB)
+summary: How to create the table used by DynamoDB persistence instead of relying on installers
+component: DynamoDB
+reviewed: 2026-08-19
+related:
+- persistence/dynamodb
+- nservicebus/operations/installers
+- nservicebus/aws/local-development
+---
+
+When [installers](/nservicebus/operations/installers.md) are enabled, an endpoint creates the table it needs at startup. This is convenient during development, but it requires the endpoint to run with permission to create and configure tables, which it never needs once the table exists.
+
+In production, the deployment process usually creates the table before the endpoint is deployed. This article describes the table that DynamoDB persistence expects, and shows how to create it with the AWS CLI. CloudFormation, Terraform, or any other deployment mechanism can create the same table.
+
+## Installers
+
+An endpoint only creates tables at startup when installers are enabled:
+
+```csharp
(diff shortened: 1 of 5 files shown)


persistence/dynamodb/table-creation.md
```diff
@@ -0,0 +1,71 @@
...
+When the deployment process creates the table, omit `EnableInstallers`. If installers are needed for other reasons, disable only the table creation:
+
+snippet: DynamoDBDisableTableCreation
+
+## Permissions
+
+Creating the table becomes the responsibility of the deployment process, so use separate deployment and application identities where possible. The deployment identity needs permission to create and configure the table. The application identity needs only runtime data access, not the table management permissions that installers require.
+
+The minimum permission set for each is documented in [permissions](/persistence/dynamodb/#permissions).
+
```
jpalac, line 33: Is it worth adding something here for running this locally? 
That they would need to configure an AWS CLI profile with dummy local credentials.

Then the scripts would need those extra bits to make them work.

I know this is about production, but the heading is "table creation" and users might not know the different steps for local testing.

persistence/dynamodb/table-creation.md
```diff
@@ -0,0 +1,71 @@
+---
+title: Table creation (DynamoDB)
```
jpalac: It's already under DynamoDB - why do we think that we need to put the name in the brackets?

jasontaylordev: I followed the convention established in related docs, see Sagas and Outbox. Happy to remove if you feel it is not necessary.

persistence/dynamodb/table-creation.md
jasontaylordev: I think it would be worthwhile, I'll push through an update.

persistence/dynamodb/table-creation.md
jpalac: Hmm - that is weird. I've looked at other persisters and none of them do that. So if we remove it, we need to do the same from the other pages. Thoughts?

jasontaylordev: I think consistency is important. I've removed it from relevant pages.

jpalac (approved):

stevedowling (approved):

PR #8471, opened 2026-08-26
Custom Recoverability Policy docs review

stevedowling:


nservicebus/recoverability/custom-recoverability-policy.md
```diff
@@ -66,10 +66,6 @@ The information provided in the configuration is static and will not change betw
 
 ## Implement a custom policy
 
-> [!NOTE]
```
jpalac, line 69: is this no longer applicable?

PhilBastian: if it is, I question why it wasn't implemented using versioned snippets

stevedowling: That was the thing - I actually started to create snippets and couldn't figure out what they would be for 6.0 so I decided that it was out of support and not worth going further with and that the note was just confusing

PhilBastian (approved):

